PR #3, opened 2023-07-11
VFB 1 - Create table component

ChloeYiuSoftwire:
Make a table component
Add demo in Client Page
Created cypress component test for table

### DIFF
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -1,6 +1,6 @@
 import supabase, { Schema } from "@/supabase";
 import { Metadata } from "next";
-
+import Table from "@/components/Tables/Table";
 import React from "react";
 
 const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
@@ -10,6 +10,23 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 
 const Clients: () => Promise<React.ReactElement> = async () => {
     const data = await dataFetch();
+
+    if (data === null) {
+        return (
+            <main>
+                <h1> Clients Page </h1>
+                <p> Error: Data not found</p>
+            </main>
+        );
+    }
+
+    const headers: { [key: string]: string } = {
+        full_name: "Name",
+        phone_number: "Phone Number",
+        address_1: "Address",
+        address_2: "Address 2",
+        address_3: "Address 3",
+    };
 
     return (
         <main>
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -3,13 +3,13 @@ import { Metadata } from "next";
 import { Datum } from "@/components/Tables/Table";
 import React from "react";
 
-const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
+const dataFetch: () => Promise<Schema["clients"][]> = async () => {
     const response = await supabase.from("clients").select();
-    return response.data;
+    return response.data ?? [];
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    const data: Datum[] = (await dataFetch()) ?? [];
+    const data: Datum[] = await dataFetch();
 
     return (
         <main>
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -9,13 +9,9 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    let data: Datum[] | null = await dataFetch();
+    const data: Datum[] | null = (await dataFetch()) ?? [];
 
-    if (data === null) {
-        data = [];
-    }
-
-    const headers: { [key: string]: string } = {
+    const headers = {
         full_name: "Name",
         phone_number: "Phone Number",
         address_1: "Address",
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -1,6 +1,6 @@
 import supabase, { Schema } from "@/supabase";
 import { Metadata } from "next";
-import Table from "@/components/Tables/Table";
+import Table, { Datum } from "@/components/Tables/Table";
 import React from "react";
 
 const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
@@ -9,15 +9,10 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    const data = await dataFetch();
+    let data: Datum[] | null = await dataFetch();
 
     if (data === null) {
-        return (
-            <main>
-                <h1> Clients Page </h1>
-                <p> Error: Data not found</p>
-            </main>
-        );
+        data = [];
     }
 
     const headers: { [key: string]: string } = {
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -1,6 +1,6 @@
 import supabase, { Schema } from "@/supabase";
 import { Metadata } from "next";
-import Table, { Datum } from "@/components/Tables/Table";
+import { Datum } from "@/components/Tables/Table";
 import React from "react";
 
 const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -11,14 +11,6 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 const Clients: () => Promise<React.ReactElement> = async () => {
     const data: Datum[] = (await dataFetch()) ?? [];
 
-    const headers = {
-        full_name: "Name",
-        phone_number: "Phone Number",
-        address_1: "Address",
-        address_2: "Address 2",
-        address_3: "Address 3",
-    };
-
     return (
         <main>
             <h1> Clients Page </h1>
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -9,7 +9,7 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    const data: Datum[] | null = (await dataFetch()) ?? [];
+    const data: Datum[] = (await dataFetch()) ?? [];
 
     const headers = {
         full_name: "Name",
--- a/src/components/Tables/Table.cy.tsx
+++ b/src/components/Tables/Table.cy.tsx
@@ -65,20 +65,7 @@ describe("<Table />", () => {
         },
     ];
 
-    const smallerData = [
-        {
-            full_name: "Tom",
-            phone_number: 123456,
-        },
-        {
-            full_name: "Sam",
-            phone_number: 999,
-        },
-        {
-            full_name: "Harper Garrett",
-            phone_number: 2171786554,
-        },
-    ];
+    const smallerData = data.slice(0, 3);
 
     const headers = {
         full_name: "Name",
--- a/src/components/Tables/Table.cy.tsx
+++ b/src/components/Tables/Table.cy.tsx
@@ -92,7 +92,7 @@ describe("<Table />", () => {
     it("sorting is correct", () => {
         cy.mount(<Table data={data} headers={headers} />);
         cy.get("div").contains("Name").parent().click();
-        cy.get("div[data-column-id='1'][role='cell']").as("table");
+        cy.get("div[data-column-id='2'][role='cell']").as("table");
         cy.get("@table").eq(0).contains("Adrian Key");
         cy.get("@table").eq(1).contains("Agnes Rosales");
         cy.get("div").contains("Name").parent().click();
@@ -110,7 +110,7 @@ describe("<Table />", () => {
 
     it("pagination page change is working", () => {
         cy.mount(<Table data={data} headers={headers} />);
-        cy.get("div[data-column-id='1'][role='cell']").as("table");
+        cy.get("div[data-column-id='2'][role='cell']").as("table");
         cy.get("@table").eq(0).contains("Tom");
         cy.get("button[id='pagination-next-page']").click();
         cy.get("@table").eq(0).contains("Forbes Doyle");
@@ -119,7 +119,7 @@ describe("<Table />", () => {
     it("pagination number of items is working", () => {
         cy.mount(<Table data={data} headers={headers} />);
         cy.get("select[aria-label='Rows per page:']").select("15");
-        cy.get("div[data-column-id='1'][role='cell']").as("table");
+        cy.get("div[data-column-id='2'][role='cell']").as("table");
         cy.get("@table").eq(14).contains("Chloe");
         cy.get("@table").eq(15).should("not.exist");
     });
--- a/src/components/Tables/Table.cy.tsx
+++ b/src/components/Tables/Table.cy.tsx
@@ -77,7 +77,7 @@ describe("<Table />", () => {
         {
             full_name: "Harper Garrett",
             phone_number: 2171786554,
-        }
+        },
     ];
 
     const headers = {
@@ -146,7 +146,7 @@ describe("<Table />", () => {
         cy.get("input[aria-label='Select row 0']").should("be.checked");
         cy.get("input[aria-label='Select row 0']").click();
         cy.get("input[aria-label='Select row 0']").should("not.be.checked");
-    })
+    });
 
     it("checkbox toggles only the selected one", () => {
         cy.mount(<Table data={data} headers={headers} />);
@@ -162,7 +162,7 @@ describe("<Table />", () => {
         cy.get("input[aria-label='Select row 0']").should("not.be.checked");
         cy.get("input[aria-label='Select row 1']").should("not.be.checked");
         cy.get("input[aria-label='Select row 2']").should("be.checked");
-    })
+    });
 
     it("filtering does not affect checkbox", () => {
         cy.mount(<Table data={data} headers={headers} />);
@@ -171,7 +171,7 @@ describe("<Table />", () => {
         cy.get("input[placeholder='Filter by Name']").type("Sam");
         cy.get("input[placeholder='Filter by Name']").clear();
         cy.get("input[aria-label='Select row 0']").should("be.checked");
-    })
+    });
 
     it("changing page does not affect checkbox", () => {
         cy.mount(<Table data={data} headers={headers} />);
@@ -180,7 +180,7 @@ describe("<Table />", () => {
         cy.get("button[id='pagination-next-page']").click();
         cy.get("button[id='pagination-previous-page']").click();
         cy.get("input[aria-label='Select row 0']").should("be.checked");
-    })
+    });
 
     it("pagination limit does not affect checkbox", () => {
         cy.mount(<Table data={data} headers={headers} />);
@@ -199,7 +199,7 @@ describe("<Table />", () => {
 
         cy.get("input[aria-label='Select row 0']").should("be.checked");
         cy.get("input[aria-label='Select row 14']").should("be.checked");
-    })
+    });
 
     it("sorting does not affect checkbox", () => {
         cy.mount(<Table data={data} headers={headers} />);
@@ -213,10 +213,10 @@ describe("<Table />", () => {
         cy.get("div").contains("Name").parent().click();
         cy.get("input[aria-label='Select row 2']").should("be.checked");
         cy.get("div").contains("Name").parent().click();
-        
+
         cy.get("input[aria-label='Select row 0']").should("be.checked");
         cy.get("input[aria-label='Select row 2']").should("be.checked");
-    })
+    });
 
     it("checkall box toggles all data", () => {
         cy.mount(<Table data={smallerData} headers={headers} />);
@@ -224,7 +224,7 @@ describe("<Table />", () => {
         cy.get("input[aria-label='Select row 0']").should("be.checked");
         cy.get("input[aria-label='Select row 1']").should("be.checked");
         cy.get("input[aria-label='Select row 2']").should("be.checked");
-    })
+    });
 
     it("uncheck one checkbox uncheck the checkall box", () => {
         cy.mount(<Table data={smallerData} headers={headers} />);
@@ -235,14 +235,14 @@ describe("<Table />", () => {
         cy.get("input[aria-label='Select row 1']").should("be.checked");
         cy.get("input[aria-label='Select row 2']").should("be.checked");
         cy.get("input[aria-label='Select all rows']").should("not.be.checked");
-    })
+    });
 
     it("check all checkboxes check the checkall box", () => {
         cy.mount(<Table data={smallerData} headers={headers} />);
         cy.get("input[aria-label='Select row 0']").click();
         cy.get("input[aria-label='Select row 1']").click();
         cy.get("input[aria-label='Select row 2']").click();
-        
+
         cy.get("input[aria-label='Select all rows']").should("be.checked");
-    })
+    });
 });
--- a/src/components/Tables/Table.cy.tsx
+++ b/src/components/Tables/Table.cy.tsx
@@ -226,7 +226,7 @@ describe("<Table />", () => {
         cy.get("input[aria-label='Select row 2']").should("be.checked");
     });
 
-    it("uncheck one checkbox uncheck the checkall box", () => {
+    it("uncheck one row unticks the checkall box", () => {
         cy.mount(<Table data={smallerData} headers={headers} />);
         cy.get("input[aria-label='Select all rows']").click();
         cy.get("input[aria-label='Select row 0']").click();
@@ -237,7 +237,7 @@ describe("<Table />", () => {
         cy.get("input[aria-label='Select all rows']").should("not.be.checked");
     });
 
-    it("check all checkboxes check the checkall box", () => {
+    it("check all rows ticks the checkall box", () => {
         cy.mount(<Table data={smallerData} headers={headers} />);
         cy.get("input[aria-label='Select row 0']").click();
         cy.get("input[aria-label='Select row 1']").click();
--- a/src/components/Tables/Table.cy.tsx
+++ b/src/components/Tables/Table.cy.tsx
@@ -1,0 +1,126 @@
+import React from "react";
+import Table from "./Table";
+
+describe("<Table />", () => {
+    const data = [
+        {
+            full_name: "Tom",
+            phone_number: 123456,
+        },
+        {
+            full_name: "Sam",
+            phone_number: 999,
+        },
+        {
+            full_name: "Harper Garrett",
+            phone_number: 2171786554,
+        },
+        {
+            full_name: "Adrian Key",
+            phone_number: 3650099130,
+        },
+        {
+            full_name: "Harrell Wallace",
+            phone_number: 4650047935,
+        },
+        {
+            full_name: "Oneill Curtis",
+            phone_number: 7058491995,
+        },
+        {
+            full_name: "Herring Rutledge",
+            phone_number: 1440882899,
+        },
+        {
+            full_name: "Eloise Rowland",
+            phone_number: 2580325390,
+        },
+        {
+            full_name: "Cathryn Burks",
+            phone_number: 7136166489,
+        },
+        {
+            full_name: "Paopao",
+            phone_number: 7136166469,
+        },
+        {
+            full_name: "Forbes Doyle",
+            phone_number: 1377097191,
+        },
+        {
+            full_name: "Agnes Rosales",
+            phone_number: 3334796379,
+        },
+        {
+            full_name: "Jan Orr",
+            phone_number: 1526538148,
+        },
+        {
+            full_name: "Colleen Lowery",
+            phone_number: 3980156139,
+        },
+        {
+            full_name: "Chloe",
+            phone_number: 4567894522,
+        },
+    ];
+
+    const headers = {
+        full_name: "Name",
+        phone_number: "Phone Number",
+    };
+
+    it("renders", () => {
+        cy.mount(<Table data={data} headers={headers} />);
+    });
+
+    it("can display data", () => {
+        cy.mount(<Table data={data} headers={headers} />);
+        cy.contains("Tom");
+        cy.contains("Sam");
+        cy.contains("123456");
+        cy.contains("999");
+    });
+
+    it("filter is correct", () => {
+        cy.mount(<Table data={data} headers={headers} />);
+        cy.get("input[placeholder='Filter by Name']").type("Tom");
+        cy.contains("Tom");
+        cy.should("not.have.value", "Sam");
+    });
+
+    it("sorting is correct", () => {
+        cy.mount(<Table data={data} headers={headers} />);
+        cy.get("div").contains("Name").parent().click();
+        cy.get("div[data-column-id='1'][role='cell']").as("table");
+        cy.get("@table").eq(0).contains("Adrian Key");
+        cy.get("@table").eq(1).contains("Agnes Rosales");
+        cy.get("div").contains("Name").parent().click();
+        cy.get("@table").eq(0).contains("Tom");
+        cy.get("@table").eq(1).contains("Sam");
+    });
+
+    it("clear button is working", () => {
+        cy.mount(<Table data={data} headers={headers} />);
+        cy.get("input[placeholder='Filter by Name']").type("Tom");
+        cy.get("button").contains("Clear").click();
+        cy.contains("Sam");
+        cy.get("input[placeholder='Filter by Name']").should("have.value", "");
+    });
+
+    it("pagination page change is working", () => {
+        cy.mount(<Table data={data} headers={headers} />);
+        cy.get("div[data-column-id='1'][role='cell']").as("table");
+        cy.get("@table").eq(0).contains("Tom");
+        cy.get("button[id='pagination-next-page']").click();
+        cy.get("@table").eq(0).contains("Forbes Doyle");
+    });
+
+    it("pagination number of items is working", () => {
+        cy.mount(<Table data={data} headers={headers} />);
+        cy.get("select[aria-label='Rows per page:']").select("15");
+        cy.get("div[data-column-id='1'][role='cell']").as("table");
+        cy.get("@table").eq(14).contains("Chloe");
+        cy.get("@table").eq(15).should("not.exist");
+    });
+});
--- a/src/components/Tables/Table.cy.tsx
+++ b/src/components/Tables/Table.cy.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import Table from "./Table";
+import Table from "@/components/Tables/Table";
 
 describe("<Table />", () => {
     const data = [
--- a/src/components/Tables/Table.cy.tsx
+++ b/src/components/Tables/Table.cy.tsx
@@ -65,6 +65,21 @@ describe("<Table />", () => {
         },
     ];
 
+    const smallerData = [
+        {
+            full_name: "Tom",
+            phone_number: 123456,
+        },
+        {
+            full_name: "Sam",
+            phone_number: 999,
+        },
+        {
+            full_name: "Harper Garrett",
+            phone_number: 2171786554,
+        }
+    ];
+
     const headers = {
         full_name: "Name",
         phone_number: "Phone Number",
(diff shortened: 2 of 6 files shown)


src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,99 @@
+"use client";
+
+import React, { useState, useMemo } from "react";
+import DataTable, { TableColumn } from "react-data-table-component";
+
+interface Props {
+    data: any[];
```
DanWilkins2107, line 7: Worth adding an interface for data or no? To use instead of any[].

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,99 @@
...
+                        value={filterText[key]}
+                        placeholder={`Filter by ${value}`}
+                        onChange={(e) => onFilter(e, key)}
+                    />
+                );
+            })}
+            <button onClick={onClear}>Clear</button>
+        </>
+    );
+};
```
daniel-conlan-softwire: Probably best to have this in a separate "FilterComponent" file instead of grouping the implementation with the Table component. You can also then rewrite "FilterComponentProps" as simply "Props" within the new file. Would improve clarity and modularity of code

src/components/Tables/Table.cy.tsx
daniel-conlan-softwire, line 1: Perhaps adding some additional tests to check that pagination works? Could add a test that mounts a large table (more items than page size) and then ensure that only N items are displayed, where N is the page size.

src/app/clients/page.tsx
```diff
@@ -11,14 +16,33 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
...
         <>
             <h1> Clients Page </h1>
 
-            <p> Testing Supabase fetching </p>
-
             {/* This should be a separate component which is passed data via props */}
-            <pre>{JSON.stringify(data, null, 4)}</pre>
+            {/* <pre>{JSON.stringify(data, null, 4)}</pre> */}
+            <Table data={dataWithoutKey} headers={headers} />
         </>
```
daniel-conlan-softwire, line 46: Should wrap the logic for rendering the `Table `component in an else block if `data `is `null`. Currently will render both, even if `data` is `null`.

daniel-conlan-softwire (commented):
Will finish review tomorrow.

src/components/Tables/Table.tsx
ChloeYiuSoftwire: We have replaced it with a more specific type (while generic enough for all JSON from Schema)

data: { [key: string]: string | number | null }[];

src/app/clients/page.tsx
```diff
@@ -11,14 +16,33 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
...
+    });
     return (
         <>
             <h1> Clients Page </h1>
 
-            <p> Testing Supabase fetching </p>
-
             {/* This should be a separate component which is passed data via props */}
-            <pre>{JSON.stringify(data, null, 4)}</pre>
+            {/* <pre>{JSON.stringify(data, null, 4)}</pre> */}
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: We have removed it just now

src/app/clients/page.tsx
ChloeYiuSoftwire: We have implemented the logic

src/app/clients/page.tsx
```diff
@@ -11,14 +16,33 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 const Clients: () => Promise<React.ReactElement> = async () => {
     const data = await dataFetch();
 
+    if (data === null) {
+        <>
+            <h1> Clients Page </h1>
+            <p> Error: Data not found</p>
+        </>;
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: That is probably a different ticket - so we will leave it in the future, but we agree it should be a separate component

src/app/clients/page.tsx
```diff
@@ -11,14 +16,33 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
...
+    }
+
+    const headers: { [key: string]: string } = {
+        full_name: "Name",
+        phone_number: "Phone Number",
+        address_1: "Address",
+    };
+
+    const dataWithoutKey: DataForTable[] = data!.map((datum: Schema["clients"]) => {
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Data cannot be null as we have caught the error in if (data === null) { return ... } part

src/components/Tables/Table.tsx
ChloeYiuSoftwire: We have separated out the component as recommended.

src/components/Tables/Table.cy.tsx
ChloeYiuSoftwire: We added 2 more tests for pagination.

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+interface Props {
+    data: { [key: string]: string | number | null }[];
+    headers: { [key: string]: string | null };
+}
+
+const Table: React.FC<Props> = (props) => {
+    const [filterText, setFilterText] = useState<{ [key: string]: string }>({});
+
+    // temporary solution to fix hydration error
```
DanWilkins2107, line 15: Is this still the case? Can a more permanent solution be found?

DanWilkins2107: Or can comment be removed?

paopaosoftwire: We have tried to find multiple solutions and all working solution suggests forcing rendering the component on the client side which is what that line of code is doing. We commented it as a temporary solution as this seems to be a workaround rather than a solution.

As this method is mentioned as the first method here, https://nextjs.org/docs/messages/react-hydration-error . I think We can leave it in.

src/components/Tables/Table.cy.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const headers = {
+        full_name: "Name",
+        phone_number: "Phone Number",
+    };
+
+    it("renders", () => {
+        cy.mount(<Table data={data} headers={headers} />);
+    });
+
+    it("display is correct", () => {
```
SamYap-Softwire, line 77: I don't think the name of the test matches what it does - perhaps you can make the data only contain the 2 records that you are testing? 

(Since the test currently doesn't ensure the display is correct even if it is passed)

src/components/Tables/Table.cy.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    it("pagination page change is working", () => {
+        cy.mount(<Table data={data} headers={headers} />);
+        cy.get("div[data-column-id='1'][role='cell']").as("table");
+        cy.get("@table").eq(0).contains("Tom");
+        cy.get("button[id='pagination-next-page']").click();
+        cy.get("@table").eq(0).contains("Forbes Doyle");
+    });
+
+    it("pagination number of items is working", () => {
```
SamYap-Softwire, line 119: I think you can also test that there is no record on the 16th row

SamYap-Softwire (approved):
My comments are only minor things. I think overall it's good to go.

src/components/Tables/Table.cy.tsx
ChloeYiuSoftwire: We have edited it in our newer version

src/components/Tables/Table.cy.tsx
ChloeYiuSoftwire: We have added it in our newer version

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,79 @@
...
+                }}
+                onClear={handleClear}
+                filterText={filterText}
+                headers={props.headers}
+            />
+        );
+    }, [filterText, props.headers]);
+
+    const filteredItems = props.data.filter((item) => {
+        let filterCondition = true;
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: We have refactored this as suggested in the newer version

src/app/themes.tsx
```diff
@@ -29,6 +29,10 @@ const StyleManager: React.FC<Props> = ({ children }) => {
         return <>{styles}</>;
     });
 
+    if (typeof window !== "undefined") {
```
TomCusack-Softwire, line 32: can you rebase to (current) main to get rid of these sort of changes - which are cherrypicked when you needed them? :smile:

src/components/Tables/Table.cy.tsx
```diff
@@ -0,0 +1,126 @@
...
+        {
+            full_name: "Tom",
+            phone_number: 123456,
+        },
+        {
+            full_name: "Sam",
+            phone_number: 999,
+        },
+        {
+            full_name: "Harper Garrett",
```
TomCusack-Softwire, line 15: where did these names come from? I love it 😆

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+interface Props {
+    data: { [key: string]: string | number | null }[];
+    headers: { [key: string]: string | null };
+}
+
+const Table: React.FC<Props> = (props) => {
+    const [filterText, setFilterText] = useState<{ [key: string]: string }>({});
+
+    // temporary solution to fix hydration error
```
TomCusack-Softwire, line 15: Can you remove 'temporary' - just `Fixes hydration error by re-rendering when the DOM is ready` would be better.
Potentially even leaving the link to the NextJS docs you got this from, or why it is necessary

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const [domLoaded, setDomLoaded] = useState(false);
+    useEffect(() => {
+        setDomLoaded(true);
+    }, []);
+
+    const subHeaderComponentMemo = useMemo(() => {
```
TomCusack-Softwire, line 21: What is `useMemo` and why did you use this here, as opposed to having this be outside the component and passing in the state as props?

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,80 @@
...
+        const handleClear: () => void = () => {
+            if (filterText) {
+                setFilterText({});
+            }
+        };
+        return (
+            <TableFilterBar
+                onFilter={(e: React.ChangeEvent<HTMLInputElement>, filterField: string): void => {
+                    const filterTextCopy = { ...filterText };
```
TomCusack-Softwire, line 30: is this equivalent to `{ ...filtertext, filterField: e.target.value }` ?

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,80 @@
...
+                filterText={filterText}
+                headers={props.headers}
+            />
+        );
+    }, [filterText, props.headers]);
+
+    const filteredItems = props.data.filter((item) => {
+        for (const key of Object.keys(props.headers)) {
+            if (
+                !(item[key] ?? "")
```
TomCusack-Softwire, line 44: not a fan of the autoformatting making this multiple lines - this suggests it's quite a complex `if` statement.
Can this be extracted as a function where you pass it both `item[key]` and `filterText[key]` and it returns a boolean?

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,80 @@
...
+                return false;
+            }
+        }
+        return true;
+    });
+
+    const columns: TableColumn<any>[] = Object.entries(props.headers).map(([key, value]) => {
+        return {
+            name: value,
+            selector: (row: any) => row[key],
```
TomCusack-Softwire, line 58: can this `any` be more specific - potentially something like 'an object which has the `key` variable as an attribute'?

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,80 @@
...
+            sortable: true,
+        };
+    });
+
+    return (
+        <div>
```
TomCusack-Softwire, line 64: What's this outer `div` doing?

src/components/Tables/TableFilterBar.tsx
```diff
@@ -0,0 +1,31 @@
+"use client";
+
+import React from "react";
+
+interface Props {
+    filterText: { [key: string]: string };
+    onFilter: (e: React.ChangeEvent<HTMLInputElement>, filterField: string) => void; // eslint-disable-line no-unused-vars
```
TomCusack-Softwire, line 7: should be fixed by rebasing on more recent main - try and remove this comment and see? 🤞

TomCusack-Softwire (changes requested):

TomCusack-Softwire:
also that's a _lot_ of commits! There's nothing wrong with this, but just a note to say you don't have to commit every small change separately! Potentially just committing when you're happy and it 'works' (or at the end of the day etc) may waste less of your time :P

src/components/Tables/Table.cy.tsx
paopaosoftwire: https://json-generator.com/ lol

src/components/Tables/Table.tsx
paopaosoftwire: Done!

src/components/Tables/Table.tsx
paopaosoftwire: I have moved it outside of the component.

src/components/Tables/Table.tsx
paopaosoftwire: From my understanding, the two are not equivalent.

The current code: takes in a filterField as a string then sets the corresponding field in filterText to a value of e.target.value
eg. console.log(filterText["full_name"]) ---> Tom Jerry
 
The suggested code: sets the field call filterFied in filterText to a value of e.target.value
eg. console.log(filterText["filterField"]) ---> Tom Jerry

src/components/Tables/Table.tsx
paopaosoftwire: Nothing...

Fixed it!

src/components/Tables/Table.tsx
ChloeYiuSoftwire: resolved as suggested

src/app/themes.tsx
ChloeYiuSoftwire: Rebased as suggested!

src/components/Tables/TableFilterBar.tsx
ChloeYiuSoftwire: Rebased as suggested!

src/components/Tables/Table.tsx
TomCusack-Softwire: That's completely correct!
Can this be solved with square brackets, like
`{ ...filtertext, [filterField]: e.target.value }`
?

TomCusack-Softwire (changes requested):
Almost there, just a small number of comments now!

Also, with the interactive rebase, you can name your commits something slightly nicer than 'This is a combination of X commits' - try another interactive rebase to make these names nice and descriptive!
You'll be a Git wizard in no time! 🙂

src/components/Tables/Table.tsx
paopaosoftwire: Thats a valid syntax. Thanks!

src/app/clients/page.tsx
```diff
@@ -11,14 +11,27 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 const Clients: () => Promise<React.ReactElement> = async () => {
     const data = await dataFetch();
 
+    if (data === null) {
+        return (
+            <main>
+                <h1> Clients Page </h1>
+                <p> Error: Data not found</p>
```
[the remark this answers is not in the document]
paopaosoftwire: Fixed!

src/components/DataViewer/SampleDataViewerWithButton.tsx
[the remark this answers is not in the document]
paopaosoftwire: I think it was in there by accident. Perhaps through some git magic. It had been removed.

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,122 @@
...
+    filterText: Headers;
+    setFilterText: React.Dispatch<
+        React.SetStateAction<{
+            [key: string]: string;
+        }>
+    >;
+    headers: Headers;
+}
+
+const subHeaderComponent: React.FC<SubHeaderComponentProps> = (props) => {
```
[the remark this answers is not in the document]
paopaosoftwire: Yes that is possible. I have moved it.

src/app/clients/page.tsx
```diff
@@ -9,16 +9,24 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    const data = await dataFetch();
+    let data: Datum[] | null = await dataFetch();
```
TomCusack-Softwire, line 12: potentially better as
`let data: Datum[] = await dataFetch() ?? []`

TomCusack-Softwire: (which then also absorbs the if statement below)

src/app/clients/page.tsx
```diff
@@ -9,16 +9,24 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
...
+
+    if (data === null) {
+        data = [];
+    }
+
+    const headers: { [key: string]: string } = {
```
TomCusack-Softwire, line 18: this type is probably not needed :)
only functions and things which use custom interfaces need type annotations!

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,100 @@
...
+                .includes((filterText[key] ?? "").toLowerCase())
+        ) {
+            return false;
+        }
+    }
+    return true;
+};
+
+const filteredNoNullItems = (headers: Headers, data: Datum[], filterText: FilterText): Row[] => {
+    return data.reduce((filteredNoNullItems: Row[], item: Datum) => {
```
TomCusack-Softwire, line 38: can the loop variable not be the same name as the function please?

src/components/Tables/TableFilterBar.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import React from "react";
+
+export interface FilterText {
+    [key: string]: string;
+}
+
+interface Props {
+    filterText: FilterText;
+    onFilter: (e: React.ChangeEvent<HTMLInputElement>, filterField: string) => void;
```
TomCusack-Softwire, line 11: no single letter variables! (event?)

src/components/Tables/TableFilterBar.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return (
+        <>
+            {Object.entries(props.headers).map(([key, value]) => {
+                return (
+                    <input
+                        key={key}
+                        type="text"
+                        value={props.filterText[key] ?? ""}
+                        placeholder={`Filter by ${value}`}
+                        onChange={(e) => props.onFilter(e, key)}
```
TomCusack-Softwire, line 26: no single letter variables! (event?)

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,100 @@
...
+            sortable: true,
+            title: value,
+        };
+    });
+
+    const onFilter = (e: React.ChangeEvent<HTMLInputElement>, filterField: string): void => {
+        setFilterText({ ...filterText, [filterField]: e.target.value });
+    };
+
+    const handleClear: () => void = () => {
```
TomCusack-Softwire, line 72: this can be `const handleClear = (): void => {` which I think is clearer

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,100 @@
...
+        if (filterText) {
+            setFilterText({});
+        }
+    };
+
+    return (
+        <>
+            {domLoaded && (
```
TomCusack-Softwire, line 80: I'd rather have an `if` statement separately, e.g.

```
if (domLoaded) {
    return <></>;
}
return <DataTable etc/>;
```

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,100 @@
...
+    const columns: TableColumn<Row>[] = Object.entries(props.headers).map(([key, value]) => {
+        return {
+            name: value,
+            selector: (row: Row) => row[key],
+            sortable: true,
+            title: value,
+        };
+    });
+
+    const onFilter = (e: React.ChangeEvent<HTMLInputElement>, filterField: string): void => {
```
TomCusack-Softwire, line 68: may want to just CTRL-F for `e:`

TomCusack-Softwire (changes requested):

src/app/clients/page.tsx
ChloeYiuSoftwire: Refactored as suggested

src/app/clients/page.tsx
ChloeYiuSoftwire: Refactored as suggested

src/components/Tables/Table.tsx
ChloeYiuSoftwire: Resolved as suggested

src/components/Tables/TableFilterBar.tsx
ChloeYiuSoftwire: refactored as suggested

src/components/Tables/Table.tsx
ChloeYiuSoftwire: refactored as suggested

src/components/Tables/Table.tsx
ChloeYiuSoftwire: refactored as suggested

src/components/Tables/TableFilterBar.tsx
ChloeYiuSoftwire: refactored as suggested

src/components/Tables/Table.tsx
ChloeYiuSoftwire: refactored as suggested

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,149 @@
...
+        ) {
+            return false;
+        }
+    }
+    return true;
+};
+
+const filteredNoNullItems = (headers: Headers, data: Datum[], filterText: FilterText): Row[] => {
+    let row = 0;
+    return data.reduce((filteredNoNullList: Row[], item: Datum, currentIndex: number) => {
```
SamYap-Softwire, line 40: This function is a little hard to follow - is it possible to use a filter and a map? Or just a for loop?

SamYap-Softwire (approved):
This seems fine. Just some minor suggestions / changes

src/app/clients/page.tsx
```diff
@@ -9,16 +9,20 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    const data = await dataFetch();
+    const data: Datum[] | null = (await dataFetch()) ?? [];
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Our function assumes the input data to have type array so that we can perform array-related functions like indexing for instance. Null object will throw an error but empty array won't so we have to keep it

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,149 @@
+"use client";
+
+import React, { useState, useEffect } from "react";
+import DataTable, { TableColumn } from "react-data-table-component";
+import TableFilterBar, { FilterText } from "./TableFilterBar";
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Refactored as suggested

src/components/Tables/Table.cy.tsx
```diff
@@ -0,0 +1,126 @@
+import React from "react";
+import Table from "./Table";
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Refactored as suggested

src/components/Tables/Table.tsx
ChloeYiuSoftwire: There are too many operations if we split into map and filter thats why we use reduce instead

TomCusack-Softwire: are there a maximum number of operations you are allowed to do?
I'd rather have cleaner, longer code than shorter, weird code.
It sounds like @SamYap-Softwire doesn't really know what's going on here, so it's probably best to revert to the filter/map way.

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,149 @@
+"use client";
+
+import React, { useState, useEffect } from "react";
+import DataTable, { TableColumn } from "react-data-table-component";
+import TableFilterBar, { FilterText } from "@/components/Tables/TableFilterBar";
+
+interface Row {
+    checkBoxKey: number;
```
TomCusack-Softwire, line 8: potentially better as `rowId`, which describes what the index _is_ rather than what it _does_

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+const filteredNoNullItems = (headers: Headers, data: Datum[], filterText: FilterText): Row[] => {
+    let row = 0;
+    return data.reduce((filteredNoNullList: Row[], item: Datum, currentIndex: number) => {
+        if (itemIncludesFilterText(headers, item, filterText)) {
+            const noNullItem: Row = { checkBoxKey: currentIndex };
+            for (const key of Object.keys(headers)) {
+                noNullItem[key] = item[key] ?? "";
+            }
+            noNullItem.id = row;
```
TomCusack-Softwire, line 46: does `id` ever get used? I assume we replaced it with `currentIndex` - can we get rid of the `row` variable etc?

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,149 @@
...
+    }, []);
+
+    useEffect(() => {
+        let toggleFlag = true;
+        selectCheckBox.forEach((eachCheckBox) => {
+            if (!eachCheckBox) {
+                toggleFlag = false;
+            }
+        });
+        setSelectAllCheckBox(toggleFlag);
```
TomCusack-Softwire, line 85: can we check for `toggleFlag === selectAllCheckbox` so it doesn't rerender unless it has to?

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,149 @@
...
+const Table: React.FC<Props> = (props) => {
+    const [filterText, setFilterText] = useState<FilterText>({});
+
+    const [selectCheckBox, setSelectCheckBox] = useState(
+        new Array<boolean>(props.data.length).fill(false)
+    );
+    const [selectAllCheckBox, setSelectAllCheckBox] = useState(false);
+
+    const toggleAllCheckBox = (): void => {
+        setSelectCheckBox(new Array<boolean>(props.data.length).fill(!selectAllCheckBox));
```
TomCusack-Softwire, line 63: does this work with pagination? (can you make a test for this? you should test the checkbox stuff _really thoroughly_ because it's something we've made, rather than just trusting that the library we `npm install`ed worked.

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,149 @@
...
+        setSelectAllCheckBox(!selectAllCheckBox);
+    };
+    const toggleOwnCheckBox = (checkBoxKey: number): void => {
+        const selectCheckBoxCopy = [...selectCheckBox];
+        selectCheckBoxCopy[checkBoxKey] = !selectCheckBoxCopy[checkBoxKey];
+        setSelectCheckBox(selectCheckBoxCopy);
+    };
+
+    // Fixes hydration error by re-rendering when the DOM is ready - https://nextjs.org/docs/messages/react-hydration-error
```
TomCusack-Softwire, line 72: is this still needed if we have another `useEffect` in use?

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const [domLoaded, setDomLoaded] = useState(false);
+    useEffect(() => {
+        setDomLoaded(true);
+    }, []);
+
+    useEffect(() => {
+        let toggleFlag = true;
+        selectCheckBox.forEach((eachCheckBox) => {
```
TomCusack-Softwire, line 80: is this the same as `const toggleFlag = selectCheckBox.all();`?

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,149 @@
...
+                aria-label="Select all rows"
+                checked={selectAllCheckBox}
+                onClick={toggleAllCheckBox}
+            />
+        ),
+        cell: (row: Row) => (
+            <input
+                type="checkbox"
+                aria-label={`Select row ${row.checkBoxKey}`}
+                checked={selectCheckBox[Number(row.checkBoxKey)]}
```
TomCusack-Softwire, line 109: why are these `Number(...)` wrapping them? they should just be good?

TomCusack-Softwire (changes requested):

src/components/Tables/Table.tsx
ChloeYiuSoftwire: the idea of row id might be a bit conflicting with the parameter id
id describes after filtering, which specific row it is located in (which is called in the bottom as row.id for each row object)
for checkBoxKey, it is specifically just a unique identifier of each row object that is the same as the unfiltered data (props.data) index, it is irrelevant to which row a particular datum is in the table hence we avoid doing so

src/components/Tables/Table.tsx
ChloeYiuSoftwire: Refactored as suggested

src/components/Tables/Table.tsx
ChloeYiuSoftwire: Yes it is needed (just checked by removing it) - the issue isn't about whether there is useEffect in the function or not, the issue is the table should not be returned until the dom is loaded, which still has to be checked by an initial useEffect (as useEffect can only be run after the dom is loaded)

src/components/Tables/Table.tsx
ChloeYiuSoftwire: Seems like array.prototype.all() doesn't exist but there is array.prototype.every( ... pass condition to check )

src/components/Tables/Table.tsx
ChloeYiuSoftwire: It was complaining just now but it didn't complain now (probably because at first we mistakenly used row.id) -> so refactored as suggested

src/app/clients/page.tsx
```diff
@@ -9,16 +9,20 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
...
 
     return (
         <main>
             <h1> Clients Page </h1>
-
-            <p> Testing Supabase fetching </p>
-
-            {/* This should be a separate component which is passed data via props */}
-            <pre>{JSON.stringify(data, null, 4)}</pre>
+            <Table data={data} headers={headers} />
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Ok we will remove it before our final one then!

src/components/Tables/Table.tsx
ChloeYiuSoftwire: nevermind, resolved as suggested

src/components/Tables/Table.tsx
ChloeYiuSoftwire: nevermind, resolved as suggested

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,150 @@
...
+        return filteredNoNullList;
+    }, []);
+};
+
+const Table: React.FC<Props> = (props) => {
+    const [filterText, setFilterText] = useState<FilterText>({});
+
+    const [selectCheckBox, setSelectCheckBox] = useState(
+        new Array<boolean>(props.data.length).fill(false)
+    );
```
daniel-conlan-softwire, line 62: The current naming of `selectCheckBox` implies that this is a single value, although it is actually an array of the check boxes for each row in the table.

Perhaps to `rowSelectBoxes` or `selectRowBoxes`? You'll need to decide what you think is best.

If you do decide to rename, then you'll need to also rename the other related variables

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    // Fixes hydration error by re-rendering when the DOM is ready - https://nextjs.org/docs/messages/react-hydration-error
+    const [domLoaded, setDomLoaded] = useState(false);
+    useEffect(() => {
+        setDomLoaded(true);
+    }, []);
+
+    useEffect(() => {
+        let toggleFlag = true;
+        selectCheckBox.forEach((eachCheckBox) => {
```
daniel-conlan-softwire: The variable `eachCheckBox` can probably just be renamed to `checkBox`

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,150 @@
...
+        ),
+        cell: (row: Row) => (
+            <input
+                type="checkbox"
+                aria-label={`Select row ${row.checkBoxKey}`}
+                checked={selectCheckBox[Number(row.checkBoxKey)]}
+                onClick={() => toggleOwnCheckBox(Number(row.checkBoxKey))}
+            />
+        ),
+        width: "47px",
```
daniel-conlan-softwire, line 113: Is there a reason for this specific `width`?

src/components/Tables/Table.tsx
paopaosoftwire: Made

src/components/Tables/Table.cy.tsx
```diff
@@ -0,0 +1,248 @@
...
+        cy.get("input[aria-label='Select all rows']").click();
+        cy.get("input[aria-label='Select row 0']").click();
+
+        cy.get("input[aria-label='Select row 0']").should("not.be.checked");
+        cy.get("input[aria-label='Select row 1']").should("be.checked");
+        cy.get("input[aria-label='Select row 2']").should("be.checked");
+        cy.get("input[aria-label='Select all rows']").should("not.be.checked");
+    });
+
+    it("check all checkboxes check the checkall box", () => {
```
SamYap-Softwire, line 240: help TT

SamYap-Softwire (approved):
Some dodgy test names but overall fine

src/components/Tables/Table.cy.tsx
paopaosoftwire: Checked!

DanWilkins2107 (approved):

src/app/clients/page.tsx
```diff
@@ -9,7 +9,7 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    const data = await dataFetch();
+    const data: Datum[] = (await dataFetch()) ?? [];
```
TomCusack-Softwire, line 12: can you move the `?? []` to the `dataFetch`, so that it _can't_ return `null`?
e.g.
```tsx
const dataFetch = async (): Promise<Schema["clients"][]> => {
    // ...
    return response.data ?? [];
}
```

src/components/Tables/Table.cy.tsx
```diff
@@ -0,0 +1,248 @@
...
+            full_name: "Colleen Lowery",
+            phone_number: 3980156139,
+        },
+        {
+            full_name: "Chloe",
+            phone_number: 4567894522,
+        },
+    ];
+
+    const smallerData = [
```
TomCusack-Softwire, line 68: can this not just be `const smallerData = data.slice(0, 3);`? I'd rather not copy and paste 'code', where possible :)

src/components/Tables/Table.cy.tsx
```diff
@@ -0,0 +1,248 @@
...
+        cy.get("input[aria-label='Select row 0']").should("be.checked");
+
+        cy.get("input[aria-label='Select row 2']").click();
+        cy.get("input[aria-label='Select row 2']").should("be.checked");
+
+        cy.get("div").contains("Name").parent().click();
+        cy.get("input[aria-label='Select row 2']").should("be.checked");
+        cy.get("div").contains("Name").parent().click();
+
+        cy.get("input[aria-label='Select row 0']").should("be.checked");
```
TomCusack-Softwire, line 217: are the `aria-label`s always the same when sorted? what counts as 'row 0'?

TomCusack-Softwire (changes requested):
make sure you do resolve each comment (or leave a comment on it) before clicking 'resolve'!
otherwise only a small number of minor changes :)

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,150 @@
...
+    [key: string]: string | number | null;
+}
+
+interface Props {
+    data: Datum[];
+    headers: Headers;
+}
+
+const itemIncludesFilterText = (headers: Headers, item: Datum, filterText: FilterText): boolean => {
+    for (const key of Object.keys(headers)) {
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: done

src/components/Tables/Table.tsx
```diff
@@ -0,0 +1,150 @@
...
+export interface Datum {
+    [key: string]: string | number | null;
+}
+
+interface Props {
+    data: Datum[];
+    headers: Headers;
+}
+
+const itemIncludesFilterText = (headers: Headers, item: Datum, filterText: FilterText): boolean => {
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: done

src/app/clients/page.tsx
ChloeYiuSoftwire: resolved

src/components/Tables/Table.cy.tsx
ChloeYiuSoftwire: done

src/components/Tables/Table.cy.tsx
ChloeYiuSoftwire: oh yes it is based on the rowId (which is the order of the raw props.data from the supabase) rather than the actual position on the table

src/components/Tables/Table.cy.tsx
```diff
@@ -0,0 +1,235 @@
...
+    it("checkall box toggles all data", () => {
+        cy.mount(<Table data={smallerData} headers={headers} />);
+        cy.get("input[aria-label='Select all rows']").click();
+        cy.get("input[aria-label='Select row 0']").should("be.checked");
+        cy.get("input[aria-label='Select row 1']").should("be.checked");
+        cy.get("input[aria-label='Select row 2']").should("be.checked");
+    });
+
+    it("uncheck one row unticks the checkall box", () => {
+        cy.mount(<Table data={smallerData} headers={headers} />);
```
TomCusack-Softwire, line 217: do you actually need `smallerData` here? why does `data` itself not work?

TomCusack-Softwire (approved):
Nice, good job!

src/components/Tables/Table.cy.tsx
ChloeYiuSoftwire: It makes the toggle check box tests less lengthy

PR #7, opened 2023-07-12
VFB-2 Add A Calendar Component

Var1377:
A semi-interactive calendar component.

 - [x] Make sure you've verified it works via npm run dev
 - [x] Make sure you've linted via npm run lint
 - [x] Make sure you've tested via npm run test


ChloeYiuSoftwire:
In day and week view, my browser in my wide monitor let me scroll down to very bottom and there is a lot of unnecessary empty space below 11pm
In all day, week and month view, I think Google Calendar doesn't even let user scroll down and they just fit the whole calendar in one screen - which might be a better practice as I have to scroll down for the calendar when I am using a large screen

paopaosoftwire:
Something to note is Google Calendar has an Agenda and Print button. Perhaps, we would still want to keep these functionalities? However, I can see that these were not specified in the ticket so perhaps it should be in a separate ticket.

src/app/calendar/clientPage.tsx
```diff
@@ -0,0 +1,41 @@
...
+            start: new Date("2023-07-13T12:30:00"),
+            textColor: "pink",
+        },
+    ];
+
+    return <StyledCalendar initialEvents={sampleEvents} editable={true} />;
+};
+
+const StyledCalendar = styled(Calendar)`
+    height: calc(100vh - 120px);
```
DanWilkins2107, line 38: Could the calendar height be specified in the actual calendar component? Not sure it is necessary tbf but just a thought.

src/app/calendar/page.tsx
```diff
@@ -1,12 +1,18 @@
 import { Metadata } from "next";
+import SampleCalendar from "./clientPage";
```
DanWilkins2107, line 2: Also I thought for a sec that this page was the client page and that code for the calendar was on the client page, maybe worth changing the name of the file?

DanWilkins2107 (approved):

src/app/calendar/page.tsx
```diff
@@ -1,12 +1,18 @@
...
 export const metadata: Metadata = {
     title: "Calendar",
 };
 
-export default Calendar;
+const CalendarPage: React.FC<{}> = () => {
+    return (
+        <>
+            <h1>Calendar Page</h1>
+            <SampleCalendar />;
```
daniel-conlan-softwire, line 13: There's an extra semi-colon on this line that's getting rendered at the bottom of the calendar

src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,116 @@
...
+                <ModalInner>
+                    <h2>View Event</h2>
+                    <p>Event Title: {eventClick?.title}</p>
+                    <p>
+                        Start:
+                        {new Date(eventClick?.startStr!).toLocaleString("en-GB", dateFormatOptions)}
+                    </p>
+                    <p>
+                        End:
+                        {new Date(eventClick?.endStr!).toLocaleString("en-GB", dateFormatOptions)}
```
daniel-conlan-softwire, line 93: Seems that the "End: <date>" on events lasting a single day doesn't display properly. Each `End:` section on an event in the SampleCalendar just renders `End: Invalid Date`

src/app/calendar/clientPage.tsx
```diff
@@ -0,0 +1,41 @@
...
+            startRecur: "2023-07-12",
+            endRecur: "2023-08-25",
+            borderColor: "yellow",
+        },
+        {
+            id: "c",
+            title: "event3",
+            start: new Date("2023-07-13T12:30:00"),
+            textColor: "pink",
+        },
```
daniel-conlan-softwire, line 31: The text colour doesn't actually change on this event - it's still black. Also, it seems to be rendered different than the other events. Maybe because it doesn't have a specified end date?

Edit: Just realised it's because some events are "All Day" events, and others are not

src/app/calendar/clientPage.tsx
```diff
@@ -0,0 +1,41 @@
+"use client";
+
+import React from "react";
+import Calendar, { CalendarEvent as CalendarEvent } from "../../components/Calendar/Calendar";
+import { styled } from "styled-components";
+
+const SampleCalendar: React.FC<{}> = () => {
+    const sampleEvents: CalendarEvent[] = [
```
daniel-conlan-softwire, line 8: Are "All Day" events only implicit upon omitting the times from the `start` and `end` props? I know this is probably how the library works, but it would be nice if all day events were declared explicitly by a boolean `allDay` prop.

daniel-conlan-softwire (changes requested):
Just a couple of small things that might need addressed before merging. But looks like a good start for the component

TomCusack-Softwire:
minor point (will do a full review when Daniel's comments are resolved) - can the PR title be in imperative case and have a verb? It means the Git history is consistent and easy-to-read - eg `VFB-123 Add props to component`
(imperative meaning it fits in the sentence 'this commit will _')

src/app/calendar/clientPage.tsx
SamYap-Softwire: We want the calendar component to be general in styling to not limit the ways it can be used. (For example it might fill up the entire page in one use case, but only be in a small section in another)

src/components/Calendar/Calendar.tsx
SamYap-Softwire: Okay, we made start and end props to be required when adding an event. This shouldn't be a problem anymore.

src/app/calendar/clientPage.tsx
SamYap-Softwire: yes, but we have now changed end date to be required.

src/app/calendar/clientPage.tsx
SamYap-Softwire: we have added an optional allDay props.

SamYap-Softwire: if allDay is set to True but the start or end has time, the time will be ignored.

src/app/calendar/clientPage.tsx
SamYap-Softwire: the text colour and border colour are also there from what we can see. Can you try again to see if it works? (it might not be very easy to see).

daniel-conlan-softwire (approved):

TomCusack-Softwire (changes requested):

src/app/calendar/samplePage.tsx
```diff
@@ -0,0 +1,59 @@
...
+            textColor: "black",
+            allDay: true,
+        },
+        {
+            id: "b",
+            title: "event2",
+            start: new Date("2023-07-12"),
+            end: new Date("2023-07-13"),
+            daysOfWeek: [3],
+            startRecur: "2023-07-12",
```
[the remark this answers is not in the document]
SamYap-Softwire: the recur field is only used when there is a "daysofweek" field which allows events to be repeated. For example in this case, the event happens on every Wednesday. But the first occurrence of the event is on 12 July, and not any earlier.

src/app/calendar/samplePage.tsx
```diff
@@ -0,0 +1,59 @@
...
+`;
+
+const SampleCalendar: React.FC<{}> = () => {
+    const sampleEvents: CalendarEvent[] = [
+        {
+            id: "a",
+            title: "event1",
+            start: new Date("2023-07-11"), // default Date constructor assumes UTC by default
+            end: new Date("2023-07-12"),
+            backgroundColor: "red",
```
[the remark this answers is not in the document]
SamYap-Softwire: This is just a sample page so that anyone who's using the component can have something to refer to, so I don't think we need to have them as themed colours. We could, however, add a function so that events belonging to a particular group will have a particular colour. Right now we are leaving that to whoever is using the component. Do you think that is something we should implement in the component instead? @TomCusack-Softwire

src/app/calendar/samplePage.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { styled } from "styled-components";
+
+const Centerer = styled.div`
+    display: flex;
+    justify-content: center;
+    align-items: center;
+`;
+const SizedAspectRatio = styled.div`
+    width: 100vmin;
+    height: 100vmin;
```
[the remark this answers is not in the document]
SamYap-Softwire: aspect-ratio: 1 would work for most dimensions but not in cases like our wide monitor screens, where the width is much greater than the height, and so you would have to scroll to see the entire calendar (e.g. the month view). 
We take into account these cases (where the width and height differ greatly) by letting both width and height be the minimum of the two. 
 
side note: the samplePage is not a part of the component.

src/app/calendar/samplePage.tsx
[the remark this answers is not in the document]
SamYap-Softwire: It will be deleted when we actually implement the calendar. It is currently there just to make testing easier and for others to reference it when making the Calendar page.

src/app/themes.tsx
```diff
@@ -34,7 +35,7 @@ const StyleManager: React.FC<Props> = ({ children }) => {
     }
 
     return (
-        <StyleSheetManager sheet={serverStyleSheet.instance}>
+        <StyleSheetManager sheet={serverStyleSheet.instance} shouldForwardProp={isPropValid}>
```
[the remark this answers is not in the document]
SamYap-Softwire: yes. It is a styled-component related bug fix.

src/app/calendar/samplePage.tsx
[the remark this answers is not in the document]
TomCusack-Softwire: where is startRecur / endRecur actually used, though? I didn't see anything for it

(no, the names are good enough for there not to be comments!)

src/app/calendar/samplePage.tsx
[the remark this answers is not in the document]
TomCusack-Softwire: ah ok this will be deleted since it's the sample page, this is fair!
it's more that the 'usual' way to do this is to fit a div to the available space and make the calendar 'fit' within it, rather than setting the height based on the viewport

src/components/Calendar/Calendar.cy.tsx
```diff
@@ -0,0 +1,86 @@
+import React from "react";
+import Calendar, { CalendarEvent } from "./Calendar";
+
+describe("<CalendarComponent />", () => {
+    let today: Date;
+    let currentMonth: string;
+    let currentYear: number;
+
+    beforeEach(() => {
```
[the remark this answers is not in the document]
SamYap-Softwire: In some of the individual tests we are setting the "today" variable (e.g. to a previous month or previous day). This won't be possible if we make it a const.
The beforeEach is there to reset "today" to the actual current date. 

Another option would be to define the variables (with const) in the specific tests that we need them in. Would that be preffered? @TomCusack-Softwire

src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,133 @@
...
+    daysOfWeek?: number[];
+    startRecur?: string;
+    endRecur?: string;
+    backgroundColor?: string;
+    borderColor?: string;
+    textColor?: string;
+    allDay?: boolean;
+}
+
+const StyledDialog = styled(Dialog)`
```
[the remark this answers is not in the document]
Var1377: Yep probably, that can be unified in another ticket as ours doesn't look exquisite either.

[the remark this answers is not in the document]
SamYap-Softwire: I think they are also being used in data-view component. But they are done with react-modal instead of dialog. I think we should make it a separate component (using either dialog or react-modal for consistency). @ChloeYiuSoftwire @paopaosoftwire 

I think dialog would be better since mui is already used in the icons component.

src/app/calendar/page.tsx
```diff
@@ -1,16 +1,18 @@
 import { Metadata } from "next";
+import SampleCalendar from "./samplePage";
 import React from "react";
 
-const Calendar: React.FC<{}> = () => {
+export const metadata: Metadata = {
+    title: "Calendar",
```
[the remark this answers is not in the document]
SamYap-Softwire: I think we were doing stuff and it just happened. Not intended but shouldn't matter.

src/app/calendar/samplePage.tsx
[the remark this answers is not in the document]
SamYap-Softwire: It's used by full calendar (as an Event prop)

package.json
```diff
@@ -62,6 +69,7 @@
     "nyc": "^15.1.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "react-modal": "^3.16.1",
```
TomCusack-Softwire, line 72: are you using react-modal any more?

src/app/calendar/page.tsx
```diff
@@ -1,16 +1,18 @@
...
+};
+
+const CalendarPage: React.FC<{}> = () => {
     return (
         <main>
-            <h1> Calendar Page </h1>
+            <h1>Calendar Page</h1>
+            <SampleCalendar />
```
TomCusack-Softwire, line 13: Can you not change the Calendar page? if it's just a sample then when we make the calendar page then we will make it 'correct' - I'm happy with a sample page as a 'reference' for now but would prefer it to just be available as a component, not actually in use on the live (test) website!

src/app/calendar/samplePage.tsx
```diff
@@ -0,0 +1,59 @@
...
+`;
+
+const SampleCalendar: React.FC<{}> = () => {
+    const sampleEvents: CalendarEvent[] = [
+        {
+            id: "a",
+            title: "event1",
+            start: new Date("2023-07-11"), // unless a timezone is specified, the ISO8601 standard assumes UTC (and therefore so does the date constructer). the calendar will then display the date in the local timezone,
+            end: new Date("2023-07-12"), // which is why the date displayed is 1 hour later than the date specified when in BST. When printing to the console, a JS Date object will also be displayed in the local timezone but is internally stored as UTC.
```
TomCusack-Softwire, line 23: can these comments be a separate multi-line comment not in the JSON list? this is great context but dislike it being 'attached' to different lines

also, does the `getHours` function return the UTC hour, if it's internally stored as it, or the locale hour? what about the day (if we are within an hour of midnight / in a different timezone)?

src/app/calendar/samplePage.tsx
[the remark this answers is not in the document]
TomCusack-Softwire: do we need recurring events?
I think we will just be reading in times from the database and plotting them on the calendar, rather than scheduling recurring meetings etc

src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,173 @@
...
+        year: "numeric",
+        month: "long",
+        day: "numeric",
+        hour: "numeric",
+        minute: "numeric",
+    };
+
+    return (
+        <>
+            <StyledDialog open={eventClick !== null} onClose={() => setEventClick(null)}>
```
TomCusack-Softwire, line 132: can we split the `StyledDialog` into a separate file? this is quite long and seems like two separate parts - the 'Calendar' and the 'Modal that opens', so should be in two separate files (but can both be in `components/Calendar`!)
This may also be the file that is unified later!

TomCusack-Softwire: it doesn't have to be a split with a lot of thought, especially since this is likely to be rewritten soon, but I think it's likely we'll use this code as a reference (so should be 'pretty clean' anyway!)

src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+const Calendar: React.FC<CalendarProps> = ({
+    initialEvents,
+    view = "dayGridMonth",
+    editable = false,
+    handleDateClick,
+}) => {
+    const [eventClick, setEventClick] = useState<CalendarEvent | null>(null);
+
+    function handleEventClick(info: EventClickArg): void {
```
TomCusack-Softwire, line 116: arrow functions where possible for consistency reasons, please :)

src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,173 @@
...
+    endRecur?: string;
+    backgroundColor?: string;
+    borderColor?: string;
+    textColor?: string;
+    allDay?: boolean;
+}
+
+const StyledDialog = styled(Dialog)`
+    // root
+    & > div {
```
TomCusack-Softwire, line 36: if there's nothing in container, is this the same as `& > div > div`?

src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,173 @@
...
+        font-size: max(1.5vmin, 15px);
+    }
+
+    .fc .fc-button-primary {
+        background-color: ${(props) => props.theme.fillColor};
+        border-color: ${(props) => props.theme.fillColor};
+        color: black;
+    }
+
+    .fc .fc-button-primary:hover {
```
TomCusack-Softwire, line 94: some of these CSS names are not immediately obvious what they are - can you add a small comment inline about what they are if not absolutely obvious?

src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,173 @@
...
+            <StyledDialog open={eventClick !== null} onClose={() => setEventClick(null)}>
+                <StyledCancelButton type="button" onClick={() => setEventClick(null)}>
+                    close
+                </StyledCancelButton>
+                <ModalInner>
+                    <h2>View Event</h2>
+                    <p>Event Title: {eventClick?.title}</p>
+                    <p>
+                        Start:
+                        {eventClick &&
```
TomCusack-Softwire, line 141: this is probably better implemented as a conditional render - one to check that eventClick is non-null at the top, then we then just use the attributes without checking every time :)

TomCusack-Softwire (changes requested):

src/app/calendar/samplePage.tsx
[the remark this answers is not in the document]
SamYap-Softwire: That's right. We have removed the recurring fields.

src/components/Calendar/Calendar.tsx
TomCusack-Softwire: still unsure on what a 'primary' button is, why you need to separate this out, or if `.fc-button` is 'all visible buttons'

also, can the `:hover` be embedded in `button-primary` with an `&`? thanks :)

TomCusack-Softwire (changes requested):

src/components/Calendar/Modal.tsx
```diff
@@ -0,0 +1,66 @@
...
+    day: "numeric",
+    hour: "numeric",
+    minute: "numeric",
+};
+
+const Modal: React.FC<{
+    eventClick: CalendarEvent | null;
+    setEventClick: (data: CalendarEvent | null) => void;
+}> = ({ eventClick, setEventClick }) => {
+    return (
```
TomCusack-Softwire, line 40: potentially
```
if (eventClick !== null) {
return <></>;
}
return <StyledDialog open={true} etc/>

**Review by TomCusack-Softwire — CHANGES_REQUESTED** [review]
last one (hopefully!), it's 99% there

File: src/components/Calendar/Modal.tsx [hunk 23]
```diff
@@ -0,0 +1,65 @@
...
+
+const Modal: React.FC<{
+    eventClick: CalendarEvent | null;
+    setEventClick: (data: CalendarEvent | null) => void;
+}> = ({ eventClick, setEventClick }) => {
+    if (!eventClick) {
+        return <></>;
+    }
+    return (
+        <StyledDialog open={eventClick !== null} onClose={() => setEventClick(null)}>
```
TomCusack-Softwire, line 44: `eventClick !== null` is always true at this point

src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,102 @@
...
+    .fc .fc-toolbar-chunk {
+        margin: 0 0.5rem;
+    }
+
+    .fc .fc-button {
+        height: max(3vmin, 30px);
+        font-size: max(1.5vmin, 15px);
+        text-align: center;
+        vertical-align: middle;
+        padding: 0 10px 0 10px;
```
TomCusack-Softwire, line 53: is this equivalent to `0 10px;` ?

src/components/Calendar/Calendar.cy.tsx
```diff
@@ -0,0 +1,80 @@
+import React from "react";
+import Calendar, { CalendarEvent } from "./Calendar";
```
TomCusack-Softwire, line 2: @ import please

src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,102 @@
+"use client";
+
+import React, { useState } from "react";
+import FullCalendar from "@fullcalendar/react";
+import timeGridPlugin from "@fullcalendar/timegrid";
+import dayGridPlugin from "@fullcalendar/daygrid";
+import interactionPlugin, { DateClickArg } from "@fullcalendar/interaction";
+import { EventClickArg } from "@fullcalendar/core";
+import styled from "styled-components";
+import Modal from "./Modal";
```
TomCusack-Softwire, line 10: @ import please

TomCusack-Softwire (approved):
Very minor changes, otherwise looks great! Thanks :)

src/components/Calendar/Calendar.tsx
SamYap-Softwire: yes, that's right.

PR #8, opened 2023-07-12
VFB-3 Create modal to read JSON input

ChloeYiuSoftwire:
We have created 3 files as follows:

1. DataViewer: contains the modal itself
2. DataViewerWithButton: contains the example for using the DataViewer component. This is also used in the cypress test so please do not remove this.
3. Cypress test file

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@fullcalendar/react": "^6.1.8",
     "@fullcalendar/timegrid": "^6.1.8",
     "@istanbuljs/nyc-config-typescript": "^1.0.2",
+    "@mui/icons-material": "^5.14.0",
     "@mui/material": "^5.14.0",
     "@supabase/auth-helpers-nextjs": "^0.7.2",
     "@supabase/auth-ui-react": "^0.4.2",
@@ -47,7 +48,6 @@
     "@types/node": "20.3.1",
     "@types/react": "18.2.13",
     "@types/react-dom": "18.2.6",
-    "@types/react-fontawesome": "^1.6.5",
     "@types/styled-components": "^5.1.26",
     "@typescript-eslint/eslint-plugin": "^5.59.11",
     "@typescript-eslint/parser": "^5.59.11",
@@ -74,8 +74,5 @@
     "styled-components": "^6.0.0-rc.3",
     "supabase": "^1.68.6",
     "typescript": "5.0.4"
-  },
-  "devDependencies": {
-    "@types/react-modal": "^3.16.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -69,9 +69,13 @@
     "nyc": "^15.1.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "react-modal": "^3.16.1",
     "start-server-and-test": "^2.0.0",
     "styled-components": "^6.0.0-rc.3",
     "supabase": "^1.68.6",
     "typescript": "5.0.4"
+  },
+  "devDependencies": {
+    "@types/react-modal": "^3.16.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -48,6 +48,8 @@
     "@types/node": "20.3.1",
     "@types/react": "18.2.13",
     "@types/react-dom": "18.2.6",
+    "@types/react-fontawesome": "^1.6.5",
+    "@types/react-modal": "^3.16.0",
     "@types/styled-components": "^5.1.26",
     "@typescript-eslint/eslint-plugin": "^5.59.11",
     "@typescript-eslint/parser": "^5.59.11",
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -2,7 +2,7 @@
 import supabase, { Schema } from "@/supabase";
 import { Metadata } from "next";
 
-import React, { useEffect, useState } from "react";
+import React, { useState } from "react";
 import DataViewer from "@/components/DataViewer/DataViewer";
 
 const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
@@ -12,23 +12,20 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 
 const Clients: () => Promise<React.ReactElement> = async () => {
     const [isOpen, setIsOpen] = useState(false);
-    console.log("rendered");
-    console.log(isOpen);
 
-    useEffect(() => {
+    const openModal: () => void = () => {
         setIsOpen(true);
-        console.log("useEffect");
-    }, []);
+    };
 
     const closeModal: () => void = () => {
         setIsOpen(false);
-        console.log(isOpen);
     };
 
     const data = await dataFetch();
     return (
         <main>
             <h1> Clients Page </h1>
+            <button onClick={openModal}>Open</button>
 
             {/* This should be a separate component which is passed data via props */}
             {/* <pre>{JSON.stringify(data, null, 4)}</pre> */}
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -1,7 +1,9 @@
+"use client";
 import supabase, { Schema } from "@/supabase";
 import { Metadata } from "next";
 
-import React from "react";
+import React, { useEffect, useState } from "react";
+import DataViewer from "@/components/DataViewer/DataViewer";
 
 const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
     const response = await supabase.from("clients").select();
@@ -9,16 +11,36 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    const data = await dataFetch();
+    const [isOpen, setIsOpen] = useState(false);
+    console.log("rendered");
+    console.log(isOpen);
 
+    useEffect(() => {
+        setIsOpen(true);
+        console.log("useEffect");
+    }, []);
+
+    const closeModal: () => void = () => {
+        setIsOpen(false);
+        console.log(isOpen);
+    };
+
+    const data = await dataFetch();
     return (
         <main>
             <h1> Clients Page </h1>
 
-            <p> Testing Supabase fetching </p>
-
             {/* This should be a separate component which is passed data via props */}
-            <pre>{JSON.stringify(data, null, 4)}</pre>
+            {/* <pre>{JSON.stringify(data, null, 4)}</pre> */}
+
+            {data !== null && (
+                <DataViewer
+                    data={data[0]}
+                    title="Client details"
+                    isOpen={isOpen}
+                    onRequestClose={closeModal}
+                />
+            )}
         </main>
     );
 };
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -5,27 +5,44 @@ import { Metadata } from "next";
 import React, { useState } from "react";
 import DataViewer from "@/components/DataViewer/DataViewer";
 
+interface ViewerProps {
+    data: { [key: string]: string | null };
+}
+
 const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
     const response = await supabase.from("clients").select();
     return response.data;
 };
 
-const Clients: () => Promise<React.ReactElement> = async () => {
-    const [isOpen, setIsOpen] = useState(false);
+const Viewer: React.FC<ViewerProps> = ({ data }) => {
+    const [viewerIsOpen, setViewerIsOpen] = useState(false);
 
     const openModal: () => void = () => {
-        setIsOpen(true);
+        setViewerIsOpen(true);
     };
 
     const closeModal: () => void = () => {
-        setIsOpen(false);
+        setViewerIsOpen(false);
     };
 
+    return (
+        <>
+            <button onClick={openModal}>Open</button>
+            <DataViewer
+                data={data}
+                title="Client details"
+                isOpen={viewerIsOpen}
+                onRequestClose={closeModal}
+            />
+        </>
+    );
+};
+
+const Clients: () => Promise<React.ReactElement> = async () => {
     const data = await dataFetch();
     return (
         <main>
             <h1> Clients Page </h1>
-            <button onClick={openModal}>Open</button>
 
             {/* This should be a separate component which is passed data via props */}
             {/* <pre>{JSON.stringify(data, null, 4)}</pre> */}
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -1,8 +1,8 @@
 import supabase, { Schema } from "@/supabase";
 import { Metadata } from "next";
+import SampleDataViewerWithButton from "@/components/DataViewer/SampleDataViewerWithButton";
 
 import React from "react";
-import DataViewerWithButton from "../../components/DataViewer/DataViewerWithButton";
 
 const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
     const response = await supabase.from("clients").select();
@@ -11,21 +11,14 @@ const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
 
 const Clients: () => Promise<React.ReactElement> = async () => {
     const data = await dataFetch();
+
     return (
         <main>
             <h1> Clients Page </h1>
 
             {/* This should be a separate component which is passed data via props */}
-            {/* <pre>{JSON.stringify(data, null, 4)}</pre> */}
-
-            {data !== null && (
-                <DataViewer
-                    data={data[0]}
-                    title="Client details"
-                    isOpen={isOpen}
-                    onRequestClose={closeModal}
-                />
-            )}
+            <pre>{JSON.stringify(data, null, 4)}</pre>
+            <SampleDataViewerWithButton data={data![0]} />
         </main>
     );
 };
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -2,7 +2,7 @@ import supabase, { Schema } from "@/supabase";
 import { Metadata } from "next";
 
 import React from "react";
-import Viewer from "./Viewer";
+import DataViewerWithButton from "../../components/DataViewer/DataViewerWithButton";
 
 const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
     const response = await supabase.from("clients").select();
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -1,6 +1,6 @@
+import SampleDataViewerWithButton from "@/components/DataViewer/SampleDataViewerWithButton";
 import supabase, { Schema } from "@/supabase";
 import { Metadata } from "next";
-import SampleDataViewerWithButton from "@/components/DataViewer/SampleDataViewerWithButton";
 
 import React from "react";
 
@@ -18,7 +18,7 @@ const Clients: () => Promise<React.ReactElement> = async () => {
 
             {/* This should be a separate component which is passed data via props */}
             <pre>{JSON.stringify(data, null, 4)}</pre>
-            <SampleDataViewerWithButton data={data![0]} />
+            <SampleDataViewerWithButton data={data![0]}/>
         </main>
     );
 };
--- a/src/app/clients/page.tsx
+++ b/src/app/clients/page.tsx
@@ -1,41 +1,12 @@
-"use client";
 import supabase, { Schema } from "@/supabase";
 import { Metadata } from "next";
 
-import React, { useState } from "react";
-import DataViewer from "@/components/DataViewer/DataViewer";
-
-interface ViewerProps {
-    data: { [key: string]: string | null };
-}
+import React from "react";
+import Viewer from "./Viewer";
 
 const dataFetch: () => Promise<Schema["clients"][] | null> = async () => {
     const response = await supabase.from("clients").select();
     return response.data;
-};
-
-const Viewer: React.FC<ViewerProps> = ({ data }) => {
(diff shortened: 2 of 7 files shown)


src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    padding: "0",
+                    border: "0",
+                    width: "80%",
+                    height: "80%",
+                    boxShadow: "0px 2px 10px rgb(200, 200, 200)",
+                    top: "50%",
+                    left: "50%",
+                    transform: "translate(-50%, -50%)",
+                },
+            }}
```
daniel-conlan-softwire, line 95: Ask Tom, but you might want to replace this using styled-components by creating a custom `<StyledModal>`  component within this file. This might not be needed, and may not be worth the effort, but it's something to ask.

src/components/DataViewer/DataViewerWithButton.tsx
daniel-conlan-softwire, line 1: If this is only used for testing, then I would just define the component within the test file itself.

If it's not going to be used within the wider application, then I don't think it should be included as a component.

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: yep, wherever possible (and it is possible here!) I don't want `style=` set, and I would like `styled-components` to be used.
Potentially ask @SamYap-Softwire / @Var1377 for help, they both did some styled component stuff with modals!

src/components/DataViewer/DataViewerWithButton.tsx
ChloeYiuSoftwire: Hmmmm it is a sample file to show how is data viewer implemented though because it has to pass in some useState stuff from the file using it but maybe I will make it clearer that it is a sample file?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,104 @@
...
+interface DataViewerProps {
+    data: { [key: string]: string | number | null };
+    title: string;
+    isOpen: boolean;
+    onRequestClose: () => void;
+}
+
+const ModalWithStyle = styled(Modal)`
+    position: absolute;
+    inset: 50% 40px 40px 50%;
```
Var1377, line 16: Try and use max or min on the inset property here to try and make it responsive, hopefully that works

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,104 @@
...
+    justify-content: space-between;
+    background-color: #eeeeee;
+    margin-bottom: 1rem;
+    margin-top: 0;
+    padding: 0.5rem;
+`;
+
+const Content = styled.div`
+    overflow: scroll;
+    height: calc(100% - 5rem);
```
Var1377, line 41: is there a cleaner way to do this? the modal can be smaller if there's less data to show

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,104 @@
...
+    font-size: small;
+`;
+
+const Value = styled.div`
+    font-size: medium;
+`;
+
+const EachItem = styled.div`
+    display: flex;
+    flex-direction: column;
```
Var1377, line 56: it should work without display: flex and flex-direction: column as you aren't using any other flexbox properties

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const CloseButton = styled.button`
+    border: 0;
+    background-color: red;
+    color: white;
+`;
+
+const DataViewer: React.FC<DataViewerProps> = (props) => {
+    const ClearButton: React.FC = () => {
+        return <CloseButton onClick={closeModel}>X</CloseButton>;
```
Var1377, line 68: Not a big deal but now that icons are merged you could use a font-awesome icon

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,104 @@
+"use client";
+
+import React from "react";
+import styled from "styled-components";
+import Modal from "react-modal";
+
+interface DataViewerProps {
+    data: { [key: string]: string | number | null };
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: We have added comment as suggested

src/components/DataViewer/DataViewer.cy.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+        cy.contains("FULL NAME");
+    });
+
+    it("modal can be closed", () => {
+        cy.mount(<DataViewerWithButton data={data} />);
+
+        cy.get("button").click();
+
+        cy.get("button").contains("X").click();
```
DanWilkins2107, line 30: Could add a separate test for modal shows correct values?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,104 @@
...
+            onRequestClose={closeModel}
+            ariaHideApp={false}
+            contentLabel="JSON Modal"
+        >
+            <Header>
+                {props.title}
+                <ClearButton />
+            </Header>
+            <Content>{JSONContent}</Content>
+        </ModalWithStyle>
```
DanWilkins2107, line 101: Not a big deal, and idk what other people think for this, so maybe get a second opinion, but we've been calling our styled version of things Styled_____ and it would be nice for that to be consistent throughout, obvs if everyone else does your way we should change our bits tho :)

DanWilkins2107 (commented):
Good apart from Small Changes

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Agreed

daniel-conlan-softwire (approved):

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Seems like max and min isn't necessary to make it responsive (?) but we removed those px-dependent number and stick to percentage, and it is responsive.

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: The modal now is responsive to the page size and the amount of data to show. We cleaned it with display flex.

cypress/e2e/accessibility.cy.ts
```diff
@@ -0,0 +1,30 @@
+describe("Accessibility tests", () => {
```
TomCusack-Softwire, line 1: looks like the rebase didn't really work out - these changes shouldn't be on your branch!
they should be on main (which would be 'beneath' your branch), so the changes would be present, but not as an addition or as separate commits...
May need an interactive rebase (to squash some commits or delete some commits) to fix this up!

src/components/DataViewer/DataViewer.tsx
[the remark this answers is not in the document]
TomCusack-Softwire: I don't think this is necessarily needed - I wouldn't expect a nested JSON to be rendered in a single-level table at all, and the type errors 'are' the sign that it won't work! I think specifying that data is a key-value (where value is string/num/null) should be sufficient.

(Especially since we are not exporting this component or making others use it, there won't be 'downstream' users using it for weird edge cases - the component is made in a way which is intuitive and easy-to-use, and it's on the caller to format their data to fit what the table is expecting!)

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: if you do then please make a new component for it! (this is a purely optional change, I think the X is fine!)

ChloeYiuSoftwire: We will deal with this later.

src/components/DataViewer/DataViewer.tsx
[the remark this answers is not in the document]
ChloeYiuSoftwire: Ok. We have removed the comment then

src/components/DataViewer/DataViewer.cy.tsx
ChloeYiuSoftwire: We have added the test as suggested in the newer version

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: We have fixed it as suggested

cypress/e2e/accessibility.cy.ts
ChloeYiuSoftwire: resolved as suggested

package.json
```diff
@@ -62,9 +66,13 @@
...
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "react-modal": "^3.16.1",
     "start-server-and-test": "^2.0.0",
     "styled-components": "^6.0.0-rc.3",
     "supabase": "^1.68.6",
     "typescript": "5.0.4"
+  },
+  "devDependencies": {
+    "@types/react-modal": "^3.16.0"
```
TomCusack-Softwire, line 76: can this be all in dependencies (rather than dev)? we're not splitting dependencies for the time being :)

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    if (!props.isOpen) {
+        return <></>;
+    }
+
+    return (
+        <StyledModal
+            isOpen={true}
+            onRequestClose={closeModal}
+            ariaHideApp={false}
```
TomCusack-Softwire, line 99: what's `ariaHideApp` do?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,111 @@
...
+            contentLabel="JSON Modal"
```
TomCusack-Softwire, line 100: Where does this get shown? Can it be in less 'tech' language - e.g. 'data viewer'?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+const CloseButton = styled.button`
+    border: 0;
+    background-color: red;
+    color: white;
+    display: flex;
+    justify-content: center;
+`;
+
+const ClearButton: React.FC<{ closeModal: () => void }> = (props) => {
```
TomCusack-Softwire, line 69: can `<{ closeModal: () => void }>` be extracted into a `ClearButtonProps` please? :)

TomCusack-Softwire (changes requested):

package.json
paopaosoftwire: I have moved it

src/components/DataViewer/DataViewer.tsx
paopaosoftwire: "It is important for users of screenreaders that other page content be hidden (via the aria-hidden attribute) while the modal is open." 
https://reactcommunity.org/react-modal/accessibility/

I think it doesn't hurt to include the line in the code.

src/components/DataViewer/DataViewer.tsx
paopaosoftwire: contentLabel is used for accessibility reasons. Yes, I agree that the text should be changed and it has now been changed.

src/components/DataViewer/SampleDataViewerWithButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+const Title = styled.span`
+    margin-left: 0.5rem;
+`;
+
+const SampleDataViewerWithButton: React.FC<ViewerProps> = ({ data }) => {
+    const [viewerIsOpen, setViewerIsOpen] = useState(false);
+
+    const openModal: () => void = () => {
+        setViewerIsOpen(true);
+        console.log("openModal", viewerIsOpen);
```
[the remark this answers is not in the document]
paopaosoftwire: Removed!

src/components/DataViewer/DataViewer.tsx
paopaosoftwire: Done!

src/app/clients/page.tsx
```diff
@@ -15,10 +16,9 @@ const Clients: () => Promise<React.ReactElement> = async () => {
         <main>
             <h1> Clients Page </h1>
 
-            <p> Testing Supabase fetching </p>
-
             {/* This should be a separate component which is passed data via props */}
             <pre>{JSON.stringify(data, null, 4)}</pre>
+            <SampleDataViewerWithButton data={data![0]} />
```
TomCusack-Softwire, line 21: Can you remove this line? It shouldn't be on the final page (yet) - we're just making the component!

src/components/DataViewer/DataViewer.cy.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+        cy.get("button").click();
+
+        cy.contains(
+            "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz"
+        );
+        cy.contains(
+            "John With A Very Very Very Very Very Very Very Very Very Very Very Very Very Long Name"
+        );
+        cy.contains("1234567");
```
TomCusack-Softwire, line 46: should cy.contains an empty string for dietary requirements be a test? or even that dietary requirements is shown on the page too?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,100 @@
...
+    closeModal: () => void;
+}
+const ClearButton: React.FC<ClearButtonProps> = (props) => {
+    return (
+        <CloseButton onClick={props.closeModal}>
+            <CloseIcon />
+        </CloseButton>
+    );
+};
+const JSONContent: React.FC<{ [key: string]: string | number | null }> = (data) => {
```
TomCusack-Softwire, line 68: can you extract this to an interface?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,100 @@
...
+    return Object.entries(data).map(([key, value]) => {
+        return (
+            <EachItem key={key}>
+                <Key>{key.toUpperCase().replace("_", " ")}</Key>
+                <Value>{value ?? ""}</Value>
+            </EachItem>
+        );
+    });
+};
+const DataViewer: React.FC<DataViewerProps> = (props) => {
```
TomCusack-Softwire, line 78: probably also better to have DataViewerProps to be defined right above this, so you don't have to scroll too far to see what props should have in it!

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: I think it's great, I just didn't know what it did!
On that page, it looks like you should give it the root element rather than 'false':

> To allow react-modal to do this, you should call Modal.setAppElement with a query selector identifying the root of your app

```
Modal.setAppElement('#root');
```

> If you are already applying the aria-hidden attribute to your app content through other means, you can pass the ariaHideApp={false} prop to your modal to avoid getting a warning that your app element is not specified.

so I think with this you are actually turning off the warning?

src/components/DataViewer/SampleDataViewerWithButton.tsx
```diff
@@ -0,0 +1,49 @@
+"use client";
+import React, { useState } from "react";
+import DataViewer from "@/components/DataViewer/DataViewer";
+// import PhoneIcon from "../Icons/PhoneIcon";
```
TomCusack-Softwire, line 4: remove the comments please!

src/components/DataViewer/SampleDataViewerWithButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+        setViewerIsOpen(true);
+    };
+
+    const closeModal: () => void = () => {
+        setViewerIsOpen(false);
+    };
+
+    return (
+        <>
+            <button onClick={openModal}>Open</button>
```
TomCusack-Softwire, line 33: might be worth using `Button` (the styled one) from the shared components folder

TomCusack-Softwire (changes requested):

src/components/DataViewer/DataViewer.cy.tsx
paopaosoftwire: An empty string cannot be passed into cy.contains so I think a better test would be just to test if cy.contains("DIETARY REQUIREMENTS") which should be the header for that section.

src/components/DataViewer/DataViewer.tsx
paopaosoftwire: Done!

src/components/DataViewer/DataViewer.tsx
paopaosoftwire: Done!

src/components/DataViewer/SampleDataViewerWithButton.tsx
```diff
@@ -0,0 +1,49 @@
+"use client";
+import React, { useState } from "react";
+import DataViewer from "@/components/DataViewer/DataViewer";
+// import PhoneIcon from "../Icons/PhoneIcon";
+import styled from "styled-components";
+
+interface ViewerProps {
```
SamYap-Softwire, line 7: the type of data is also used in the dataviewer file. Maybe you can export the interface?

src/components/DataViewer/SampleDataViewerWithButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+    };
+
+    return (
+        <>
+            <button onClick={openModal}>Open</button>
+            <DataViewer
+                data={data}
+                header={
+                    <Header>
+                        {/* <PhoneIcon /> */}
```
SamYap-Softwire, line 38: I'm not sure what this comment is referring to

SamYap-Softwire (changes requested):

src/components/DataViewer/DataViewer.tsx
paopaosoftwire: You are completely correct. We have now added the setAppElement and set ariaHideApp to true. Thanks.

src/components/DataViewer/SampleDataViewerWithButton.tsx
paopaosoftwire: The current button takes a href which is not what we want as we need to use the onClick. I think we could change it once theres a better button implemented in a separate ticket.

src/components/DataViewer/DataViewer.cy.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import DataViewerWithButton from "./SampleDataViewerWithButton";
```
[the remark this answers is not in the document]
paopaosoftwire: DONE

src/components/DataViewer/SampleDataViewerWithButton.tsx
paopaosoftwire: DONE

src/components/DataViewer/SampleDataViewerWithButton.tsx
paopaosoftwire: It wasn't supposed to be commented out. Thanks!

src/components/DataViewer/DataViewer.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+interface DataViewerProps {
+    data: Data;
+    header: ReactElement | string;
+    isOpen: boolean;
+    onRequestClose: () => void;
+}
+
+const DataViewer: React.FC<DataViewerProps> = (props) => {
+    const closeModal: () => void = () => {
```
TomCusack-Softwire, line 86: can this have type `closeModel = (): void => {}`? typing the return (and parameters) individually rather than typing the entire function

src/components/DataViewer/SampleDataViewerWithButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+`;
+
+const Title = styled.span`
+    margin-left: 0.5rem;
+`;
+
+const SampleDataViewerWithButton: React.FC<ViewerProps> = ({ data }) => {
+    const [viewerIsOpen, setViewerIsOpen] = useState(false);
+
+    const openModal: () => void = () => {
```
TomCusack-Softwire, line 23: same here

src/components/DataViewer/SampleDataViewerWithButton.tsx
```diff
@@ -0,0 +1,49 @@
+"use client";
+import React, { useState } from "react";
+import DataViewer, { Data } from "@/components/DataViewer/DataViewer";
+import styled from "styled-components";
+import PhoneIcon from "../Icons/PhoneIcon";
```
TomCusack-Softwire, line 5: @/ import please :)

TomCusack-Softwire (approved):
Very small changes, but otherwise I'm happy.
Please do wait for @SamYap-Softwire to also approve before merging! 🙂

package.json
```diff
@@ -26,10 +26,14 @@
   },
   "dependencies": {
     "@cypress/code-coverage": "^3.10.7",
+    "@emotion/react": "^11.11.1",
```
SamYap-Softwire, line 29: image

We want the bottom install instead of the top for styled components

SamYap-Softwire (approved):
I think it's good.

package.json
paopaosoftwire: Looking at other branches which has MUI installed, such as navbar, @mui/styled-engine-sc seems to not be used at all. And when installing it via npm there are dependency errors. uninstalling everything and keeping only @mui/material and styled-components seems to work fine so it will be kept like that.

PR #10, opened 2023-07-13
Fix navigation in the navbar

Var1377:
This flag should be enabled by default as far as I can tell although setting it appears to fix the issue. Using non experimental-flag workarounds don't appear to work without completely opting out of client-side routing. Can others also test if this actually fixes things?

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -4,6 +4,9 @@ const nextConfig = {
     compiler: {
         styledComponents: true,
     },
+    experimental: {
+        newNextLinkBehavior: true,
+    }
 };
 
 module.exports = nextConfig;


SamYap-Softwire (approved):

TomCusack-Softwire (approved):
> Can others also test if this actually fixes things?

Can you get someone else to test this before it is merged? I'm happy with it otherwise :smile:

daniel-conlan-softwire:
The navigation between pages seems to be working, but if you go to an unknown URL (_e.g._ `/not_mapped`) and then go back to any of the other pages, the styling for the navbar disappears

SamYap-Softwire:
Navigation sometimes still doesn't work

PR #12, opened 2023-07-17
VFB-16 Add to themes

SamYap-Softwire:
Dark and light mode.


src/app/themes.tsx
```diff
@@ -5,13 +5,31 @@ import { useServerInsertedHTML } from "next/navigation";
...
+    accentColor: "#3889cd",
+    surfaceForegroundColor: "#000",
+    surfaceBackgroundColor: "#b3dbd1",
+    errorColor: "#ff624e"
 };
 
+
+// tertiaryColor
+// disabledColor
+// pressedColor
```
daniel-conlan-softwire: Are these comments supposed to be here?

src/app/themes.tsx
```diff
@@ -35,7 +53,7 @@ const StyleManager: React.FC<Props> = ({ children }) => {
 
     return (
         <StyleSheetManager sheet={serverStyleSheet.instance}>
-            <ThemeProvider theme={currentTheme}>
+            <ThemeProvider theme={darkTheme}>
```
daniel-conlan-softwire: Default dark theme...nice

src/components/LoginPanel.tsx
```diff
@@ -29,11 +29,11 @@ const LoginPanel: React.FC<{}> = () => {
...
-                                inputLabelText: currentTheme.foregroundColor,
-                                anchorTextColor: currentTheme.foregroundColor,
-                                brand: currentTheme.fillColor,
-                                brandAccent: currentTheme.fillColor,
-                                brandButtonText: currentTheme.foregroundColor,
+                                inputLabelText: lightTheme.foregroundColor,
+                                anchorTextColor: lightTheme.foregroundColor,
+                                brand: lightTheme.surfaceBackgroundColor,
+                                brandAccent: lightTheme.surfaceBackgroundColor,
+                                brandButtonText: lightTheme.foregroundColor,
```
daniel-conlan-softwire: This comment applies to everywhere that explicitly uses either `darkTheme` or `lightTheme`, but would it be best to have a global `currentTheme` variable that can then be toggled once instead of having to change the theme everywhere manually?

daniel-conlan-softwire: Actually...nevermind.  Just seen where you change the theme in `ThemeProvider`. But still, is there a reason why the `LoginPanel` doesn't pull from the `ThemeProvider`?

src/app/themes.tsx
```diff
@@ -6,11 +6,32 @@ import { useServerInsertedHTML } from "next/navigation";
...
+    backgroundColor: "#eeeeee",
+    primaryForegroundColor: "#000000",
+    primaryColor: "#63a036",
+    secondaryForegroundColor: "#ffffff",
+    secondaryColor: "#066940",
+    accentColor: "#609fd3",
+    accentForegroundColor: "#ffffff",
+    surfaceForegroundColor: "#000000",
+    surfaceColor: "#ffffff",
+    errorColor: "#ff624e",
```
ChloeYiuSoftwire, line 20: Better be consistent with the notation of "foreground" vs "background" color
Also maybe arrange them in terms of importance? e.g. primary should be above secondary

src/app/themes.tsx
```diff
@@ -6,11 +6,32 @@ import { useServerInsertedHTML } from "next/navigation";
...
+    foregroundColor: "#ffffff",
+    backgroundColor: "#1a1a1a",
+    secondaryForegroundColor: "#000000",
+    secondaryColor: "#63a036",
+    primaryForegroundColor: "#ffffff",
+    primaryColor: "#066940",
+    surfaceForegroundColor: "#ffffff",
+    surfaceColor: "#161414",
+    accentColor: "#1b629c",
+    accentForegroundColor: "#ffffff",
```
ChloeYiuSoftwire, line 34: Same as light theme - orders and renaming

src/components/Buttons/ButtonPost.tsx
```diff
@@ -2,11 +2,11 @@ import React from "react";
 import styled from "styled-components";
 
 const Submit = styled.input.attrs({ type: "submit" })`
-    background-color: ${(props) => props.theme.fillColor};
+    background-color: ${(props) => props.theme.secondaryColor};
```
ChloeYiuSoftwire, line 5: This resonate with my comment above, esp when it is put in "background-color"

ChloeYiuSoftwire (changes requested):

paopaosoftwire (commented):
Colours look great!

src/app/themes.tsx
```diff
@@ -30,13 +51,15 @@ const StyleManager: React.FC<Props> = ({ children }) => {
...
     if (typeof window !== "undefined") {
-        return <>{children}</>;
+        return <ThemeProvider theme={chosenTheme}>{children}</ThemeProvider>;
     }
 
     return (
         <StyleSheetManager sheet={serverStyleSheet.instance} shouldForwardProp={isPropValid}>
-            <ThemeProvider theme={currentTheme}>
+            <ThemeProvider theme={chosenTheme}>
                 <>
```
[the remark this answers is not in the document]
SamYap-Softwire: Unfortunately we do

daniel-conlan-softwire (commented):
All looks good to me. Only thing is that the theming will have to be applied to the new navigation bar once it's merged

paopaosoftwire:
> 

But Chloe thinks it could be a little lighter.

src/app/global_styles.ts
```diff
@@ -8,6 +8,12 @@ const GlobalStyle = createGlobalStyle`
     html, body {
         height: 100%;
         width: 100%;
+        font-family: Arial, Helvetica, sans-serif;
```
paopaosoftwire, line 11: Goodbye comic-sans and times new roman

paopaosoftwire (approved):
It looks fine. Please dont forget about the toggling test.

src/app/global_styles.ts
SamYap-Softwire: 😢

cypress/e2e/themes.cy.ts
ChloeYiuSoftwire, line 1: nice test

ChloeYiuSoftwire (approved):

src/app/themes.tsx
```diff
@@ -3,25 +3,65 @@
...
+    errorColor: "#ff624e",
 };
 
 interface Props {
     children: React.ReactElement;
+    theme?: CustomTheme;
 }
 
+export const ThemeUpdateContext = createContext((theme: CustomTheme) =>
+    console.error(`attempted to set theme outside of a ThemeUpdateContext.Provider: ${theme}`)
```
TomCusack-Softwire, line 57: Is it worth this throwing an error instead of just `console.error`'ing? It's very easy to just not look at `console.log` :)

src/app/themes.tsx
```diff
@@ -30,19 +70,31 @@ const StyleManager: React.FC<Props> = ({ children }) => {
         return <>{styles}</>;
     });
 
+    const [chosenTheme, setChosenTheme] = useState(theme);
+
     if (typeof window !== "undefined") {
-        return <>{children}</>;
+        return (
```
TomCusack-Softwire, line 76: can these returns be combined? there's a lot of duplicate code between them. Maybe something like:
```tsx
const childrenWithStyleManager = (typeof window !== "undefined") ? children : (
    <StyleSheetManager ...> {children} </...>
);
return // just one return using above variable
```

src/app/themes.tsx
```diff
@@ -30,19 +70,31 @@ const StyleManager: React.FC<Props> = ({ children }) => {
...
 
     return (
-        <StyleSheetManager sheet={serverStyleSheet.instance} shouldForwardProp={isPropValid}>
-            <ThemeProvider theme={currentTheme}>
-                <>
-                    <GlobalStyle />
+        <ThemeUpdateContext.Provider value={setChosenTheme}>
+            <ThemeProvider theme={chosenTheme}>
+                <GlobalStyle />
+                <StyleSheetManager
```
TomCusack-Softwire, line 90: StyleSheetManager has changed position - should it be above ThemeProvider?
(no idea which way around it should be, just you've changed it - can you show that your way is the correct way?)

src/app/themes.tsx
```diff
@@ -3,25 +3,65 @@
...
+    secondaryForegroundColor: string;
+    secondaryBackgroundColor: string;
+    accentBackgroundColor: string;
+    accentForegroundColor: string;
+    surfaceForegroundColor: string;
+    surfaceBackgroundColor: string;
+    errorColor: string;
+};
+
+export const lightTheme: CustomTheme = {
```
TomCusack-Softwire, line 23: not sure why all of these are exported - are they used anywhere else?

src/app/themes.tsx
```diff
@@ -3,25 +3,65 @@
...
+    surfaceBackgroundColor: "#161414",
+    errorColor: "#ff624e",
 };
 
 interface Props {
     children: React.ReactElement;
+    theme?: CustomTheme;
 }
 
+export const ThemeUpdateContext = createContext((theme: CustomTheme) =>
```
TomCusack-Softwire, line 56: why do you need ThemeUpdateContext? Themes were working before, just there was only one of them (light mode).
I understand using `useState` to store which theme is used, but is there no way to pass the setter down to certain children?

src/components/LoginPanel.tsx
```diff
@@ -1,23 +1,27 @@
...
 const MiddleDiv = styled.div`
     max-width: 400px;
-    border: solid 1px black;
+    border: solid 1px ${(props) => props.theme.foregroundColor};
     border-radius: 10px;
     padding: 50px;
     margin: auto;
+
+    & button:hover {
+        color: ${(props) => props.theme.secondaryForegroundColor};
```
TomCusack-Softwire, line 18: this isn't needed - use the `defaultButtonBackgroundHover` attribute in the `colors` object
(https://github.com/supabase/auth-ui/blob/main/packages/shared/src/theming/Themes.ts)

TomCusack-Softwire (changes requested):

src/app/global_styles.ts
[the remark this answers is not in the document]
Var1377: Yeah, for consistency reasons we thought the font should probably stay constant across themes.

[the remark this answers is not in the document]
Var1377: Honestly not a lot of thought went into it, they're just fonts most people have installed and are a bit nicer than default.

src/app/themes.tsx
```diff
@@ -3,25 +3,65 @@
...
+
+export const lightTheme: CustomTheme = {
+    foregroundColor: "#000000",
+    backgroundColor: "#eeeeee",
+    primaryForegroundColor: "#000000",
+    primaryBackgroundColor: "#63a036",
+    secondaryForegroundColor: "#ffffff",
+    secondaryBackgroundColor: "#066940",
+    accentBackgroundColor: "#609fd3",
+    accentForegroundColor: "#ffffff",
```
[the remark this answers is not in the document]
Var1377: It should be, but I thought it would be better to be explicit and keep to the pattern of `*backgroundColor` and `*foregroundColor`.

src/app/themes.tsx
Var1377: Theme switching only seemed to work in this configuration, not sure why though.

src/app/themes.tsx
Var1377: fixed

src/app/themes.tsx
Var1377: Using context is the easiest and cleanest solution to pass `setState` down the component tree to the navbar. Passing them through props would result in worse code that's just harder to read.

src/components/LoginPanel.tsx
Var1377: This is a `color` and not a `background-color` so this is still required to pass accessibility checks.

TomCusack-Softwire: whyyyy isn't there a `defaultButtonHoverColor`...

src/components/NavBar.tsx
```diff
@@ -42,6 +47,24 @@ const NavBar: React.FC<{}> = () => {
                 <Gap />
 
                 <ButtonPost text="Sign out" url="/auth/signout" />
+                <button
```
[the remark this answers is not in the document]
TomCusack-Softwire: also means we can CTRL-F through code for `// TODO` to see what needs fixing urgently :)

TomCusack-Softwire (approved):
After minor changes, looks good to me!

PR #20, opened 2023-07-19
VFB-3 Data View Component (Bug Fix)

daniel-conlan-softwire:
Fixed JSONContent component to be an actual JSX component instead of a function.


ChloeYiuSoftwire (approved):
Good

Var1377:
Could you also implement fixes for the problems that @DanWilkins2107 and I detailed in the JIRA comments for the original DataViewer PR?

ChloeYiuSoftwire:
> Could you also implement fixes for the problems that @DanWilkins2107 and I detailed in the JIRA comments for the original DataViewer PR?

Done as suggested
For your information, we deleted the id mentioned in JIRA because somehow it still passed the accessibility test without it existing (it was included at first for the sake of accessibility test)

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
     justify-content: space-between;
     align-items: center;
-    background-color: #eeeeee;
-    padding: 0.5rem;
-    margin-top: 0;
+    color: ${(props) => props.theme.secondaryForegroundColor};
+    background-color: ${(props) => props.theme.secondaryBackgroundColor}A0;
+    font-size: 1.5em;
+    font-weight: bolder;
+    padding: 0.75em 1em 0.75em 1em;
```
SamYap-Softwire, line 39: I think this is the same as `padding: 0.75em 1em` ?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
+
+const Header = styled.div`
     display: flex;
     justify-content: space-between;
     align-items: center;
-    background-color: #eeeeee;
-    padding: 0.5rem;
-    margin-top: 0;
+    color: ${(props) => props.theme.secondaryForegroundColor};
+    background-color: ${(props) => props.theme.secondaryBackgroundColor}A0;
```
SamYap-Softwire, line 36: Is this opacity level something that will be reused? (e.g. whenever we use a modal). If it is, do you think we should add this to `themes`? (e.g. as `blurredBackgroundColor`)

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
+    font-weight: bolder;
+    padding: 0.75em 1em 0.75em 1em;
 `;
+
 const Content = styled.div`
-    overflow: scroll;
-    margin: 1.5rem;
-    margin-top: 1rem;
+    overflow: auto;
+    margin: 1rem 1.5rem 1.5rem 1.5rem;
```
SamYap-Softwire, line 44: This can also be `margin: 1rem 1.5rem 1.5rem` I think

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
+
 const Key = styled.div`
-    color: grey;
-    font-size: small;
+    color: ${(props) => props.theme.foregroundColor};
+    background-color: ${(props) => props.theme.secondaryBackgroundColor}40;
+    display: inline-block;
+    font-size: medium;
+    border-radius: 0.7em;
+    padding: 0.2em 0.5em 0.2em 0.5em;
```
SamYap-Softwire, line 54: Same as above (`padding: 0.2em 0.5em`)

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
 `;
+
 const Value = styled.div`
     font-size: medium;
+    padding: 0.2em 0.5em 0.2em 0.5em;
```
SamYap-Softwire, line 59: Same here

(I think using the shorthand is good for consistency, but it may not be necessary)

src/components/DataViewer/DataViewer.tsx
```diff
@@ -86,15 +115,10 @@ const DataViewer: React.FC<DataViewerProps> = (props) => {
...
-    if (!props.isOpen) {
-        return <></>;
-    }
-
-    Modal.setAppElement("#modal-parent");
 
     return (
         <StyledModal
-            isOpen={true}
+            isOpen={props.isOpen}
```
SamYap-Softwire, line 121: Is there a reason for changing this? 

I think I prefer the previous implementation of 
```if (!props.isOpen) {
    return <></>;
}```
because it means the modal will not be rendered if it doesn't have to, whereas now it is still rendered in the background even if it's not closed.

**Review by SamYap-Softwire — APPROVED** [review]
I think it's good :) 
Just some minor comments on using css shorthands.

File: src/components/DataViewer/DataViewer.tsx [hunk 7]
```diff
@@ -86,15 +115,10 @@ const DataViewer: React.FC<DataViewerProps> = (props) => {
     const closeModal = (): void => {
         props.onRequestClose();
     };
-    if (!props.isOpen) {
-        return <></>;
-    }
-
-    Modal.setAppElement("#modal-parent");
```
paopaosoftwire, line 93: Might want to check https://reactcommunity.org/react-modal/accessibility/ to make sure removing this line is OK. Because I can see that it currently passes the accessibility checks which is great but I don't know it it is necessary to include this line and the div id in SampleDataWithButton to allow screen readers to actually read it.

paopaosoftwire (commented):
Love the styling of the modal!

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: agree on shorthand being a good thing!
these might be remnants from testing different values, but the final production code should definitely have shorter CSS where possible :)

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: should the modal be rendered and 'not shown' for screen-readers and accessibility reasons? having rerenders of the page may be worse than just rendering everything once an dynamically adjusting visibility rules.

(for the record, I don't know what is 'better'! we should Google and decide as a team what's 'better')

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: if it is required then can it be encapsulated at a component-level, so callers of DataViewer don't have to 'know' to put in a div with a certain ID in?

also, for reuse, this may violate certain rules like duplicate IDs... 🤔 may be worth investigating if we pass a generated ID via React or something similar!

TomCusack-Softwire:
this has also evolved past just 'theming' - can you change the PR title?
also have a verb at the start - 'This PR, when merged, will _' - e.g. `VFB-123 Add accessibility checks`

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: additionally I fully agree with not just deleting things because 'accessibility test still passes' - I might not have set up the tests fully, or this may be 'better' extended accessibility than what's currently being checked!
We should make an informed choice for this sort of thing, rather than just writing code that passes the tests :)

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: We looked into the html in the element - the only difference is if we use props.isOpen then there will be an empty react modal div generated when it is closed. It doesn't seem to bring much difference and also the code is tidier when we directly put props.isOpen inside the modal

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Refactored

src/components/DataViewer/DataViewer.tsx
daniel-conlan-softwire: Yeah, I agree on making these colours reusable; although, I was expecting this to be changed during a re-haul of the theming implementation. But, if you think it's best that it's done just now then that can be done

ChloeYiuSoftwire: Also I think rather than adding the opacity as a theme, we should make new colors that looks the same but with 100% opacity. It is because opacity changes the coloring by making it more transparent, and in dark mode it doesn't work as the background it black, it just makes things less visible

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Refactored

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Refactored

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: our newest iteration will use our general modal component which is from the mui library - hence the id is no longer needed for this new library

src/components/DataViewer/DataViewer.tsx
```diff
@@ -65,14 +92,16 @@ const ClearButton: React.FC<ClearButtonProps> = (props) => {
...
 const JSONContent: React.FC<Data> = (data) => {
-    return Object.entries(data).map(([key, value]) => {
-        return (
-            <EachItem key={key}>
-                <Key>{key.toUpperCase().replace("_", " ")}</Key>
-                <Value>{value ?? ""}</Value>
-            </EachItem>
-        );
-    });
+    return (
```
TomCusack-Softwire, line 95: rebase main to get rid of this :)

src/components/DataViewer/SampleDataViewerWithButton.tsx
```diff
@@ -2,7 +2,7 @@
 import React, { useState } from "react";
 import DataViewer, { Data } from "@/components/DataViewer/DataViewer";
 import styled from "styled-components";
-import PhoneIcon from "@/components/Icons/PhoneIcon";
+import PhoneIcon from "@mui/icons-material/Phone";
```
TomCusack-Softwire, line 5: please revert this, I would prefer to use already-styled Phone icons rather than MUIs one :)

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: This still uses React-Modal so please keep this in! We want it to be as accessible as possible at all points - even if it's getting replaced.
(Which will require adding a div with that ID back in!)

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
+    font-size: 1.5em;
+    font-weight: bolder;
+    padding: 0.75em 1em;
 `;
+
 const Content = styled.div`
-    overflow: scroll;
-    margin: 1.5rem;
-    margin-top: 1rem;
+    overflow: auto;
```
TomCusack-Softwire, line 43: why has this been changed?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
+    margin: 1rem 1.5rem 1.5rem;
+    background-color: ${(props) => props.theme.surfaceBackgroundColor};
 `;
+
 const Key = styled.div`
-    color: grey;
-    font-size: small;
+    color: ${(props) => props.theme.foregroundColor};
```
TomCusack-Softwire, line 49: can this be `secondaryForegroundColor` - it's not really that great to mix and match colours! Pick on type (secondary) and stick with it :)

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
+    background-color: ${(props) => props.theme.secondaryBackgroundColor}40;
+    display: inline-block;
+    font-size: medium;
```
TomCusack-Softwire, line 52: is medium not default? can this be removed?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
-const CloseButton = styled.button`
+
+const CloseButton = styled.button.attrs({
+    "aria-label": "Close Button",
+})`
     border: 0;
-    background-color: red;
-    color: white;
+    border-radius: 50%;
+    color: grey;
```
TomCusack-Softwire, line 71: can this be from a theme?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -15,45 +17,70 @@ const StyledModal = styled(Modal)`
...
     display: flex;
     justify-content: center;
+    aspect-ratio: 1;
+    align-items: center;
+    transition: 0.2s;
+
+    &:hover {
+        color: white;
+        background-color: darkgray;
```
TomCusack-Softwire, line 80: can these be from a theme?

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: There are likely a lot of ARIA labels which help the screen-readers etc out when it is closed - there's a reason they provide an `isOpen` attribute instead of just saying 'only have this on your page if needed'!

TomCusack-Softwire: Also apparently there may be a 'better closing animation' when you close the page - may be worth checking!

src/components/DataViewer/DataViewer.tsx
```diff
@@ -103,7 +127,7 @@ const DataViewer: React.FC<DataViewerProps> = (props) => {
                 {props.header}
                 <ClearButton closeModal={closeModal} />
             </Header>
-            <Content>{JSONContent(props.data)}</Content>
+            <Content tabIndex={0}>{JSONContent(props.data)}</Content>
```
TomCusack-Softwire, line 130: what's `tabIndex` do?

TomCusack-Softwire:
PR title is slightly misleading - there's more than just adding themes, and you don't add any accessibility **checks**! maybe `Improve CSS and Accessibility of Data View Component` would be better :)

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Refactored

src/components/DataViewer/DataViewer.tsx
daniel-conlan-softwire: `tabIndex` allows for the modal to be _focusable_ by a user of the website. This is no longer required though as we are changing the modal to use Material UI

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: This is an issue of VFB-16 and we have left a note on this for the person handling VFB-16 to work  on the color palette

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Same for this, it is an issue of VFB-16 and we have left a note on this for the person handling this ticket to refactor it once done

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Deleted as suggested

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Same for this, it is an issue of VFB-16 and we have left a note on this for the person handling this ticket to refactor it once done

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: We have migrated to use MUI Modal which has different css properties - so the newest iteration should be the css that fits the new library

src/components/DataViewer/SampleDataViewerWithButton.tsx
ChloeYiuSoftwire: We are not using the icon in our newest iteration

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: We have rebased to main

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Ok! However our newest version has already replaced with MUI library, but will bear this in mind!

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Refactored!

SamYap-Softwire (approved):
I think it'll be good after you clean up the TODO comments

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: I agree that most of the time we shouldn't leave comments in code, but in this case it's referring to another ticket which currently is in 'todo', meaning someone else will pick it up. It has a ticket (VFB-16) and this is a good way to link the Jira ticket with code, and the ticket shouldn't be finished until this is done!

I do think this is a non-trivial amount of work that _should_ be done in the separate ticket - if it was tiny then it should be done, but this is 'create a whole new palette of colours', which fits 16 better than 3! There is also a 'quick fix' here - using 40 as opacity, which is 'fine for main' but should be fixed as 'tech debt' - which is why they've left the comment.

src/components/DataViewer/DataViewer.tsx
```diff
@@ -9,23 +9,28 @@ export interface Data {
...
 `;
 
 const EachItem = styled.div`
     padding-bottom: 1em;
 `;
 
+const ContentDiv = styled.div`
+    width: 1000px;
+    max-width: 100%;
+`;
```
Var1377, line 32: use `width: min(100%, 1000px)`

Var1377 (approved):

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Hmmm it is not working - so maybe we will stick to the original one

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: Can we set this to something in themes, even if it isn't 'as good', e.g. white/black? Why does it need to be grey?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -1,78 +1,58 @@
...
+    return (Array.isArray(value) && value.length === 0) || value === "" || value === null;
+};
 
-const ClearButton: React.FC<ClearButtonProps> = (props) => {
-    return (
-        <CloseButton onClick={props.closeModal}>
-            <CloseIcon />
-        </CloseButton>
-    );
+const formatDisplayValue = (value: any): string => {
```
TomCusack-Softwire, line 38: value should be of type `string[] | string | number | boolean | null` (which may be easier to set as a type if you type it out enough lol)

src/components/DataViewer/DataViewer.tsx
```diff
@@ -1,78 +1,58 @@
...
+    if (valueIsEmpty(value)) {
+        return "-";
+    } else if (typeof value === "boolean") {
```
TomCusack-Softwire, line 41: `else` is not needed if there's a `return` - the code should just be like
```tsx
if (valueIsEmpty(value)) {
   return "-";
}
if (/* ... */) {
   return;
}
return;
```

src/components/Modal/Modal.cy.tsx
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+import Modal from "@/components/Modal/Modal";
+
+describe("General Modal Component", () => {
+    it("renders", () => {
+        cy.mount(
+            <Modal header="Modal Header" ariaLabel="Modal Label" isOpen={true} onClose={() => {}}>
```
TomCusack-Softwire, line 7: Is `ariaLabel` something that we should be implementing in `DataViewer.tsx`?

src/components/Modal/Modal.cy.tsx
```diff
@@ -0,0 +1,47 @@
...
+        const onCloseSpy = cy.spy().as("onCloseSpy");
+
+        cy.mount(
+            <Modal header="Modal Header" isOpen={true} onClose={onCloseSpy}>
+                <h1>Modal Content</h1>
+            </Modal>
+        );
+
+        cy.get("svg").parent("button").click();
+        cy.get("@onCloseSpy").should("have.been.called");
```
TomCusack-Softwire, line 45: is `have.been.calledOnce` a thing? slightly stricter as an assertion
(but great tests!)

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    &:hover {
+        // TODO VFB-16 Change to using theme with palettes of colours
+        color: white;
+        background-color: darkgray;
+    }
+`;
+
+interface ModalProps {
+    header: ReactNode;
```
TomCusack-Softwire, line 63: I prefer `ReactElement` over `ReactNode`, but try `React.FC` (the best but fiddliest option) first!

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    display: flex;
+    justify-content: center;
+    aspect-ratio: 1;
+    align-items: center;
+    transition: 0.2s;
+    margin-left: 1em;
+
+    &:hover {
+        // TODO VFB-16 Change to using theme with palettes of colours
```
TomCusack-Softwire, line 56: As mentioned somewhere else, I don't want you to push hardcoded colours because 'it will be changed later' - main branch should be good to revert to at all times!
I'd much rather you use a themed colour (that might not be the same) and put in the comment exactly what you want this colour to look more like (e.g. `color: ...theme.foregroundColor // TODO VFB-16 Add 'grey' to palette of colours`)

TomCusack-Softwire (changes requested):

src/components/DataViewer/DataViewer.tsx
daniel-conlan-softwire: Yeah, me and Chloe agreed on this point, but thought that you would prefer the explicit `else` cases

src/components/DataViewer/DataViewer.tsx
ChloeYiuSoftwire: Refactored

src/components/Modal/Modal.cy.tsx
ChloeYiuSoftwire: We make it optional

src/components/Modal/Modal.tsx
daniel-conlan-softwire: I think `ReactElement` is best for this situtation as not all things passed to the modal will be a functional component - could be a `string` for instance

src/components/Modal/Modal.tsx
ChloeYiuSoftwire: Refactored

src/components/DataViewer/DataViewer.tsx
TomCusack-Softwire: cool - I think `else` is a relatively 'unclean' practice! It's a sign that the if could be extracted to functions with `return`s instead :)

src/components/DataViewer/DataViewer.tsx
```diff
@@ -1,78 +1,64 @@
...
-        </CloseButton>
-    );
+const formatDisplayValue = (value: valueType): string => {
+    if (valueIsEmpty(value)) {
+        return "-";
+    }
+
+    if (typeof value === "boolean") {
+        const booleanString = value.toString();
+        return booleanString[0].toUpperCase() + booleanString.slice(1);
```
TomCusack-Softwire, line 49: This is very general when it's literally just a boolean - `return value ? "True" : "False";` also works!
Also is true/false 'technical language' which may be better phrased as 'Yes/No'?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -1,78 +1,64 @@
...
+    }
+
+    return value!.toString();
```
TomCusack-Softwire, line 52: why is the exclamation mark needed? is there a type which 'gets through' the type system and passes through `valueIsEmpty` such that `value` could be 'empty' at this point? (`undefined`?)

Maybe `valueIsEmpty` is better defined as `(array bit) || !value;`? does this catch too much?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -84,31 +70,23 @@ interface DataViewerProps {
     header: ReactElement | string;
     isOpen: boolean;
     onRequestClose: () => void;
+    ariaLabelledBy: string;
```
TomCusack-Softwire, line 73: did you say aria label is optional? doesn't look it here.
(can we set it by default to something if you want it optional, or just make it non-optional and force devs to put a label? happy with either)

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,97 @@
...
+    width: 100%;
+`;
+
+const Content = styled.div`
+    overflow: auto;
+    padding: 1em 1.5em 1.5em;
+    width: 100%;
+`;
+
+const CloseButton = styled.button.attrs({
```
TomCusack-Softwire, line 41: is a closed button common enough to be it's own file? potentially under Button?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,97 @@
...
+    display: flex;
+    justify-content: center;
+    aspect-ratio: 1;
+    align-items: center;
+    transition: 0.2s;
+    margin-left: 1em;
+
+    &:hover {
+        // TODO VFB-16 Either change to 'white' or the corresponding 'secondaryBackgroundColor'
+        color: ${(props) => props.theme.secondaryBackgroundColor}; // white
```
TomCusack-Softwire, line 64: end `// white` not needed since it's covered in the TODO

TomCusack-Softwire (changes requested):
Almost there!

src/components/DataViewer/DataViewer.tsx
daniel-conlan-softwire: Yep, that's a very good point. Much better solution

src/components/DataViewer/DataViewer.tsx
daniel-conlan-softwire: Doesn't look like the type system can infer that the return value from `valueIsEmpty` cannot be `null`. That's why we had to add the exclamation mark

src/components/DataViewer/DataViewer.tsx
daniel-conlan-softwire: Because the modal has an aria role of `dialog` an `aria-labelledby` property (different to `aria-label`) has to be passed to the `Modal` component. If this is not setup on the modal, then accessibility standards are violated - that's why it hasn't been set as optional

src/components/Modal/Modal.tsx
daniel-conlan-softwire: Sorry these were just artefacts, didn't mean to commit them

src/components/Modal/Modal.tsx
daniel-conlan-softwire: At this time, I don't think it's worth it at this point for the close button to be its own file just yet. Do you still want me to refactor it out into its own component?

TomCusack-Softwire (approved):
Happy once 'ariaLabelledBy' has been renamed!

PR #21, opened 2023-07-19
VFB-6 Add a lists page

Var1377:
This PR implements the lists page


src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    const dataAndTooltips = rawData?.map((row) => {
+        const data: ListRow = {
+            item_name: row.item_name,
+        };
+        const tooltips: ListRow = {};
+
+        for (let i = 1; i <= 10; i++) {
+            const header = `${i}_quantity`;
+            // type cast required as the type system can't infer that the header is a key of row
```
TomCusack-Softwire, line 39: would this be fixed by changing L26 (init to empty list) to be
```tsx
const rawData: ListRow[] = [];
```
I really don't like `as` unless absolutely necessary, since we should be encoding types 'well enough' at all points. I think this is just TS going 'empty list could be a list of anything'.

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const StyledTable = styled(Table)`
+    width: 100%;
+    background-color: transparent;
+`;
+
+type ListRow = { [headerKey: string]: string };
+
+const ListsDataView = async (): Promise<React.ReactElement> => {
+    // const rawData = (await supabase.from("lists").select("*")).data;
```
TomCusack-Softwire, line 25: the brackets here suggest that two 'logical' things are being done - so should be two constants!
(There's no 'cost' on variables, especially since this will likely compile down to the same thing, so please do write nice clean code!)
e.g. `const response = await supabase...select(); const rawData = response.data;`

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const rawData = [];
+    for (let i = 0; i < 100; i++) rawData.push({ item_name: `Item ${i}` });
```
TomCusack-Softwire, line 27: is it really that much more effort to type `index` instead of `i`? :P

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    console.log("fetched data");
+
+    const dataAndTooltips = rawData?.map((row) => {
```
TomCusack-Softwire, line 31: this is such a massive function I'd be tempted to extract to a different (named) function outside of the component

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,82 @@
...
+            tooltips[header] = (row as ListRow)[`${i}_notes`];
+        }
+
+        return {
+            data,
+            tooltips,
+        };
+    });
+
+    if (dataAndTooltips === null || dataAndTooltips === undefined || dataAndTooltips.length === 0) {
```
TomCusack-Softwire, line 50: is this covered by `if (!dataAndTooltips)`?

src/app/lists/loading.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { styled } from "styled-components";
+
+const Centerer = styled.div`
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 80vh;
+`;
+
+export const Loading: React.FC<{}> = () => {
```
TomCusack-Softwire, line 14: can this just be the main Loading page? I don't see why we need a separate component for lists when we can just improve the root one :)

src/app/lists/page.tsx
```diff
@@ -1,10 +1,12 @@
 import { Metadata } from "next";
 import React from "react";
+import ListsDataView from "./dataview";
```
TomCusack-Softwire, line 3: @ syntax!

src/app/themes.tsx
```diff
@@ -21,6 +21,7 @@ type CustomTheme = {
     accentForegroundColor: string;
     surfaceForegroundColor: string;
     surfaceBackgroundColor: string;
+    disabledColor: string;
```
TomCusack-Softwire, line 24: what's disabled colour? is this not just 'normal colour with `<input ... disabled/>`, in which it sets it to some greyed out colour?

src/app/themes.tsx
```diff
@@ -31,10 +32,12 @@ const lightTheme: CustomTheme = {
     primaryBackgroundColor: "#63a036",
     secondaryForegroundColor: "#ffffff",
     secondaryBackgroundColor: "#066940",
-    accentBackgroundColor: "#609fd3",
+    accentBackgroundColor: "#1b629c",
```
TomCusack-Softwire, line 35: why is this changed here? this should be in a new ticket (attached to the Themes VFB ticket number) if it's a themeing change - it's not part of 'creating a list page' necessarily!

src/app/themes.tsx
```diff
@@ -57,6 +61,8 @@ interface Props {
     theme?: CustomTheme;
 }
 
+let forceRerenderBecauseImTotallySane = 0;
```
TomCusack-Softwire, line 64: lol does this actually do anything? is there an implicit bug that needs to be fixed?

src/components/Icons/Icon.tsx
```diff
@@ -15,12 +15,62 @@ const StyledFontAwesomeIcon = styled(FontAwesomeIcon)`
...
     return (
-        <StyledFontAwesomeIcon icon={props.icon} title={props.onHoverText} color={props.color} />
+        <>
+            {props.onHoverText && hovered ? (
+                <Popper
+                    onMouseEnter={show}
+                    onMouseLeave={hide}
+                    onMouseDown={show}
+                    onMouseUp={hide}
+                    onTouchStart={show}
```
TomCusack-Softwire, line 54: is it worth having the `touch` mobile events be toggles on any tap, rather than just while long holding? I know my iPhone will immediately go to a 'different view' if you long hold for too long.

(also might be better to have multiple open at the same time! might want to just mess around with different approaches and see what 'feels' most natural 🙂)

src/components/Icons/SpeechBubbleIcon.tsx
```diff
@@ -4,9 +4,10 @@ import { faComment } from "@fortawesome/free-solid-svg-icons";
 
 interface Props {
     onHoverText?: string;
+    color?: string;
 }
 const SpeechBubbleIcon: React.FC<Props> = (props) => {
-    return <Icon icon={faComment} onHoverText={props.onHoverText} color="blue" />;
+    return <Icon icon={faComment} onHoverText={props.onHoverText} color={props.color} />;
```
TomCusack-Softwire, line 10: why this change? I can't see where you've used this, but I don't see why it 'shouldn't be blue - can understand making this the _theme_ blue (rather than 'hardcoded' blue) but not having it just be 'any colour specified'

src/components/Tables/Table.tsx
```diff
@@ -3,9 +3,15 @@
 import React, { useState, useEffect } from "react";
 import DataTable, { TableColumn } from "react-data-table-component";
 import TableFilterBar, { FilterText } from "@/components/Tables/TableFilterBar";
+import { styled } from "styled-components";
```
TomCusack-Softwire, line 6: should be `import styled from "styled-components";`

surprised it exports styled as a prop as well as the default tbf

src/components/Tables/Table.tsx
```diff
@@ -3,9 +3,15 @@
...
+import { NoSsr } from "@mui/material";
```
TomCusack-Softwire, line 7: nice, didn't know this was a thing, was planning on finding a `NoSsr` thing at some point!

src/components/Tables/Table.tsx
```diff
@@ -3,9 +3,15 @@
...
+import SpeechBubbleIcon from "../Icons/SpeechBubbleIcon";
```
TomCusack-Softwire, line 8: @ syntax please, just makes moving files and navigating the repo etc a lot easier :)

src/components/Tables/Table.tsx
```diff
@@ -3,9 +3,15 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faAnglesUp, faAnglesDown } from "@fortawesome/free-solid-svg-icons";
 
 export interface Datum {
-    [headerKey: string]: string | number | null;
+    data: { [headerKey: string]: string | number | null | undefined };
+    tooltips?: { [headerKey: string]: string | null };
```
TomCusack-Softwire, line 14: at what point is it better to structure data as 
```
{ headerkey: {data: xyz, tooltip?: abc} }
```
rather than
```
{ data: {headerKey: xyz}, tooltip?: {headerKey: abc} }
```
?

(I think this is good, just wanted to spawn a conversation about 'higher-level design', especially if it helps the type system!)

src/components/Tables/Table.tsx
```diff
@@ -86,41 +109,83 @@ const Table: React.FC<Props> = (props) => {
...
+                            const newData = [...data];
+                            const temp = newData[row.rowId];
+                            newData[row.rowId] = newData[row.rowId - 1];
+                            newData[row.rowId - 1] = temp;
+                            return newData;
+                        });
+                    }} icon={faAnglesUp} />
+                    <StyledIcon onClick={() => {
+                        if (row.rowId === data.length - 1) return;
+                        setData(data => {
```
TomCusack-Softwire, line 176: potential to export this setData to a new function?

src/components/Tables/Table.tsx
```diff
@@ -132,28 +197,140 @@ const Table: React.FC<Props> = (props) => {
...
-            subHeaderComponent={
-                <TableFilterBar
-                    filterText={filterText}
-                    onFilter={onFilter}
-                    handleClear={handleClear}
-                    headers={props.headers}
+        <Styling>
+            <NoSsr>
+                <StyledDataTable
+                    columns={columns as any}
```
TomCusack-Softwire, line 206: `as any`?

src/components/Tables/Table.tsx
```diff
@@ -132,28 +197,140 @@ const Table: React.FC<Props> = (props) => {
         }
     };
 
-    if (!domLoaded) {
-        return <></>;
-    }
+    const filters = (props.filters) ? Object.fromEntries(props.filters.map(filter => [filter, props.headers[filter]])) : props.headers;
```
TomCusack-Softwire, line 200: what does this do? potential to split out a little bit to be more clear - even if the actual 'code content' is still long, because JS is sometimes just bad!!

src/components/Tables/Table.tsx
TomCusack-Softwire, line 1: remember to add more tests for anything you write yourself! the client needs guarantees that things work the way we want them to!

src/components/Tables/Table.tsx
```diff
@@ -132,28 +197,140 @@ const Table: React.FC<Props> = (props) => {
...
+
+        padding: 4px 1px 4px 8px;
+
+        border-radius: 10px;
+        width: 10rem;
+        border: solid 1px ${(props) => props.theme.disabledColor};
+    }
+`;
+
+const StyledDataTable = styled(DataTable)`
```
TomCusack-Softwire, line 325: how much of the above `Styled` can be moved into `StyledDataTable`?

TomCusack-Softwire (changes requested):
I know this was stylistic only but I couldn't resist commenting on the 'style' of the code :D

src/app/lists/dataview.tsx
DanWilkins2107: Changed

src/app/lists/dataview.tsx
DanWilkins2107: as has been removed

src/app/lists/dataview.tsx
DanWilkins2107: Changed

src/app/lists/page.tsx
```diff
@@ -1,10 +1,11 @@
...
 
 const Lists: React.FC<{}> = () => {
     return (
         <main>
-            <h1> Lists Page </h1>
```
paopaosoftwire, line 7: Need a h1 tag to pass accessibility test.

"Ensure that the page, or at least one of its frames contains a level-one heading" - cypress

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,92 @@
...
+        ["4_quantity", "Family of 4"],
+        ["5_quantity", "Family of 5"],
+        ["6_quantity", "Family of 6"],
+        ["7_quantity", "Family of 7"],
+        ["8_quantity", "Family of 8"],
+        ["9_quantity", "Family of 9"],
+        ["10_quantity", "Family of 10+"],
+    ];
+
+    const toggleableHeaders = headers.map(([key, _value]) => key);
```
paopaosoftwire, line 69: Nice

src/components/Tables/TableFilterBar.tsx
```diff
@@ -1,33 +1,70 @@
...
-                    />
-                );
-            })}
-            <button onClick={props.handleClear}>Clear</button>
+            <ClearButton onClick={props.handleClear}>Clear</ClearButton>
+            {props.headers
+                .filter(([key, _value]) => props.filterKeys?.includes(key) ?? true)
+                .map(([key, value]) => {
+                    return (
+                        <StyledFilterBar
```
paopaosoftwire, line 53: I tried the Filter and its currently doing nothing. Is this currently being fixed because the table filter functionality was working before.

paopaosoftwire (changes requested):
Nice styling

src/components/Tables/Table.tsx
Var1377: Done

src/components/Tables/Table.tsx
Var1377: Fixed

src/components/Tables/TableFilterBar.tsx
Var1377: Fixed

ChloeYiuSoftwire (commented):
General comments from me and Daniel:

1. Would be nice to have a select all/none button for the select column

2. Would be nice if the column selector doesn't look like a text box

3. On my screen, family of 10 is shown as 'family of ...' - is it something intended? Feels like columns header should be fully shown

4. Would be nice to have larger text (both the content and the comments - size in the "Clear" button seems nice), and would be nice if the style of the heading is slightly different from the text (e.g. making them bold) 

5. Would be nice when there is less column, rather than making the whole table span the whole page and make each column have a lot of white space, maybe just make the table narrower and have less space between each column (with the table centered in the page)

6. Somehow the column heading on the left hand side is slight left to the content beneath, but the column heading on the right hand side is slight right to the content beneath - might need some fix

7. Would be a nice feature if I can close the select column thingy by clicking elsewhere rather than exactly the select column button

8. Perhaps adding responsiveness to clear and add buttons on hover and on click might be nice (currently the old navbar has this feature too!)

src/app/themes.tsx
Var1377: It was a theme specific grey, but for now I've changed all instances to grey and added a todo message for VFB16

src/app/themes.tsx
Var1377: reverted

src/components/Tables/Table.tsx
Var1377: If designing from scratch, something along the lines of what you're suggesting would be preferable, however this made things easier to integrate with the existing codebase.

Tbh if I was designing it from scratch we would be using arrays and not objects as I feel that makes more sense for something data oriented.

Var1377:
> General comments from me and Daniel:
> 
>     1. Would be nice to have a select all/none button for the select column
> 
>     2. Would be nice if the column selector doesn't look like a text box (e.g. adding an inverted triangle that signals something will pop below when clicked)
> 
>     3. On my screen, family of 10 is shown as 'family of ...' - is it something intended? Feels like columns header should be fully shown
> 
>     4. Would be nice to have larger text (both the content and the comments - size in the "Clear" button seems nice), and would be nice if the style of the heading is slightly different from the text (e.g. making them bold)
> 
>     5. Would be nice when there is less column, rather than making the whole table span the whole page and make each column have a lot of white space, maybe just make the table narrower and have less space between each column (with the table centered in the page)
> 
>     6. Somehow the column heading on the left hand side is slight left to the content beneath, but the column heading on the right hand side is slight right to the content beneath - might need some fix
> 
>     7. Would be a nice feature if I can close the select column thingy by clicking elsewhere rather than exactly the select column button
> 
>     8. Perhaps adding responsiveness to clear and add buttons on hover and on click might be nice (currently the old navbar has this feature too!)
> 
>     9. Can consider using the FreeFormTextInput component created in VFB-5 for the "Filter by Description" box

1. I feel this would be overkill for <= 10 options but I would be keen to hear other people's opinions.
2. Good idea
3. This is not intended, will look for a fix.
4. This would be down to the user changing browser scaling, we're using a 1rem font size.
5. This isn't easily doable with react-data-table unfortunately.
6. Will have a look
7. It's an MUI accordion, will look into it but if it isn't native behaviour it's probably not worth implementing.
8. These are MUI buttons, this sort of thing would be part of theming
9. Will have a look.

@ChloeYiuSoftwire @daniel-conlan-softwire

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,202 @@
...
+    const [errorMsg, setErrorMsg] = useState<string | null>(null);
+
+    if (rawData === null) {
+        throw new Error("No data found");
+    }
+
+    const remapTooltips = (
+        row: ListRow
+    ): Datum & { unmappedTooltips: { [key: string]: string | null } } => {
+        const data: { [key: string]: string | null } = {
```
paopaosoftwire, line 80: { [key: string]: string | null }, is being used across both files multiple times, Might be better to put this into a type?

src/app/lists/edit_modal.tsx
```diff
@@ -0,0 +1,155 @@
+"use client";
```
paopaosoftwire, line 1: The file name is in snake_case but the others are done in camelCase. Is there some sort convention? I'm curious!

src/components/Tables/Table.cy.tsx
```diff
@@ -1,76 +1,61 @@
 import React from "react";
-import Table from "@/components/Tables/Table";
+import Table, { Datum } from "./Table";
```
paopaosoftwire, line 2: @ import please.

src/components/Tables/Table.tsx
```diff
@@ -135,28 +258,164 @@ const Table: React.FC<Props> = (props) => {
...
-            }
-            pagination
-            persistTableHead
-        />
+            </NoSsr>
+        </Styling>
     );
 };
 
+const EditandReorderArrowDiv = styled.div`
```
paopaosoftwire, line 293: Again, might wanna group of styled components together.

src/components/Tables/Table.tsx
```diff
@@ -135,28 +258,164 @@ const Table: React.FC<Props> = (props) => {
...
+        > nav {
+            background-color: transparent;
+            color: ${(props) => props.theme.surfaceForegroundColor};
+            border-top: solid 1px ${(props) => props.theme.foregroundColor};
+        }
+    }
+
+    // the icons in the pagination bar
+    & svg {
+        // TODO: VFB-16 different greys for different theme modes
```
paopaosoftwire, line 341: There are quite a few TODOs and comments. Does it need to all be here?

src/app/lists/edit_modal.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        let operation;
+
+                        if (data === null) {
+                            operation = supabase.from("lists").insert(toSubmit);
+                        } else {
+                            operation = supabase
+                                .from("lists")
+                                .update(toSubmit)
+                                .eq("primary_key", toSubmit.primary_key)
+                                .select();
```
paopaosoftwire, line 111: When I want to insert a new row, I got Failed to load resource: the server responded with a status of 409. Is that just my laptop?

paopaosoftwire: Also it seems like I cant edits the rows as well. Is this still in development?

paopaosoftwire (changes requested):
Love the look of all this! Not all the things I mentioned need to be change just comment where appropriate. Lmk when you need me to look at it again:)

src/app/lists/edit_modal.tsx
Var1377: That's an issue with the database, modifications were working previously, I'll have a look into it

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,202 @@
...
+const ListsDataView: React.FC<Props> = ({ data: rawData }) => {
+    // null => add, undefined => modal closed
+    const [modal, setModal] = useState<Datum | null | undefined>(undefined);
+    const [toDelete, setToDelete] = useState<number | null>(null);
+    // need another setState otherwise the modal content changes before the close animation finishes
+    const [toDeleteModalOpen, setToDeleteModalOpen] = useState<boolean>(false);
+    const [errorMsg, setErrorMsg] = useState<string | null>(null);
+
+    if (rawData === null) {
+        throw new Error("No data found");
```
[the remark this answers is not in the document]
Var1377: Throwing an error here is intended, when throwing an error in next it shows an error page

src/components/Tables/Accordion.tsx
```diff
@@ -0,0 +1,113 @@
...
+    height: 3rem;
+    overflow: visible;
+    z-index: 2;
+    margin-bottom: 0.5rem;
+    margin-top: 0.5rem;
+    & .MuiPaper-root {
+        background-color: ${(props) => props.theme.surfaceBackgroundColor};
+        border: 1px solid ${(props) => props.theme.secondaryBackgroundColor};
+        box-shadow: 0 0 1px ${(props) => props.theme.secondaryBackgroundColor};
+        // TODO: VFB16 change to a theme specific grey
```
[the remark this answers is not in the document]
Var1377: These are for Chloe to do when she does theming, these aren't my own.

src/components/Tables/Table.tsx
Var1377: The comments are the only thing that make the css maintainable and understandable as I'm targeting components coming from a library. The todos are for Chloe who's doing theming.

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,202 @@
...
+    background-color: transparent;
+`;
+
+const StyledAddButton = styled.button`
+    margin: 10px 5px 5px 0;
+    height: 2rem;
+    width: 5rem;
+`;
+
+type Props = {
```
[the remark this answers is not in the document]
daniel-conlan-softwire: I'd pop the `Props` type right above the definition of the main component of this file so that they're grouped

daniel-conlan-softwire (approved):
Looks fine to me - seems as though you've already sorted the issue with the `.only` in the Cypress tests

src/app/lists/dataview.tsx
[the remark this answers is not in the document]
paopaosoftwire: It depends on personal preference I guess!

src/components/Tables/Accordion.tsx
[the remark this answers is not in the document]
paopaosoftwire: Then do we prefer to put the TODOs on main or on the JIRA ticket or both? Just want to know incase something similar comes up in my ticket.

paopaosoftwire (approved):
Looks good, Thanks for clarifying a lot of the stuff!

src/components/Tables/Table.tsx
```diff
@@ -135,28 +249,175 @@ const Table: React.FC<Props> = (props) => {
...
+
+const EditandReorderArrowDiv = styled.div`
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
+    width: 100%;
+    // this transform is necessary to make the buttons visually consistent with the rest of the table without redesigning the layout
+    transform: translateX(-0.8rem);
+`;
+
+const StyledIcon = styled(FontAwesomeIcon)`
```
TomCusack-Softwire, line 303: Should this use Icon component over FontAwesome directly?

src/components/Tables/Table.tsx
```diff
@@ -89,41 +125,119 @@ const Table: React.FC<Props> = (props) => {
...
+                                onClick={() => swapRows(row.rowId, row.rowId + 1)}
+                                aria-label="reorder row downwards"
+                            >
+                                <StyledIcon icon={faAnglesDown} />
+                            </IconButton>
+                        ) : (
+                            <></>
+                        )}
+                        {props.onDelete ? (
+                            <IconButton
```
TomCusack-Softwire, line 226: same here - how much does our existing Icon component cover?

src/components/Tables/TableFilterBar.tsx
```diff
@@ -1,33 +1,74 @@
 "use client";
 
 import React from "react";
+import { styled } from "styled-components";
+import FilterAccordion from "@/components/Tables/Accordion";
 
-export interface FilterText {
+export type FilterText = {
```
TomCusack-Softwire, line 7: why did you change this? what's the advantage of interface over type?
(I think I prefer interface over type unless needed, but if there's other benefits I can be convinced!)

src/components/Tables/TableFilterBar.tsx
```diff
@@ -1,33 +1,74 @@
...
+    width: 6rem;
+    font-size: 1rem;
+`;
+
+const StyledFilterBar = styled.input`
+    height: 3rem;
+    font-size: 1rem;
+    width: 15rem;
+    overflow: visible;
+    border: 1px solid ${(props) => props.theme.secondaryBackgroundColor}!important;
```
TomCusack-Softwire, line 30: please no `!important`, it is bad practice

TomCusack-Softwire (changes requested):

src/components/Tables/Accordion.tsx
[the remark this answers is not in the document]
Var1377: I was under the impression that accordion was a well-known term even outside of UI programming, similar to something like "Dropdown Menu". I would guess that changing this may harm understandability.

src/components/Tables/TableFilterBar.tsx
Var1377: Types are (imo) strictly better. You can compose them with & and |; they can alias primitives, arrays, tuples, functions; and many of typescripts utility types (https://www.typescriptlang.org/docs/handbook/utility-types.html) only work on types. In this case both would work, but interfaces have a strict subset of functionality and (again imo) should never really be used.

(also not sure why but `type` has superior intellisense in vscode, not sure how intended that is)

src/components/Tables/Table.tsx
```diff
@@ -58,15 +86,23 @@ const dataToRows = (data: Datum[], headers: Headers): Row[] => {
...
 };
 
 const Table: React.FC<Props> = (props) => {
+    const [shownHeaderKeys, setShownHeaderKeys] = useState(
+        props.defaultShownHeaders ?? props.headers.map(([key]) => key)
+    );
+
+    const shownHeaders = props.headers.filter(([key]) => shownHeaderKeys.includes(key));
+
+    const [data, setData] = useState(props.data);
```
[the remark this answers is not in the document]
Var1377: Row reordering isn't yet persisted anywhere, I'll look into keeping it in localstorage for now. Could you look into implementing some sort of database level ordering?

src/components/Tables/Table.tsx
Var1377: Our icon component adds unwanted colour and padding

src/components/Tables/Table.tsx
Var1377: Our Icon component adds styling that isn't necessary when used with a material icon button so it's easier and cleaner to not use it.

src/components/Tables/Table.tsx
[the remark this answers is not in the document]
Var1377: I've had a look into it and it doesn't seem worth it to implement if we're gonna persist it on the database anyway. Could this be a separate ticket instead.

src/components/Tables/TableFilterBar.tsx
TomCusack-Softwire: image
disagree slightly, I think `type` can alias strings etc better but should only 'really' be used as an alias for existing types, whereas interfaces describe 'classes'/objects better (since often they are used in OOP) and the 'strict subset' stuff is usually desired.

I would much rather you stick to consistent use of interfaces wherever possible, but if you can't do a problem using interfaces then revert to types, but ideally I'd like to only use interfaces!

src/components/Tables/Table.tsx
TomCusack-Softwire: I would much rather we update the Icon and use it wherever possible (for consistency, and for if we need to change every Icon quickly)

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,199 @@
...
+    const [toDeleteModalOpen, setToDeleteModalOpen] = useState<boolean>(false);
+    const [errorMsg, setErrorMsg] = useState<string | null>(null);
+
+    if (rawData === null) {
+        throw new Error("No data found");
+    }
+
+    const remapTooltips = (
+        row: ListRow
+    ): Datum & { unmappedTooltips: { [key: string]: string } } => {
```
TomCusack-Softwire, line 59: Can you extract these all to interfaces? (potentially using 'extends')
`{[key: string]: string}` is enough on one line to be a separate interface, in my opinion.

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+    return (
+        <>
+            <ConfirmDialog
+                message={`Are you sure you want to delete ${
+                    toDelete ? rawData[toDelete].item_name : ""
+                }?`}
+                isOpen={toDeleteModalOpen}
+                onConfirm={async () => {
+                    if (toDelete !== null) {
```
TomCusack-Softwire, line 114: can this be extracted to it's own (named) function?

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,199 @@
...
+                            .delete()
+                            .match({ primary_key: data.primary_key });
+
+                        if (Math.floor(response.status / 100) !== 2) {
+                            setErrorMsg(
+                                `${response.status}: ${response.statusText} -- data: ${response.data}`
+                            );
+                        } else {
+                            setToDeleteModalOpen(false);
+                            window.location.reload();
```
TomCusack-Softwire, line 127: why are we reloading here?

src/app/lists/editModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+        return {};
+    };
+
+    const [toSubmit, setToSubmit] = useState<{ [key: string]: string | null }>(
+        generateDefaultData()
+    );
+
+    const [errorMsg, setErrorMsg] = useState<string | null>(null);
+
+    const headersAndTooltips: [[string, string], [string, string]][] = [];
```
TomCusack-Softwire, line 54: again can you extract complex types into an interface? (or probably a type here, since it's just an alias for string lists)

src/app/lists/editModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const headersThatHaveTooltips = headers.filter(([key]) => key !== "item_name");
+
+    for (let index = 0; index < headersThatHaveTooltips.length; index++) {
+        headersAndTooltips.push([headersThatHaveTooltips[index], tooltips[index]]);
```
TomCusack-Softwire, line 58: can this be some sort of one-line filter? not sure how complex that would be

src/app/lists/editModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    return (
+        <Modal header="Edit List" headerId="editList" isOpen={data !== undefined} onClose={onClose}>
+            <ModalInner>
+                <h3>Description</h3>
+                <TextInput
+                    defaultValue={toSubmit.item_name ?? ""}
+                    onChange={(event) => {
+                        const text = event.target.value as string;
```
TomCusack-Softwire, line 68: this is used enough that it may be worth extracting into a separate function - take specifically a text input on change event and the key to change, and set the state of the object

src/app/lists/editModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+                        }
+
+                        const response = await operation;
+
+                        if (Math.floor(response.status / 100) !== 2) {
+                            setErrorMsg(
+                                `${response.status}: ${response.statusText} -- data: ${response.data}`
+                            );
+                        } else {
+                            window.location.reload();
```
TomCusack-Softwire, line 121: do you think it's worth commonising setting an error message from supabase in `supabase.ts`?
(again reload too)

src/app/lists/page.tsx
```diff
@@ -1,10 +1,23 @@
 import { Metadata } from "next";
-import React from "react";
+import React, { ReactElement } from "react";
+import ListsDataView from "@/app/lists/dataview";
+import supabase from "@/supabase";
+import { Database } from "@/database_types_file";
+
+export const revalidate = 0;
+
+const fetchData = async (): Promise<Database["public"]["Tables"]["lists"]["Row"][] | null> => {
```
TomCusack-Softwire, line 9: `{Schema} from "@/supabase"` please :)

src/components/Icons/Icon.tsx
```diff
@@ -9,18 +9,74 @@ interface Props {
...
+
+    const hide = (): void => {
+        setHovered(false);
+    };
+
+    const onTap = (): void => {
+        setHovered(true);
+        setTimeout(() => {
+            setHovered(false);
+        }, 2000);
```
TomCusack-Softwire, line 49: can this be a little longer, for longer text? maybe 5000, idk

TomCusack-Softwire (changes requested):
Only minor changes now!

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+        const editData = remapTooltips(row!);
+
+        setModal({
+            data: editData.data,
+            tooltips: editData.unmappedTooltips,
+        });
+    };
+
+    const onDelete = async (index: number): Promise<void> => {
```
SamYap-Softwire, line 101: Why is this an async function?

SamYap-Softwire (commented):
Will get back to this later today, but so far just a few questions for clarification :)

src/app/lists/dataview.tsx
Var1377: So we refetch data so the updated data is reflected in the UI. For some reason, `useRouter().refresh()` doesn't seem to work...

src/app/lists/page.tsx
```diff
@@ -1,10 +1,23 @@
...
+    supabase.channel("lists").subscribe();
```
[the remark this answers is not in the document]
Var1377: I was trying to do that at some point but just resorted to reloading the page on edit. This is now removed

src/components/Tables/TableFilterBar.tsx
Var1377: Switched

src/app/lists/dataview.tsx
Var1377: Fixed, turns out there was one but I wasn't using it

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,199 @@
...
+import Table, { Datum } from "@/components/Tables/Table";
+import React, { useState } from "react";
+import { styled } from "styled-components";
+import EditModal from "@/app/lists/editModal";
+import supabase, { Schema } from "@/supabase";
+import ConfirmDialog from "@/components/Modal/Confirm";
+import Snackbar from "@mui/material/Snackbar/Snackbar";
+import Alert from "@mui/material/Alert/Alert";
+
+export type ListRow = Schema["lists"];
```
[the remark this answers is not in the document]
Var1377: good catch

src/app/lists/dataview.tsx
```diff
@@ -0,0 +1,199 @@
...
+    ["4_quantity", "Family of 4"],
+    ["5_quantity", "Family of 5"],
+    ["6_quantity", "Family of 6"],
+    ["7_quantity", "Family of 7"],
+    ["8_quantity", "Family of 8"],
+    ["9_quantity", "Family of 9"],
+    ["10_quantity", "Family of 10+"],
+];
+
+export const tooltips: [string, string][] = [
```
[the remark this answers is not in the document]
Var1377: For clarity I feel it makes more sense here, although I'm not 100% convinced either way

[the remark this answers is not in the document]
SamYap-Softwire: Same here, I don't mind either way so I think what you have now is fine :)

src/app/lists/Dataview.tsx
```diff
@@ -0,0 +1,195 @@
...
+    const onConfirmDeletion = async (): Promise<void> => {
+        if (toDelete !== null) {
+            const data = rawData[toDelete];
+            const response = await supabase
+                .from("lists")
+                .delete()
+                .match({ primary_key: data.primary_key });
+
+            // this is not a normal js response object, so checking response.error (and using response.error.message) is fine
+            if (response.error) {
```
SamYap-Softwire, line 113: If you're not using the rest of the `response` (i.e. `data`, `status`, ...) you can just do `const { error } = ...` on line 107 instead :)

src/app/lists/Dataview.tsx
```diff
@@ -0,0 +1,195 @@
...
+                    defaultShownHeaders={["item_name", ...toggleableHeaders]}
+                    data={dataAndTooltips}
+                    reorderable
+                    headerFilters={["item_name"]}
+                    pagination={false}
+                    onEdit={onEdit}
+                    onDelete={onDeleteButtonClick}
+                    sortable={false}
+                />
+                <StyledAddButton onClick={() => setModal(null)}>+ Add</StyledAddButton>
```
SamYap-Softwire, line 162: I would prefer just `Add` and not `+ Add` 👉 👈 , but this is your call.

SamYap-Softwire (commented):
Sorry, could only finish reviewing `Dataview.tsx` but will get back to this!

src/app/lists/Dataview.tsx
```diff
@@ -0,0 +1,195 @@
...
+    ["5_notes", "Family of 5"],
+    ["6_notes", "Family of 6"],
+    ["7_notes", "Family of 7"],
+    ["8_notes", "Family of 8"],
+    ["9_notes", "Family of 9"],
+    ["10_notes", "Family of 10+"],
+];
+
+const ListsDataView: React.FC<Props> = ({ data: rawData }) => {
+    // null => add, undefined => modal closed
```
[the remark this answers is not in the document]
SamYap-Softwire: Just saw that this seems similar to the `Props` interface in `EditModal` so maybe you can export that instead? (That also has an `onClose` property which might be unwanted though.

src/app/lists/Dataview.tsx
```diff
@@ -0,0 +1,195 @@
...
+            item_name: row.item_name,
+            primary_key: row.primary_key,
+        };
+        const tooltips: RowData = {};
+        const unmappedTooltips: RowData = {};
+
+        for (const [key, value] of Object.entries(row)) {
+            if (key.endsWith("quantity") && value) {
+                data[key] = value;
+            } else if (key.endsWith("notes") && value) {
```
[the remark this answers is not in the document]
Var1377: We could but I feel this would harm clarity as these are mutually exclusive branches

src/app/lists/Dataview.tsx
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        return {
+            data,
+            tooltips,
+            unmappedTooltips,
+        };
+    };
+
+    const dataAndTooltips = rawData?.map(remapTooltips);
```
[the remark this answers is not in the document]
Var1377: we don't, nice catch

src/app/lists/Dataview.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+    // extract header keys
+    const toggleableHeaders = headers.map(([key]) => key);
+    // removw description header
```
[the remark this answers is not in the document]
Var1377: Yes, it's not immediately obvious what shift is doing

src/app/lists/Dataview.tsx
```diff
@@ -0,0 +1,195 @@
...
+    toggleableHeaders.shift();
+
+    const onEdit = (index: number): void => {
+        const row = rawData[index];
+
+        const editData = remapTooltips(row!);
```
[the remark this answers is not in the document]
Var1377: we don't, nice catch

src/app/lists/Dataview.tsx
Var1377: I could, although I feel that this would harm readability

src/components/Modal/Modal.tsx
```diff
@@ -3,7 +3,8 @@
 import React, { ReactNode } from "react";
 import styled from "styled-components";
 import { Dialog } from "@mui/material";
-import CloseIcon from "@mui/icons-material/Close";
+import { faClose } from "@fortawesome/free-solid-svg-icons";
+import Icon from "../Icons/Icon";
```
SamYap-Softwire, line 7: @ import

src/components/Tables/FilterAccordion.tsx
```diff
@@ -0,0 +1,104 @@
+import React from "react";
+import { Accordion, Checkbox, AccordionSummary, AccordionDetails } from "@mui/material";
+import { styled } from "styled-components";
```
SamYap-Softwire, line 3: curly brackets

src/components/Tables/FilterAccordion.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { faChevronDown } from "@fortawesome/free-solid-svg-icons";
+import Icon from "../Icons/Icon";
```
SamYap-Softwire, line 5: @ import

src/components/Tables/TableFilterBar.tsx
```diff
@@ -1,33 +1,74 @@
 "use client";
 
 import React from "react";
+import { styled } from "styled-components";
```
SamYap-Softwire, line 4: curly brackets

SamYap-Softwire (changes requested):

src/app/lists/EditModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+type HeadersAndTooltips = [[string, string], [string, string]][];
+
+const ModalInner = styled.div`
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+`;
+
+const DataWithTooltipDiv = styled.div`
```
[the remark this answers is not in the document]
Var1377: Not sure, my intuition says they should stay separate as they represent totally different elements but again I'm not convinced either way

src/app/lists/EditModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+        return [header, tooltips[index]];
+    });
+
+    const setKey = (event: React.ChangeEvent<HTMLInputElement>, key: string): void => {
+        const text = event.target.value as string;
+        setToSubmit({ ...toSubmit, [key]: text });
+    };
+
+    const onSubmit = async (): Promise<void> => {
+        let operation;
```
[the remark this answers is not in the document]
Var1377: amended

src/app/lists/page.tsx
```diff
@@ -1,10 +1,22 @@
 import { Metadata } from "next";
-import React from "react";
+import React, { ReactElement } from "react";
+import ListsDataView from "@/app/lists/Dataview";
+import supabase, { Schema } from "@/supabase";
+
+export const revalidate = 0;
```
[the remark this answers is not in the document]
Var1377: yep, it's a next specific option that disables caching

src/app/lists/EditModal.tsx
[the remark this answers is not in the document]
SamYap-Softwire: yeah I'm fine with how it currently is for this.

src/app/lists/Dataview.tsx
SamYap-Softwire, line 1: Can you rename this to ListsDataView to be consistent with the `export default`?

src/app/lists/Dataview.tsx
```diff
@@ -62,9 +62,14 @@ const ListsDataView: React.FC<Props> = ({ data: rawData }) => {
         const unmappedTooltips: RowData = {};
 
         for (const [key, value] of Object.entries(row)) {
-            if (value && key.endsWith("quantity")) {
+            if (!value) {
```
SamYap-Softwire, line 65: Sorry... more opinions.... 
I don't really like `continue` -> what do you think about `if (value)` instead?

src/app/lists/EditModal.tsx
```diff
@@ -43,49 +43,34 @@ const DisplayContents = styled.div`
...
-    const generateDefaultData = (): { [key: string]: string | null } => {
-        if (data !== null && data !== undefined) {
-            return { ...data.data, ...data.tooltips };
-        }
-        return {};
-    };
-
     const [toSubmit, setToSubmit] = useState<{ [key: string]: string | null }>(
-        generateDefaultData()
+        data !== null && data !== undefined ? { ...data.data, ...data.tooltips } : {}
```
SamYap-Softwire, line 47: You can also do `!!data` :)

src/app/lists/EditModal.tsx
```diff
@@ -43,49 +43,34 @@ const DisplayContents = styled.div`
...
-            operation = supabase.from("lists").insert(toSubmit);
-        } else {
-            operation = supabase
-                .from("lists")
-                .update(toSubmit)
-                .eq("primary_key", toSubmit.primary_key)
-                .select();
-        }
+        const table = supabase.from("lists");
+        const operation =
```
SamYap-Softwire, line 65: Just curious: is there a reason for having a `const table` and a `const operation` and then `await operation` instead of just `await supabase.from ...`? Or is it a preference thing?

src/app/lists/page.tsx
```diff
@@ -6,8 +6,8 @@ import supabase, { Schema } from "@/supabase";
 export const revalidate = 0;
 
 const fetchData = async (): Promise<Schema["lists"][]> => {
-    const response = await supabase.from("lists").select("*");
-    return response.data ?? [];
+    const { data } = await supabase.from("lists").select();
```
SamYap-Softwire, line 9: thanks (for this and the rest) !!

SamYap-Softwire (changes requested):
Just reviewed the changes from the last commit - and have a few questions. Will review the entire thing again later :)

src/app/lists/Dataview.tsx
Var1377: You've been outnumbered...

Var1377: Although I was also outnumbered

src/app/lists/EditModal.tsx
SamYap-Softwire: I see, cool :)

src/app/lists/ListDataview.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+const ListsDataView: React.FC<Props> = ({ data: rawData }) => {
+    const [modal, setModal] = useState<EditModalState>();
+    const [toDelete, setToDelete] = useState<number | null>(null);
+    // need another setState otherwise the modal content changes before the close animation finishes
+    const [toDeleteModalOpen, setToDeleteModalOpen] = useState<boolean>(false);
+    const [errorMsg, setErrorMsg] = useState<string | null>(null);
+
+    if (rawData === null) {
+        throw new Error("No data found");
```
SamYap-Softwire, line 53: How is this error handled? Is it catched?

src/components/Icons/SpeechBubbleIcon.tsx
```diff
@@ -1,12 +1,24 @@
+"use client";
```
SamYap-Softwire, line 1: Does this need to be "use client"?

src/components/Tables/FilterAccordion.tsx
```diff
@@ -0,0 +1,104 @@
+import React from "react";
+import { Accordion, Checkbox, AccordionSummary, AccordionDetails } from "@mui/material";
+import styled from "styled-components";
```
SamYap-Softwire, line 3: Might need a "use-client" here :)

SamYap-Softwire (changes requested):
Final comments -> once these are all resolved it'll be approved immediately (by me)!

src/app/lists/ListDataview.tsx
Var1377: It'll show the nextjs error page

SamYap-Softwire (approved):
Looks great!

src/app/lists/EditModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+        flex-direction: row;
+    }
+`;
+
+const DisplayContents = styled.div`
+    display: contents;
+`;
+
+const EditModal: React.FC<Props> = ({ data, onClose }) => {
+    const [toSubmit, setToSubmit] = useState<{ [key: string]: string | null }>(
```
TomCusack-Softwire, line 46: interface this please :)

src/app/lists/dataview.tsx
TomCusack-Softwire: hmmm this may be a sign that a new page can do the editting/adding directly, with the modal only viewing, so navigating back to the data page 'automatically' updates...
but this is a fine solution for the time being!

src/app/lists/EditModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+        data ? { ...data.data, ...data.tooltips } : {}
+    );
+
+    const [errorMsg, setErrorMsg] = useState<string | null>(null);
+
+    const headersThatHaveTooltips = headers.filter(([key]) => key !== "item_name");
+
+    const headersAndTooltips: HeadersAndTooltips = headersThatHaveTooltips.map((header, index) => {
+        return [header, tooltips[index]];
```
TomCusack-Softwire, line 55: FYI you don't _need_ a `{ return ... }` - if you use single brackets then you can just have a value - e.g. `.map((h,i) => [h, t[i]]);` should be sufficient

src/app/lists/EditModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+                            </DataWithTooltipDiv>
+                        </DisplayContents>
+                    );
+                })}
+                <Button variant="contained" color="primary" onClick={onSubmit}>
+                    Submit
+                </Button>
+                <Snackbar
+                    message={errorMsg}
+                    autoHideDuration={3000}
```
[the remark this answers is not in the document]
TomCusack-Softwire: applies to anything with `duration` as well, within reason

TomCusack-Softwire (changes requested):

src/app/clients/page.tsx
```diff
@@ -9,7 +8,7 @@ const dataFetch: () => Promise<Schema["clients"][]> = async () => {
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    const data: Datum[] = await dataFetch();
+    const data = await dataFetch();
```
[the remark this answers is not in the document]
Var1377: This was a result of a find-and-replace to try and make some types more specific. It's not related, but it's still a worthwhile change

src/app/lists/EditModal.tsx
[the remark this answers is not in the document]
Var1377: I'll change it for the icon

src/components/Tables/Table.cy.tsx
```diff
@@ -225,11 +275,55 @@ describe("<Table />", () => {
...
+        cy.mount(
+            <ThemeProvider theme={mockTheme}>
+                <Table data={smallerData} headerKeysAndLabels={headers} checkboxes={false} />
+            </ThemeProvider>
+        );
+
+        cy.get("input[placeholder*='Filter by Name']").should("exist");
+        cy.get("input[placeholder*='Filter by Phone Number']").should("exist");
+        cy.get("div.MuiAccordionSummary-root").contains("Select Columns");
+    });
```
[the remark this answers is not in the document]
Var1377: I'd rather leave themeing tests until the new themes have been merged.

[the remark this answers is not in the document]
TomCusack-Softwire: I'd rather you make 'some' that work on current main, but with the intention that they will be changed 'trivially' by converting the `const mockTheme` into a new format and changing the test a tiny amount, rather than the theme person (Chloe?) having to remember to write a test and having to know all the context

src/app/lists/ListDataview.tsx
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    const remapTooltips = (row: ListRow): Datum & { unmappedTooltips: RowData } => {
```
TomCusack-Softwire, line 56: interface this too!

src/components/Tables/Table.cy.tsx
```diff
@@ -103,96 +109,132 @@ describe("<Table />", () => {
...
+        );
         cy.get("select[aria-label='Rows per page:']").select("15");
         cy.get("input[aria-label='Select row 14']").should("not.be.checked");
 
         cy.get("input[aria-label='Select row 14']").click();
         cy.get("input[aria-label='Select row 14']").should("be.checked");
 
-        cy.get("input[aria-label='Select row 0']").click();
-        cy.get("input[aria-label='Select row 0']").should("be.checked");
+        cy.get("input[aria-label='Select row 9']").click();
```
TomCusack-Softwire, line 218: why change 0 -> 9?

src/components/Tables/Table.tsx
```diff
@@ -89,41 +149,109 @@ const Table: React.FC<Props> = (props) => {
...
-            };
+    const columns: TableColumn<Row>[] = (
+        props.toggleableHeaders ? shownHeaders : props.headerKeysAndLabels
+    ).map(([headerKey, headerName]) => {
+        return {
+            name: headerName,
+            selector: (row) => row.data[headerKey],
+            minWidth: "12rem",
+            maxWidth: "20rem",
+            sortable: props.sortable ?? true,
```
TomCusack-Softwire, line 160: I think I'd prefer setting a default value in props as much as possible, so you can easily inspect what the default is (rather than defaulting to `undefined` and setting _somewhere_ in code!)

src/app/lists/EditModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+import Alert from "@mui/material/Alert/Alert";
+import Button from "@mui/material/Button/Button";
+
+interface Props {
+    onClose: () => void;
+    /** null => add, undefined => modal closed, Datum => edit */
+    data: EditModalState;
+}
+
+// null => add, undefined => modal closed
```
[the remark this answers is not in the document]
Var1377: It had been but @SamYap-Softwire suggested an enum instead, I'll change it back

package-lock.json
```diff
@@ -3361,6 +3362,15 @@
         "tslib": "^2.4.0"
       }
     },
+    "node_modules/@types/cypress": {
```
TomCusack-Softwire, line 3365: can you revert the `package-lock.json` (or regenerate with `npm i`) now you have removed `@types/cypress`?

src/components/Tables/Table.cy.tsx
```diff
@@ -103,74 +123,74 @@ describe("<Table />", () => {
...
 
     it("checkbox select toggles", () => {
-        cy.mount(<Table data={data} headers={headers} />);
-        cy.get("input[aria-label='Select row 0']").should("not.be.checked");
-        cy.get("input[aria-label='Select row 0']").click();
-        cy.get("input[aria-label='Select row 0']").should("be.checked");
-        cy.get("input[aria-label='Select row 0']").click();
-        cy.get("input[aria-label='Select row 0']").should("not.be.checked");
+        cy.mount(<Component />);
+        cy.get("input[aria-label='Select row 2']").should("not.be.checked");
```
TomCusack-Softwire, line 151: should all of these tests have their numbers changed?

src/app/lists/EditModal.tsx
TomCusack-Softwire: Personally disagree, I think this would be better if `await`s are attached to the asynchronous thing they are waiting for - e.g. in this example `data === null ? await table.insert(...) : await table.update(...)`
especially when you put the new lines in like this, it makes it very clear what's going on.
but I'm very happy to keep it like this! it's not an awful solution either way.

src/app/lists/ListDataview.tsx
```diff
@@ -0,0 +1,199 @@
...
+    const [toDelete, setToDelete] = useState<number | null>(null);
+    // need another setState otherwise the modal content changes before the close animation finishes
+    const [toDeleteModalOpen, setToDeleteModalOpen] = useState<boolean>(false);
+    const [errorMsg, setErrorMsg] = useState<string | null>(null);
+
+    if (rawData === null) {
+        throw new Error("No data found");
+    }
+
+    const remapTooltips = (row: ListRow): RemappedData => {
```
TomCusack-Softwire, line 60: are both mapped and unmapped tooltips used in code?

src/app/themes.tsx
```diff
@@ -124,6 +124,25 @@ const StyleManager: React.FC<Props> = ({ children, theme = lightTheme }) => {
                     },
                 },
             },
+            MuiCheckbox: {
```
TomCusack-Softwire, line 127: assuming these are unavoidable? maybe I should separate MUI themes into a separate file... (not on you to do!)

src/components/Modal/Confirm.tsx
```diff
@@ -0,0 +1,41 @@
...
+const Text = styled.p`
+    font-size: 1.5rem;
+`;
+
+const ConfirmDialog: React.FC<Props> = ({ isOpen, message, onCancel, onConfirm }) => {
+    return (
+        <Modal isOpen={isOpen} onClose={onCancel} header="Confirm" headerId="confirmDialog">
+            <ModalInner>
+                <Text>{message}</Text>
+                <Button type="button" onClick={onConfirm} variant="contained" color="primary">
```
TomCusack-Softwire, line 33: is `type="button"` a thing? don't recall seeing it on any other MUI button

src/components/NavBar/NavigationBar.tsx
```diff
@@ -147,10 +147,10 @@ const ResponsiveAppBar: React.FC = () => {
                     </LogoElementContainer>
                     <DesktopButtonContainer>
                         {pages.map(([page, link]) => (
-                            <>
+                            <React.Fragment key={page}>
```
TomCusack-Softwire, line 150: ideally changes completely unrelated would be in a separate PR (with a separate title) - it makes the history really clear and random bits don't change when you do large PRs. Imagine the NavBar stopped working because of this - someone could use Git to find out it's in this PR, but then would have to review a lot of code to find out that is was here where it was changed!
I _much_ prefer lots of short PRs doing 'one semantic chunk' of things. (but not enough to make you change it - just remember this for the future!)

src/components/Tables/FilterAccordion.tsx
```diff
@@ -0,0 +1,106 @@
...
+    height: 3rem;
+    overflow: visible;
+    z-index: 2;
+    margin-bottom: 0.5rem;
+    margin-top: 0.5rem;
+    & .MuiPaper-root {
+        background-color: ${(props) => props.theme.surfaceBackgroundColor};
+        border: 1px solid ${(props) => props.theme.secondaryBackgroundColor};
+        box-shadow: 0 0 1px ${(props) => props.theme.secondaryBackgroundColor};
+        // TODO: VFB16 change to a theme specific grey
```
TomCusack-Softwire, line 27: TODO Format:
`// TODO VFB-XXX Short description which is capitalised`
(sorry, I'm a little particular about this sort of thing!)

TomCusack-Softwire (approved):
Once minor comments are addressed (and make sure they are all done!) then I'm fine for you to merge. Finally this massive ticket will be done! Thanks for the work :)

src/app/lists/ListDataview.tsx
Var1377: Yep, one is used in the onEdit handler so the edit modal can have the unmapped tooltips, while the table component uses the mapped tooltips.

src/app/themes.tsx
Var1377: Turns out these actually did nothing, they have now been removed.

(They were made in an attempt to get the checkboxes to change colour, which we couldn't figure out how to do in an MUI way)

src/components/Modal/Confirm.tsx
Var1377: It's the default, it's now been removed.

src/components/NavBar/NavigationBar.tsx
Var1377: Noted

PR #25, opened 2023-07-20
Add type checking to checks and fixed the bugs found by it

TomCusack-Softwire:
Now GitHub actions should fail if TypeScript doesn't compile :)


src/components/Tables/Table.tsx
```diff
@@ -48,7 +48,8 @@ const dataToRows = (data: Datum[], headers: Headers): Row[] => {
         const row: Row = { rowId: currentIndex };
 
         for (const headerKey of Object.keys(headers)) {
-            row[headerKey] = datum[headerKey] ?? "";
+            const databaseValue = datum[headerKey] ?? "";
+            row[headerKey] = Array.isArray(databaseValue) ? databaseValue.join() : databaseValue;
```
daniel-conlan-softwire, line 52: Should the `.join()` function be replaced with `.join(", ")`? Might look a bit nicer with a space between list elements, but it may not matter too much

daniel-conlan-softwire (approved):
All looks good

src/components/Tables/Table.tsx
paopaosoftwire: Is the new section arises from the database changes? Should this be in a separate PR if it does not relate to type checking? Just afraid that this might effect others that are using the Table component.

TomCusack-Softwire: It maybe should be in another PR, but it does directly relate to the type checking - without this, the type checker fails!
Previously it didn't check types 'that well' so when the database types changed it allowed it - but there is failing code on main! :eyes:

PR #27, opened 2023-07-26
Set up screenshot tests

TomCusack-Softwire:
Shows a picture of exactly what has changed visually since the last main!
Does require you to actively run `npm run test:regenerate_snapshots` if you change anything.
(Will eventually be how we verify different browsers all have identical outputs! But I'm not there yet 😆)


cypress/e2e/PAGES.ts
```diff
@@ -0,0 +1,12 @@
+interface Page {
+    friendly_name: string;
+    url: string;
+}
+
+const PAGES: Page[] = [
+    { friendly_name: "Clients", url: "/clients" },
+    { friendly_name: "Calendar", url: "/calendar" },
+    { friendly_name: "Lists", url: "/lists" },
```
DanWilkins2107, line 9: Do other pages such as add-clients need to be added? Or is that for us to do as we make the pages?

DanWilkins2107 (approved):

cypress/e2e/PAGES.ts
TomCusack-Softwire: both - if you make a new page then you should add to PAGES to add to all tests, but at the moment these are the only pages on main!

Var1377 (approved):

TomCusack-Softwire:
is not generating the same screenshots on run, so setting to draft for now

PR #28, opened 2023-07-31
Vfb 8 components

SamYap-Softwire:
Split the add-clients page into smaller components.


src/components/Form/formFunctions.ts
```diff
@@ -0,0 +1,281 @@
...
+
+type Field = Fields[keyof Fields];
+type OnChange = (event: React.ChangeEvent<HTMLInputElement> | SelectChangeEvent) => void;
+type OnChangeCheckbox = (event: React.ChangeEvent<HTMLInputElement>) => void;
+export type ErrorSetter = (errorKey: string, errorType: Error) => void;
+export type FieldSetter = (fieldKey: string, newFieldValue: Field) => void;
+type PersonType = Database["public"]["Enums"]["gender"];
+type FamilyDatabaseRecord = InsertSchema["families"];
+export type ClientDatabaseRecord = InsertSchema["clients"];
+export enum Error {
```
DanWilkins2107, line 15: What is the benefit of using enum here as opposed to a regular const?

src/components/Form/formStyling.tsx
```diff
@@ -0,0 +1,94 @@
...
+    }
+`;
+
+export const ErrorText = styled(Text)`
+    color: ${(props) => props.theme.errorColor};
+    margin-bottom: 3em;
+    text-align: center;
+`;
+
+export const StyledButton = styled.button`
```
DanWilkins2107, line 62: This is quite a generic name if it's being exported, for example I know I've used similar names throughout stuff I've coded.

src/components/Form/formStyling.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export const Text = styled.p`
+    padding-bottom: 1em;
+`;
+
+export const Subheading = styled.h2`
+    padding-bottom: 1em;
+`;
+
+export const Heading = styled.h1`
```
DanWilkins2107, line 44: Same thing as below for all of these

DanWilkins2107 (approved):
Small Changes but looks good :)

src/components/Form/deliveryInstructionsCard.tsx
```diff
@@ -0,0 +1,22 @@
...
+    errorSetter: ErrorSetter;
+    fieldSetter: FieldSetter;
+}
+
+const deliveryInstructionsCard: React.FC<CardProps> = ({ errorSetter, fieldSetter }) => {
+    return genericFormCard(
+        "Delivery Instructions",
+        false,
+        <FreeFormTextInput
+            label="E.g. The doorbell does not work. Use the door code: xxxx."
```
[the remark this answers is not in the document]
SamYap-Softwire: Yeah, they actually have a prompt "do they have a door code? do they live upstairs and cannot come downstairs?" in their current form, which is why i've included it :)

src/components/Form/formFunctions.ts
SamYap-Softwire: I think it makes more sense to treat the errors as "a group of constants belonging to the `Error` family" rather than "a constant (object/interface) with these properties". 

Other than that, there's not much difference with using enums and consts, except that an enum can be used as a type and the values that each member are initialised to are fixed and different from each other (which I think is suitable for use here as each error is different).

src/components/Form/addressCard.tsx
```diff
@@ -0,0 +1,62 @@
+import FreeFormTextInput from "@/components/DataInput/FreeFormTextInput";
```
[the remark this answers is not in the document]
SamYap-Softwire: Alright, i'll move the specific components to the page, but feel free to move them into `components\Form` :)

src/app/clients/add/addClientForm.tsx
ChloeYiuSoftwire, line 1: Since it is a tsx component, for consistency should we put it as a capitalized start (?) I think the ParcelCalendar.tsx file in the calendar page has the same formatting

src/app/clients/add/formSections/babyProductCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+import GenericFormCard from "@/components/Form/genericFormCard";
+import { selectChangeEventHandler } from "@/components/DataInput/inputHandlerFactories";
+
+interface CardProps {
+    formErrors: FormErrors;
+    errorSetter: ErrorSetter;
+    fieldSetter: FieldSetter;
+    fields: Fields;
+}
+
```
ChloeYiuSoftwire, line 23: I noticed that CardProps seems to be used among all the formSections except with a bit of variation - would it be more efficient if you define one and import it to all these formSections?

```
interface CardProps {
    formErrors?: FormErrors;
    errorSetter?: ErrorSetter;
    fieldSetter: FieldSetter;
    fields?: Fields;
}
```

src/components/Calendar/Calendar.cy.tsx
ChloeYiuSoftwire, line 1: Just to double check if you simply rebase will these appear haha I don't know if this is like Dan's one when he created tonnes of file or is it just normal?

src/app/clients/add/formSections/babyProductCard.tsx
SamYap-Softwire: Yep, I've considered that but I thought that if we make certain fields optional, the necessary props might not be passed in for a specific card. 
Alternatively, we can make all of them required and just pass them in but not use them - but right now eslint doesn't allow unused variables.

src/components/Calendar/Calendar.cy.tsx
SamYap-Softwire: Unfortunately it's not normal TT

DanWilkins2107 (approved):
No more changes to suggest from me :)

src/app/clients/add/AddClientForm.tsx
```diff
@@ -0,0 +1,180 @@
...
+    useEffect(() => {
+        if (fields.numberChildren <= fields.children.length) {
+            fieldSetter("children", fields.children.slice(0, fields.numberChildren));
+            return;
+        }
+
+        const extraChildren: Person[] = Array(fields.numberChildren - fields.children.length)
+            .fill(0)
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            .map((_item) => {
```
daniel-conlan-softwire: Does this `// eslint-disable-next-line @typescript-eslint/no-unused-vars` still need to be here?

src/app/clients/add/AddClientForm.tsx
daniel-conlan-softwire, line 1: This is looking really good by the way - much easier to read

src/app/clients/add/formSections/babyProductCard.tsx
SamYap-Softwire: Have now implemented what you suggested ! 😄

src/app/clients/add/AddClientForm.tsx
```diff
@@ -0,0 +1,194 @@
...
+                return {
+                    personType: "child",
+                    age: null,
+                };
+            });
+        fieldSetter("children", [...fields.children, ...extraChildren]);
+    }, [fields.numberChildren]); // eslint-disable-line react-hooks/exhaustive-deps
+
```
daniel-conlan-softwire, line 119: Same as above, does this `// eslint-disable-line react-hooks/exhaustive-deps` need to be here?

src/components/DataInput/CheckboxGroupInput.tsx
daniel-conlan-softwire, line 1: Thanks for changing this

src/components/Form/formFunctions.ts
```diff
@@ -0,0 +1,181 @@
...
+    feminineProducts: booleanGroup;
+    babyProducts: boolean | null;
+    nappySize: string;
+    petFood: booleanGroup;
+    otherItems: booleanGroup;
+    deliveryInstructions: string;
+    extraInformation: string;
+    attentionFlag: boolean;
+    signpostingCall: boolean;
+}
```
daniel-conlan-softwire, line 69: I don't think these interfaces should be in the general `Form/formFunctions.ts` file as they are rather specific to the `AddClientsForm`. I'd perhaps move them into the `AddClientsForm.tsx` file and make general `Field` and `Error` interfaces in this folder. You can then say that the interfaces defined in the `AddClientsForm.tsx` file `extend` the more general interfaces defined in `formFunctions.ts`

src/components/Form/formFunctions.ts
```diff
@@ -0,0 +1,181 @@
...
+    return (event) => {
+        const input = event.target.value;
+        const errorType = getErrorType(input, required, regex, additionalCondition);
+        errorSetter(key, errorType);
+        if (errorType === Error.none) {
+            const newValue = formattingFunction ? formattingFunction(input) : input;
+            fieldSetter(key, newValue);
+        }
+    };
+};
```
daniel-conlan-softwire, line 125: I would perhaps rename this function to something along the lines of `getFormValueOnChange` (or something like that) just to make it specific to the use-case of within forms explicit. I found that code like `onChange={onChange(...)}` isn't immediately obvious as to what it's doing

src/components/Form/formFunctions.ts
```diff
@@ -0,0 +1,181 @@
...
+
+export const onChangeRadioGroup = (
+    fieldSetter: FieldSetter,
+    key: string
+): selectChangeEventHandler => {
+    return (event) => {
+        const input = event.target.value;
+        fieldSetter(key, input === "Yes");
+    };
+};
```
daniel-conlan-softwire, line 146: Perhaps something similar with these functions as mentioned above

daniel-conlan-softwire (approved):
Looks really good, Sam. Very little to change, just a couple of naming suggestions for functions and where to move some interfaces to.

src/app/clients/add/AddClientForm.tsx
SamYap-Softwire: For some reason it doesn't pass the linting test (both locally & on github) if it's not there - even though it has already been disabled.

src/app/clients/add/AddClientForm.tsx
SamYap-Softwire: Yep, this rule is not disabled globally (and probably shouldn't be?) but in this case we only want the `useEffect `to depend on `fields.numberChildren` even though it uses the other `useStates `as functions.

Var1377 (changes requested):
Looking much better with functionality split up, just a few changes/suggestions left from me

Var1377 (approved):

src/components/Form/formFunctions.ts
```diff
@@ -0,0 +1,156 @@
...
+    required?: boolean,
+    regex?: RegExp,
+    additionalCondition?: (value: string) => boolean
+): Error => {
+    if (required && input === "") {
+        return Error.required;
+    }
+    if (
+        (regex !== undefined && !input.match(regex)) ||
+        (additionalCondition !== undefined && !additionalCondition(input))
```
TomCusack-Softwire, line 75: can you use `!regex` and `!additionalCondition`?

src/components/Form/formFunctions.ts
```diff
@@ -0,0 +1,156 @@
...
+        }
+        if (error === Error.initial) {
+            amendedErrorTypes = {
+                ...amendedErrorTypes,
+                [errorKey]: Error.required,
+            };
+        }
+    }
+    if (errorExists) {
+        errorSetter({ ...amendedErrorTypes });
```
TomCusack-Softwire, line 153: why `({... obj})` over `(obj)`?

TomCusack-Softwire (changes requested):
looks pretty good, just some minor comments left!

src/app/clients/add/formSections/NumberAdultsCard.tsx
```diff
@@ -0,0 +1,86 @@
...
+        } else if (input.match(numberRegex)) {
+            newValue[personIndex].quantity = parseInt(input);
+        } else {
+            newValue[personIndex].quantity = -1;
+        }
+
+        fieldSetter("adults", newValue);
+
+        const invalidAdultEntry = newValue.filter((value) => value.quantity === -1);
+        const nonZeroAdultEntry = newValue.filter((value) => value.quantity! > 0);
```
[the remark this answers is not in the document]
SamYap-Softwire: `quantity` is an optional property in the `Person` interface since `children` have `quantity = undefined` (this is then set to `1` when the form is submitted). However, since we're dealing with `adults` here, their quantity will always be a number and so we can assert that it is never undefined in this case.

src/components/Form/formFunctions.ts
```diff
@@ -0,0 +1,156 @@
...
+import { Database } from "@/database_types_file";
+import {
+    changeEventHandler,
+    selectChangeEventHandler,
+} from "@/components/DataInput/inputHandlerFactories";
+
+type Field = Fields[keyof Fields];
+export type ErrorSetter = (errorKey: string, errorType: Error) => void;
+export type FieldSetter = (fieldKey: string, newFieldValue: Field) => void;
+export type PersonType = Database["public"]["Enums"]["gender"];
```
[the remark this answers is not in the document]
SamYap-Softwire: I'm not sure I get what you mean by simplifying this as `Database["public"]["Enums"]["gender"]` seems like the only way to get to the type: unless you mean _creating_ a new type in `supabase.ts` called `DatabaseEnums` and then calling that `["gender"]`?

src/components/Form/formFunctions.ts
SamYap-Softwire: Would that not mean the opposite? (Since I want it to be `!== undefined` instead of `=== undefined`?

src/app/clients/add/AddClientForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        try {
+            const ids = await insertClient(clientRecord);
+            try {
+                await insertFamily([...fields.adults, ...fields.children], ids.family_id);
+                router.push("/clients");
+            } catch (error) {
+                await deleteFailedInsert(ids.primary_key);
+                throw error;
```
SamYap-Softwire, line 192: Just a quick note here that I ended up using a try inside a try because I prefer this over making `ids` a `let`.

src/app/clients/add/formSections/NumberAdultsCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+    errorText,
+    FieldSetter,
+    numberRegex,
+    Person,
+    PersonType,
+} from "@/components/Form/formFunctions";
+import GenericFormCard from "@/components/Form/GenericFormCard";
+import { selectChangeEventHandler } from "@/components/DataInput/inputHandlerFactories";
+
+const setQuantity = (input: string): number => {
```
TomCusack-Softwire, line 17: not really 'setting' anything - it's parsing / getting
(function naming)

src/components/Form/formFunctions.ts
TomCusack-Softwire: very true, can you do `regex` or `!!regex`?

src/components/Form/formFunctions.ts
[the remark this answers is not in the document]
TomCusack-Softwire: yep, export a new type in `supabase.ts`:
`export type DatabaseEnums = Database["public"]["Enums"]`
and use that here!

src/app/clients/add/AddClientForm.tsx
```diff
@@ -176,20 +177,27 @@ const AddClientForm: React.FC = () => {
...
-        if (!insertSuccess[0]) {
-            setSubmitError(Error.external);
-            setSubmitErrorMessage(insertSuccess[1]);
-            setSubmitDisabled(false);
-            return;
+
+        try {
+            const ids = await insertClient(clientRecord);
+            try {
+                await insertFamily([...fields.adults, ...fields.children], ids.family_id);
```
TomCusack-Softwire, line 188: shame supabase doesn't support 'transactions', I can get something to work via postgres but I think this is fine for now

TomCusack-Softwire (approved):
I'm happy as soon as the database changes go through!

src/components/Form/formFunctions.ts
TomCusack-Softwire: talked in person, this is better

src/database_types_file.ts
```diff
@@ -104,23 +104,20 @@ export interface Database {
         Row: {
           age: number | null
           family_id: string
-          person_type: Database["public"]["Enums"]["gender"]
+          gender: Database["public"]["Enums"]["gender"]
```
TomCusack-Softwire, line 107: this shouldn't be on your branch - rebase main! :)

PR #30, opened 2023-07-31
Add events table and implement assorted database changes

TomCusack-Softwire:
Implements events and parcel information I forgot about 💀


src/database_types_file.ts
```diff
@@ -22,11 +22,13 @@ export interface Database {
           extra_information: string
           family_id: string
           feminine_products: string[]
+          flag_attention: boolean
           full_name: string
           other_items: string[]
           pet_food: string[]
           phone_number: string
           primary_key: string
+          signposting_call: boolean
```
daniel-conlan-softwire, line 31: Should the fields `flag_attention` and `signposting_call` not be for each individual parcel? Also, would field names such as `flagged_for_attention` or `signposting_call_required` be a bit more clear?

src/database_types_file.ts
```diff
@@ -58,14 +62,44 @@ export interface Database {
...
           pet_food?: string[]
           phone_number?: string
           primary_key?: string
+          signposting_call?: boolean
         }
         Relationships: []
       }
+      events: {
+        Row: {
+          event_name: string
```
daniel-conlan-softwire, line 77: I presume `event_name` is the description of a parcel event - for example, "Shopping List Printed"?

daniel-conlan-softwire (approved):
Thanks for this, looks all good

src/database_types_file.ts
TomCusack-Softwire: sure will change

PR #34, opened 2023-08-02
Update families table

TomCusack-Softwire:
By popular request, I've fixed the families table - now the gender is _just_ male/female/other, the age is nullable (but should be set for kids) and the quantity is no longer - add multiple rows!

### DIFF
--- a/src/database_types_file.ts
+++ b/src/database_types_file.ts
@@ -104,23 +104,20 @@ export interface Database {
         Row: {
           age: number | null
           family_id: string
-          person_type: Database["public"]["Enums"]["gender"]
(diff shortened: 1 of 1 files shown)


paopaosoftwire (approved):
YES!

Var1377 (changes requested):
Looking much better, could we get a doc comment explaining what age === null implies?

TomCusack-Softwire:
this file is completely autogenerated - where would you want docstring?

Var1377:
> this file is completely autogenerated - where would you want docstring?

Ah ok I didn't realise

Var1377 (approved):

PR #35, opened 2023-08-02
Add function that checks for congestion charge

TomCusack-Softwire:
callable via `const response = await supabase.functions.invoke("check-congestion-charge", {body: { postcodes: [...] }});`
some of this is auto-generated via supabase!


SamYap-Softwire (approved):

supabase/seed.sql
daniel-conlan-softwire, line 1: I presume that this file is supposed to be empty?

supabase/functions/check-congestion-charge/index.ts
```diff
@@ -0,0 +1,47 @@
...
+        global: {
+            headers: {
+                Authorization: req.headers.get("Authorization")!,
+            },
+        },
+    });
+
+    const { data, error } = await supabase.storage
+        .from("public")
+        .download("public/congestion-zone-postcodes.txt");
```
daniel-conlan-softwire, line 33: Do you think it would be worth it to have this data stored within the database itself for performance in future? I don't think it needs to be done just now, I just wanted to know what you thought

daniel-conlan-softwire: Actually, as long as a set of postcodes are processed in bulk, I think this should be fine

daniel-conlan-softwire (commented):
Looks all good

PR #36, opened 2023-08-03
VFB-20 Implement the status/action bar for the clients page

Var1377:
This implements the bar that allows the user to add events to the database when parcels are selected.

(no longer a draft)

### DIFF
--- a/src/app/clients/ActionBar.cy.tsx
+++ b/src/app/clients/ActionBar.cy.tsx
@@ -1,7 +1,7 @@
 import { Schema } from "@/supabase";
 import ActionBar from "@/app/clients/ActionBar";
 import React from "react";
-import StyleManager from "../themes";
+import StyleManager from "@/app/themes";
 import { LocalizationProvider } from "@mui/x-date-pickers";
 import { AdapterDayjs } from "@mui/x-date-pickers/AdapterDayjs";
 import "dayjs/locale/en-gb";
@@ -30,9 +30,11 @@ describe("Clients - Action Bar", () => {
         );
     };
 
-    it("should render", () => {
+    beforeEach(() => {
         cy.mount(<Component />);
     });
+
+    it("should render", () => {});
 
     [
         ["status", "Request Denied"],
@@ -40,13 +42,11 @@ describe("Clients - Action Bar", () => {
     ].forEach(([type, example]) => {
         describe(`${type} menu`, () => {
             it(`should open the menu when the ${type} button is clicked`, () => {
-                cy.mount(<Component />);
                 cy.get(`#${type}-button`).click();
                 cy.get(`#${type}-menu`).should("exist");
             });
 
             it("should open the modal when an item is selected", () => {
-                cy.mount(<Component />);
                 cy.get(`#${type}-button`).click();
                 cy.get(`#${type}-menu`).should("exist");
                 cy.get(`#${type}-menu`).contains(example).click();
@@ -54,7 +54,6 @@ describe("Clients - Action Bar", () => {
             });
 
             it("should close the modal when the close button is clicked", () => {
-                cy.mount(<Component />);
                 cy.get(`#${type}-button`).click();
                 cy.get(`#${type}-menu`).should("exist");
                 cy.get(`#${type}-menu`).contains(example).click();
@@ -64,7 +63,6 @@ describe("Clients - Action Bar", () => {
             });
 
             it("should have a modal with date and time pickers", () => {
-                cy.mount(<Component />);
                 cy.get(`#${type}-button`).click();
                 cy.get(`#${type}-menu`).should("exist");
                 cy.get(`#${type}-menu`).contains(example).click();
--- a/src/app/clients/ActionBar.cy.tsx
+++ b/src/app/clients/ActionBar.cy.tsx
@@ -20,9 +20,24 @@ describe("Clients - Action Bar", () => {
             parcelId: "123456789",
             requiresFollowUpPhoneCall: false,
         },
+        {
+            addressPostcode: "AB1 aaaa2CD",
+            collectionCentre: "Centraaaae 1",
+            collectionDatetime: new Date().toISOString(),
+            congestionChargeApplies: false,
+            familyCategory: "Familaaaay 1",
+            flaggedForAttention: false,
+            fullName: "John Smaaaaith",
+            lastStatusUpdate: new Date().toISOString(),
+            lastStatus: "Deliveaaaared",
+            packingDatetime: new Date().toISOString(),
+            packingTimeLabel: "10aaaa:00",
+            parcelId: "123456aaaa789",
+            requiresFollowUpPhoneCall: false,
+        },
     ];
 
-    const selectedIndices = [0];
+    const selectedIndices = [0, 1];
 
     const MockActionBar: React.FC<{}> = () => {
         return (
@@ -78,6 +93,26 @@ describe("Clients - Action Bar", () => {
                     minute: "2-digit",
                 });
                 cy.get("input[value='" + timeString + "']").should("exist");
+            });
+
+            it("should have a modal that contains the selected data", () => {
+                cy.get(`#${type}-button`).click();
+                cy.get(`#${type}-menu`).should("exist");
+                cy.get(`#${type}-menu`).contains(example).click();
+                cy.get(`#${type}-modal-header`).should("exist");
+                cy.get(".MuiPaper-root").contains("Date");
+                cy.get(".MuiPaper-root").contains("Time");
+                const dateString = new Date().toLocaleDateString("en-GB");
+                cy.get("input[value='" + dateString + "']").should("exist");
+                const timeString = new Date().toLocaleTimeString("en-GB", {
+                    hour: "2-digit",
+                    minute: "2-digit",
+                });
+                cy.get("input[value='" + timeString + "']").should("exist");
+                selectedIndices.forEach((index) => {
+                    cy.get(".MuiPaper-root").contains(mockData[index].collectionCentre);
+                    cy.get(".MuiPaper-root").contains(mockData[index].fullName);
+                });
             });
         });
     });
--- a/src/app/clients/ActionBar.cy.tsx
+++ b/src/app/clients/ActionBar.cy.tsx
@@ -1,0 +1,33 @@
+import { Schema } from "@/supabase";
+import ActionBar from "@/app/clients/ActionBar";
+
+describe("Clients - Action Bar", () => {
+    const mockData: Schema["clients"][] = [
+        {
+            address_1: "",
+            address_2: "",
+            address_county: "",
+            address_postcode: "",
+            address_town: "",
+            baby_food: true,
+            delivery_instructions: "",
+            dietary_requirements: [],
+            extra_information: "",
+            family_id: "",
+            feminine_products: [],
+            flagged_for_attention: false,
+            full_name: "",
+            other_items: [],
+            pet_food: [],
+            phone_number: "",
+            primary_key: "",
+            signposting_call_required: false,
+        },
+    ];
+
+    const selectedIndices = [0];
+
+    it("should render", () => {
+        cy.mount(<ActionBar data={mockData} selected={selectedIndices} />);
+    });
+});
--- a/src/app/clients/ActionBar.cy.tsx
+++ b/src/app/clients/ActionBar.cy.tsx
@@ -34,88 +34,51 @@ describe("Clients - Action Bar", () => {
         cy.mount(<Component />);
     });
 
-    it("should open the menu when the status button is clicked", () => {
-        cy.mount(<Component />);
-        cy.get("#status-button").click();
-        cy.get("#status-menu").should("exist");
-    });
+    [
+        ["status", "Request Denied"],
+        ["action", "Print Shopping List"],
+    ].forEach(([type, example]) => {
+        describe(`${type} menu`, () => {
+            it(`should open the menu when the ${type} button is clicked`, () => {
+                cy.mount(<Component />);
+                cy.get(`#${type}-button`).click();
+                cy.get(`#${type}-menu`).should("exist");
+            });
 
-    it("should update the button text when statuses are selected", () => {
-        cy.mount(<Component />);
-        cy.get("#status-button").click();
-        cy.get("#status-menu").should("exist");
-        cy.get("#status-menu").contains("Request Denied").click();
-        cy.get("#status-modal-button").contains("Apply 1 status to 1 item");
-        cy.get(".MuiBackdrop-root").trigger("keydown", { key: "Escape", force: true });
-    });
+            it("should open the modal when an item is selected", () => {
+                cy.mount(<Component />);
+                cy.get(`#${type}-button`).click();
+                cy.get(`#${type}-menu`).should("exist");
+                cy.get(`#${type}-menu`).contains(example).click();
+                cy.get(`#${type}-modal-header`).should("exist");
+            });
 
-    it("should open the modal when the modal button is clicked when an item is selected", () => {
-        cy.mount(<Component />);
-        cy.get("#status-button").click();
-        cy.get("#status-menu").should("exist");
-        cy.get("#status-menu").contains("Request Denied").click();
-        cy.get(".MuiBackdrop-root").trigger("keydown", { key: "Escape", force: true });
-        cy.get("#status-modal-button").click();
-        cy.get("#status-modal-header").should("exist");
-    });
+            it("should close the modal when the close button is clicked", () => {
+                cy.mount(<Component />);
+                cy.get(`#${type}-button`).click();
+                cy.get(`#${type}-menu`).should("exist");
+                cy.get(`#${type}-menu`).contains(example).click();
+                cy.get(`#${type}-modal-header`).should("exist");
+                cy.get("[aria-label='Close Button']").click();
+                cy.get(`#${type}-modal-header`).should("not.exist");
+            });
 
-    it("should not open the modal when the modal button is clicked when no items are selected", () => {
-        cy.mount(<Component />);
-        cy.get("#status-modal-button").click();
-        cy.get("#status-modal-header").should("not.exist");
-    });
-
-    it("should contain all of the selected statuses in the modal", () => {
-        const statuses = [
-            "Request Denied",
-            "Pending More Info",
-            "Ready to Dispatch",
-            "Delivery Failed",
-            "Delivery Cancelled",
-        ];
-
-        cy.mount(<Component />);
-        cy.get("#status-button").click();
-        cy.get("#status-menu").should("exist");
-        statuses.forEach((status) => {
-            cy.get("#status-menu").contains(status).click();
+            it("should have a modal with date and time pickers", () => {
+                cy.mount(<Component />);
+                cy.get(`#${type}-button`).click();
+                cy.get(`#${type}-menu`).should("exist");
+                cy.get(`#${type}-menu`).contains(example).click();
+                cy.get(`#${type}-modal-header`).should("exist");
+                cy.get(".MuiPaper-root").contains("Date");
+                cy.get(".MuiPaper-root").contains("Time");
+                const dateString = new Date().toLocaleDateString("en-GB");
+                cy.get("input[value='" + dateString + "']").should("exist");
+                const timeString = new Date().toLocaleTimeString("en-GB", {
+                    hour: "2-digit",
+                    minute: "2-digit",
+                });
+                cy.get("input[value='" + timeString + "']").should("exist");
+            });
         });
-        cy.get(".MuiBackdrop-root").trigger("keydown", { key: "Escape", force: true });
-        cy.get("#status-modal-button").click();
-        cy.get("#status-modal-header").should("exist");
-        statuses.forEach((status) => {
-            cy.get(".MuiPaper-root").contains(status);
-        });
-    });
-
-    it("should close the modal when the close button is clicked", () => {
-        cy.mount(<Component />);
-        cy.get("#status-button").click();
-        cy.get("#status-menu").should("exist");
-        cy.get("#status-menu").contains("Request Denied").click();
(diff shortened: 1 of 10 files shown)


src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,289 @@
...
+                    type="button"
+                    id="action-button"
+                >
+                    Actions
+                </Button>
+            </OuterDiv>
+        </>
+    );
+};
+
```
SamYap-Softwire, line 186: What do you think about splitting this into 2 files here?

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,289 @@
...
+const ModalInner = styled.div`
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+    align-items: stretch;
+`;
+
+const StatusText = styled.p`
+    margin-left: 1rem;
+    border-top: 1px solid darkgrey;
```
SamYap-Softwire, line 213: Is this going to be a colour from `themes.tsx`?

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,289 @@
...
+                </div>
+                <p>To</p>
+                <div>
+                    {props.data.map((parcel, index) => {
+                        return (
+                            <StatusText key={index}>
+                                {parcel.collection_centre}: (
+                                {parcel.collection_datetime ? (
+                                    new Date(parcel.collection_datetime).toLocaleString()
+                                ) : (
```
SamYap-Softwire, line 272: Can we use `&&` here? (only if we don't need the empty fragment)

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,289 @@
...
+                                    <></>
+                                )}
+                                )
+                            </StatusText>
+                        );
+                    })}
+                </div>
+                {props.errorText ? <small>{props.errorText}</small> : <></>}
```
SamYap-Softwire, line 280: `&&` here as well please :)

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,289 @@
...
+const StatusText = styled.p`
+    margin-left: 1rem;
+    border-top: 1px solid darkgrey;
+    padding: 1rem 0;
+    &:last-child {
+        border-bottom: 1px solid darkgrey;
+    }
+`;
+
+const SharedModal: React.FC<SharedModalProps> = (props) => {
```
paopaosoftwire, line 220: Should the SharedModal and ActionBar be in separate files?

src/app/clients/page.tsx
```diff
@@ -9,7 +8,7 @@ const dataFetch: () => Promise<Schema["clients"][]> = async () => {
 };
 
 const Clients: () => Promise<React.ReactElement> = async () => {
-    const data: Datum[] = await dataFetch();
+    const data = await dataFetch();
```
paopaosoftwire, line 11: Why was the type removed?

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+    return (
+        <Modal {...props} header={props.header} headerId={props.headerId}>
+            <ModalInner>
+                <Row>
+                    <p>Date: </p>
+                    <DatePicker
+                        value={date}
+                        defaultValue={date}
+                        onChange={(newDate) =>
```
paopaosoftwire, line 235: Is it worth putting these onChange function outside of the component? The function is quite big but its only used once so its up to you:)

paopaosoftwire (changes requested):

src/app/clients/ActionBar.cy.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    const selectedIndices = [0];
+
+    const Component: React.FC<{}> = () => {
+        return (
+            <LocalizationProvider adapterLocale="en-gb" dateAdapter={AdapterDayjs}>
+                <StyleManager>
+                    <ActionBar data={mockData} selected={selectedIndices} />
+                </StyleManager>
+            </LocalizationProvider>
```
DanWilkins2107, line 29: Shouldn't the localization provider already be added in themes? or is it different because its a cypress file?

src/app/clients/ActionBar.cy.tsx
```diff
@@ -0,0 +1,84 @@
...
+            collection_centre: "Clapham",
+            collection_datetime: null,
+            voucher_number: null,
+            packing_datetime: null,
+        },
+    ];
+
+    const selectedIndices = [0];
+
+    const Component: React.FC<{}> = () => {
```
DanWilkins2107, line 23: Could have a better name than component, but also that is me being pedantic lol

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,289 @@
...
+import dayjs, { Dayjs } from "dayjs";
+
+interface Props {
+    selected: number[];
+    data: Schema["parcels"][];
+}
+
+const statuses = [
+    "Request Denied",
+    // probably shouldn't be a status (?)
```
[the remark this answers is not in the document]
Var1377: No Status

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,289 @@
...
+    "No Status",
```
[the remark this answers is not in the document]
Var1377: No clue, I just copied the original site.

src/app/clients/page.tsx
Var1377: type inference gives it a more specific type and results in no type errors, so having no explicit type would be preferred.

src/app/clients/ActionBar.tsx
DanWilkins2107: I would agree with this tbf

DanWilkins2107 (changes requested):

src/app/clients/ActionBar.cy.tsx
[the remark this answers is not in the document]
Var1377: We don't but @TomCusack-Softwire says we should keep it there specifically for React.FC

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,289 @@
...
+    gap: 1rem;
+    border-radius: 0.5rem;
+    background-color: ${(props) => props.theme.surfaceBackgroundColor};
+    margin: 1rem;
+`;
+
+const ActionBar: React.FC<Props> = ({ selected, data }) => {
+    const selectedData = Array.from(selected.map((index) => data[index]));
+
+    const [statusAnchorEl, setStatusAnchorEl] = React.useState<null | HTMLElement>(null);
```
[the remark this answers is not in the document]
Var1377: good catch

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,186 @@
...
+                    setModalError(null);
+                }}
+                data={selectedData}
+                status={selectedAction}
+                header="Perform Action"
+                headerId="action-modal-header"
+                onSubmit={submitAction}
+                errorText={modalError}
+            />
+            <Menu
```
paopaosoftwire, line 120: Seems like this section and the section below is very similar, would it be better to have a component? or at least have the onClick function separated out.

paopaosoftwire (commented):
Looking good!

src/app/clients/ActionBar.tsx
Var1377: My instinct would suggest that it would be better to keep it inline although other people's comments/opinions are welcome

paopaosoftwire: I get what you mean. Happy to keep it as it is!

paopaosoftwire (commented):
Looks good. Just some questions and suggestions!

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+        if (error) {
+            setModalError(error.message);
+        } else {
+            setStatusModal(false);
+            setModalError(null);
+        }
+    };
+
+    const submitAction = async (): Promise<void> => {};
```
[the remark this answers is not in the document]
Var1377: For whenever we implement submitting actions. I'll add a TODO

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+interface Props {
+    selected: number[];
+    data: Schema["parcels"][];
+}
+
+const statuses = [
+    "Request Denied",
+    // probably shouldn't be a status (?)
+    "No Status",
```
DanWilkins2107, line 21: Idk if this would make more sense as `"Reset Status"` if that is what is meant to be? If `No Status` is what it's meant to be that's fine though

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,188 @@
...
+    "Delete Request",
+];
+
+const OuterDiv = styled.div`
+    display: flex;
+    padding: 1rem;
+    gap: 1rem;
+    border-radius: 0.5rem;
+    background-color: ${(props) => props.theme.surfaceBackgroundColor};
+    margin: 1rem;
```
DanWilkins2107, line 50: Why use both padding and margin? Just cos surely `padding: 2rem` would have the same effect. Or is this just conventional to use both?

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,188 @@
...
+`;
+
+const ActionBar: React.FC<Props> = ({ selected, data }) => {
+    const selectedData = Array.from(selected.map((index) => data[index]));
+
+    const [statusAnchorEl, setStatusAnchorEl] = useState<null | HTMLElement>(null);
```
DanWilkins2107, line 56: Is it necessary to shorten this to El? Like I'm guessing it means Element but I don't see why you wouldn't just have it as `statusAnchorElement` for clarity.

src/app/clients/ActionBarModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+    gap: 1rem;
+    align-items: stretch;
+`;
+
+const StatusText = styled.p`
+    margin-left: 1rem;
+    border-top: 1px solid darkgrey;
+    padding: 1rem 0;
+    &:last-child {
+        border-bottom: 1px solid darkgrey;
```
DanWilkins2107, line 38: Can these be colours from the theme?

DanWilkins2107 (changes requested):
Some small stylistic changes to make, but fine other than that.

src/app/clients/ActionBar.tsx
Var1377: "No Status" was copied from their current implementation, the comment was an expression of my own confusion

src/app/clients/ActionBar.tsx
Var1377: They aren't the same with a `background-color` or `background-image`, the decoration is applied to the padding area but not the margin area.

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    const [statusModal, setStatusModal] = useState(false);
+    const [actionModal, setActionModal] = useState(false);
+
+    const [modalError, setModalError] = useState<string | null>(null);
+
+    const submitStatus = async (date: Dayjs): Promise<void> => {
+        const toInsert = selectedData
+            .map((parcel) => {
+                const event_name = selectedStatus!;
```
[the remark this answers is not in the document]
Var1377: muscle memory...

nice catch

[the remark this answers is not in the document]
Var1377: Nevermind, it's because those are the same names of the database columns

src/app/clients/ActionBarModal.tsx
Var1377: I'll add a TODO for VFB-16

src/app/clients/ActionBar.tsx
```diff
@@ -142,9 +142,9 @@ const ActionBar: React.FC<Props> = ({ selected, data }) => {
                 </MenuList>
             </Menu>
             <Menu
-                open={actionAnchorEl !== null}
-                onClose={() => setActionAnchorEl(null)}
```
DanWilkins2107: Try to fix the navBar taking up the whole screen on open and close :)

DanWilkins2107 (changes requested):

src/app/clients/ActionBar.tsx
Var1377: I can't reproduce this on any browser under any viewport size...

paopaosoftwire (approved):
This looks fine now! Great stuff. I see some comments for VFB16 but thats already merged so I think that should be removed if you rebase main. And I think Tom doesn't like TODOs with a colon 😂. But I'll approve it :)

src/app/clients/ActionBar.cy.tsx
```diff
@@ -0,0 +1,84 @@
+import ActionBar from "@/app/clients/ActionBar";
+import React from "react";
+import StyleManager from "@/app/themes";
+import { ClientsTableRow } from "./getClientsTableData";
```
SamYap-Softwire, line 4: @ import please :)

src/app/clients/ActionBar.tsx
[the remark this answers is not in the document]
SamYap-Softwire: might be worth looking into it?

src/app/clients/ActionBar.tsx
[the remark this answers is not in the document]
SamYap-Softwire: Yea I don't know if it should be a status 🤷 (not a very helpful comment ...)

src/app/clients/ActionBar.tsx
```diff
@@ -0,0 +1,193 @@
...
+        if (error) {
+            setModalError(error.message);
+        } else {
+            setStatusModal(false);
+            setModalError(null);
+        }
+    };
+
+    const submitAction = async (): Promise<void> => {
+        // TODO: implement whatever actions are meant to do
```
SamYap-Softwire, line 96: Is this a note to self or for a future ticket?

src/app/clients/ActionBarModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+`;
+
+const ModalInner = styled.div`
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+    align-items: stretch;
+`;
+
+// TODO VFB-16 Make these border colours change on dark and light themes
```
SamYap-Softwire, line 33: Now that themes has been merged into main, can you rebase and update this?

src/app/clients/ActionBarModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+                <Row>
+                    <p>Time: </p>
+                    <TimePicker value={date} onChange={onTimeChange} disableFuture />
+                </Row>
+                <p>Applying:</p>
+                <div>
+                    <StatusText key={props.status}>{props.status}</StatusText>
+                </div>
+                <p>To</p>
+                <div>
```
SamYap-Softwire, line 86: what's the difference between a `<div></div>` and a `<></>`?

src/app/clients/ActionBarModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+            date
+                .set("hour", newDate?.hour() ?? date.hour())
+                .set("minute", newDate?.minute() ?? date.minute())
+        );
+
+    return (
+        <Modal {...props} header={props.header} headerId={props.headerId}>
+            <ModalInner>
+                <Row>
+                    <p>Date: </p>
```
SamYap-Softwire, line 69: Can this work without a `<p>` tag? (is it `<p>` by default?)

src/app/clients/ClientsPage.tsx
```diff
@@ -19,6 +19,7 @@ import { Schema } from "@/supabase";
 import TableSurface from "@/components/Tables/TableSurface";
 import { CenterComponent } from "@/components/Form/formStyling";
 import Button from "@mui/material/Button";
+import ActionBar from "./ActionBar";
```
SamYap-Softwire, line 22: @ import

src/components/Tables/Table.tsx
```diff
@@ -151,10 +153,19 @@ const Table: React.FC<Props> = ({
 
     const [filterText, setFilterText] = useState<FilterText>({});
 
-    const [selectCheckBoxes, setSelectCheckBoxes] = useState(
+    const [selectCheckBoxes, _setSelectCheckBoxes] = useState(
         new Array<boolean>(data.length).fill(false)
     );
 
+    const setSelectCheckBoxes = (newSelection: boolean[]): void => {
```
SamYap-Softwire, line 160: (continuing from the previous comment...) and rename this?

src/components/Tables/Table.tsx
```diff
@@ -151,10 +153,19 @@ const Table: React.FC<Props> = ({
 
     const [filterText, setFilterText] = useState<FilterText>({});
 
-    const [selectCheckBoxes, setSelectCheckBoxes] = useState(
+    const [selectCheckBoxes, _setSelectCheckBoxes] = useState(
```
SamYap-Softwire, line 156: can we keep this as `setSelectCheckBoxes` because I think we've set underscore at the start of variable names to mean unused variable.

SamYap-Softwire (changes requested):
Minor changes

src/app/clients/ActionBarModal.tsx
Var1377: By having it as an element, it affects the flex layout (specifically where the flex gap appears)

src/components/Tables/Table.tsx
DanWilkins2107: Also generally why would you rename it (like is there a reason why this is stylistically better)

src/components/Tables/Table.tsx
```diff
@@ -164,7 +175,8 @@ const Table: React.FC<Props> = ({
     };
 
     const toggleAllCheckBox = (): void => {
-        setSelectCheckBoxes(new Array<boolean>(data.length).fill(!selectAllCheckBox));
+        const newSelection = new Array<boolean>(data.length).fill(!selectAllCheckBox);
+        setSelectCheckBoxes(newSelection);
```
DanWilkins2107, line 179: Is this just a new function? Or is it the same as the one with the underscore at the front?

DanWilkins2107 (changes requested):

src/app/clients/ActionBarModal.tsx
Var1377: It can, they've been removed although I'm not sure if it's now better or not

SamYap-Softwire: i prefer it this way (removed) so thanks :)

SamYap-Softwire (approved):
looks good!

DanWilkins2107 (approved):

src/app/clients/ActionBar.cy.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+            it("should have a modal with date and time pickers", () => {
+                cy.get(`#${type}-button`).click();
+                cy.get(`#${type}-menu`).should("exist");
+                cy.get(`#${type}-menu`).contains(example).click();
+                cy.get(`#${type}-modal-header`).should("exist");
+                cy.get(".MuiPaper-root").contains("Date");
+                cy.get(".MuiPaper-root").contains("Time");
+                const dateString = new Date().toLocaleDateString("en-GB");
+                cy.get("input[value='" + dateString + "']").should("exist");
```
TomCusack-Softwire, line 90: can these all be template strings - `` `input[value='${dateString}']` ``

src/app/clients/ActionBarModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+    margin-left: 1rem;
+    border-top: 1px solid darkgrey;
+    padding: 1rem 0;
+    &:last-child {
+        border-bottom: 1px solid darkgrey;
+    }
+`;
+
+const ActionBarModal: React.FC<SharedModalProps> = (props) => {
+    const [date, setDate] = useState(dayjs(Date()));
```
TomCusack-Softwire, line 43: do you want `new Date()` when calling each of these Date constructors?

src/app/clients/ClientsPage.tsx
```diff
@@ -166,6 +168,7 @@ const ClientsPage: React.FC<Props> = (props) => {
                     columnStyleOptions={clientTableColumnStyleOptions}
                     onRowClick={onClientTableRowClick}
                     checkboxes={true}
+                    onRowSelection={setSelected}
```
TomCusack-Softwire, line 171: how does this work with pagination? can you select rows 1 and 20 if only 10 are shown at once?

src/app/themes.tsx
```diff
@@ -185,11 +188,13 @@ const StyleManager: React.FC<Props> = ({ children, theme = lightTheme }) => {
         );
 
     return (
-        <ThemeUpdateContext.Provider value={handleThemeChange}>
-            <ThemeProvider theme={chosenTheme}>
-                <MaterialAndGlobalStyle>{themedChildren}</MaterialAndGlobalStyle>
-            </ThemeProvider>
-        </ThemeUpdateContext.Provider>
+        <LocalizationProvider dateAdapter={AdapterDayjs} adapterLocale="en-gb">
```
TomCusack-Softwire, line 191: don't think this should be in `themes.tsx` but instead `layout.tsx`

TomCusack-Softwire (approved):

src/app/clients/ClientsPage.tsx
Var1377: With the current implementation of pagination it works fine (it uses rowIdx which is always unique even between pages) but this might break once @paopaosoftwire changes how pagination works

PR #39, opened 2023-08-07
Fix problems on deployed main

TomCusack-Softwire:
- Add a content div for the nav bar, so content _must_ be underneath the bar (instead of underneath, which only happens on prod??)
- try and patch a couple of security vulnerabilities


paopaosoftwire (approved):
Looking good.

DanWilkins2107 (approved):
Looks fine to me

PR #40, opened 2023-08-07
Fix e2e testing and (hopefully) all routing issues

Var1377:
Hopefully this works, it also fixes login redirect and sign-out redirects


Var1377:
Fixes auth flow but breaks all the tests...

cypress/e2e/auth.cy.ts
```diff
@@ -6,7 +6,7 @@ describe("Authentication tests", () => {
         }
     });
 
-    it("Get to clients page", () => {
+    it.only("Get to clients page", () => {
```
DanWilkins2107, line 9: Should this now be removed?

DanWilkins2107 (approved):

src/components/NavBar/SignOutButton.tsx
```diff
@@ -1,21 +1,22 @@
+"use client";
```
SamYap-Softwire, line 1: Why do we need "use client" here?

src/app/layout.tsx
```diff
@@ -3,18 +3,21 @@ import { Metadata } from "next";
 
 import React from "react";
 import NavigationBar from "@/components/NavBar/NavigationBar";
+import { AuthRouting } from "@/app/auth";
 
 interface Props {
-    children: React.ReactElement;
+    children: React.ReactNode;
```
SamYap-Softwire, line 9: Can you explain why this is `React.ReactNode` instead of `React.ReactElement`?

SamYap-Softwire (commented):
Seems good. Will approve once my questions are answered :)

src/components/NavBar/SignOutButton.tsx
Var1377: I think we need it for `createClientComponentClient`, I'm not totally sure though I'll have a look into it

src/app/layout.tsx
Var1377: React.ReactNode is the type we should be using for children. React.ReactElement means we can't use strings/numbers etc. and is causing unneccessary fragments across the codebase

cypress/e2e/accessibility.cy.ts
```diff
@@ -29,11 +29,11 @@ describe("Accessibility tests in light mode", () => {
...
     it("Checks login page", () => {
-        cy.visit("/clients");
+        cy.visit("/");
         cy.get("button[aria-label='Sign Out Button']").click();
-        cy.url().should("include", "/login");
-        // Forces Page to Render Fully
-        cy.contains("Sign in").should("be.visible");
+        cy.visit("/login");
+
+        cy.get("#login-panel[data-loaded='true']", { timeout: 10000 }).should("exist");
```
SamYap-Softwire, line 35: What does `data-loaded="true"` check for?

cypress/e2e/auth.cy.ts
```diff
@@ -20,4 +43,19 @@ describe("Authentication tests", () => {
 
         cy.url().should("include", "/login");
```
SamYap-Softwire, line 44: Is it worth checking that `cy.visit("/clients")` actually redirects to clients page before clicking the sign out button? 
(in case it stays on the login page despite `cy.visit("/clients")`)

cypress/e2e/auth.cy.ts
```diff
@@ -20,4 +43,19 @@ describe("Authentication tests", () => {
...
     });
+
+    it("CSR Redirected to login page", () => {
+        cy.visit("/login");
+
+        // wait for hydration
+        cy.get("#login-panel[data-loaded='true']", extendedTimeout).should("exist");
+
+        for (const url of paths) {
```
SamYap-Softwire, line 53: Is this in a sign out state initially?

daniel-conlan-softwire (changes requested):
Looks good in general, just a small bug in the dark mode accessibility tests I believe

cypress/e2e/accessibility.cy.ts
```diff
@@ -67,19 +65,18 @@ describe("Accessibility tests in dark mode", () => {
...
     it("Checks login page", () => {
-        cy.visit("/clients");
-        cy.get("label[aria-label='Theme Switch']").click();
-        cy.get("button[aria-label='Sign Out Button']").click();
-        cy.url().should("include", "/login");
-        // Forces Page to Render Fully
-        cy.contains("Sign in").should("be.visible");
+        cy.visit("/login");
+
+        // wait for hydration
```
SamYap-Softwire, line 78: Do we need `cy.get("label[aria-label='Theme Switch']").click();` here?

cypress/support/commands.ts
```diff
@@ -14,23 +14,44 @@ import { Result } from "axe-core";
...
 
+const loginWithRetry = (iteration: number = 0): void => {
+    if (iteration >= 5) {
+        // This only ever seems to happen on GH Actions on the first test, so we can just log and move on
+        // Redirects are tested by the auth spec
+        cy.log("Login redirect failed");
+        return;
+    }
+
+    cy.get("h1").then(($h1) => {
```
SamYap-Softwire, line 25: Can you explain what the `$` is for?

cypress/support/commands.ts
```diff
@@ -14,23 +14,44 @@ import { Result } from "axe-core";
 import "@cypress/code-coverage/support";
 import "cypress-axe";
 
+const loginWithRetry = (iteration: number = 0): void => {
+    if (iteration >= 5) {
```
SamYap-Softwire, line 18: Why 5?

cypress.config.ts
```diff
@@ -28,11 +28,11 @@ export default defineConfig({
             return config;
         },
         baseUrl: "http://localhost:3200",
-        video: false,
+        video: true,
         videoUploadOnPasses: false,
-        screenshotOnRunFailure: false,
+        screenshotOnRunFailure: true,
+        videoCompression: 1,
```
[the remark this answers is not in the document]
Var1377: So we know why tests fail. You can get to them by clicking on checks and scrolling down to artifacts.

src/components/DataInput/DataInput.cy.tsx
```diff
@@ -167,9 +167,13 @@ describe("Data Input Components", () => {
             cy.get("li[data-value='a']").click();
             cy.get("@onChangeSpy").should("have.been.calledWith", "a");
 
+            cy.get("div[class^='MuiSelect']").click();
```
SamYap-Softwire, line 170: Why do we need this?

src/components/LoginPanel.tsx
```diff
@@ -37,8 +37,14 @@ const LoginPanel: React.FC<{}> = () => {
     const supabase = createClientComponentClient<DatabaseAutoType>();
     const theme = useTheme();
 
+    const [loaded, setLoaded] = React.useState(false);
+
+    useEffect(() => {
+        setLoaded(true);
```
SamYap-Softwire, line 43: Why do we need this? Is it different from just setting `data-loaded={true}`?

SamYap-Softwire (changes requested):
Sorry, quite a few things that I don't understand and need clarification for before I approve this. 👉 👈

cypress/e2e/auth.cy.ts
```diff
@@ -1,11 +1,34 @@
 describe("Authentication tests", () => {
+    const email: string = Cypress.env("TEST_USER");
+    const password: string = Cypress.env("TEST_PASS");
+
+    const extendedTimeout = { timeout: 10000 };
```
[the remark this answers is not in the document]
Var1377: The issues between locally testing and testing on actions are partially down to the performance difference, especially in terms of compiling the page on very first load and in terms of time until hydration. This just accounts for this.

cypress/e2e/accessibility.cy.ts
Var1377: `useEffect` only runs after hydration, at which point the data-loaded attribute is set to true. With a long timeout, this has the effect of waiting until hydration happens, and means we don't have to wait an arbitrary amount of time and hoping that it'll have hydrated by then.

cypress/support/commands.ts
Var1377: This is a jquery idiom (cypress uses it under the hood). I'm not sure if we want to keep them or not, but this is generally what jquery code looks like.

https://jquery.com/
(jquery was a very commonly used library that is now largely redundant due to the improvement of browsers)

cypress/support/commands.ts
Var1377: It was arbitrary, I just wanted it to happen within 10 seconds although 0.5 + 1 + 2 + 4 + 8 > 10 so I did my maths wrong ;-;

src/components/DataInput/DataInput.cy.tsx
Var1377: This test breaks on actions about 1/8 times, adding a redundant get statement adds some more time and hopefully means reliability improves.

src/components/LoginPanel.tsx
Var1377: data-loaded={true} would mean on a server-side-rendered document, it would still be true. The key thing here is that `useEffect` only runs after hydration, so `loaded === true` implies that hydration has happened and themeing + routing infrastructure should be in place.

src/components/DataInput/DataInput.cy.tsx
SamYap-Softwire: Is there no other way to do this?

cypress/support/commands.ts
SamYap-Softwire: I think I prefer the variable naming to be clearer and without the `$` sign.

cypress/support/commands.ts
```diff
@@ -69,3 +90,14 @@ Cypress.Commands.add("checkColorContrast", () => {
     cy.injectAxe();
     cy.checkA11y(undefined, { runOnly: { type: "tag", values: ["wcag2aa"] } }, terminalLog);
```
DanWilkins2107, line 91: How come we're not using `cy.checkAccessibility`

src/components/DataInput/DataInput.cy.tsx
```diff
@@ -167,9 +167,13 @@ describe("Data Input Components", () => {
...
+
             cy.get("li[data-value='b']").click();
             cy.get("@onChangeSpy").should("have.been.calledWith", "b");
 
+            cy.get("div[class^='MuiSelect']").click();
```
DanWilkins2107, line 175: So my understanding of this is wrong I think, but I thought this div targeted like the header bit of the dropdown menu. If this is the case, would this not open and shut the menu? Does seem to work though so I think I'm wrong.

DanWilkins2107 (approved):
A few questions, don't see any major issues though

cypress/e2e/accessibility.cy.ts
```diff
@@ -1,64 +1,62 @@
 describe("Accessibility tests in light mode", () => {
-    beforeEach(() => {
-        cy.login();
-    });
-
     it("Checks clients page", () => {
+        cy.login();
```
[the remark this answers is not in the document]
Var1377: Because we don't want it to run when checking the login page and it's better than logging in and the signing out as that takes more time

cypress/support/commands.ts
Var1377: It's used in accessibility.cy.ts

src/components/DataInput/DataInput.cy.tsx
Var1377: There probably is, but I'm personally happy with this solution as it's concern is that the `Change handler for dropdown works`. The current issue is an asynchronously fired DOM change and not an issue with the dropdown. Feel free to experiment around with another solution, although it's tough to know what works when the failure happens so infrequently

Var1377: I also don't know if it's actually a solution, it's tough to tell but I haven't had any test failures on this specific test since I added this.

Var1377: Just experimented with the dropdown, and when you click a list item it closes. Not sure how the other test was passing but this is how the interaction would happen

src/components/NavBar/SignOutButton.tsx
Var1377: Yep we do :)

daniel-conlan-softwire (approved):

cypress.config.ts
TomCusack-Softwire, line 34: did you figure out how to set a really small retention period? (e.g. delete after a day)

cypress/e2e/auth.cy.ts
```diff
@@ -1,11 +1,35 @@
 describe("Authentication tests", () => {
+    const email: string = Cypress.env("TEST_USER");
```
TomCusack-Softwire, line 2: is `: string` really needed? can this just be automatically inferred (potentially with a `!` in case of null)?

cypress/support/commands.ts
```diff
@@ -14,23 +14,44 @@ import { Result } from "axe-core";
 import "@cypress/code-coverage/support";
 import "cypress-axe";
 
+const loginWithRetry = (iteration: number = 0): void => {
+    if (iteration >= 4) {
+        // This only ever seems to happen on GH Actions on the first test, so we can just log and move on
+        // Redirects are tested by the auth spec
+        cy.log("Login redirect failed");
```
TomCusack-Softwire, line 21: I see this being very easily missed - is there a way to make it a warning?

cypress/support/commands.ts
```diff
@@ -14,23 +14,44 @@ import { Result } from "axe-core";
...
+        return;
+    }
+
+    cy.get("h1").then((heading) => {
+        if (heading.text().includes("Login")) {
+            cy.get("button[type='submit']").then((submitButton) => {
+                if (submitButton.not(":disabled")) {
+                    submitButton.trigger("click");
```
TomCusack-Softwire, line 29: is this the same as `submitButton.click()`?

cypress/e2e/auth.cy.ts
```diff
@@ -16,8 +40,24 @@ describe("Authentication tests", () => {
     it("Sign out", () => {
         cy.login();
         cy.visit("/clients");
+        cy.url().should("include", "/clients");
         cy.get("button[aria-label='Sign Out Button']").click();
 
         cy.url().should("include", "/login");
     });
+
+    it("CSR Redirected to login page", () => {
```
TomCusack-Softwire, line 49: is this not the same as `Redirected to login page` tests? especially if we don't start off in a logged in state

cypress/e2e/auth.cy.ts
```diff
@@ -1,11 +1,35 @@
...
 
+    it("Redirected to clients after login", () => {
+        cy.visit("/login");
+
+        cy.get("#login-panel[data-loaded='true']", extendedTimeout).should("exist");
+
+        cy.get("input#email").type(email);
+        cy.get("input#password").type(password);
+
+        cy.get("#email").should("have.value", email);
```
TomCusack-Softwire, line 25: why sometimes `#email` and sometimes `input#email`?

TomCusack-Softwire: can this be 'better' semantically? e.g. testing for the thing that has 'email' in the label - idk what you can access

src/app/auth.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const [loggedIn, setLoggedIn] = React.useState<boolean>();
+
+    // This fixes issues with navigation when the user has signed in but then signed out and caching still allows them to
+    // access pages. It also fixes authentication flow issues on sign in and sign out
+    const toRedirectTo = (): string | null => {
+        if (loggedIn === undefined) {
+            return null;
+        }
+
+        if (loggedIn) {
```
TomCusack-Softwire, line 21: can these be combined to `if (loggedIn && ...) return; if (!loggedIn && ...) return;`? I hate else

src/app/auth.tsx
```diff
@@ -0,0 +1,66 @@
+"use client";
```
TomCusack-Softwire, line 1: are there any issues with doing auth 'client side' - can a malicious client control this? assuming not?

TomCusack-Softwire (changes requested):

cypress.config.ts
Var1377: yep, it's what the `retention-days` key does in `checks.yml`

cypress/e2e/auth.cy.ts
Var1377: It's inferred as any without it (which is fine, it's just nice to have everything be typed)

cypress/support/commands.ts
Var1377: No, if the button is disabled and we try `submitButton.click()`, it would result in a test failure.

cypress/e2e/auth.cy.ts
Var1377: No, because we wait for hydration and click `Link`s.

This ensures it uses client side routing and covers our bases better. With how it's currently implemented there shouldn't be a difference, but having the test ensures it stays that way.

src/app/auth.tsx
Var1377: Middleware and database level security still protect us from those without authentication. The one thing that concerns me about a malicious client is whether sensitive data remains in cache after an authenticated user has logged out although I'm not sure if that's something we should be concerned with.

src/app/auth.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    // This fixes issues with navigation when the user has signed in but then signed out and caching still allows them to
+    // access pages. It also fixes authentication flow issues on sign in and sign out
+    const toRedirectTo = (): string | null => {
+        if (loggedIn === undefined) {
+            return null;
+        }
+
+        if (loggedIn) {
+            return window.location.pathname.startsWith("/login") ? "/clients" : null;
```
TomCusack-Softwire, line 22: I'd prefer 
```ts
if (loggedIn && ...("/login")) {
    return "/clients";
}
if (!loggedIn && ... "/login") {
    return "/login";
}
return null;
```
just reads a _lot_ better in my opinion

TomCusack-Softwire: very similar to how I wrote: https://github.com/VauxhallFoodbank/vauxhall-foodbank/blob/a6347c23f729be7903dd0192438f3214795a0dfa/src/middleware.ts#L13

TomCusack-Softwire (approved):

cypress/support/commands.ts
Var1377: submitButton is a jquery element, and `click` is deprecated in jquery in favour of `.trigger("click")`

I could also do `cy.wrap(submitButton).click()` if you prefer that?

cypress/support/commands.ts
Var1377: Unfortunarely cy.warn isn't a thing, it's either this or console.warn but the latter isn't guaranteed to be displayed.

PR #43, opened 2023-08-09
VFB-19 remove old pdf

paopaosoftwire:
As mentioned in the standup meeting. I have decided to change the way we will be implementing the pdf pages from using jsPDF to react-pdf/renderer.

The differences between the two is:

jsPDF: Use to save a currently rendered react component to a pdf. This is useful if we want to print out current components eg table or modal. However, this means that any responsive changes to screen size will be printed out as well and a preview of the pdf is needed.

react-pdf/renderer: Define how the pdf should look using the supplied specific components and styling from the library and downloads a pdf when the button is clicked. This is useful when we want to generate a pdf without the need of previewing or existing react components. The screen-size will not affect how the pdf is saved.

All I have done in this PR is removing the old functionality of jsPDF and npm installed react-pdf/renderer.

refs:
https://www.npmjs.com/package/jspdf
https://react-pdf.org/


DanWilkins2107 (approved):

DanWilkins2107 (approved):
Just checking, all the bits are removed where the PDF stuff is used?

paopaosoftwire:
> Just checking, all the bits are removed where the PDF stuff is used?

Yes everything is removed including the test. I think the test will need to be rewritten with each pdf component.

TomCusack-Softwire (approved):
PR title should potentially be `Replace PDF packages`
Do you not need any button or anything similar to make a PDF? no infrastructure? or will this be a separate PR?

PR #48, opened 2023-08-10
Implement minor styling hotfixes

Var1377:
Fixes the styling issues we saw in retro (aside from the tooltip issue which I can't reproduce in dev or release mode)

### DIFF
--- a/src/app/global_styles.tsx
+++ b/src/app/global_styles.tsx
@@ -16,7 +16,7 @@ const GlobalStyle = createGlobalStyle`
         height: 100%;
         width: 100%;
         font-family: Helvetica, Arial, sans-serif;
-        overflow: auto;
+        overflow: hidden;
     }
 
     body {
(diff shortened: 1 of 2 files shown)


SamYap-Softwire (approved):

src/app/global_styles.tsx
```diff
@@ -16,7 +16,7 @@ const GlobalStyle = createGlobalStyle`
         height: 100%;
         width: 100%;
         font-family: Helvetica, Arial, sans-serif;
-        overflow: auto;
+        overflow: hidden;
```
DanWilkins2107, line 19: Is this to fix the scroll bar issue? When I tried similar it didn't work but seems to be fine on yours :)

DanWilkins2107 (approved):

src/app/global_styles.tsx
Var1377: Yep, not sure why it didn't work when you tried it

PR #50, opened 2023-08-11
Change navbar and modal theming

Var1377:
Navbar is now white by popular request

(Changing the MUI primary colour also improves the look of other components as it was closer to what it was developed with)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,9 @@ next-env.d.ts
 # webstorm
 .idea
 
+# vscode
+.vscode
+
 # code coverage files
 coverage
 .nyc_output
--- a/src/components/Buttons/LinkButton.tsx
+++ b/src/components/Buttons/LinkButton.tsx
@@ -19,7 +19,7 @@ type LinkButtonProps = {
 
 const LinkButton: React.FC<LinkButtonProps> = (props) => {
     const pathname = usePathname();
-    const active = pathname.startsWith(props.link);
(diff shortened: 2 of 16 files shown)


DanWilkins2107 (approved):

src/components/Buttons/LinkButton.tsx
```diff
@@ -10,10 +10,6 @@ const UnstyledLink = styled(Link)`
     display: contents;
 `;
 
-const UpperCaseButton = styled(Button)`
-    text-transform: uppercase;
```
TomCusack-Softwire, line 14: why is this deleted? I thought people wanted it uppercase (even if it doesn't _yet_ work on `prod` build!)

TomCusack-Softwire (approved):

.gitignore
```diff
@@ -37,6 +37,9 @@ next-env.d.ts
 # webstorm
 .idea
 
+# vscode
+.vscode
```
DanWilkins2107, line 41: What acc is this?

src/components/Buttons/LinkButton.tsx
DanWilkins2107: Obviously correct me if I'm wrong on this, why couldn't we just type in uppercase if we want the button to be uppercase instead of having a separate component?

DanWilkins2107 (approved):
Couple Qs from me, nothing to change though I don't think

PR #52, opened 2023-08-14
VFB-30: detect system theme

Var1377:
OS Theme detection now works.

This is still a draft as I still want to implement persistence.


SamYap-Softwire (approved):

paopaosoftwire (approved):

paopaosoftwire (approved):

src/app/themes.tsx
```diff
@@ -169,10 +183,25 @@ const StyleManager: React.FC<Props> = ({ children, theme = lightTheme }) => {
         return <>{styles}</>;
     });
 
-    const [chosenTheme, setChosenTheme] = useState(theme);
+    const [themePreference, setThemePreference] = useLocalStorage<ThemePreference>(
+        "theme",
+        "system"
+    );
+
+    const [systemTheme, setSystemTheme] = useState<"dark" | "light">("light");
```
TomCusack-Softwire, line 191: type should be `<SystemTheme>`

src/components/Hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,21 @@
+import { useState, useEffect } from "react";
+
+const useLocalStorage = <T>(key: string, initialValue: T): [T, (newValue: T) => void] => {
+    const [value, setValue] = useState<T>(initialValue);
```
TomCusack-Softwire, line 4: is `<T>` not inferred? I don't know how good TS is at this sort of thing, but assume not - just wanted to check!

src/components/Hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    useEffect(() => {
+        const value = window.localStorage.getItem(key);
```
TomCusack-Softwire, line 7: don't shadow nonlocal variables! name it something other than `value` - it's very easy to accidently be using the wrong variable like this!
maybe 'existingValue'? or you can come up with a better name :)

TomCusack-Softwire (approved):
Minor changes - but otherwise looks good!
I'm surprised React/Next doesn't ship with it's own Local Storage hooks - it seems like this may be a common problem/solution!

src/components/Hooks/useLocalStorage.ts
Var1377: Yep it is

PR #56, opened 2023-08-15
VFB 9 Day Overview Component

ChloeYiuSoftwire:
Day Overview PDF that displays information of parcels to be collected at a particular centre on a particular day
Test for downloading is not written as it is expected to integrate with all existing PDF in the future

Overflowing table looks like this:
[4efbef52-1ed6-4050-abd9-468227d2aa77.pdf](https://github.com/VauxhallFoodbank/vauxhall-foodbank/files/12358595/4efbef52-1ed6-4050-abd9-468227d2aa77.pdf)


src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    return data;
+};
+
+interface Props {
+    text: string;
+}
+
+const DayOverview = async ({ text }: Props): Promise<React.ReactElement> => {
```
DanWilkins2107, line 45: can this not just be passed in as `text` not `{text}`? Would also remove the need for the props interface.

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+}
+
+interface DayOverviewerCardProps extends DayOverviewerPdfProps {}
+
+interface GeneralCellStyle {
+    borderStyle: "solid" | "dashed" | "dotted" | undefined;
+    borderWidth: number;
+    borderLeftWidth: number;
+    borderTopWidth: number;
+    padding: string;
```
DanWilkins2107, line 21: Probably best to just have these as optional entries, so that for example borderLeftWidth doesn't have to be set to 0

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+}
+
+interface CustomSVGProps {
+    icon: IconDefinition;
+    color: string;
+    fill: boolean;
+}
+
+const generalCellStyle: GeneralCellStyle = {
```
DanWilkins2107, line 30: So how it is done in the docs/ how we've done it elsewhere is very different. Instead of using consts, we combine classes like `[styles.class1, styles.class2, { any inline styles}]`. Have a look at mine or Sam's code perhaps for how we've done it.

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+        margin: "0px 40px 20px",
+    },
+    table: {
+        borderStyle: "solid",
+        borderWidth: 1,
+        borderRightWidth: 0,
+        borderBottomWidth: 0,
+    },
+    title: {
+        fontFamily: "Helvetica-Bold",
```
DanWilkins2107, line 54: If Helvetica-Bold is used everywhere it should just be set in a higher level component

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+        margin: "0 1px",
+        padding: "0 1px",
+    },
+    row: { display: "flex", flexDirection: "row" },
+    bold: { fontFamily: "Helvetica-Bold" },
+
+    // headers
+    header_flag: { flex: 1, fontFamily: "Helvetica-Bold", ...generalCellStyle },
+    header_name: { flex: 4, fontFamily: "Helvetica-Bold", ...generalCellStyle },
+    header_postcode: { flex: 4, fontFamily: "Helvetica-Bold", ...generalCellStyle },
```
DanWilkins2107, line 76: Don't actually think you should need to do it like this once the above bits are sorted, should be able to do `[styles.generalCellStyle , styles.classThatSetsTheFlexToWhatever]`

src/app/pdfDownloader/page.tsx
```diff
@@ -12,6 +13,10 @@ const PdfDownloader = async (): Promise<ReactElement> => {
                 {/* @ts-ignore */}
```
Var1377, line 13: rebase main, we no longer need a @ts-ignore

src/app/pdfDownloader/page.tsx
```diff
@@ -12,6 +13,10 @@ const PdfDownloader = async (): Promise<ReactElement> => {
...
                 <ShippingsLabel text="Download Shipping Labels" />
             </Centerer>
+            <Centerer>
+                {/* @ts-ignore */}
```
Var1377, line 17: same here

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import supabase, { Schema } from "@/supabase";
+import DayOverviewButton from "@/components/DayOverview/DayOverviewButton";
+
+export type ParcelOfSpecificDateLocation = {
+    collection_datetime: Schema["parcels"]["collection_datetime"];
+    clients: {
```
Var1377, line 7: Can use `Pick` here instead

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,62 @@
...
+        flagged_for_attention: Schema["clients"]["flagged_for_attention"];
+        full_name: Schema["clients"]["full_name"];
+        address_postcode: Schema["clients"]["address_postcode"];
+        delivery_instructions: Schema["clients"]["delivery_instructions"];
+    } | null;
+};
+
+export const getParcelsOfSpecificDateLocation = async (
```
Var1377, line 15: Could you rename this to getParcelsOfSpecificDate**And**Location

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,62 @@
...
+    text: string;
+}
+
+const DayOverview = async ({ text }: Props): Promise<React.ReactElement> => {
+    const date = new Date("2023-07-17");
+    const location = "Delivery";
+    const parcelsOfSpecificDate = await getParcelsOfSpecificDateLocation(date, location);
+
+    return (
+        <main>
```
Var1377, line 51: Don't use a main tag in a component, that should be down to the user of said component

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+                <Text style={styles.bold}>Instructions</Text>
+            </View>
+        </View>
+    );
+};
+
+// Database return type has null
+// But we know a client foreign key must map to an existing client
+// And collection_date is specified to be within a range
+// Hence both cannot be null
```
DanWilkins2107, line 144: Where do these comments refer to?

src/components/DayOverview/DayOverviewButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface DayOverviewerButtonProps {
+    date: Date;
+    location: string;
+    data: ParcelOfSpecificDateLocation[];
+    text: string;
+}
+const DayOverviewButton: React.FC<DayOverviewerButtonProps> = ({ date, location, data, text }) => {
+    return (
+        <>
```
Var1377, line 17: This fragment isn't needed

DanWilkins2107 (changes requested):
Some stylistic changes :)

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+            strokeWidth={40}
+            d={icon.icon[4] as string}
+        />
+        {/*IconDefinition[4] is defined as string | string[] but in reality there is no use case*/}
+        {/*with string[] as svgPathData*/}
+    </Svg>
+);
+
+const DayOverviewerDate: React.FC<{ date: Date }> = ({ date }) => {
+    const dateString = date.toLocaleDateString("sv-SE");
```
Var1377, line 108: is this Swedish? (and is this meant to be there?)

Var1377 (changes requested):

src/components/DayOverview/DayOverview.tsx
SamYap-Softwire, line 15: does this need to be exported?

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const { data, error } = await supabase
+        .from("parcels")
+        .select(
+            "collection_datetime, clients(flagged_for_attention, full_name, address_postcode, delivery_instructions)"
+        )
+        .gte("collection_datetime", startDateString)
+        .lt("collection_datetime", endDateString)
+        .eq("collection_centre", collectionCentre)
+        .order("collection_datetime");
+
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Prettier helps me split into lines - if not it will be far too long :/

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,62 @@
...
+    if (error) {
+        throw new Error(`Database error: ${JSON.stringify(error)}`);
+    }
+
+    return data;
+};
+
+interface Props {
```
SamYap-Softwire, line 41: Can you move this to the top of the file, thanks hehe

src/components/DayOverview/DayOverview.tsx
SamYap-Softwire, line 45: Can this just be `async (text: string): ...` ? 
I personally would prefer that over using `Props`  interface - but tbf I think others might disagree :(

src/components/DayOverview/DayOverviewButton.tsx
SamYap-Softwire, line 17: Do you need the empty fragments here?

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+import { faFlag, faSquare, IconDefinition } from "@fortawesome/free-solid-svg-icons";
+
+import { ParcelOfSpecificDateLocation } from "@/components/DayOverview/DayOverview";
+interface DayOverviewerPdfProps {
+    date: Date;
+    location: string;
+    data: ParcelOfSpecificDateLocation[];
+}
+
+interface DayOverviewerCardProps extends DayOverviewerPdfProps {}
```
SamYap-Softwire: Can you explain why we need this interface?
(Seems like we can just replace the usage below with `DayOverviewedPdfProps`)

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+interface CustomSVGProps {
+    icon: IconDefinition;
+    color: string;
+    fill: boolean;
+}
+
+const generalCellStyle: GeneralCellStyle = {
+    borderStyle: "solid",
+    borderWidth: 1,
```
SamYap-Softwire, line 25: Can you make them consistent? Either strings with units or numbers (I don't mind either, but I did `" _ pt"` for mine. 
Just because it might not be clear that numbers default to `pt` as their unit

SamYap-Softwire: Same for the ones below, thanks :)

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+    borderLeftWidth: 0,
+    borderTopWidth: 0,
+    padding: "3px",
```
SamYap-Softwire, line 28: I think `px` is not a valid unit in the docs - they use `pt` (although it did work when I tried it too...) 
[https://react-pdf.org/styling](url)

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+};
+
+const styles = StyleSheet.create({
+    page: {
+        display: "flex",
+        flexDirection: "column",
+        fontSize: "12px",
+    },
+    card: {
+        margin: "0px 40px 20px",
```
SamYap-Softwire, line 38: probably don't need units for 0

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+        marginBottom: "5px",
+    },
+    svg: {
+        margin: "0 1px",
+        padding: "0 1px",
+    },
+    row: { display: "flex", flexDirection: "row" },
+    bold: { fontFamily: "Helvetica-Bold" },
+
+    // headers
```
SamYap-Softwire: I think it's quite clear from the variable names that they are the headers so there's probably no need for a comment here?

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+    header_flag: { flex: 1, fontFamily: "Helvetica-Bold", ...generalCellStyle },
+    header_name: { flex: 4, fontFamily: "Helvetica-Bold", ...generalCellStyle },
+    header_postcode: { flex: 4, fontFamily: "Helvetica-Bold", ...generalCellStyle },
+    header_time: { flex: 2, fontFamily: "Helvetica-Bold", ...generalCellStyle },
+    header_collection: { flex: 4, fontFamily: "Helvetica-Bold", ...generalCellStyle },
+    header_six: { flex: 10, fontFamily: "Helvetica-Bold", ...generalCellStyle },
+
+    // cells
```
SamYap-Softwire, line 71: Same as above :)

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: I removed it due to the changes of combining classes

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: I replaced most to use multi class with `styles.bold`, but I feel like it seems better specifically for the title one to be specified bold, a bit like how `h1` turns things bold automatically without extra specification of bold in separate class styling

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: `DayOverviewRow`

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+    return new Date(datetime).getHours() <= 11 ? "AM" : "PM";
+};
+
+const CustomSVG: React.FC<CustomSVGProps> = ({ icon, color, fill }) => (
+    <Svg width={12} height={12} viewBox="0 0 512 512" style={styles.svg}>
+        <Path
+            fill={fill ? color : undefined}
+            stroke={color}
+            strokeWidth={40}
+            d={icon.icon[4] as string}
```
SamYap-Softwire, line 90: is there a better way to do this that is not type casting? (not a big issue, just curious)

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+            </View>
+        </View>
+    );
+};
+
+// Database return type has null
+// But we know a client foreign key must map to an existing client
+// And collection_date is specified to be within a range
+// Hence both cannot be null
+const DayOverviewerRow: React.FC<{ row: ParcelOfSpecificDateLocation }> = ({ row }) => {
```
SamYap-Softwire: Can this be `React.FC<ParcelOfSpecificDateLocation> = (row) ...` ? 
(again, not an issue - just personal preference)

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,202 @@
...
+    return (
+        <View style={styles.row}>
+            <View style={styles.cell_logo}>
+                <View style={styles.row}>
+                    <CustomSVG icon={faSquare} color="black" fill={false} />
+                    {row.clients!.flagged_for_attention ? (
+                        <CustomSVG icon={faFlag} color="orange" fill={true} />
+                    ) : (
+                        <></>
```
SamYap-Softwire, line 145: do you want the empty fragment there (i.e. a "gap" sort of if you're not rendering the icon)? if not, can you make this conditional rendering (`&&`)?

src/app/pdfDownloader/page.tsx
ChloeYiuSoftwire: Done

src/app/pdfDownloader/page.tsx
ChloeYiuSoftwire: Done

SamYap-Softwire (changes requested):

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverviewButton.tsx
ChloeYiuSoftwire: Removed

src/components/DayOverview/DayOverviewButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+            <NoSsr>
+                <PDFDownloadLink
+                    document={<DayOverviewPdf date={date} location={location} data={data} />}
+                    fileName="DayOverview.pdf"
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: I made a function that renders name based on date and location - see if it looks good (?)

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Not sure what language is it
But the format looks good as I want something like `2023-08-15` and this returns it

src/app/pdfDownloader/page.tsx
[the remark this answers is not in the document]
ChloeYiuSoftwire: Rebased

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,62 @@
...
+import DayOverviewButton from "@/components/DayOverview/DayOverviewButton";
+
+export type ParcelOfSpecificDateLocation = {
+    collection_datetime: Schema["parcels"]["collection_datetime"];
+    clients: {
+        flagged_for_attention: Schema["clients"]["flagged_for_attention"];
+        full_name: Schema["clients"]["full_name"];
+        address_postcode: Schema["clients"]["address_postcode"];
+        delivery_instructions: Schema["clients"]["delivery_instructions"];
+    } | null;
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Varun just told me something called `Pick` that simplifies it

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export const getParcelsOfSpecificDateLocation = async (
+    date: Date,
+    collectionCentre: string
+): Promise<ParcelOfSpecificDateLocation[]> => {
+    const startDateString = date.toISOString();
+    const endDate = new Date();
+    endDate.setDate(date.getDate() + 1);
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Seems like this does not happen
However, I updated the function to fix other bugs

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Oh yes, forgot I updated the pages and now it doesn't need to be exported

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Done ^^

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Sure - this has been totally removed btw so it is no longer relevant

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: same as above

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: good question, I ain't sure why is `IconDefinition[4]` casted as `string | string[]` at first at all - looked through most of them and don't seem there is any array at all

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Don't think it is allowed - putting `<ParcelOfSpecificDateLocation>` without `{}` implies it is the props having type `ParcelOfSpecificDateLocation`, but not `props.row`

ChloeYiuSoftwire: But there is only one props here so it seems a bit redundant to make an interface just for it

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: It doesn't matter as the column width should be sufficient for two icons - but using `&&` seems tidier so refactored

src/components/DayOverview/DayOverview.tsx
SamYap-Softwire: Nvm, just realized Dan has the same comment

src/components/DayOverview/DayOverviewPdf.tsx
SamYap-Softwire: (the comment is outdated for some reason so i'll just resolve this and re-comment on the latest commit)

SamYap-Softwire (changes requested):
Added some more comments, will review it again once you respond to the comments from last review as well :)

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    if (error) {
+        throw new Error(`Database error: ${JSON.stringify(error)}`);
+    }
+
+    return data;
+};
+
+const DayOverview = async ({ text }: Props): Promise<React.ReactElement> => {
+    const date = new Date("2023-07-17");
```
paopaosoftwire, line 44: Why is this hardcoded? Will we have a function to handle this later on or will this be more like exporting today's documents.

src/components/DayOverview/DayOverviewButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <NoSsr>
+            <PDFDownloadLink
+                document={<DayOverviewPdf date={date} location={location} data={data} />}
+                fileName={`DayOverview_${dateString}_${collectionCentreToAbbreviation(
+                    location
+                )}.pdf`}
+            >
+                {text}
+            </PDFDownloadLink>
+            <PDFViewer width="400" height="500">
```
paopaosoftwire, line 56: Dont merge the viewer into main. We only need the download link as we are not doing pdf preview. The download link component will be combined to a generic component later.

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+    icon: IconDefinition;
+    color: string;
+    fill: boolean;
+}
+
+const styles = StyleSheet.create({
+    page: {
+        display: "flex",
+        flexDirection: "column",
+        fontSize: "12px",
```
paopaosoftwire, line 26: Can we use pt for fontSize? Its a little more standard.

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+        borderLeftWidth: 0,
+        borderTopWidth: 0,
+        padding: "3px",
+    },
+    row: { display: "flex", flexDirection: "row" },
+    bold: { fontFamily: "Helvetica-Bold" },
+
+    // cells
+    cell_logo: { flex: 1 },
+    cell_name: { flex: 4 },
```
paopaosoftwire, line 64: A lot of the cell here has the same thing eg. flex: 4, can we combine this?

src/components/DayOverview/DayOverviewPdf.tsx
paopaosoftwire: In that case should we just manually do it via, getDate, getFullYear and concat it manually? sv-SE is not clear for other developers and could cause problems later on. I agree that toLocalDateString is more concise but might not fit the use case sometimes.

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+const DayOverviewDate: React.FC<{ date: Date }> = ({ date }) => {
+    const dateString = date.toLocaleDateString("sv-SE");
+    return (
+        <View style={styles.row}>
+            <Text style={styles.bold}>Date: </Text> <Text>{dateString}</Text>
+        </View>
+    );
+};
+
+const DayOverviewLocation: React.FC<{ location: string }> = ({ location }) => (
```
paopaosoftwire, line 97: I think tom likes the interface to be separated out even if its small.

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+    <View style={styles.row}>
```
paopaosoftwire, line 98: This whole black be just

`<Text style={styles.row}><Text style={styles.bold}>Location: </Text>{location}</Text>`

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+            <View style={[styles.cell_postcode, styles.bold, styles.cell]}>
+                <Text style={styles.bold}>Postcode</Text>
+            </View>
+            <View style={[styles.cell_time, styles.bold, styles.cell]}>
+                <Text style={styles.bold}>Time</Text>
+            </View>
+            <View style={[styles.cell_instructions, styles.bold, styles.cell]}>
+                <Text style={styles.bold}>Instructions</Text>
+            </View>
+        </View>
```
paopaosoftwire, line 119: Similarly to the one above, you could condense this down by not wrapping one text in one view. (I did not try this out on your code particularly but I condensed my code using this so I think it works in this case as well).

src/components/DayOverview/DayOverviewPdf.tsx
paopaosoftwire: Does it have to be here? I think by putting a ! as a reader I kinda assume that it cannot be null already?

src/components/DayOverview/DayOverviewPdf.tsx
DanWilkins2107: Think you should be able to change this to something like gb-EN or something? Seems wrong to have swedish in there (also might end up being a problem w time zones!)

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+    );
+};
+
+// Database return type has null
+// But we know a client foreign key must map to an existing client
+// And collection_date is specified to be within a range
+// Hence both cannot be null
+const DayOverviewRow: React.FC<{ row: ParcelOfSpecificDateLocation }> = ({ row }) => {
+    return (
+        <View style={styles.row}>
```
paopaosoftwire, line 129: you might want to add wrap={false} to prevent text from overflowing the page when the boxes get too large etc.

https://react-pdf.org/advanced

paopaosoftwire (changes requested):
Cool stuff!

src/components/DayOverview/DayOverview.tsx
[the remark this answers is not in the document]
DanWilkins2107: Don't get me wrong, some of the line splits are helpful, but generally if there's one function I feel like it is easier to read if there aren't load of extra lines everywhere

src/components/DayOverview/DayOverview.tsx
DanWilkins2107: So we talked about this in person, but on second thoughts you can defo do it as Sam mentioned below idk why I thought for a second you couldn't

src/components/DayOverview/DayOverviewButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+        case "Waterloo - Oasis":
+            return "WAT_OA";
+        case "Waterloo - St George the Martyr":
+            return "WAT_SG";
+        case "Waterloo - St Johns":
+            return "WAT_SJ";
+        case "Delivery":
+            return "Delivery";
+        default:
+            return "-";
```
DanWilkins2107, line 40: Do we want this as the default as opposed to an error message? Probably not needed tbf just something to think about

DanWilkins2107: Something like "Invalid Collection Centre" idk

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+        display: "flex",
+        flexDirection: "column",
+        fontSize: "12px",
+    },
+    card: {
+        margin: "0px 40px 20px",
+    },
+    table: {
+        borderStyle: "solid",
+        borderWidth: 1,
```
DanWilkins2107, line 33: Can just set `borderTopWidth` and `borderLeftWidth`

src/components/DayOverview/DayOverviewPdf.tsx
DanWilkins2107: Makes sense, just looked bad with it put everywhere!

src/components/DayOverview/DayOverviewPdf.tsx
DanWilkins2107: same as above???

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+const DayOverviewLocation: React.FC<{ location: string }> = ({ location }) => (
+    <View style={styles.row}>
+        <Text style={styles.bold}>Location: </Text> <Text>{location}</Text>
+    </View>
+);
+
+const DayOverviewHeader: React.FC<{}> = () => {
+    return (
+        <View style={styles.row}>
```
DanWilkins2107, line 105: Should be able to just put `styles.bold` in here instead of in every single view. (This is more what I meant when saying to set it in a higher level component)

src/components/DayOverview/DayOverviewPdf.tsx
DanWilkins2107: Are they needed?

DanWilkins2107: / can they be put inside so they are clearer

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+    );
+};
+
+const DayOverviewLocation: React.FC<{ location: string }> = ({ location }) => (
+    <View style={styles.row}>
+        <Text style={styles.bold}>Location: </Text> <Text>{location}</Text>
+    </View>
+);
+
+const DayOverviewHeader: React.FC<{}> = () => {
```
DanWilkins2107, line 103: Do you need a `wrap: false` somewhere to stop a section overflowing to the next page?

DanWilkins2107 (changes requested):

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Removed

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,182 @@
...
+import { faFlag, faSquare, IconDefinition } from "@fortawesome/free-solid-svg-icons";
+
+import { ParcelOfSpecificDateLocation } from "@/components/DayOverview/DayOverview";
+interface DayOverviewPdfProps {
+    date: Date;
+    location: string;
+    data: ParcelOfSpecificDateLocation[];
+}
+
+interface DayOverviewCardProps extends DayOverviewPdfProps {}
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: It shares the same prop as `DayOverviewPdfProp` at this moment
But I want to make it clear like this is just the prop of `DayOverviewCardProps` (don't want to use `DayOverviewPdfProp` as sharing props seems messy
Also it seems more maintainable if other people want to extend it to add more props to the overview card function

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Currently I want to set a specific date and specific location.
It is probably a future ticket issue if the button is designed to be the login user current date and the location the user belongs to
It is possible that people might also want to download the list for e.g. tomorrow - so I want to make it more flexible first
In the future maybe this can be extended to a panel letting user select a date on the calendar and a location (if they have the privilege to see others location for example?)

src/components/DayOverview/DayOverviewButton.tsx
ChloeYiuSoftwire: Sorry, will remove before merge - the current branch doesn't have this page anyways but it is helpful for now

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Probably might not, since the number are not technically supposed to have any correlation (it is just coincidentally 4 looks nice for those columns) - so if I change one column's flex number it should not change the other columns one if future developers want to improve the layout

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Done for both

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Re the comment: if y'all find it easy to understand already I will just remove the comment - I was just worried if you guys will question why do I explicitly add a ! to it

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: That is only valid for functions. This is a component, and props have to be passed in as an object

ChloeYiuSoftwire: same as above (I think shipping label used the exact same syntax)

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: You are right

src/components/DayOverview/DayOverviewButton.tsx
ChloeYiuSoftwire: Let me consult Tom on this.
I assume the user using it will either be only granted privilege to access one location, or there is a drop down list to choose, so no matter what it shouldn't reach the default case
It's just if the program crashed or sth, should I let the program throw an error or return dash

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import supabase, { Schema } from "@/supabase";
+import DayOverviewButton from "@/components/DayOverview/DayOverviewButton";
+
+export type ParcelOfSpecificDateLocation = {
+    collection_datetime: Schema["parcels"]["collection_datetime"];
+    clients: {
+        flagged_for_attention: Schema["clients"]["flagged_for_attention"];
```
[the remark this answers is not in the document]
DanWilkins2107: I disagree with this, in that I don't think there will be changes in the schema, and I think it would be better readability wise to have the types explicitly written out. See what others think tho.

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    if (error) {
+        throw new Error(`Database error: ${JSON.stringify(error)}`);
+    }
+
+    return data;
+};
+
+const DayOverview = async ({ text }: Props): Promise<React.ReactElement> => {
+    const date = new Date("2023-07-17");
```
DanWilkins2107, line 44: Seems odd to not have this passed in as part of the props as opposed to being hard coded. In any use case people would choose the date from the page, not here.

DanWilkins2107: Oh wait nvm if you don't have a page that is fine. However, what is this page used for? Like is it a dummy page? In which case it should probably not exist when merging and the interfaces etc should be contained in the DayOverviewButton File

src/components/DayOverview/DayOverviewButton.tsx
DanWilkins2107: Yeah no that was what I was thinking, as it is something going wrong I think you should throw an error potentially

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { faFlag, faSquare, IconDefinition } from "@fortawesome/free-solid-svg-icons";
+
+import { ParcelOfSpecificDateAndLocation } from "@/components/DayOverview/DayOverview";
+
+interface DayOverviewDateProps {
+    date: Date;
+}
+
+interface DayOverviewLocationProps {
+    location: string;
```
DanWilkins2107, line 14: Do all of these need to be interfaces? Instead of just having it written out once in the props?

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,205 @@
...
+    page: {
+        display: "flex",
+        flexDirection: "column",
+        fontSize: "11pt",
+    },
+    card: {
+        margin: "0 30pt 15pt",
+    },
+    table: {
+        borderStyle: "solid",
```
DanWilkins2107, line 45: Is there borderLeftStyle? Not sure if there is tho so please ignore if not! (In which case could be added inline)

src/components/DayOverview/DayOverviewPdf.tsx
DanWilkins2107: I agree I would leave these separate!

DanWilkins2107 (changes requested):

src/components/DayOverview/DayOverviewButton.tsx
ChloeYiuSoftwire: Ok I changed it to throwing errors

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Oh wait actually I realise you and @paopaosoftwire are right. I should move it to the dummy `/pdfDownloader` page instead because `DayOverview` is not a dummy page - as `DayOverviewButton` is on client side but data fetching should be done on the async side

src/components/DayOverview/DayOverviewPdf.tsx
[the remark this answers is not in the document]
ChloeYiuSoftwire: I rethink about the implementation. I think like it's the implementation of Pdf and Card that coincidentally shares the same props, not like they are supposed to have the same props. Hence, I shouldn't even use the term extend, because that implies it must have everything that pdf has.

Usually we have one prop per function - seem to make things clearer in general

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: I think Tom likes this format more

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: There is borderLeft. Refactored as one-line (also the same for `styles.table`!)

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,58 @@
...
+        .select(
+            "collection_datetime, clients(flagged_for_attention, full_name, address_postcode, delivery_instructions)"
+        )
+        .gte("collection_datetime", startDateString)
+        .lt("collection_datetime", endDateString)
+        .eq("collection_centre", collectionCentre)
+        .order("collection_datetime");
+
+    if (error) {
+        throw new Error(`Database error: ${JSON.stringify(error)}`);
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Refactored

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,58 @@
...
+    clients: Pick<
+        Schema["clients"],
+        "flagged_for_attention" | "full_name" | "address_postcode" | "delivery_instructions"
+    > | null;
+};
+
+const getParcelsOfSpecificDateAndLocation = async (
+    date: Date,
+    collectionCentre: string
+): Promise<ParcelOfSpecificDateAndLocation[]> => {
```
DanWilkins2107, line 22: This name seems a bit long and verbose, would `SpecificParcel` not do?

src/components/DayOverview/DayOverviewButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+            throw new Error("Invalid location");
+    }
+};
+
+const getCurrentDate = (date: Date): string => {
+    const year = date.getFullYear();
+    const month = (date.getMonth() + 1).toString().padStart(2, "0");
+    const day = date.getDate().toString().padStart(2, "0");
+
+    return `${year}${month}${day}`;
```
DanWilkins2107, line 49: I mentioned on Paopao's too, would it be better to have some sort of / or _ or - so that users can see that it is a date?

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,204 @@
...
+        borderRightWidth: 1,
+        borderBottomWidth: 1,
+        padding: "2.25pt",
+    },
+    margin: { height: "37.5pt" },
+    row: { display: "flex", flexDirection: "row" },
+    bold: { fontFamily: "Helvetica-Bold" },
+
+    cell_logo: { flex: 1 },
```
DanWilkins2107, line 73: Can we standardise the naming conventions here to either camelCase or kebab-case

src/components/DayOverview/DayOverviewButton.tsx
DanWilkins2107: Also there is an almost identical function elsewhere

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,204 @@
...
+    cell_name: { flex: 4 },
+    cell_postcode: { flex: 4 },
+    cell_time: { flex: 2 },
+    cell_collection: { flex: 4 },
+    cell_instructions: { flex: 10 },
+});
+
+const getCurrentDate = (date: Date): string => {
+    const year = date.getFullYear();
```
DanWilkins2107, line 82: Here is the almost identical function mentioned earlier

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Specific parcels doesn't tell us that it is specific based on date and location (i.e. the filter criteria)
Also it is an internal function not a reusable component so it should be fine to be lengthy but clear about what's its function?

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,204 @@
...
+const DayOverviewLocation: React.FC<DayOverviewLocationProps> = ({ location }) => (
+    <Text>
+        <Text style={styles.bold}>Location: </Text>
+        {location}
+    </Text>
+);
+
+const DayOverviewHeader: React.FC<{}> = () => {
+    return (
+        <View style={[styles.row, styles.bold, { borderTop: "1 solid black" }]} fixed={true}>
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: `fixed` makes thing persist on every page
If you look at the overflow pdf with 3 pages in the PR description, you can see the top margin and the header appears in every single page
Removed true btw

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,204 @@
...
+            <View style={[styles.cell_logo, styles.cell]}>
+                <View style={styles.row}>
+                    <CustomSVG icon={faSquare} color="black" fill={false} />
+                    {row.clients!.flagged_for_attention && (
+                        <CustomSVG icon={faFlag} color="orange" fill={true} />
+                    )}
+                </View>
+            </View>
+            <View style={[styles.cell_name, styles.cell]}>
+                <Text>{row.clients!.full_name}</Text>
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: In parcels table, the setting ensures `client_id` must be a foreign key mapping to client table's primary id. Hence, when I do a join function to client, there must be an existing client and the info cannot be null.
However, typescript does not know it and will complain client might possibly be null, which is impossible.

Also, for the `collection_date`, since a range filter is used, it also cannot be null, hence there is a ! too

src/components/DayOverview/DayOverview.tsx
DanWilkins2107: Sure if you think it is better that is fine!

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,58 @@
...
+import DayOverviewButton from "@/components/DayOverview/DayOverviewButton";
+
+interface Props {
+    text: string;
+    date: Date;
+    location: string;
+}
+
+export type ParcelOfSpecificDateAndLocation = {
+    collection_datetime: Schema["parcels"]["collection_datetime"];
```
DanWilkins2107, line 12: Is this not just a date? (or a string idk which lol)

src/components/DayOverview/DayOverviewButton.tsx
DanWilkins2107: Sure that's fine, could those two be combined though? Like with say a boolean prop called hyphens where if true hyphens are included else they are not?

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,204 @@
...
+        padding: "0 0.75pt",
+    },
+    cell: {
+        borderStyle: "solid",
+        borderRightWidth: 1,
+        borderBottomWidth: 1,
+        padding: "2.25pt",
+    },
+    margin: { height: "37.5pt" },
+    row: { display: "flex", flexDirection: "row" },
```
DanWilkins2107, line 70: flex-direction row is default

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,204 @@
...
+}
+
+interface DayOverviewCardProps {
+    date: Date;
+    location: string;
+    data: ParcelOfSpecificDateAndLocation[];
+}
+
+interface CustomSVGProps {
+    icon: IconDefinition;
```
DanWilkins2107, line 34: Could changing here help with later problem maybe?

src/components/DayOverview/DayOverviewPdf.tsx
DanWilkins2107: I wonder if changing the interface value could help with this? Not sure (mentioned above as to where)

src/components/DayOverview/DayOverviewPdf.tsx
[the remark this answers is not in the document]
DanWilkins2107: True is still there?

DanWilkins2107 (changes requested):

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: I tested and I think column is default

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Don't think I can because I didn't define the interface `IconDefinition` myself. It is something from `fortawesome`

src/components/DayOverview/DayOverviewPdf.tsx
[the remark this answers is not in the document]
ChloeYiuSoftwire: Removed

src/components/DayOverview/DayOverviewButton.tsx
```diff
@@ -0,0 +1,66 @@
+"use client";
```
paopaosoftwire, line 1: This is looking good! My only comment would be is it possible for you to make this component look something like `src\components\ShippingLabels\ShippingLabelsButton.tsx` where there are no data logic in the file at all. So things like date and location should be passed in from inside the data props.

There's absolutely nothing wrong with the current implementation. Its just this file will be changed to a generic PDF button component which is implemented in VFB-11, and your current implementation will not be able to use that directly.

paopaosoftwire (commented):
Just FYI, I tried your thing with 512 rows of data and it works fine, the pdf generation is a little slow but I think its fine.

src/components/DayOverview/DayOverviewButton.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverview.tsx
DanWilkins2107: I would say to change this, I think it is implied that it is the same through the name of the variable.

DanWilkins2107: Also the fact I don't immediately know what the type is isn't great

src/components/DayOverview/DayOverviewPdf.tsx
DanWilkins2107: I mean idk if there's some inheritance somehow due to React-pdf, but row is definitely default.

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,185 @@
...
+    },
+    cell: {
+        borderStyle: "solid",
+        borderRightWidth: 1,
+        borderBottomWidth: 1,
+        padding: "2.25pt",
+    },
+    margin: { height: "37.5pt" },
+    row: { display: "flex", flexDirection: "row" },
+    bold: { fontFamily: "Helvetica-Bold" },
```
DanWilkins2107, line 68: If this is all that is in styles.bold it can be added inline (dw tho if you think this is nicer)

DanWilkins2107 (approved):
Final changes, just needs a couple changes.

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Feels like styles.bold is nicer because if e.g, future developer want to change fonts we can just change one fontFamily and not change all the inline code

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Not sure. But if everything breaks below because I remove the line `flexDirection: row` then it means it has to be kept 
Maybe because the document is set at column (?)

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: But if this is the case then the `Pick` stuff also doesn't immediately tell us the type
But hard coding all the types doesn't seem like a good idea
The idea is we should refer to the Schema to check the type? I ain't sure - let's see how others think

Var1377: I think @ChloeYiuSoftwire is correct here, I would keep it as it is

paopaosoftwire (approved):
Apart from my unresolved comment in src/components/DayOverview/DayOverviewPdf.tsx I think everything is fine now. If you dont agree with that comment I am fine with you not doing it so you can resolve that comment. I will now approve this.

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Done

src/app/pdfDownloader/page.tsx
SamYap-Softwire, line 1: can you rebase main? I don't think this folder exists on main anymore

SamYap-Softwire (changes requested):
I noticed you haven't rebased main so I'll let you do that first before continuing the review :)

src/app/pdfDownloader/page.tsx
SamYap-Softwire: I don't think we need this page to be merged anymore :)

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+export type ParcelOfSpecificDateAndLocation = Pick<Schema["parcels"], "collection_datetime"> & {
+    clients: Pick<
+        Schema["clients"],
+        "flagged_for_attention" | "full_name" | "address_postcode" | "delivery_instructions"
+    > | null;
+};
+
+export type DayOverviewData = {
```
SamYap-Softwire, line 19: can you explain why you've used type for this instead of interface?

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,99 @@
...
+    date: Date;
+    location: string;
+    data: ParcelOfSpecificDateAndLocation[];
+};
+
+export const getCurrentDate = (date: Date, hyphen: boolean = false): string => {
```
SamYap-Softwire, line 25: can we not use JS Date (toLocaleString) for this?

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,99 @@
...
+        .from("parcels")
+        .select(
+            "collection_datetime, clients(flagged_for_attention, full_name, address_postcode, delivery_instructions)"
+        )
+        .gte("collection_datetime", startDateString)
+        .lt("collection_datetime", endDateString)
+        .eq("collection_centre", collectionCentre)
+        .order("collection_datetime");
+
+    if (error) {
```
SamYap-Softwire, line 52: might be worth either adding a TODO for VFB 22 or changing this to fit the new error handling messages - depending on which gets merged first ^^

src/components/DayOverview/DayOverviewPdf.tsx
SamYap-Softwire: for some reason i have the same "issue" of docs saying row is default but whenever i try it it always seems like column is default TT

src/components/DayOverview/DayOverviewPdf.tsx
SamYap-Softwire: not sure if there needs to be something in place for handling `icon.icon[4]` when it is a `string[]` - maybe even just a ternary / if else - because I doubt they'll put that in the docs if there isn't a use case (although tbf, they might)

I'll leave this for you to decide though :)

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,171 @@
...
+        </View>
+    );
+};
+
+const DayOverviewRow: React.FC<DayOverviewRowProps> = ({ row }) => {
+    return (
+        <View style={styles.row} wrap={false}>
+            <View style={[styles.cellLogo, styles.cell, styles.row]}>
+                <CustomSVG icon={faSquare} color="black" fill={false} />
+                {row.clients!.flagged_for_attention && (
```
SamYap-Softwire, line 123: will there be a case when `clients` is null?

SamYap-Softwire (changes requested):
looks good, just some minor things

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Changed back to interface

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Sure I updated it

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: I think the database by default need a parcel to contain a foreign client key and if a client is deleted, then the parcels will be removed by cascade too (?) 
I have to double check with Tom if the foreign client key can be null though but I don't think so?

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: I have tried to implement some logic to ensure it is definitely string (just in case I overlooked the library)
See if you find it ok or not?

src/components/DayOverview/DayOverviewPdf.tsx
SamYap-Softwire: yeah no that makes sense :)

src/app/pdfDownloader/page.tsx
ChloeYiuSoftwire: Yep I will keep it until Tom has approved my changes because this thing here will help me do future edits if Tom requested changes

SamYap-Softwire (approved):

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,111 @@
...
+): Promise<ParcelOfSpecificDateAndLocation[]> => {
+    const startDateString = date.toISOString();
+    const endDate = new Date(date);
+    endDate.setDate(date.getDate() + 1);
+    const endDateString = endDate.toISOString();
+
+    const { data, error } = await supabase
+        .from("parcels")
+        .select(
+            "collection_datetime, clients(flagged_for_attention, full_name, address_postcode, delivery_instructions)"
```
TomCusack-Softwire, line 53: can this be split over multiple lines, one line for each column? this sort of format looks tidier to me :)
image

src/components/DayOverview/DayOverview.tsx
```diff
@@ -0,0 +1,111 @@
...
+    if (error) {
+        throw new Error(
+            "We were unable to fetch the parcels with the specified collection date and location."
+        );
+    }
+
+    return data;
+};
+
+const collectionCentreToAbbreviation = (
```
TomCusack-Softwire, line 71: assuming this will be changed in the generalise collection centre ticket?

TomCusack-Softwire (changes requested):
minor changes

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Yep

ChloeYiuSoftwire: I think most likely this ticket will be merged earlier so i will work on this when it is done

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -0,0 +1,169 @@
...
+    return new Date(datetime).getHours() <= 11 ? "AM" : "PM";
+};
+
+const CustomSVG: React.FC<CustomSVGProps> = ({ icon, color, fill }) => (
+    <Svg width={11} height={11} viewBox="0 0 512 512" style={styles.svg}>
+        <Path
+            fill={fill ? color : undefined}
+            stroke={color}
+            strokeWidth={40}
+            d={typeof icon.icon[4] === "string" ? icon.icon[4] : icon.icon[4][0]}
```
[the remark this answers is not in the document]
ChloeYiuSoftwire: Done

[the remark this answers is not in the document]
ChloeYiuSoftwire: I looked into the faIcon library and the SVG path is stored in this position

src/components/DayOverview/DayOverview.tsx
```diff
@@ -23,19 +23,14 @@ export interface DayOverviewData {
...
+        day: "2-digit",
+    });
 
-    const year = formattedDate[2];
-    const month = formattedDate[1];
-    const day = formattedDate[0];
-
-    return hyphen ? `${year}-${month}-${day}` : `${year}${month}${day}`;
+    const separator = hyphen ? "-" : "";
+    return formattedDate.replaceAll("-", separator);
```
TomCusack-Softwire, line 33: this 'does nothing' in one of the branches - because you have now changed it to `en-CA`, it's just replacing hyphens with hyphens, which is inefficient.

Maybe return the formatted date if hyphen is false, and formattedDate with hyphens replaced if hyphen is true? (ternary?)

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -79,16 +79,19 @@ const dateTimeToAMPM = (datetime: string): string => {
...
-);
+const CustomSVG: React.FC<CustomSVGProps> = ({ icon, color, fill }) => {
+    const path = icon.icon[4];
+    return (
+        <Svg width={11} height={11} viewBox="0 0 512 512" style={styles.svg}>
+            <Path
+                fill={fill ? color : undefined}
+                stroke={color}
+                strokeWidth={40}
+                d={typeof path === "string" ? path : path[0]}
```
TomCusack-Softwire, line 90: can this typeof also be converted to a variable? ideally we'd want just `d={variableName}` and very little logic in the TSX bit of code

src/components/DayOverview/DayOverviewPdf.tsx
```diff
@@ -79,16 +79,19 @@ const dateTimeToAMPM = (datetime: string): string => {
...
-        <Path
-            fill={fill ? color : undefined}
-            stroke={color}
-            strokeWidth={40}
-            d={typeof icon.icon[4] === "string" ? icon.icon[4] : icon.icon[4][0]}
-        />
-    </Svg>
-);
+const CustomSVG: React.FC<CustomSVGProps> = ({ icon, color, fill }) => {
+    const path = icon.icon[4];
```
TomCusack-Softwire, line 83: not sure what `path` refers to, can you think of a better name?

TomCusack-Softwire (changes requested):

src/components/DayOverview/DayOverview.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: Done

src/components/DayOverview/DayOverviewPdf.tsx
ChloeYiuSoftwire: See if `svgPath` is better because the literal meaning of those string is just SVG path

TomCusack-Softwire (approved):
remember to delete temp files!

PR #57, opened 2023-08-15
Speed up automated testing

Var1377:
7 minutes is an eternity, they now run in just over 3 minutes. Screenshots and videos are also disabled now that we don't have a mismatch between test results on actions and local testing.


DanWilkins2107 (approved):

paopaosoftwire (approved):

SamYap-Softwire (commented):
The code seems fine, though I don't think we should be disabling screenshot and video test.
But I can be convinced (I think?) so we can discuss it later :)

.github/workflows/checks.yml
```diff
@@ -43,21 +46,8 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Compile
+        run: npm run build
+
       - name: Run e2e tests
         run: npm run test:e2e
-
-      - name: Upload Screenshots
```
TomCusack-Softwire, line 49: is it ever worth to keep screenshots, or not really?

Var1377: Not sure, until screenshot tests are introduce I'm not sure it's needed although I imagine it actually doesn't add too much time to the test run. It's your call

TomCusack-Softwire (approved):
It's probably best if we leave the GitHub Actions stuff in tact - if fail, upload everything - but turn off screenshots / videos in Cypress (for now) so it doesn't 'actually' do anything!

I'm happy to leave screenshots off for now. :)

PR #60, opened 2023-08-15
VFB-28 Add comment box

DanWilkins2107:
Added a comment box on lists. This should fetch the current comment on the database, then allow the user to alter the comment and submit it. Also added is a reset comment feature.

### DIFF
--- a/src/app/lists/CommentBox.tsx
+++ b/src/app/lists/CommentBox.tsx
@@ -2,8 +2,7 @@
 
 import FreeFormTextInput from "@/components/DataInput/FreeFormTextInput";
 import supabase from "@/supabase";
-import { Button } from "@mui/material";
-import { set } from "cypress/types/lodash";
+import Button from "@mui/material/Button";
 import React, { useState } from "react";
 import styled from "styled-components";
 
--- a/src/app/lists/CommentBox.tsx
+++ b/src/app/lists/CommentBox.tsx
@@ -25,6 +25,9 @@ const ConfirmAndCancelButtons = styled(Button)`
     width: 60%;
     max-width: 10rem;
     margin: 0.5rem;
+    @media (min-width: 800px) {
+        width: 75%;
+    }
 `;
 
 const Header = styled.p`
@@ -64,6 +67,7 @@ const ButtonContainer = styled.div`
     @media (min-width: 800px) {
         flex-direction: column;
         align-items: center;
+        max-width: 25rem;
     }
 `;
 
--- a/src/app/lists/CommentBox.tsx
+++ b/src/app/lists/CommentBox.tsx
@@ -101,6 +101,7 @@ const CommentContainer: React.FC<CommentProps> = (props) => {
             </HeaderAndButtonContainer>
             <CommentBoxContainer>
                 <FreeFormTextInput
+                    label="Comment"
                     value={value}
                     onChange={onChangeSetValue}
                     multiline
--- a/src/app/lists/CommentBox.tsx
+++ b/src/app/lists/CommentBox.tsx
@@ -3,6 +3,7 @@
 import FreeFormTextInput from "@/components/DataInput/FreeFormTextInput";
 import supabase from "@/supabase";
 import { Button } from "@mui/material";
+import { set } from "cypress/types/lodash";
 import React, { useState } from "react";
 import styled from "styled-components";
 
@@ -38,9 +39,9 @@ const CommentBoxContainer = styled.div`
     width: 100%;
     margin: 1rem;
     min-width: 15rem;
-    & > * {
-        width: 100%;
-    }
+    display: flex;
+    flex-direction: column;
+    align-items: stretch;
     @media (min-width: 800px) {
         width: 75%;
     }
@@ -73,6 +74,7 @@ const SuccessText = styled.p`
 `;
 
 const CommentContainer: React.FC<CommentProps> = (props) => {
+    const [originalComment, setOriginalComment] = useState(props.originalComment);
     const [value, setValue] = useState(props.originalComment);
     const [errorMessage, setErrorMessage] = useState("");
     const [successMessage, setSuccessMessage] = useState("");
@@ -87,6 +89,7 @@ const CommentContainer: React.FC<CommentProps> = (props) => {
             setSuccessMessage("");
         } else {
             setErrorMessage("");
+            setOriginalComment(value);
             setSuccessMessage("Comment successfully updated.");
         }
     };
@@ -95,7 +98,7 @@ const CommentContainer: React.FC<CommentProps> = (props) => {
             <HeaderAndButtonContainer>
                 <h2>Comments</h2>
                 <ButtonContainer>
-                    <Button variant="outlined" onClick={() => setValue(props.originalComment)}>
+                    <Button variant="outlined" onClick={() => setValue(originalComment)}>
                         Reset
                     </Button>
                     <Button variant="contained" onClick={onSubmit}>
@@ -107,7 +110,11 @@ const CommentContainer: React.FC<CommentProps> = (props) => {
                 <FreeFormTextInput
                     label="Comments"
                     value={value}
-                    onChange={onChangeSetValue}
+                    onChange={(newValue) => {
+                        onChangeSetValue(newValue);
+                        setSuccessMessage("");
+                        setErrorMessage("");
+                    }}
                     multiline
                     maxRows={4}
                     minRows={4}
--- a/src/app/lists/CommentBox.tsx
+++ b/src/app/lists/CommentBox.tsx
@@ -72,9 +72,9 @@ const SuccessText = styled.p`
     font-size: 0.8rem;
 `;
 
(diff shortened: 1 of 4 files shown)


Var1377 (changes requested):
image

Mobile probably needs a bit more top padding, and see how reducing the size of the text box a little changes the look.

image

ChloeYiuSoftwire (changes requested):
UI comments
- Would it be better to just use `Reset` and `Submit` for the buttons? It looks fine on desktop, but the button looks a bit large in mobile when the words are split into two lines

src/app/lists/CommentBox.tsx
```diff
@@ -0,0 +1,117 @@
...
+    };
+    const onSubmit = async (): Promise<void> => {
+        const table = supabase.from("website_data");
+        const { error } = await table.update({ value: value }).eq("name", "lists_text");
+        if (error) {
+            setErrorMessage("There was an error uploading your changes to the database.");
+        } else {
+            setErrorMessage("");
+        }
+    };
```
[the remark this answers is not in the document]
DanWilkins2107: There is now a success message!

src/app/lists/CommentBox.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const HeaderAndButtonContainer = styled.div`
+    display: flex;
+    align-items: center;
+    width: 100%;
+    min-width: 8rem;
+    flex-direction: column;
+    gap: 1rem;
+    justify-content: space-between;
+    @media (min-width: 800px) {
```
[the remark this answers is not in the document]
DanWilkins2107: Don't think that is needed, it is quite common to use the same number repeatedly in media queries. Also it's not like that number is going to be wanted to be changed.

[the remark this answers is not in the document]
Var1377: @ChloeYiuSoftwire the standard practice for this sort of thing is to have predefined screen breakpoints (see something like how TailwindCSS manages it) so that actually would make sense. On the other hand, I was never too sure about why they should be standardized as I've never really come across an example where that would've made my life easier.

src/app/lists/CommentBox.tsx
```diff
@@ -0,0 +1,122 @@
...
+    padding: 0.5rem;
+`;
+
+const CommentBoxContainer = styled.div`
+    width: 100%;
+    margin: 1rem;
+    min-width: 15rem;
+    & > * {
+        width: 100%;
+    }
```
Var1377, line 43: try and use `display: flex` and `align-items: stretch`, I think that may work

src/app/lists/CommentBox.tsx
```diff
@@ -0,0 +1,122 @@
...
+    };
+    const onSubmit = async (): Promise<void> => {
+        const table = supabase.from("website_data");
+        const { error } = await table.update({ value: value }).eq("name", "lists_text");
+        if (error) {
+            setErrorMessage("There was an error uploading your changes to the database.");
+            setSuccessMessage("");
+        } else {
+            setErrorMessage("");
+            setSuccessMessage("Comment successfully updated.");
```
ChloeYiuSoftwire, line 90: Cool! However, it does not disappear when I type something else, and even when I reset it
Probably the text should only exist after I pressed success, and not when I do any more actions like clicking buttons or typing text

src/components/DataInput/FreeFormTextInput.tsx
```diff
@@ -17,6 +22,10 @@ const FreeFormTextInput: React.FC<Props> = (props) => {
             label={props.label}
             defaultValue={props.defaultValue}
             onChange={props.onChange}
+            value={props.value}
+            multiline={props.multiline}
+            maxRows={props.maxRows}
+            minRows={props.minRows}
```
Var1377, line 28: you can just `{...props}` instead of listing everything in props

Var1377 (changes requested):

src/app/lists/CommentBox.tsx
ChloeYiuSoftwire: Also, might be a helpful feature to have a message if the current comment display differs from the one on the server, so the user knows that it has been edited

src/app/lists/CommentBox.tsx
DanWilkins2107: Rip forgot flex-direction column

ChloeYiuSoftwire (changes requested):
UI comments
- In mobile mode, the gap between the buttons seem a bit too large
- In mobile mode, the top margin seems too small (Just a guess: maybe in desktop mode, your margin is set to be the gap between your comment box and the top, which looks good, but in mobile mode, it is the margin between the "Comment" header and the top)
- Let me know your preference for whether the desktop mode button should be vertical or horizontal (feels like vertical has too much gaps on the side, but horizontal has too much gaps on the top and bottom - I guess vertical might be better, but I haven't tried!)

src/app/lists/CommentBox.tsx
DanWilkins2107: But will add the first one.

src/app/lists/CommentBox.tsx
```diff
@@ -0,0 +1,129 @@
+"use client";
+
+import FreeFormTextInput from "@/components/DataInput/FreeFormTextInput";
+import supabase from "@/supabase";
+import { Button } from "@mui/material";
+import { set } from "cypress/types/lodash";
```
ChloeYiuSoftwire, line 6: Just a question what's this

src/app/lists/page.tsx
```diff
@@ -12,13 +12,22 @@ const fetchData = async (): Promise<ListRow[]> => {
...
+    const data = await supabase.from("website_data").select().eq("name", "lists_text");
+    if (data.data) {
+        return data.data![0].value;
+    }
+    return "";
+};
+
 const Lists = async (): Promise<ReactElement> => {
     const data = await fetchData();
+    const comment = await fetchComment();
```
Var1377, line 25: `Promise.all`

Var1377 (approved):

paopaosoftwire (approved):
Looking good!

src/app/lists/CommentBox.tsx
```diff
@@ -0,0 +1,128 @@
...
+const HeaderAndButtonContainer = styled.div`
+    display: flex;
+    align-items: center;
+    width: 100%;
+    min-width: 8rem;
+    flex-direction: column;
+    gap: 1rem;
+    justify-content: space-between;
+    @media (min-width: 800px) {
+        width: 15%;
```
TomCusack-Softwire, line 22: does this look wildly different at 799px and 801px?

src/app/lists/CommentBox.tsx
```diff
@@ -0,0 +1,128 @@
...
+    font-size: 0.8rem;
+`;
+
+const SuccessText = styled.p`
+    color: ${(props) => props.theme.primary.background[3]};
+    margin: 1rem 0 0;
+    font-size: 0.8rem;
+`;
+
+const CommentContainer: React.FC<CommentProps> = (props) => {
```
TomCusack-Softwire, line 75: I prefer converting `props` to `{originalComment}`, unless there's _lots_ of variables!

src/app/lists/page.tsx
```diff
@@ -7,18 +7,26 @@ import Title from "@/components/Title/Title";
...
 
-const fetchData = async (): Promise<ListRow[]> => {
-    const { data } = await supabase.from("lists").select();
-    return data ?? [];
+interface Props {
+    data: ListRow[];
+    comment: string;
+}
+
+const fetchData = (): Promise<Props> => {
```
TomCusack-Softwire, line 15: fetchData should be async - and I think I prefer the await syntax.
in this case, await the Promise.all and save that to a const, then return the object you want 'normally'
(also, what happens in case of error from db?)

TomCusack-Softwire (changes requested):

src/app/lists/CommentBox.tsx
DanWilkins2107: 799px - snip
800px+ snip

DanWilkins2107: Bottom one is 800px+

DanWilkins2107: Needs to be different as otherwise it looks odd when the comment box is really short

src/app/lists/page.tsx
DanWilkins2107: Error handling was not implemented before I picked up the ticket, will sort on VFB-22. Have changed to await and async.

src/app/lists/CommentBox.tsx
TomCusack-Softwire: this looks great!

src/app/lists/page.tsx
TomCusack-Softwire: I'd rather it be merged with 'ok' error handling which is then improved in VFB-22 - but I do think that the `?? []` does that fine for now!

TomCusack-Softwire (approved):

PR #62, opened 2023-08-16
Extend auth timeout

Var1377:
Normal login and logout times even locally were exceeding 500ms and causes an unpleasant white flash as themes update when not using client side routing so I've extended the timeout

### DIFF
--- a/src/app/auth.tsx
+++ b/src/app/auth.tsx
@@ -35,13 +35,12 @@ export const AuthRouting: React.FC<{ children: React.ReactNode }> = ({ children 
             router.push(redirect);
 
             // if Client Side Routing is not working, fallback to changing the window location
-            // This seemingly only happens on GH Actions
             setTimeout(() => {
                 const redirect = toRedirectTo();
                 if (redirect) {
                     window.location.pathname = redirect;
                 }
-            }, 500);
+            }, 2500);
         }
     };
 


DanWilkins2107 (approved):

daniel-conlan-softwire (approved):
Looks fine to me. Good job on this PR - lots of work done

paopaosoftwire (approved):

src/app/auth.tsx
```diff
@@ -35,13 +35,12 @@ export const AuthRouting: React.FC<{ children: React.ReactNode }> = ({ children
...
 
             // if Client Side Routing is not working, fallback to changing the window location
-            // This seemingly only happens on GH Actions
             setTimeout(() => {
                 const redirect = toRedirectTo();
                 if (redirect) {
                     window.location.pathname = redirect;
                 }
-            }, 500);
+            }, 2500);
```
TomCusack-Softwire, line 43: to confirm, is this 'after 2.5s, if not already redirected, redirect manually'?
why does this cause a white flash?

TomCusack-Softwire (approved):

src/app/auth.tsx
Var1377: Yep that's correct. When not using client side routing, Next does the same thing that it does on an initial page load and sends an SSRed page, which then gets styled and picks up themeing in local storage etc. all on hydration

PR #67, opened 2023-08-18
Add to button component

SamYap-Softwire:
NavBarButton - used in the navBar (changes styling for selected page)
LinkButton - a button that brings you to a different page 
PopUpButton - a button that opens a modal 

### DIFF
--- a/src/components/Buttons/LinkButton.tsx
+++ b/src/components/Buttons/LinkButton.tsx
@@ -3,7 +3,7 @@ import Button from "@mui/material/Button";
 import { ButtonWrap, UnstyledLink } from "@/components/Buttons/GeneralButtonParts";
 
 interface Props {
-    children: string;
+    children?: React.ReactNode;
     link: string;
 }
 
--- a/src/components/Buttons/LinkButton.tsx
+++ b/src/components/Buttons/LinkButton.tsx
@@ -1,31 +1,21 @@
-"use client";
-
 import React from "react";
-import Link from "next/link";
-import styled from "styled-components";
 import Button from "@mui/material/Button";
-import { usePathname } from "next/navigation";
+import { ButtonWrap, UnstyledLink } from "@/components/Buttons/GeneralButtonParts";
 
-const UnstyledLink = styled(Link)`
-    text-decoration: none;
-    display: contents;
-`;
-
-type LinkButtonProps = {
+interface Props {
+    children: string;
     link: string;
-    page: string;
-    onClick?: () => void;
-};
+}
 
-const LinkButton: React.FC<LinkButtonProps> = (props) => {
-    const pathname = usePathname();
-    const active = pathname?.startsWith(props.link) ?? false;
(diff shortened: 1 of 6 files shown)


src/components/Buttons/LinkButton.tsx
```diff
@@ -1,31 +1,21 @@
...
-    const active = pathname?.startsWith(props.link) ?? false;
+const LinkButton: React.FC<Props> = ({ children, link }) => {
     return (
-        <UnstyledLink key={props.page} href={props.link} prefetch={false}>
-            <Button color="primary" variant={active ? "contained" : "outlined"}>
-                {props.page}
-            </Button>
-        </UnstyledLink>
+        <ButtonWrap>
+            <UnstyledLink href={link} prefetch={false}>
```
Var1377, line 13: Do we always want `prefetch={false}`

src/components/Buttons/LinkButton.tsx
```diff
@@ -1,31 +1,21 @@
...
-    const pathname = usePathname();
-    const active = pathname?.startsWith(props.link) ?? false;
+const LinkButton: React.FC<Props> = ({ children, link }) => {
     return (
-        <UnstyledLink key={props.page} href={props.link} prefetch={false}>
-            <Button color="primary" variant={active ? "contained" : "outlined"}>
-                {props.page}
-            </Button>
-        </UnstyledLink>
+        <ButtonWrap>
```
Var1377, line 12: Could we just style the button instead of adding another div into the dom?

Var1377 (changes requested):

src/components/Buttons/LinkButton.tsx
SamYap-Softwire: discussed in person - will change all the prefetch to be true (by default)

src/components/Buttons/LinkButton.tsx
ChloeYiuSoftwire: Seems like some buttons don't use the wrap though

ChloeYiuSoftwire (approved):
Looks good

src/components/Buttons/LinkButton.tsx
SamYap-Softwire: this button wrap is to add margin so you can't directly style the button for that (especially since it's wrapped inside a link)

ChloeYiuSoftwire: I am thinking maybe it is also possible to pass in margin as a boolean prop rather than a separate div (?)

Var1377 (approved):

src/components/Buttons/LinkButton.tsx
TomCusack-Softwire: can you not add the margin on the UnstyledLink (and only have ButtonWrap for PopUpButton, meaning it can live in the same file and not have to be exported)?

TomCusack-Softwire (approved):

PR #70, opened 2023-08-18
VFB-22 Add error handling

DanWilkins2107:
Implementing Error Handling. Currently implemented is a global error page to catch error in the root layout, a general error page, and a not found error page.


src/app/error.tsx
```diff
@@ -0,0 +1,47 @@
...
+    return (
+        <ErrorCenterer>
+            <ErrorPanel elevation={5}>
+                <ErrorLargeText>OOPS!</ErrorLargeText>
+                <ErrorSecondaryText>
+                    We could not connect to the database at this time.
+                </ErrorSecondaryText>
+                <Button
+                    variant="outlined"
+                    onClick={() => {
```
Var1377, line 35: `onClick={retry}`

src/app/error.tsx
```diff
@@ -0,0 +1,47 @@
...
+    reset: () => void;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const Error: React.FC<ErrorProps> = ({ reset }) => {
+    const [ErrorMessage, SetErrorMessage] = useState("");
+
+    const retry = (): void => {
+        SetErrorMessage("Retrying...");
+        setTimeout(() => {
```
Var1377, line 22: Why wait for this?

src/app/error.tsx
```diff
@@ -0,0 +1,47 @@
+"use client";
+import React, { useState } from "react";
+import Button from "@mui/material/Button";
+import {
+    ErrorCenterer,
+    ErrorPanel,
+    ErrorLargeText,
+    ErrorSecondaryText,
+    ErrorRetryText,
+} from "./errorPageStyling";
```
Var1377, line 10: @ import

src/app/not-found.tsx
```diff
@@ -1,13 +1,19 @@
-import Link from "next/link";
+"use client";
 import React from "react";
+import { ErrorCenterer, ErrorPanel, ErrorLargeText, ErrorSecondaryText } from "./errorPageStyling";
+import LinkButton from "@/components/Buttons/LinkButton";
 
-const NotFound: React.FC<{}> = () => {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Var1377, line 6: You don't need this

src/app/not-found.tsx
```diff
@@ -1,13 +1,19 @@
-import Link from "next/link";
+"use client";
 import React from "react";
+import { ErrorCenterer, ErrorPanel, ErrorLargeText, ErrorSecondaryText } from "./errorPageStyling";
```
Var1377, line 3: @ import

Var1377 (changes requested):

src/app/clients/getClientsTableData.ts
```diff
@@ -49,8 +49,10 @@ const getProcessingData = async () => {
         .order("packing_datetime", { ascending: false })
         .order("timestamp", { ascending: false, foreignTable: "events" })
         .limit(1, { foreignTable: "events" });
-
-    return response.data ?? [];
+    if (error) {
+        throw new Error();
```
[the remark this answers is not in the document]
DanWilkins2107: Have added an error message!

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -43,8 +43,10 @@ export const getRawClientDetails = async (parcelId: string) => {
         )
         .eq("primary_key", parcelId)
         .single();
-
-    return response.data;
+    if (error) {
+        throw new Error();
```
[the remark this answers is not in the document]
DanWilkins2107: ^

src/app/error.tsx
```diff
@@ -0,0 +1,47 @@
...
+        setTimeout(() => {
+            reset();
+        }, 1000);
+    };
+    return (
+        <ErrorCenterer>
+            <ErrorPanel elevation={5}>
+                <ErrorLargeText>OOPS!</ErrorLargeText>
+                <ErrorSecondaryText>
+                    We could not connect to the database at this time.
```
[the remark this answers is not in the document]
DanWilkins2107: Changed :)

src/app/error.tsx
DanWilkins2107: So it has the text onscreen for a second, however this is after the error page rerenders, so should probably change the text to make it more clear.

daniel-conlan-softwire (changes requested):
Looks good in general, Dan - there's just some notes about the naming of some things

src/app/clients/add/databaseFunctions.ts
```diff
@@ -23,8 +23,8 @@ export const insertFamily = async (peopleArray: Person[], familyID: string): Pro
     const { status, error } = await supabase.from("families").insert(familyRecords);
 
     if (error !== null || Math.floor(status / 100) !== 2) {
-        throw Error(
-            `Error occurred whilst inserting into Families table. HTTP Code: ${status}, PostgreSQL Code: ${error?.code}. `
+        throw new Error(
+            "We could not connect to the database at this time. Please try again later."
```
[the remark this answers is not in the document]
DanWilkins2107: Good point!

src/app/not-found.tsx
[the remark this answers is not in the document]
DanWilkins2107: There is a not-found page, which is the page for when a page doesn't exist, then an error page for when an Error is thrown. They are very consistently styled, but are different pages!

src/app/error.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    const retry = (): void => {
+        SetErrorMessage("Retrying...");
+        setTimeout(() => {
+            reset();
+        }, 1000);
+    };
+    return (
+        <ErrorCenterer>
+            <ErrorPanel elevation={5}>
```
ChloeYiuSoftwire, line 28: The error modal is overflowing in mobile mode - maybe gotta set something like `width=max(...px, ...vw)` stuff

src/app/not-found.tsx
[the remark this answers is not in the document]
DanWilkins2107: Ah nvm just understood what you meant, will change.

src/app/error.tsx
```diff
@@ -0,0 +1,40 @@
...
+                <ErrorLargeText>OOPS!</ErrorLargeText>
+                <ErrorSecondaryText>{error.message}</ErrorSecondaryText>
```
ChloeYiuSoftwire, line 30: Just thinking, what if instead of general database error, there is an error that pops out because e.g. we missed an edge case then the program crashed in runtime, rather than error at these locations
In this case, I ain't sure but either the error returns a ugly formatted message, or it doesn't return
Would you want to handle these cases too?

src/app/errorPageStyling.tsx
```diff
@@ -0,0 +1,32 @@
...
+`;
+export const ErrorPanel = styled(Paper)`
+    max-width: 450px;
+    border-radius: 2rem;
+    padding: 5rem;
+    gap: 2rem;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    background-color: ${(props) => props.theme.main.background[0]};
```
ChloeYiuSoftwire, line 17: This is unnecessary - as I think by default in `global_styles.tsx` we have set all paper background and foreground to use `main.background[1]`
For consistency sake with all the papers maybe just use the default color and not set it?
P.S if you still want to use background[0] and override the setting, you should also add `color: ${(props) => props.theme/main/foreground[0]};` in this

src/app/lists/ListDataview.tsx
```diff
@@ -71,7 +71,7 @@ const ListsDataView: React.FC<Props> = (props) => {
     const [errorMsg, setErrorMsg] = useState<string | null>(null);
 
     if (props.data === null) {
-        throw new Error("No data found");
+        throw new Error("We could not find any data for the lists table.");
```
ChloeYiuSoftwire, line 74: This might be a bit misleading - because an empty list is a valid data but it still means there is no data found for this list table 
The data object can only be null when there is problem fetching data from the database, so maybe be more specific like "We could not fetch the list table data at this time. Please try again later"

src/app/not-found.tsx
```diff
@@ -1,13 +1,23 @@
...
+const Error: React.FC<{}> = () => {
     return (
-        <>
-            <p> No page found. </p>
-            <Link href="/clients"> Home </Link>
-        </>
+        <ErrorCenterer>
+            <ErrorPanel elevation={5}>
+                <ErrorLargeText>OOPS!</ErrorLargeText>
+                <ErrorSecondaryText>404 Error || This page does not exist</ErrorSecondaryText>
```
ChloeYiuSoftwire, line 16: Maybe just `This page does not exist` ? To make it less "techy" for general users

ChloeYiuSoftwire (changes requested):

src/app/error.tsx
DanWilkins2107: Good point, have changed.

src/app/error.tsx
DanWilkins2107: From the docs: global-error.js is the least granular error UI and can be considered "catch-all" error handling for the whole application. It is unlikely to be triggered often as root components are typically less dynamic, and other error.js boundaries will catch most errors. 
I think this means global-error should catch any edge cases.

src/app/errorPageStyling.tsx
DanWilkins2107: Removed

src/app/lists/ListDataview.tsx
DanWilkins2107: So when there is a problem fetching data from the database there is already an error thrown I think? So in theory this would only be for an occasion where all the data has been removed from the lists table I think

src/app/not-found.tsx
DanWilkins2107: 404 Error is common on error pages, if you google page not found error page that is usually what comes up!

src/app/clients/add/submitFormHelpers.ts
```diff
@@ -225,12 +208,16 @@ const revertClientInsert = async (primaryKey: string): Promise<void> => {
...
-// TODO VFB-22: Handle error from Family Updates
 const revertClientUpdate = async (initialRecords: ClientDatabaseUpdateRecord): Promise<void> => {
-    await supabase
+    const { error } = await supabase
         .from("clients")
         .update(initialRecords)
         .eq("primary_key", initialRecords.primary_key);
+    if (error) {
+        throw new Error(
+            "We could not revert the incomplete client update at this time. Please try again later."
```
SamYap-Softwire, line 218: Would this be user friendly? Since most of them might not know why client update needs to be reverted

src/app/errorPageStyling.tsx
```diff
@@ -0,0 +1,31 @@
+import Paper from "@mui/material/Paper";
+import styled from "styled-components";
```
SamYap-Softwire, line 2: "use client"?

src/app/global-error.tsx
```diff
@@ -0,0 +1,26 @@
...
+    ErrorLargeText,
+    ErrorSecondaryText,
+} from "@/app/errorPageStyling";
+import LinkButton from "@/components/Buttons/LinkButton";
+
+const GlobalError: React.FC<{}> = () => {
+    return (
+        <ErrorCenterer>
+            <ErrorPanel elevation={5}>
+                <ErrorLargeText>OOPS!</ErrorLargeText>
```
SamYap-Softwire, line 15: haha love this

SamYap-Softwire (commented):
Just a skim through the code. Seems good but will properly take a look at it tomorrow!

src/app/errorPageStyling.tsx
DanWilkins2107: Shouldn't be needed this is just a styling page!

src/app/global-error.tsx
```diff
@@ -0,0 +1,26 @@
+"use client";
```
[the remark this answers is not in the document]
DanWilkins2107: Error components must be client components according to next docs

src/app/not-found.tsx
```diff
@@ -1,13 +1,23 @@
-import Link from "next/link";
+"use client";
```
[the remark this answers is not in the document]
DanWilkins2107: ^^

SamYap-Softwire (approved):
Looks great! Just a couple of tiny comments but i think it's good to go once tests pass :)

src/app/error.tsx
```diff
@@ -0,0 +1,40 @@
...
+                <Button variant="outlined" onClick={retry}>
+                    Try again
+                </Button>
```
paopaosoftwire, line 33: Assuming this page will be shown every time an error is thrown. In that case would it make more sense for the retry button to be optional for each page? For example, if the add-client fails it is possible that we will have wrong data in the database so we probably don't want them to try again right away?

src/app/errorPageStyling.tsx
```diff
@@ -0,0 +1,31 @@
+import Paper from "@mui/material/Paper";
```
paopaosoftwire, line 1: Just want to know if you need a use client on this page?

src/app/clients/getClientsTableData.ts
```diff
@@ -52,8 +52,12 @@ const getProcessingData = async () => {
...
         .order("timestamp", { ascending: false, foreignTable: "events" })
         .limit(1, { foreignTable: "events" });
-
-    return response.data ?? [];
+    if (error) {
+        throw new Error(
+            "We could not fetch the clients' data at this time. Please try again later."
+        );
+    }
+    return data ?? [];
```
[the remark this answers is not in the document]
DanWilkins2107: I was thinking on the very odd situation there actually is no data in the table?

src/app/error.tsx
DanWilkins2107: I think we should be able to, as if it is a temporary loss of connection allowing the user to try again will save a user going to a developer!

src/app/errorPageStyling.tsx
DanWilkins2107: Nope as it is just a styling page, can't see why I would?

SamYap-Softwire: yeah i was confused yesterday for a second, but you need use client for styled components

SamYap-Softwire (changes requested):
missed something yesterday hehe ^^

src/app/clients/add/submitFormHelpers.ts
```diff
@@ -100,9 +94,8 @@ const updateChildren = async (children: Person[]): Promise<void> => {
             .eq("primary_key", child.primaryKey);
 
         if (errorExists(error, status)) {
-            // TODO VFB-22 Standardize Error Handling
-            throw Error(
-                `Error occurred whilst updating children in the Families table. HTTP Code: ${status}, PostgreSQL Code: ${error?.code}. `
+            throw new Error(
+                "We could not process the request at this time. Please try again later."
```
ChloeYiuSoftwire, line 98: This statement is used multiple times across the same file. Would it be nice to make it a const like `genericErrorMessage` or something alike?

ChloeYiuSoftwire (approved):
Seems nice ^^ Just some (maybe optional) minor changes then all good

daniel-conlan-softwire (approved):

src/app/calendar/parcelCalendarFunctions.ts
```diff
@@ -15,11 +15,9 @@ export const getParcelsWithCollectionDate = async (): Promise<ParcelWithClientNa
         .from("parcels")
         .select("*, clients ( full_name )")
         .not("collection_datetime", "is", null);
-
     if (error) {
-        throw new Error("Database error");
+        throw new Error("We were unable to fetch the parcels with the specified collection date.");
```
ChloeYiuSoftwire, line 19: Wait just realise it is wrong like the getParcelsWithCollectionDate only filtered those that have null collection_date (there is no specified date if it makes sense)

ChloeYiuSoftwire: Also do you want to standardize to have "... at this time. Please try again later."

paopaosoftwire (approved):

src/app/clients/add/submitFormHelpers.ts
DanWilkins2107: I agree, will do!

src/app/clients/add/submitFormHelpers.ts
```diff
@@ -116,10 +108,7 @@ const deleteChildren = async (children: Person[]): Promise<void> => {
             .eq("primary_key", child.primaryKey);
 
         if (errorExists(error, status)) {
-            // TODO VFB-22 Standardize Error Handling
-            throw Error(
-                `Error occurred whilst deleting children in the Families table. HTTP Code: ${status}, PostgreSQL Code: ${error?.code}. `
-            );
+            throw new Error(RequestErrorMessage);
```
Var1377, line 111: For these probably just make a `RequestError` class that extends error so you can just throw that. It also means you can define a more meaningful constructor and unify formatting for error messages, and users elsewhere can specifically check for and catch that error.

Var1377 (changes requested):
The UI parts look good, but you should make it easier for other devs to create consistent errors by encapsulating them into classes, especially as downstream code can specifically catch certain errors if they want to. Something like a `DatabaseError` and/or `RequestError` etc.

src/app/errorStylingandMessages.tsx
```diff
@@ -0,0 +1,34 @@
+import Paper from "@mui/material/Paper";
```
SamYap-Softwire, line 1: might be good to have "use client" here 
(though it's technically implicitly set to client side already)

src/app/themes.tsx
```diff
@@ -152,7 +152,9 @@ interface Props {
 
 export const ThemeUpdateContext = createContext((dark: boolean): void => {
     throw new Error(
-        `attempted to set theme outside of a ThemeUpdateContext.Provider: Dark: ${dark}`
+        `Attempted to set theme to ${
```
SamYap-Softwire, line 155: is this something shown to users? (what do you think of having a message on error pages for users only (user friendly) and maybe a troubleshooting icon that opens up to dev-friendly error messages?)

SamYap-Softwire (approved):
looks good! just a few comments :)

src/app/not-found.tsx
DanWilkins2107: I kinda like the two lines ngl :/

src/app/themes.tsx
DanWilkins2107: It is, but also should never really happen? Tbh I'm not sure when it would ever come up, but also it was there initially so I didn't want to remove it!

src/app/clients/add/submitFormHelpers.ts
```diff
@@ -80,11 +75,7 @@ const deleteAdultMembers = async (
         .limit(count);
 
     if (errorExists(error, status)) {
-        // TODO VFB-22 Standardize Error Handling
-        console.error(error);
-        throw Error(
-            `Error occurred whilst updating adults in the Families table. HTTP Code: ${status}, PostgreSQL Code: ${error?.code}. `
-        );
+        throw new RequestError();
```
Var1377, line 78: Make a database error class that takes in a supabase error in the constructor so that the constructed error message can take values from that object. It's also clearer for downstream devs

Var1377 (changes requested):

src/app/themes.tsx
Var1377: @SamYap-Softwire we generally don't want this, usually in production something like this would use a library like Sentry (https://docs.sentry.io/platforms/javascript/) so that we know when and where exceptions happen. (This error specifically should never be shown to users, we have a context provider at the top of the tree)

Var1377 (approved):

src/app/errorClasses.ts
```diff
@@ -0,0 +1,10 @@
+export class DatabaseError extends Error {
+    constructor(errorCategory: DatabaseOperation, faultyArea = "") {
```
TomCusack-Softwire, line 2: surprised faultyArea didn't want to be typed as a string

src/app/lists/page.tsx
```diff
@@ -12,12 +13,25 @@ interface Props {
...
+    }
+    return data;
+};
+
+const fetchComment = async (): Promise<string> => {
+    const { data, error } = await supabase.from("website_data").select().eq("name", "lists_text");
+    if (error) {
+        throw new DatabaseError("fetch", "lists comment");
+    }
+    return data?.[0]?.value ?? "";
```
TomCusack-Softwire, line 29: the `?.` should be unnecessary - data should be non-null if error is null

src/app/lists/page.tsx
```diff
@@ -12,12 +13,25 @@ interface Props {
...
+};
+
 const fetchData = async (): Promise<Props> => {
-    const values = await Promise.all([
-        supabase.from("lists").select(),
-        supabase.from("website_data").select().eq("name", "lists_text"),
-    ]);
-    return { data: values[0].data ?? [], comment: values[1].data![0].value ?? "" };
+    const values = await Promise.all([fetchList(), fetchComment()]);
```
TomCusack-Softwire, line 33: is it worth unpacking this? e.g. `const {data, comment} = await ...`

src/pdf/ShoppingList/databaseFetch.ts
```diff
@@ -1,54 +1,55 @@
...
 
 export const fetchComment = async (): Promise<string> => {
     const { data, error } = await supabase.from("website_data").select().eq("name", "lists_text");
-    if (error !== null) {
-        throw Error(`${error.code}`);
+    if (error) {
+        throw new DatabaseError("fetch", "lists comment");
     }
-    return data![0].value ?? "";
+    return data?.[0]?.value ?? "";
```
TomCusack-Softwire, line 54: same here - `?.` _should_ be redundant, leaving `data[0]?.value`

TomCusack-Softwire (approved):
Looks great, thanks Dan!

src/app/errorClasses.ts
DanWilkins2107: Same tbf

PR #72, opened 2023-08-18
Refactor AddClients Form

SamYap-Softwire:
Refactored AddClients form so that it will be easy to make an EditClients form

Some functions might not be used right now but will be in the Edit Clients

### DIFF
--- a/src/app/clients/add/AddClientForm.tsx
+++ b/src/app/clients/add/AddClientForm.tsx
@@ -4,7 +4,6 @@ import React, { useEffect, useState } from "react";
 import { useRouter } from "next/navigation";
 import { booleanGroup } from "@/components/DataInput/inputHandlerFactories";
 import {
-    checkboxGroupToArray,
     checkErrorOnSubmit,
     Errors,
     Fields,
@@ -34,15 +33,17 @@ import DeliveryInstructionsCard from "@/app/clients/add/formSections/DeliveryIns
 import ExtraInformationCard from "@/app/clients/add/formSections/ExtraInformationCard";
 import AttentionFlagCard from "@/app/clients/add/formSections/AttentionFlagCard";
 import SignpostingCallCard from "@/app/clients/add/formSections/SignpostingCallCard";
-import {
-    ClientDatabaseInsertRecord,
-    deleteFailedInsert,
-    insertClient,
-    insertFamily,
-} from "@/app/clients/add/databaseFunctions";
 import Button from "@mui/material/Button";
+import { submitFormAddClients, submitFormEditClients } from "@/app/clients/add/databaseFunctions";
 
-interface AddClientFields extends Fields {
+interface Props {
+    initialFields: Fields;
+    initialFormErrors: FormErrors;
+    editMode: boolean;
+    clientID?: string;
+}
+
+export interface AddClientFields extends Fields {
     fullName: string;
     phoneNumber: string;
     addressLine1: string;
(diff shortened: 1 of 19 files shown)


src/app/clients/add/formSections/NumberAdultsCard.tsx
```diff
@@ -64,24 +71,27 @@ const NumberAdultsCard: React.FC<CardProps> = ({
...
                 />
                 <FreeFormTextInput
                     error={errorExists(formErrors.adults)}
                     helperText={errorText(formErrors.adults)}
                     label="Prefer Not To Say"
+                    defaultValue={getNumberAdultsDefault(fields.adults, "other")}
                     onChange={getNumberAdults(fieldSetter, errorSetter, fields.adults, "other")}
                 />
-            </GappedDiv>
+            </>
```
DanWilkins2107, line 94: How come we don't need the gapped div anymore out of interest?

DanWilkins2107 (approved):
Just a couple small changes but looks good :)

src/app/clients/add/databaseFunctions.ts
```diff
@@ -1,11 +1,20 @@
...
 type FamilyDatabaseInsertRecord = InsertSchema["families"];
+type FamilyDatabaseUpdateRecord = UpdateSchema["families"];
+
+export type SubmitFormHelper = (
+    fields: Fields,
+    router: AppRouterInstance,
+    primaryKey?: string
+) => void;
 export type ClientDatabaseInsertRecord = InsertSchema["clients"];
+export type ClientDatabaseUpdateRecord = UpdateSchema["clients"];
```
ChloeYiuSoftwire, line 14: Yee I was referring to the type above - a quick glimpse of `database_types_file.ts` kinda show me they are fairly similar (for client it is identical, for other forms maybe some param become optional in update?) - just thinking is there ways to combine them

src/app/clients/add/formSections/NumberAdultsCard.tsx
```diff
@@ -13,7 +13,14 @@ import {
 } from "@/components/Form/formFunctions";
 import GenericFormCard from "@/components/Form/GenericFormCard";
 import { selectChangeEventHandler } from "@/components/DataInput/inputHandlerFactories";
-import { GappedDiv } from "@/components/Form/formStyling";
+
+const getNumberAdultsDefault = (adults: Person[], gender: string): string | undefined => {
+    const personIndex = adults.findIndex((person) => person.gender === gender);
+    return adults[personIndex].quantity === 0
+        ? undefined
```
ChloeYiuSoftwire, line 20: Just want to check 
What if it is actively being put as a gender being 0 in a household - do you still want to display nothing or you want to explicitly display it is 0

ChloeYiuSoftwire: Or it is just non differentiable in our current database because all number are initialized as 0 and there is no "null"?

ChloeYiuSoftwire: Would explicitly showing it is 0 important though - otherwise people might thought it is just not filled rather than assuming maybe a household has no specific gender

ChloeYiuSoftwire (changes requested):

src/app/clients/add/formSections/NumberAdultsCard.tsx
SamYap-Softwire: deleted by accident :)

src/app/clients/add/databaseFunctions.ts
```diff
@@ -45,6 +54,190 @@ export const insertClient = async (
...
+
+const countAdultMember = async (familyID: string, gender: string): Promise<number> => {
+    const { count, error } = await supabase
+        .from("families")
+        .select("*", { count: "exact", head: true })
+        .eq("family_id", familyID)
+        .eq("gender", gender)
+        .is("age", null);
+
+    if (error !== null || count === null) {
```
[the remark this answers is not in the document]
SamYap-Softwire: if there are no adult members, it will return count = 0 instead of count === null (which is returned when there is an error like the table doesn't exist). The reason I have both `error !== null` and `count === null`.

src/app/clients/add/formSections/BabyProductCard.tsx
```diff
@@ -27,6 +28,13 @@ const getBaby = (fieldSetter: FieldSetter, errorSetter: ErrorSetter): selectChan
     };
 };
 
+const getBabyDefaultValue = (value: boolean | null): string => {
+    if (value) {
+        return "Yes";
+    }
+    return value === null ? "Don't Know" : "No";
+};
+
```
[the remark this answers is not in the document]
SamYap-Softwire: I think this is a little hard to read for me

src/app/clients/add/formSections/NumberAdultsCard.tsx
SamYap-Softwire: 0 and not filled are basically the same in this case, because it defaults to 0. And the form makes it so that there will be at least one adult in the entire household. And 0 means there won't be any records added to the database

SamYap-Softwire: do let me know if you need more explanation

src/app/clients/add/databaseFunctions.ts
SamYap-Softwire: I think it's better to assume they maynot always be identical in the future

src/app/clients/add/AddClientForm.tsx
```diff
@@ -183,13 +128,10 @@ const AddClientForm: React.FC<{}> = () => {
...
-            try {
-                await insertFamily([...fields.adults, ...fields.children], ids.family_id);
-                router.push("/clients");
-            } catch (error) {
-                await deleteFailedInsert(ids.primary_key);
-                throw error;
+            if (editMode) {
+                await submitFormEditClients(fields, router, clientID);
+            } else {
+                await submitFormAddClients(fields, router);
```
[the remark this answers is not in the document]
SamYap-Softwire: I don't think that is necessary as the only difference is where the "if statement" will be in. I personally prefer them to be different functions and have the if statement in the addclientform because that will make it easy to change the specific functions

ChloeYiuSoftwire (approved):

SamYap-Softwire:
requested changes done. this PR will be included in PR 78 (edit client form)

PR #74, opened 2023-08-18
Add Parcels button on Clients Page

SamYap-Softwire:
Currently, when the client's row in the popup is clicked, it will redirect to clients/[id] but there is no page for that yet. 

Also had to make some minor changes for the ActionBar which uses parcel.collectionDatetime but that wasn't in ClientsTableRow.

### DIFF
--- a/src/app/clients/AddParcelsButton.tsx
+++ b/src/app/clients/AddParcelsButton.tsx
@@ -1,0 +1,79 @@
+"use client";
+
+import React from "react";
+import { ClientsTableRow } from "@/app/clients/getClientsTableData";
+import { useRouter } from "next/navigation";
+import { useState } from "react";
+import PopUpButton from "@/components/Buttons/PopUpButton";
+import LinkButton from "@/components/Buttons/LinkButton";
+import Button from "@mui/material/Button";
+import TableSurface from "@/components/Tables/TableSurface";
+import Table, { TableHeaders } from "@/components/Tables/Table";
+import styled from "styled-components";
+
+interface Props {
+    data: ClientsTableRow[];
+}
+
+const PopUp = styled.div`
+    display: flex;
+    flex-direction: column;
+    text-align: center;
+    padding: 2rem;
+`;
+
+const headers: TableHeaders = [
+    ["fullName", "Name"],
+    ["familyCategory", "Family"],
+    ["addressPostcode", "Postcode"],
+];
+
+const styleOptions = {
+    fullName: {
+        minWidth: "8rem",
+    },
+    familyCategory: {
+        hide: 550,
+    },
+    addressPostcode: {
+        hide: 800,
+    },
+};
+
+const AddParcelsButton: React.FC<Props> = ({ data }) => {
+    const [existingClientsView, setExistingClientsView] = useState(false);
+    const router = useRouter();
+
+    return (
+        <PopUpButton displayText="Add Parcel">
+            <>
+                {!existingClientsView ? (
+                    <PopUp>
+                        <LinkButton link="/clients/add">New Client</LinkButton>
+                        <Button onClick={() => setExistingClientsView(true)}>
+                            Existing Client
+                        </Button>
+                    </PopUp>
+                ) : (
+                    <>
+                        <TableSurface>
+                            <Table
+                                data={data}
+                                headerKeysAndLabels={headers}
+                                onRowClick={(row) => router.push(`/clients/${row.data.primaryKey}`)}
+                                sortable={true}
+                                pagination={false}
+                                checkboxes={false}
+                                columnStyleOptions={styleOptions}
+                                headerFilters={["fullName"]}
+                            />
+                        </TableSurface>
+                        <Button onClick={() => setExistingClientsView(false)}>Back</Button>
+                    </>
+                )}
+            </>
+        </PopUpButton>
+    );
+};
+
+export default AddParcelsButton;
--- a/src/app/clients/AddParcelsButton.tsx
+++ b/src/app/clients/AddParcelsButton.tsx
@@ -71,32 +71,28 @@ const AddParcelsButton: React.FC<Props> = ({ data }) => {
 
     return (
         <PopUpButton displayText="Add Parcel">
-            <>
-                {!existingClientsView ? (
-                    <PopUp>
-                        <LinkButton link="/clients/add">New Client</LinkButton>
-                        <Button onClick={() => setExistingClientsView(true)}>
-                            Existing Client
-                        </Button>
-                    </PopUp>
-                ) : (
-                    <>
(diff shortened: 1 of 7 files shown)


src/app/clients/AddParcelsButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const showClients = (data: ClientsTableRow[]): ClientsListRow[] => {
+    const primaryKeys: string[] = [];
+    const clientsData: ClientsListRow[] = [];
+
+    for (const datum of data) {
+        if (!primaryKeys.includes(datum.primaryKey)) {
+            const { primaryKey, fullName, familyCategory, addressPostcode } = datum;
+            clientsData.push({ primaryKey, fullName, familyCategory, addressPostcode });
```
DanWilkins2107, line 58: can this not be `clients.Data.push(datum)`

src/app/clients/AddParcelsButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+                                pagination={false}
+                                checkboxes={false}
+                                columnStyleOptions={styleOptions}
+                                headerFilters={["fullName"]}
+                            />
+                        </TableSurface>
+                        <Button onClick={() => setExistingClientsView(false)}>Back</Button>
+                    </>
+                )}
+            </>
```
DanWilkins2107, line 99: Is this fragment needed?

src/app/clients/AddParcelsButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+                        </Button>
+                    </PopUp>
+                ) : (
+                    <>
+                        <TableSurface>
+                            <Table
+                                data={clientData}
+                                headerKeysAndLabels={headers}
+                                onRowClick={(row) => router.push(`/clients/${row.data.primaryKey}`)}
+                                sortable={true}
```
DanWilkins2107, line 89: Can remove the `={true}` (and also maybe remove all the `={false}` bits and just not have anything there!)

DanWilkins2107 (approved):
A couple of stylistic changes only.

src/app/clients/AddParcelsButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+                                pagination={false}
```
paopaosoftwire, line 90: Dont you want pagination to be true? Im assume it will show all existing clients and over time this data is probably gonna be more than 10 rows?

src/app/clients/AddParcelsButton.tsx
```diff
@@ -0,0 +1,104 @@
+"use client";
```
paopaosoftwire, line 1: Does this have to be a use client? Its not currently using any client side hooks or is useRouter a client only thing?

src/app/clients/AddParcelsButton.tsx
SamYap-Softwire: no, because I only want some of the propersties of datum and not the rest

src/app/clients/AddParcelsButton.tsx
SamYap-Softwire: nope, nice catch

src/app/clients/AddParcelsButton.tsx
SamYap-Softwire: it's using styled components so i think it does

src/app/layout.tsx
```diff
@@ -32,4 +32,6 @@ export const metadata: Metadata = {
     description: "Providing foodbank services to clients in south London.",
 };
 
+export const dynamic = "force-dynamic";
```
TomCusack-Softwire, line 35: try removing this and see if it still works? main has this implemented in _just_ the PDF downloader... which hopefully fixes things!

src/app/clients/getClientsTableData.ts
```diff
@@ -57,25 +59,26 @@ const getProcessingData = async () => {
...
             requiresFollowUpPhoneCall: client.signposting_call_required,
             fullName: client.full_name,
             familyCategory: familyCountToFamilyCategory(client.family.length),
             addressPostcode: client.address_postcode,
             collectionCentre: parcel.collection_centre ?? "-",
-            collectionDatetime: parcel.collection_datetime,
             congestionChargeApplies: congestionChargeDetails[index].congestionCharge,
             packingDate: formatDatetimeAsDate(parcel.packing_datetime),
             packingTimeLabel: datetimeToPackingTimeLabel(parcel.packing_datetime),
+            collectionDatetime: parcel.collection_datetime ?? "-",
```
TomCusack-Softwire, line 81: why the reorganisation?

src/app/clients/AddParcelsButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const AddParcelsButton: React.FC<Props> = ({ data }) => {
+    const [existingClientsView, setExistingClientsView] = useState(false);
+    const router = useRouter();
+
+    const clientData = showClients(data);
+
+    return (
+        <PopUpButton displayText="Add Parcel">
+            {!existingClientsView ? (
```
TomCusack-Softwire, line 74: can you flip this to ne just `existingClientsView ? ... : ...`? (should be able to do it in your IDE)

src/app/clients/AddParcelsButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+                <PopUp>
+                    <LinkButton link="/clients/add">New Client</LinkButton>
+                    <Button onClick={() => setExistingClientsView(true)}>Existing Client</Button>
```
TomCusack-Softwire, line 77: surprised you didn't just use a new page to link to, but this also works (quite well from what I remember!)

TomCusack-Softwire (approved):
Nothing major that I can see, good job!

src/app/clients/AddParcelsButton.tsx
TomCusack-Softwire: I assumed that in HTML if you don't specify it goes to false and if you do it goes to true, e.g.
`<HTML sortable>` vs `<HTML>` for a sortable=true/false - could be wrong!

src/app/clients/getClientsTableData.ts
SamYap-Softwire: there was a bug in Actions and status bar which was on main - the collectionDatetime prop was called there but was not defined here, so i just added it (for lint_fix to work)

src/app/clients/AddParcelsButton.tsx
SamYap-Softwire: no particular reason for that, I just _feel_ like they belong in the same page. wouldn't mind changing that in the future though :)

PR #83, opened 2023-08-21
Refactor Table and Table Filters

Var1377:
The table is now fully typed, and the API has slightly changed.


src/app/lists/ListDataview.tsx
```diff
@@ -19,49 +14,81 @@ import TooltipCell from "@/app/lists/TooltipCell";
...
-            );
-        };
-        // Column Style Function
-        listsColumnStyleOptions[headerKey] = {
-            minWidth: "9rem",
-            center: true,
-        };
-    }
-});
+export const headerKeysAndLabels = [
```
ChloeYiuSoftwire, line 41: The original one has a type `[string, string][]` ? Why is it not needed now

src/app/lists/ListDataview.tsx
```diff
@@ -19,49 +14,81 @@ import TooltipCell from "@/app/lists/TooltipCell";
...
+    "1": QuantityAndNotes;
+    "2": QuantityAndNotes;
+    "3": QuantityAndNotes;
+    "4": QuantityAndNotes;
+    "5": QuantityAndNotes;
+    "6": QuantityAndNotes;
+    "7": QuantityAndNotes;
+    "8": QuantityAndNotes;
+    "9": QuantityAndNotes;
+    "10": QuantityAndNotes;
```
ChloeYiuSoftwire, line 28: Is it "10+" or "10" because it is 10 or more right

src/app/lists/ListDataview.tsx
```diff
@@ -74,6 +101,23 @@ const ListsDataView: React.FC<Props> = (props) => {
...
+            "1": { quantity: row["1_quantity"], notes: row["1_notes"] },
+            "2": { quantity: row["2_quantity"], notes: row["2_notes"] },
+            "3": { quantity: row["3_quantity"], notes: row["3_notes"] },
+            "4": { quantity: row["4_quantity"], notes: row["4_notes"] },
+            "5": { quantity: row["5_quantity"], notes: row["5_notes"] },
+            "6": { quantity: row["6_quantity"], notes: row["6_notes"] },
+            "7": { quantity: row["7_quantity"], notes: row["7_notes"] },
+            "8": { quantity: row["8_quantity"], notes: row["8_notes"] },
+            "9": { quantity: row["9_quantity"], notes: row["9_notes"] },
+            "10": { quantity: row["10_quantity"], notes: row["10_notes"] },
```
ChloeYiuSoftwire, line 116: Same here, is it `10+_quantity`

src/app/clients/ActionBar.cy.tsx
```diff
@@ -9,34 +9,39 @@ describe("Clients - Action Bar", () => {
...
+                requiresFollowUpPhoneCall: false,
+                flaggedForAttention: false,
+            },
+            voucherNumber: "123456789",
         },
         {
             primaryKey: "primaryKey2",
             addressPostcode: "AB1 aaaa2CD",
-            collectionCentre: "Centraaaae 1",
+            deliveryCollection: { collectionCentre: "Centraaaae 1", congestionChargeApplies: true },
```
DanWilkins2107, line 32: How come congestionChargeApplies isn't just optional props?

DanWilkins2107: Also what was wrong with `Centre 2`?

src/app/clients/ClientsPage.tsx
```diff
@@ -48,24 +48,25 @@ const collectionCentreToAbbreviation = (
...
     ["deliveryCollection", ""],
-    ["packingDate", "Packing Date"],
+    ["packingDatetime", "Packing Date"],
     ["packingTimeLabel", "Time"],
     ["lastStatus", "Last Status"],
+    ["voucherNumber", "Voucher"],
 ];
 
-const toggleableHeaders = [
+const toggleableHeaders: readonly (keyof ClientsTableRow)[] = [
```
DanWilkins2107, line 62: what does readonly do that isn't the same as const?

src/app/clients/ClientsPage.tsx
```diff
@@ -48,24 +48,25 @@ const collectionCentreToAbbreviation = (
...
 
-export const clientTableHeaderKeysAndLabels: TableHeaders = [
-    ["iconsColumn", ""],
+export const clientTableHeaderKeysAndLabels: TableHeaders<ClientsTableRow> = [
     ["fullName", "Name"],
     ["familyCategory", "Family"],
     ["addressPostcode", "Postcode"],
     ["deliveryCollection", ""],
-    ["packingDate", "Packing Date"],
+    ["packingDatetime", "Packing Date"],
```
DanWilkins2107, line 56: Why do we have packingDateTime if we also have packingTimeLabel?

src/components/Tables/Table.cy.tsx
```diff
@@ -67,14 +72,14 @@ const data: Datum[] = [
...
-];
+] as const;
 
 interface TestTableProps {
     checkboxes?: boolean;
-    toggleableHeaders?: string[];
+    toggleableHeaders?: (keyof TestData)[];
 }
 
 const Component: React.FC<TestTableProps> = ({ checkboxes = true, toggleableHeaders }) => {
```
ChloeYiuSoftwire, line 85: Why do you have to passs in checkboxes and toggleableHeaders as a prop? I don't see you feeding in checkboxes and toggleableHeaders in any of the `<Component />` that you have mounted below?

src/components/Tables/Table.tsx
```diff
@@ -256,11 +238,10 @@ const Table: React.FC<Props> = ({
         });
     }
 
-    // TODO VFB-23 Implement conditional styling: center icon when only option selected, grid otherwise
-    if (reorderable || onEdit || onDelete) {
+    if (reorderable || onEdit) {
         columns.unshift({
-            name: "",
-            cell: (row: Row) => {
+            name: <p>Sort</p>,
```
ChloeYiuSoftwire, line 243: Why is it named sort? I think the column does more than sorting though there is also edit and delete

ChloeYiuSoftwire (changes requested):
The comment box becomes so short and the scrollbar disappeared?

src/app/lists/ListDataview.tsx
```diff
@@ -19,49 +14,81 @@ import TooltipCell from "@/app/lists/TooltipCell";
...
+    minWidth: "10rem",
+    center: true,
+};
+
+const listsColumnStyleOptions = {
+    itemName: {
+        minWidth: "8rem",
+    },
+    "1": defaultStyleOptions,
+    "2": defaultStyleOptions,
```
DanWilkins2107, line 82: Same as above

src/app/lists/ListDataview.tsx
```diff
@@ -74,6 +101,23 @@ const ListsDataView: React.FC<Props> = (props) => {
         throw new Error("No data found");
     }
 
+    const rows = props.data.map((row) => {
+        const newRow: ListRow = {
+            itemName: row.item_name,
+            "1": { quantity: row["1_quantity"], notes: row["1_notes"] },
+            "2": { quantity: row["2_quantity"], notes: row["2_notes"] },
+            "3": { quantity: row["3_quantity"], notes: row["3_notes"] },
```
DanWilkins2107, line 109: Same as above

src/components/Tables/FilterAccordion.tsx
```diff
@@ -5,17 +5,17 @@ import { Accordion, Checkbox, AccordionSummary, AccordionDetails } from "@mui/ma
 import styled from "styled-components";
 import { faChevronDown } from "@fortawesome/free-solid-svg-icons";
 import Icon from "@/components/Icons/Icon";
+import { TableHeaders } from "./Table";
```
DanWilkins2107, line 8: @ import?

src/components/Tables/Filters.tsx
```diff
@@ -0,0 +1,100 @@
...
+    key: N
+): string => {
+    return headers.find(([headerKey]) => headerKey === key)![1];
+};
+
+const StyledFilterBar = styled.input`
+    font-size: 14px;
+    width: 15rem;
+    overflow: visible;
+    box-shadow: none;
```
DanWilkins2107, line 41: Is this needed?

src/components/Tables/Filters.tsx
```diff
@@ -0,0 +1,100 @@
...
+};
+
+export const headerLabelFromKey = <Data, N extends keyof Data>(
+    headers: TableHeaders<Data>,
+    key: N
+): string => {
+    return headers.find(([headerKey]) => headerKey === key)![1];
+};
+
+const StyledFilterBar = styled.input`
```
DanWilkins2107, line 37: Any reason for using the normal input not FreeFormTextInput?

src/app/clients/ClientsPage.tsx
```diff
@@ -105,52 +109,49 @@ const ClientsPage: React.FC<Props> = (props) => {
...
+                    <CollectionIcon
+                        color={theme.main.largeForeground[0]}
+                        collectionPoint={data.collectionCentre}
+                    />
+                    {collectionCentreToAbbreviation(data.collectionCentre)}
+                </>
+            );
+        },
+        packingDatetime: (data: ClientsTableRow["packingDatetime"]): string => {
+            return data ? new Date(data).toLocaleDateString() : "";
```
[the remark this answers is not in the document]
Var1377: The types claim `string | null`so that's what I'm going of.

src/components/Tables/Table.tsx
```diff
@@ -13,22 +13,21 @@ import {
...
 import IconButton from "@mui/material/IconButton/IconButton";
 import Icon from "@/components/Icons/Icon";
+import { Filter, headerLabelFromKey, textFilter } from "@/components/Tables/Filters";
 
-export interface Datum {
-    [headerKey: string]: string | number | boolean | null;
-}
-
-export type TableHeaders = [string, string][];
+export type TableHeaders<Data> = readonly (readonly [keyof Data, string])[];
```
DanWilkins2107, line 18: Why readonly (readonly ?

src/components/Tables/Table.tsx
```diff
@@ -13,22 +13,21 @@ import {
...
 }
 
-export type ColumnDisplayFunction = (row: Row) => ReactNode;
-export type TableColumnDisplayFunctions = { [headerKey: string]: ColumnDisplayFunction };
+export type ColumnDisplayFunction<T> = (data: T) => ReactNode;
 
-export type OnRowClickFunction = (row: Row, e: React.MouseEvent<Element, MouseEvent>) => void;
+export type OnRowClickFunction<Data> = (
+    row: Row<Data>,
+    e: React.MouseEvent<Element, MouseEvent>
```
DanWilkins2107, line 29: can we call this event instead of e? or is there a specific reason?

DanWilkins2107: Also surprised there isn't a linting error lol

src/components/Tables/Table.tsx
```diff
@@ -41,86 +40,59 @@ export interface ColumnStyleOptions {
     allowOverflow?: boolean;
     hide?: number;
 }
-export type TableColumnStyleOptions = { [headerKey: string]: ColumnStyleOptions };
 
-interface Props {
-    data: Datum[];
-    headerKeysAndLabels: TableHeaders;
+export interface SortOptions<Data, N extends keyof Data> {
+    key: N;
```
DanWilkins2107, line 45: What is N?

src/components/Tables/Table.tsx
```diff
@@ -41,86 +40,59 @@ export interface ColumnStyleOptions {
...
+    cmp: (a: Data[N], b: Data[N]) => number;
```
DanWilkins2107, line 46: What is cmp?

src/components/Tables/Table.tsx
```diff
@@ -41,86 +40,59 @@ export interface ColumnStyleOptions {
...
+    const element = (
         <>
             {columnDisplayFunctions[headerKey]
-                ? columnDisplayFunctions[headerKey](row)
+                ? columnDisplayFunctions[headerKey]!(row.data[headerKey])
                 : row.data[headerKey]}
         </>
     );
+
+    if (!React.isValidElement(element)) {
```
DanWilkins2107, line 87: Confused as to why we need this?

DanWilkins2107: Like in what case would we not have a valid element?

src/app/lists/ListDataview.tsx
Var1377: That would be the wrong type as this is now [keyof ListRow, string]. Have a read of https://blog.logrocket.com/complete-guide-const-assertions-typescript/ if you're unsure of what the `as const` does here.

src/components/Tables/Table.tsx
```diff
@@ -206,7 +189,7 @@ const Table: React.FC<Props> = ({
         return {
             name: headerName,
             selector: (row) => row.data[headerKey] ?? "",
-            sortable: sortable,
+            sortable: sortable?.some((sortKey) => sortKey === headerKey),
```
DanWilkins2107, line 192: Do we need a description of what this is in a comment or no? Tbf it's easy to understand but did need a google from me

src/app/lists/ListDataview.tsx
Var1377: no, these are the keys on the schema and there it's called `10_quantity` and `10_notes`

src/components/Tables/Table.tsx
```diff
@@ -311,35 +292,37 @@ const Table: React.FC<Props> = ({
...
-    };
+    const rows = data.map((data, index) => ({ rowId: index, data }));
 
-    const handleClear = (): void => {
-        if (filterText) {
-            setFilterText({});
-        }
+    const shouldFilterRow = (row: Row<Data>): boolean => {
+        return filters.every((filter) => !filter.shouldFilter(row.data, filter.state));
+        // return true;
```
DanWilkins2107, line 299: is this comment needed? I have no clue what it is referring to so if so it needs more description

src/components/Tables/TableFilterBar.tsx
```diff
@@ -1,70 +1,46 @@
 "use client";
 
 import React from "react";
-import styled from "styled-components";
 import FilterAccordion from "@/components/Tables/FilterAccordion";
 import Button from "@mui/material/Button";
+import { TableHeaders } from "./Table";
```
DanWilkins2107, line 6: @ imports

src/components/Tables/Table.cy.tsx
Var1377: Ah right I hadn't pushed those changes yet

src/components/Tables/Table.tsx
Var1377: It feels like the most intuitive name, feel free to suggest another

DanWilkins2107 (changes requested):

src/app/admin/usersTable/UsersTable.tsx
```diff
@@ -23,13 +23,13 @@ export const OptionButtonDiv = styled.div`
...
 
-const usersTableHeaderKeysAndLabels: TableHeaders = [
+const usersTableHeaderKeysAndLabels = [
     ["id", "User ID"],
     ["email", "Email"],
     ["userRole", "Role"],
     ["createdAt", "Created At"],
     ["updatedAt", "Updated At"],
-];
+] as const;
```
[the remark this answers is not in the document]
Var1377: https://blog.logrocket.com/complete-guide-const-assertions-typescript/, it's required so that the type isn't inferred to be `[string, string][]` but instead to a subtype of `[keyof Data, string][]`

src/app/clients/ActionBar.cy.tsx
Var1377: Nothing, no thought went into that because the name doesn't really matter.

Congestion charge is required for the column as the red C icon is displayed in that column

src/app/clients/ClientsPage.tsx
Var1377: `packingDatetime` is rendered as just a date, but keeping the data intact means sorting works. The `packingLabel` column just renders as AM or PM. In these tables it doesn't really matter if there's data duplication as this should just be used for rendering and sorting logic. If you want a way to trace back to the original data, just add a primary key to it.

src/app/clients/ClientsPage.tsx
Var1377: It asserts that this is immutable and allows me to accept data declared `as const`

src/app/lists/ListDataview.tsx
Var1377: The database naming scheme calls it `10_(quantity|notes)` and these keys aren't used for rendering so I think it's better to stick with this.

src/app/lists/ListDataview.tsx
```diff
@@ -19,49 +14,81 @@ import TooltipCell from "@/app/lists/TooltipCell";
...
+    ["9", "Family of 9"],
+    ["10", "Family of 10+"],
+] as const;
+
+const displayQuantityAndNotes = (data: QuantityAndNotes): React.ReactElement => {
+    return <TooltipCell cellValue={data.quantity} tooltipValue={data.notes ?? ""} />;
+};
+
+const listDataViewColumnDisplayFunctions = {
+    "1": displayQuantityAndNotes,
```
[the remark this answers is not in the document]
Var1377: No but it's cleaner this way as opposed to having a global-scope for-loop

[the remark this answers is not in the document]
Var1377: This does need to exist otherwise react will try and render an Object which can't be done

src/components/Tables/Filters.tsx
Var1377: I can't get it to look consistent, and this is what it always was

src/components/Tables/Filters.tsx
Var1377: Yep, this was all there before it was just moved

src/components/Tables/Table.tsx
Var1377: I need to be able to accept deeply immutable data here to accept `as const`

src/components/Tables/Table.tsx
Var1377: This shouldn't've existed

src/components/Tables/Table.tsx
Var1377: If there is no column display function and the data itself isn't a valid ReactElement, react will try to display an object

src/components/Tables/Table.tsx
Var1377: I reckon it's probably fine

src/components/Tables/Table.tsx
Var1377: nope, that was for debugging I'll get rid of it

src/app/lists/ListDataview.tsx
SamYap-Softwire: yeah I agree (and I think similarly for the ones below as well)

src/app/lists/ListDataview.tsx
```diff
@@ -74,7 +101,26 @@ const ListsDataView: React.FC<Props> = (props) => {
...
+            "8": { quantity: row["8_quantity"], notes: row["8_notes"] },
+            "9": { quantity: row["9_quantity"], notes: row["9_notes"] },
+            "10": { quantity: row["10_quantity"], notes: row["10_notes"] },
+        };
+        return newRow;
+    });
+
+    const toggleableHeaders = headerKeysAndLabels
+        .map(([key]) => key)
+        .filter((key) => key !== "itemName");
```
SamYap-Softwire: do you think it's worth replacing the `.map.filter` with just `.filter(([key]) => key !== "itemName")`?

src/components/Tables/Filters.tsx
```diff
@@ -0,0 +1,100 @@
+import React from "react";
+import { styled } from "styled-components";
+import { TableHeaders } from "@/components/Tables/Table";
+
+export interface Filter<Data, State> {
+    shouldFilter(data: Data, state: State): boolean;
```
SamYap-Softwire, line 6: Ooo - haven't seen this syntax before - didn't know you can have functions in JS interfaces. I've always just done something like: `shouldFilter: (data: Data, state: State) => boolean;`
just curious - but is it common to do this in an interface? or would it be more of a class thing?

src/components/Tables/Filters.tsx
```diff
@@ -0,0 +1,100 @@
...
+    filterComponent(state: State, setState: (state: State) => void): React.ReactElement;
+    state: State;
+    initialState: State;
+    clear(state: State, setState: (state: State) => void): void;
+}
+
+export interface KeyedFilter<Data, N extends keyof Data, State> extends Filter<Data, State> {
```
SamYap-Softwire: `N`?
(is this intentional?)

SamYap-Softwire: same for the ones below

src/components/Tables/Filters.tsx
```diff
@@ -0,0 +1,100 @@
...
+        initialState: initialValue,
+        shouldFilter: (data, state) => {
+            let string = defaultToString(data[key]);
+
+            if (!caseSensitive) {
+                string = string.toLowerCase();
+                state = state.toLowerCase();
+            }
+
+            return !string.includes(state);
```
SamYap-Softwire, line 81: Do you want it to return True if `string === ""`? (just checking)

src/components/Tables/Table.tsx
```diff
@@ -41,86 +40,59 @@ export interface ColumnStyleOptions {
...
-    headerKey: string;
+interface CellProps<Data> {
+    row: Row<Data>;
+    columnDisplayFunctions: { [headerKey in keyof Data]?: ColumnDisplayFunction<Data[headerKey]> };
+    headerKey: keyof Data;
 }
 
-const CustomCell: React.FC<CellProps> = ({ row, columnDisplayFunctions, headerKey }) => {
-    return (
+const CustomCell = <Data extends unknown>({
```
SamYap-Softwire, line 69: can you explain why Data extends unknown?

src/components/Tables/Table.tsx
```diff
@@ -41,86 +40,59 @@ export interface ColumnStyleOptions {
...
             {columnDisplayFunctions[headerKey]
-                ? columnDisplayFunctions[headerKey](row)
+                ? columnDisplayFunctions[headerKey]!(row.data[headerKey])
                 : row.data[headerKey]}
         </>
     );
+
+    if (!React.isValidElement(element)) {
+        throw new Error(
+            `${element} is not a valid JSX element, add a column display function for ${String(
```
SamYap-Softwire, line 84: (genuinely) not sure if the error message is consistent with VFB-22 so maybe add a TODO VFB-22 if yours is merged before that, or change it (if necessary)?

src/components/Tables/TableFilterBar.tsx
SamYap-Softwire: (line 7 too - in case you miss it)

SamYap-Softwire (changes requested):
mostly questions - looks good :)

src/app/clients/getClientsTableData.ts
```diff
@@ -1,20 +1,25 @@
...
     fullName: Schema["clients"]["full_name"];
     familyCategory: string;
     addressPostcode: Schema["clients"]["address_postcode"];
-    collectionCentre: string;
-    congestionChargeApplies: boolean;
+    deliveryCollection: {
+        collectionCentre: string;
+        congestionChargeApplies: boolean;
+    };
     packingTimeLabel: string;
```
ChloeYiuSoftwire, line 14: Would it be helpful to make it more specific so it can only be like a valid time string rather than random gibberish like "!123456789876543e4565"

src/app/clients/getClientsTableData.ts
```diff
@@ -1,20 +1,25 @@
...
-    collectionDatetime: string;
+    collectionDatetime: Schema["parcels"]["collection_datetime"];
     lastStatus: string;
```
ChloeYiuSoftwire, line 16: Similarly, we have a list of status right - should we ensure they are limited to those that we can choose in the client page?

src/app/clients/getClientsTableData.ts
```diff
@@ -1,20 +1,25 @@
...
-    flaggedForAttention: boolean;
-    requiresFollowUpPhoneCall: boolean;
+    primaryKey: Schema["clients"]["primary_key"];
     fullName: Schema["clients"]["full_name"];
     familyCategory: string;
     addressPostcode: Schema["clients"]["address_postcode"];
-    collectionCentre: string;
-    congestionChargeApplies: boolean;
+    deliveryCollection: {
+        collectionCentre: string;
```
ChloeYiuSoftwire, line 11: Collection Centre is `Schema["parcels"]["collection_centre"]` ?

src/app/clients/getClientsTableData.ts
```diff
@@ -1,20 +1,25 @@
...
+        congestionChargeApplies: boolean;
+    };
     packingTimeLabel: string;
-    collectionDatetime: string;
+    collectionDatetime: Schema["parcels"]["collection_datetime"];
     lastStatus: string;
+    voucherNumber: Schema["parcels"]["voucher_number"];
+    iconsColumn: {
+        flaggedForAttention: boolean;
```
ChloeYiuSoftwire, line 19: Should this be `Schema["clients"]["flagged_for_attention"]`

src/app/clients/getClientsTableData.ts
```diff
@@ -1,20 +1,25 @@
...
+        requiresFollowUpPhoneCall: boolean;
```
ChloeYiuSoftwire, line 20: And this `Schema["clients"]["signposting_call_required"]`?

src/app/clients/getClientsTableData.ts
ChloeYiuSoftwire: Although it is just a boolean - but not sure if it is more restrictive in terms of the meaning behind it

src/app/lists/ListDataview.tsx
```diff
@@ -19,49 +14,81 @@ import TooltipCell from "@/app/lists/TooltipCell";
 import TableSurface from "@/components/Tables/TableSurface";
 import CommentBox from "@/app/lists/CommentBox";
 
-export type ListRow = Schema["lists"] & Datum;
+interface ListRow {
+    itemName: string;
```
ChloeYiuSoftwire, line 18: Should this be `Schema["lists"]["item_name"]`?

src/components/Tables/Filters.tsx
ChloeYiuSoftwire: If using this check, I think simply write `if (!string)` will do?

ChloeYiuSoftwire (changes requested):
Also is there a reason why the filter bar and accordion one becomes taller in height? Now the clear button looks a bit stretched TT

src/components/Tables/Filters.tsx
Var1377: probably not? not sure what expected behaviour would be here

DanWilkins2107 (approved):
Happy after everyone's changes + my comment are resolved

src/components/Tables/Table.tsx
Var1377: unknown is a fundamental type in typescript, and the type hierarchy goes `never <- unknown <- everythingElse`. It's essentially the same as any except with any you can perform operations on it like it's any other type.

In type parameters for arrow functions, you have to put some sort of type bound and saying that `Data` extends unknown means that there are no type constraints on `Data` aside from the fact that it can't be `never`

src/components/Tables/Table.tsx
Var1377: This shouldn't ever be displayed to the user, this is a broken API invariant. This could probably be fixed with more exotic types but that would come at the expense of code readability so this is what I thought the best solution would be

src/app/lists/ListDataview.tsx
Var1377: I would probably want a type error here if it was changed to something that wasn't a string.

src/app/clients/getClientsTableData.ts
Var1377: I wanted to decouple what's in the UI from the database schema, so if the schema ever changed it would raise a type error at the point at which we want the error to occur

src/components/Tables/Table.cy.tsx
Var1377: They are now used

SamYap-Softwire (approved):

ChloeYiuSoftwire (approved):
Apart from noting down the status part (which should only be limited to those available), I think everything else looks good now

src/components/Tables/Filters.tsx
```diff
@@ -0,0 +1,100 @@
...
+`;
+
+interface TextFilterProps<Data, Key extends keyof Data> {
+    key: Key;
+    label: string;
+    caseSensitive?: boolean;
+    initialValue?: string;
+}
+
+const defaultToString = (value: any): string => {
```
paopaosoftwire, line 56: Can this not be an any?

paopaosoftwire (approved):

DanWilkins2107 (commented):
Make sure to fix the family size column sort!

ChloeYiuSoftwire (changes requested):
UI comments
- Clear button display seems buggy (the width, the location and the icon color with respect to the text color)
Capture
- In general the view from width 5XX - 8XX px seems to be overflowing
- Date picker in filter shouldn't be red when it is deselected as it is an optional field?
- Font size in "Additional Filter" differs from the placeholder in the text box which might not look too tidy
- When I shrink the page to mobile field the accordion is showing tick to columns that are not available in mobile view?

src/components/Tables/FilterAccordion.tsx
```diff
@@ -92,27 +110,47 @@ const FilterAccordion: React.FC<FilterAccordionProps> = ({
...
                     })}
+                    <ColumnSelectRow>
+                        <SectionLabel>Select Columns:</SectionLabel>
+                        {(toggleableHeaders ?? []).map((key) => {
+                            const headerKeyAndLabel =
+                                headers.find(([headerKey]) => headerKey === key) ?? key.toString();
+                            const headerLabel = headerKeyAndLabel[1];
+                            return (
+                                <ContainerDiv key={headerLabel}>
+                                    <label
```
TomCusack-Softwire, line 138: should label be for separate text (and not just be the checkbox itself)?

src/components/Tables/Filters.tsx
```diff
@@ -0,0 +1,189 @@
...
+}: TextFilterProps<Data, Key>): KeyedFilter<Data, Key, string> => {
+    return keyedFilter(key, label, {
+        state: initialValue,
+        initialState: initialValue,
+        shouldFilter: (data, state) => {
+            let string = defaultToString(data[key]);
+
+            if (!caseSensitive) {
+                string = string.toLowerCase();
+                state = state.toLowerCase();
```
TomCusack-Softwire, line 80: I'm surprised it lets you set `state` without a `let` explicitly, and just 'setting' the input variable

src/components/Tables/Filters.tsx
```diff
@@ -0,0 +1,189 @@
...
+    return keyedFilter(key, label, {
+        state: initialValue,
+        initialState: initialValue,
+        shouldFilter: (data, state) => {
+            const date = data[key] as Date;
+            const from = state.from;
+            const to = state.to;
+
+            if (from !== null && date < from) {
+                return true;
```
TomCusack-Softwire, line 148: is 'true' meaning 'it should be filtered [out]' rather than 'the filter should remove all falses'?

src/components/Tables/Table.cy.tsx
```diff
@@ -67,24 +72,36 @@ const data: Datum[] = [
...
+    toggleableHeaders?: (keyof TestData)[];
+    pagination?: boolean;
+    filters?: (keyof TestData)[];
+    sortable?: (keyof TestData)[];
 }
 
-const Component: React.FC<TestTableProps> = ({ checkboxes = true, toggleableHeaders }) => {
+const Component: React.FC<TestTableProps> = ({
+    checkboxes = true,
+    toggleableHeaders,
```
TomCusack-Softwire, line 90: what's the default for toggleableheaders - is undefined 'correct'?

src/components/Tables/Table.tsx
```diff
@@ -13,22 +13,21 @@ import {
...
-export interface Row {
+export interface Row<Data> {
     rowId: number;
-    data: Datum;
+    data: Data;
 }
 
-export type ColumnDisplayFunction = (row: Row) => ReactNode;
-export type TableColumnDisplayFunctions = { [headerKey: string]: ColumnDisplayFunction };
+export type ColumnDisplayFunction<T> = (data: T) => ReactNode;
```
TomCusack-Softwire, line 25: I much prefer ReactElement (and then scattering fragments around) rather than ReactNode, where possible, but might be a reason for this I'm not aware of!

src/components/Tables/Table.tsx
```diff
@@ -134,32 +107,57 @@ const defaultColumnStyleOptions: ColumnStyleOptions = {
     maxWidth: "20rem",
 };
 
-const Table: React.FC<Props> = ({
+const defaultSortFunction = <T extends unknown>(left: T, right: T): number => {
```
TomCusack-Softwire, line 110: why isn't this just default??? is this a 'javascript bad' moment?

src/components/Tables/Table.tsx
```diff
@@ -134,32 +107,57 @@ const defaultColumnStyleOptions: ColumnStyleOptions = {
...
+}: Props<Data>): ReactElement => {
     const [shownHeaderKeys, setShownHeaderKeys] = useState(
         defaultShownHeaders ?? headerKeysAndLabels.map(([key]) => key)
     );
 
     const shownHeaders = headerKeysAndLabels.filter(([key]) => shownHeaderKeys.includes(key));
 
-    const [data, setData] = useState(inputData);
+    const toFilter = (filter: Filter<Data, any> | keyof Data): Filter<Data, any> => {
+        if (filter instanceof Object) {
```
TomCusack-Softwire, line 143: why are we even accepting objects as well as Filters? can we not just enforce that people use the Filter object at all times?

src/components/Tables/Table.tsx
```diff
@@ -195,18 +193,39 @@ const Table: React.FC<Props> = ({
...
     }, [selectedCheckboxes, selectAllCheckBox]);
 
-    const columns: TableColumn<Row>[] = (
+    const sortOptions = sortable.map((sortOption) => {
+        if (sortOption instanceof Object) {
+            return sortOption;
+        }
+        return {
+            key: sortOption,
+            sortFunction: (datapoint: Data[keyof Data], otherDatapoint: Data[keyof Data]) => {
```
TomCusack-Softwire, line 202: this looks like `defaultSortFunction` could replace it

src/components/Tables/Table.tsx
```diff
@@ -311,35 +320,38 @@ const Table: React.FC<Props> = ({
...
                     keyField="rowId"
                     fixedHeader
                     subHeader
                     subHeaderComponent={
-                        <TableFilterBar
-                            filterText={filterText}
-                            filterKeys={filterKeys}
+                        <TableFilterBar<Data>
+                            handleClear={() =>
+                                setPrimaryFilters((filters) =>
```
TomCusack-Softwire, line 343: extract to a const above the TSX pls

src/components/Tables/TableFilterBar.tsx
```diff
@@ -1,70 +1,100 @@
...
+    width: 100%;
 
-    &:focus {
-        outline: none;
+    @media (min-width: 500px) {
+        flex-wrap: nowrap;
     }
 `;
 
-interface Props {
```
TomCusack-Softwire, line 30: would rather the order of the file doesn't change _too_ much, so that the git diff is more readable!

src/components/Tables/TableFilterBar.tsx
```diff
@@ -1,70 +1,100 @@
...
-                setShownHeaderKeys={props.setShownHeaderKeys}
-                headers={props.headers}
-            />
+            <FilterHeading>Filters</FilterHeading>
+            <Styling>
+                {props.filters.length > 0 && (
+                    <>
+                        {props.filters.map((filter, index) => {
+                            const onFilter = (state: any): void => {
+                                const newFilters = [...props.filters];
```
TomCusack-Softwire, line 67: you've used this 'set state of one index of a list' a lot - can we extract to a function, or just give components a setter function that 'does this for you'?

TomCusack-Softwire (changes requested):

src/components/Tables/FilterAccordion.tsx
```diff
@@ -5,19 +5,25 @@ import { Accordion, Checkbox, AccordionSummary, AccordionDetails } from "@mui/ma
...
+    headers: TableHeaders<Data>;
+    filters: Filter<Data, any>[];
+    setFilters: (filters: Filter<Data, any>[]) => void;
 }
 
 const Styling = styled.div`
     flex-grow: 1;
-    height: 39px;
-    overflow: visible;
+    place-self: center;
```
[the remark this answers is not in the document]
Var1377: This was for adding the additional filters bar

src/components/Tables/Filters.tsx
Var1377: `true` means it `shouldFilter` i.e. it shouldn't be included

src/components/Tables/Table.tsx
Var1377: Nope this is an "I'm bad" moment

PR #84, opened 2023-08-21
VFB-24 View Client Page

SamYap-Softwire:
`/clients/[id]` now displays the client information with the matching primary key


src/pdf/ShoppingList/dataPreparation.ts
TomCusack-Softwire, line 1: probably should move this out of /pdf/shoppinglist if other files are going to start using it! maybe a thing to do in the week after the demo?

src/app/clients/[id]/ViewClientPage.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const KeyText = styled.p`
+    font-weight: bold;
+`;
+
+const ViewClientPage: React.FC<Props> = ({ clientID, clientData }) => {
+    return (
+        <CenterComponent>
+            <StyledCard>
+                <TempDiv>
```
Comment (TomCusack-Softwire) [inline src/app/clients/[id]/ViewClientPage.tsx:38]: does both inner and outer need to be wrapped in TempDivs?

src/app/clients/[id]/getClientData.ts
```diff
@@ -0,0 +1,24 @@
...
+    prepareClientSummary,
+    prepareHouseholdSummary,
+    prepareRequirementSummary,
+    RequirementSummary,
+} from "@/pdf/ShoppingList/dataPreparation";
+import { fetchClients, fetchFamilies } from "@/pdf/ShoppingList/databaseFetch";
+
+export type ClientData = ClientSummary & HouseholdSummary & RequirementSummary;
+
+const getClientData = async (clientID: string): Promise<ClientData> => {
```
Comment (TomCusack-Softwire) [inline src/app/clients/[id]/getClientData.ts:13]: 😍

src/app/clients/[id]/getClientData.ts
```diff
@@ -0,0 +1,24 @@
+import {
+    ClientSummary,
+    HouseholdSummary,
+    prepareClientSummary,
+    prepareHouseholdSummary,
+    prepareRequirementSummary,
+    RequirementSummary,
+} from "@/pdf/ShoppingList/dataPreparation";
+import { fetchClients, fetchFamilies } from "@/pdf/ShoppingList/databaseFetch";
```
Comment (TomCusack-Softwire) [inline src/app/clients/[id]/getClientData.ts:9]: may want to be careful - these functions assume client-side (supabaseClient is used) but I think these are currently server-side, which may cause some weird bugs

TomCusack-Softwire (approved):

src/app/clients/[id]/ViewClientPage.tsx
Comment (SamYap-Softwire) [reply src/app/clients/[id]/ViewClientPage.tsx:38]: thanks for catching this TT 
neither of them are needed - hence the `Temp` in the name ^^

src/pdf/ShoppingList/dataPreparation.ts
SamYap-Softwire: yep agree, i was thinking of doing this but wasn't sure where I should put it in - will do after demo though!

src/app/clients/[id]/getClientData.ts
Comment (SamYap-Softwire) [reply src/app/clients/[id]/getClientData.ts:9]: temporary fix now but will do a proper fix (separating client and server) after the demo!

PR #88, opened 2023-08-22
VFB-34 Change view for callers

DanWilkins2107:
Added functionality to edit what pages are available to users based on their role. Also edited the nav bar based on this.

### DIFF
--- a/src/app/auth.tsx
+++ b/src/app/auth.tsx
@@ -1,9 +1,10 @@
 "use client";
 
-import React, { useEffect } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import { usePathname, useRouter } from "next/navigation";
 import { createClientComponentClient } from "@supabase/auth-helpers-nextjs";
 import { DatabaseAutoType } from "@/databaseUtils";
+import { RoleUpdateContext } from "@/app/roles";
 
 interface Props {
     children: React.ReactNode;
@@ -13,7 +14,8 @@ export const AuthRouting: React.FC<Props> = ({ children = <></> }) => {
     const supabase = createClientComponentClient<DatabaseAutoType>();
     const router = useRouter();
     const pathname = usePathname();
-    const [loggedIn, setLoggedIn] = React.useState<boolean>();
+    const [loggedIn, setLoggedIn] = useState<boolean>();
+    const { setRole } = useContext(RoleUpdateContext);
 
     // This fixes issues with navigation when the user has signed in but then signed out and caching still allows them to
     // access pages. It also fixes authentication flow issues on sign in and sign out
@@ -48,6 +50,14 @@ export const AuthRouting: React.FC<Props> = ({ children = <></> }) => {
         }
     };
 
+    const findUserRole: () => Promise<void> = async () => {
+        const {
+            data: { user },
+        } = await supabase.auth.getUser();
+        const userRole = user?.app_metadata.role;
+        setRole(userRole);
+    };
+
     useEffect(() => {
         const {
             data: { subscription },
@@ -60,6 +70,7 @@ export const AuthRouting: React.FC<Props> = ({ children = <></> }) => {
 
     useEffect(() => {
         onRouteChange();
+        findUserRole();
         // eslint-disable-next-line react-hooks/exhaustive-deps -- We don't need onRouteChange in the dependency array
     }, [loggedIn, pathname, router]);
 
--- a/src/app/roles.tsx
+++ b/src/app/roles.tsx
@@ -10,7 +10,7 @@ interface RolesInterface {
     [key: string]: string[];
 }
 
-export const rolesAndHiddenPages: RolesInterface = {
+export const rolesToHiddenPages: RolesInterface = {
     admin: [],
     caller: ["/admin", "/lists"],
 };
--- a/src/app/roles.tsx
+++ b/src/app/roles.tsx
@@ -10,7 +10,7 @@ interface RolesInterface {
     [key: string]: string[];
 }
 
-export const rolesToHiddenPages: RolesInterface = {
+export const roleToHiddenPages: RolesInterface = {
     admin: [],
     caller: ["/admin", "/lists"],
 };
--- a/src/app/roles.tsx
+++ b/src/app/roles.tsx
@@ -1,0 +1,28 @@
+"use client";
+
+import React, { useState, createContext } from "react";
+
+interface Props {
+    children: React.ReactNode;
+}
+
+interface RolesInterface {
+    [key: string]: string[];
+}
+
+export const rolesAndHiddenPages: RolesInterface = {
+    admin: [],
+    caller: ["/admin", "/lists"],
+};
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const RoleUpdateContext = createContext({ role: "", setRole: (role: string): void => {} });
+
+export const RoleManager: React.FC<Props> = ({ children }) => {
+    const [role, setRole] = useState("");
+
+    return (
+        <RoleUpdateContext.Provider value={{ role, setRole }}>
+            {children}
+        </RoleUpdateContext.Provider>
+    );
+};
--- a/src/app/roles.tsx
+++ b/src/app/roles.tsx
@@ -1,24 +1,41 @@
 "use client";
 
+import { DatabaseEnums } from "@/databaseUtils";
 import React, { useState, createContext } from "react";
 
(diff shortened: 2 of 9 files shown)


src/app/roles.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface RolesInterface {
+    [key: string]: string[];
+}
+
+export const rolesAndHiddenPages: RolesInterface = {
+    admin: [],
+    caller: ["/admin", "/lists"],
+};
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
SamYap-Softwire, line 17: will there still be an error if you use underscore at the front? (i.e. `_role`)

src/components/NavBar/NavigationBar.tsx
```diff
@@ -108,18 +109,39 @@ const SignOutButtonContainer = styled(NavElementContainer)`
...
+interface RoleProps {
+    children?: React.ReactNode;
+    role: string;
+    pathname: string;
+}
+
+const RoleDependent: React.FC<RoleProps> = ({ children, role, pathname }) => {
+    const hiddenPages = role
+        ? rolesAndHiddenPages[role]
+        : ["/clients", "/lists", "/calendar", "/admin"];
```
SamYap-Softwire, line 131: just curious: is this list something we have to change everytime we add a new page to the navbar?

SamYap-Softwire (approved):
Looks good :) 
Just some questions

src/app/roles.tsx
DanWilkins2107: Will check!

src/components/NavBar/NavigationBar.tsx
DanWilkins2107: Yes! So this list is for when the user is logged out, which ones to not show

src/app/roles.tsx
```diff
@@ -0,0 +1,28 @@
+"use client";
+
+import React, { useState, createContext } from "react";
+
+interface Props {
+    children: React.ReactNode;
+}
+
+interface RolesInterface {
+    [key: string]: string[];
```
ChloeYiuSoftwire, line 10: Just thinking because it is not taking general string but only "/clients" "/lists" "/calendar" and "/admin", would it worth making a type called pathname and only include these four, and share this type among all things that will use pathname? 
Similarly for role, maybe can make a type used globally that has only "admin" and "caller"?

ChloeYiuSoftwire: But I ain't sure if it is an overengineering (because it has to be updated whenever role and pathname has been created)
Let me know your thoughts

src/app/roles.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const rolesAndHiddenPages: RolesInterface = {
```
ChloeYiuSoftwire, line 13: Rename to `rolesToHiddenPages` maybe? So that I know  more clearly I "feed" it a role and it returns me what pages are hidden

ChloeYiuSoftwire (commented):
Looks nice - but just have a bit comments

src/components/NavBar/NavigationBar.tsx
paopaosoftwire: Isn't all pages except login then?

DanWilkins2107: Yes, but we only need the ones in the navbar here

src/components/NavBar/NavigationBar.tsx
```diff
@@ -108,18 +109,39 @@ const SignOutButtonContainer = styled(NavElementContainer)`
...
     children?: React.ReactNode;
 }
 
+interface RoleProps {
+    children?: React.ReactNode;
+    role: string;
+    pathname: string;
+}
+
+const RoleDependent: React.FC<RoleProps> = ({ children, role, pathname }) => {
```
daniel-conlan-softwire, line 128: I quite like this wrapper, nice idea for a component

daniel-conlan-softwire (changes requested):
Looks pretty good to me - some nice components and functionality. Just some small logic errors and perhaps a small, but not required, refactor

src/app/auth.tsx
```diff
@@ -1,18 +1,19 @@
 "use client";
 
-import React, { useEffect } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import { usePathname, useRouter } from "next/navigation";
 import { createClientComponentClient } from "@supabase/auth-helpers-nextjs";
 import { DatabaseAutoType } from "@/database_utils";
+import { RoleUpdateContext } from "@/app/roles";
 
 export const AuthRouting: React.FC<{ children: React.ReactNode }> = ({ children = <></> }) => {
```
paopaosoftwire, line 9: Extract to an interface

src/components/NavBar/NavigationBar.tsx
```diff
@@ -11,6 +11,7 @@ import LightDarkSlider from "@/components/NavBar/LightDarkSlider";
 import SignOutButton from "@/components/NavBar/SignOutButton";
 import NavBarButton from "@/components/Buttons/NavBarButton";
 import { usePathname } from "next/navigation";
+import { RoleUpdateContext, rolesToHiddenPages } from "@/app/roles";
 
 export const NavBarHeight = "4rem";
```
paopaosoftwire, line 17: I noticed that we still have background-color: ${(props) => props.theme.primaryBackgroundColor} in this file. I know its not in your ticket, but can you quickly refactor this?

paopaosoftwire (changes requested):

paopaosoftwire (approved):
<3

daniel-conlan-softwire (approved):

src/app/roles.tsx
TomCusack-Softwire: there shouldn't be... this is the `role` in the (function input) rather than the outside one!

src/components/NavBar/NavigationBar.tsx
```diff
@@ -108,18 +106,39 @@ const SignOutButtonContainer = styled(NavElementContainer)`
...
+interface RoleProps {
+    children?: React.ReactNode;
+    role: string;
+    pathname: string;
+}
+
+const RoleDependent: React.FC<RoleProps> = ({ children, role, pathname }) => {
+    const hiddenPages = role
+        ? roleToHiddenPages[role]
+        : ["/clients", "/lists", "/calendar", "/admin"];
```
TomCusack-Softwire, line 128: I _think_ I'd rather specify 'pages they CAN see' - if we add a new admin page, it shouldn't be 'everyone sees it unless you say otherwise'!

src/components/NavBar/NavigationBar.tsx
```diff
@@ -131,54 +150,52 @@ const ResponsiveAppBar: React.FC<Props> = ({ children }) => {
...
+                                </DrawerButtonWrapper>
+                            </RoleDependent>
                         ))}
                     </DrawerInner>
                 </StyledSwipeableDrawer>
             </LoginDependent>
             <AppBar>
                 <AppBarInner>
                     <MobileNavMenuContainer>
-                        <LoginDependent>
```
TomCusack-Softwire, line 159: do we not need a `LoginDependent` any more? what about a `RoleDependent`? Should mobile users only see the buttons they are meant to click?

src/middleware.ts
```diff
@@ -20,8 +21,13 @@ const middleware: NextMiddleware = async (req: NextRequest) => {
     if (user && req.nextUrl.pathname === "/login") {
         return NextResponse.redirect(new URL("/clients", req.url));
     }
-
-    return res;
+    const userRole = user?.app_metadata.role;
+    const blockedPages = roleToHiddenPages[userRole] ?? [];
+    if (blockedPages.includes(req.nextUrl.pathname)) {
+        const url = req.nextUrl;
+        url.pathname = "/404";
```
TomCusack-Softwire, line 28: is this actually the 404 page or just an 'invalid url' which then it redirected to the 404?

TomCusack-Softwire (changes requested):

src/components/NavBar/NavigationBar.tsx
DanWilkins2107: Refactored

src/middleware.ts
DanWilkins2107: Redirects to the 404 page, with lists still in the URL

PR #89, opened 2023-08-22
VFB 21 Add Parcels Technical Review Changes

DanWilkins2107:
After Technical Review, added a past date error message.

### DIFF
--- a/src/components/Form/formFunctions.ts
+++ b/src/components/Form/formFunctions.ts
@@ -17,6 +17,7 @@ export enum Errors {
     invalid = "Please enter a valid entry.",
     submit = "Please ensure all fields have been entered correctly. Required fields are labelled with an asterisk.",
     external = "Please try again later.",
+    pastDate = "Please enter a date in the future.",
 }
 
 export const numberRegex = /^\d+$/;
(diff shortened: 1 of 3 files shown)


src/components/Form/formFunctions.ts
```diff
@@ -147,6 +148,22 @@ export const onChangeDateOrTime = (
     fieldSetter(key, value);
 };
 
+export const onChangeDate = (
```
SamYap-Softwire, line 151: why do we need this function - instead of just put the additional logic (lines 158 to 164) as more if statements in `onChangeDateTime`?

SamYap-Softwire (changes requested):
More of a question instead of a request change

src/components/Form/formFunctions.ts
DanWilkins2107: Because onChangeDate is only applied to the dates

SamYap-Softwire (approved):

paopaosoftwire (approved):

src/components/Form/formFunctions.ts
```diff
@@ -17,6 +17,7 @@ export enum Errors {
     invalid = "Please enter a valid entry.",
     submit = "Please ensure all fields have been entered correctly. Required fields are labelled with an asterisk.",
     external = "Please try again later.",
+    pastDate = "Please enter a date in the future.",
```
TomCusack-Softwire, line 20: should we not be able to backdate things? if I forgot to update a parcel but it's now been collected, what should happen?

TomCusack-Softwire (approved):

src/components/Form/formFunctions.ts
DanWilkins2107: So ironically this is how it was implemented before. I think it is better to not allow it when adding a parcel as I can't imagine someone forgetting to add a parcel completely many times, and I think the number of errors it would catch makes it worth it, but I think whoever is implementing the edit parcel functionality should have some possibility to enter a past date as there needs to be some way to do it.

PR #90, opened 2023-08-23
VFB 28 Add Comment Box after Technical Review

DanWilkins2107:
I have completely reworked the functionality of add comment box in order to make sure it works as intended.

### DIFF
--- a/src/app/lists/CommentBox.tsx
+++ b/src/app/lists/CommentBox.tsx
@@ -90,6 +90,7 @@ const CommentContainer: React.FC<CommentProps> = ({ originalComment }) => {
             setErrorMessage("");
             setResetComment(value);
             setSuccessMessage("Comment successfully updated.");
+            window.location.reload();
         }
     };
     return (
--- a/src/app/lists/CommentBox.tsx
+++ b/src/app/lists/CommentBox.tsx
@@ -36,13 +36,14 @@ const Wrapper = styled.div`
 
 const CommentBoxContainer = styled.div`
     width: 100%;
-    margin: 1rem;
+    margin: 1rem 1rem 0 1rem;
     min-width: 15rem;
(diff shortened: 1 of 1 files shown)


ChloeYiuSoftwire (approved):
Looks good to me
Can you add a bit more upper margin to the Comment title
The top margin is significantly narrower than the bottom margin of the comment box in mobile view

paopaosoftwire (approved):

src/app/lists/CommentBox.tsx
```diff
@@ -88,16 +88,16 @@ const CommentContainer: React.FC<CommentProps> = ({ originalComment }) => {
             setSuccessMessage("");
         } else {
             setErrorMessage("");
-            setResetComment(value);
             setSuccessMessage("Comment successfully updated.");
```
TomCusack-Softwire, line 91: do you need a success message if we are immediately reloading?

src/app/lists/CommentBox.tsx
```diff
@@ -88,16 +88,16 @@ const CommentContainer: React.FC<CommentProps> = ({ originalComment }) => {
...
+            window.location.reload();
```
TomCusack-Softwire, line 92: just wondering if there's a way to do this with state (and forcing rerenders) instead of refreshes... it's fine to do it this way but if we can get rid of a whole page refresh it would be lovely!

src/app/lists/CommentBox.tsx
DanWilkins2107: Yes, it flashes up for a second before the reload happens!

src/app/lists/CommentBox.tsx
DanWilkins2107: If I have time I will have a look!

PR #92, opened 2023-08-23
Fix clients page having no data

TomCusack-Softwire:
server side function was using something which was being called from a client side page as well, so it was marked as client side (and a supabase client component client was being used) on the server, which doesn't work because cookies doesn't exist.

I reordered some functions and converted to a server to make it all work!

### DIFF
--- a/src/app/clients/ClientsPage.cy.tsx
+++ b/src/app/clients/ClientsPage.cy.tsx
@@ -1,8 +1,4 @@
-import {
-    CongestionChargeDetails,
-    ProcessingData,
-    RawClientDetails,
-} from "@/app/clients/fetchDataFromServer";
+import { CongestionChargeDetails, ProcessingData } from "@/app/clients/fetchDataFromServer";
 import {
     datetimeToPackingTimeLabel,
     eventToStatusMessage,
@@ -16,6 +12,7 @@ import {
     formatDatetimeAsTime,
     formatHouseholdFromFamilyDetails,
     rawDataToExpandedClientDetails,
+    RawClientDetails,
 } from "@/app/clients/getExpandedClientDetails";
 
 const sampleProcessingData: ProcessingData = [
--- a/src/app/clients/ClientsPage.cy.tsx
+++ b/src/app/clients/ClientsPage.cy.tsx
@@ -1,7 +1,11 @@
 import {
+    CongestionChargeDetails,
+    ProcessingData,
+    RawClientDetails,
+} from "@/app/clients/fetchDataFromServer";
+import {
     datetimeToPackingTimeLabel,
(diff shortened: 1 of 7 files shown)


paopaosoftwire (approved):

src/app/clients/ClientsPage.cy.tsx
```diff
@@ -78,36 +85,27 @@ const sampleRawExpandedClientDetails: RawClientDetails = {
...
-                            packingTimeLabel: "PM",
-                            lastStatus: "LAST_EVENT @ 04/08/2023",
-                        },
-                    ]);
-                })
+            const clientTableData = processingDataToClientsTableData(
+                sampleProcessingData,
+                sampleCongestionChargeData
             );
+            expect(clientTableData).to.deep.equal([
```
DanWilkins2107, line 92: How come we can't just have equal here?

src/app/clients/fetchDataFromServer.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    return JSON.parse(response.data);
+};
+
+export type ProcessingData = Awaited<ReturnType<typeof getProcessingData>>;
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export const getProcessingData = async () => {
+    const response = await supabase
```
DanWilkins2107, line 27: Do we not need some sort of error handling here?

DanWilkins2107 (changes requested):
Let me know if we don't need error handling there and I will approve :)

TomCusack-Softwire:
fyi I've only moved around existing code and minimally changed it rather than writing my own stuff :)

src/app/clients/ClientsPage.cy.tsx
TomCusack-Softwire: I think 'shallow' equals will check if they are the same object in memory, which they aren't here (because they are created at two different points) - whereas 'deep' equals checks each property recursively, meaning two objects are equal if they 'look' the same - which is what we want here

TomCusack-Softwire: image
image

src/app/clients/fetchDataFromServer.ts
TomCusack-Softwire: that's your ticket! you've already done a great job at it all
https://github.com/VauxhallFoodbank/vauxhall-foodbank/pull/70/files#diff-af4456595211b28ff643ecd6021325966085c08d00ffc74a5e3c0188c83c6c34R55-R60
this is the exact same line!

DanWilkins2107 (approved):

PR #101, opened 2023-08-24
VFB-38 Make Clients Page and Parcels Page

SamYap-Softwire:
Moved old clients page to parcels and made new clients page

- [ ] Make sure you've verified it works on dev via `npm run dev`
- [ ] Make sure you've verified it works on prod via `npm run build` and `npm run start`


src/app/parcels/ParcelsPage.tsx
SamYap-Softwire, line 1: There isn't any change here too! Just file moving

SamYap-Softwire, line 1: This is not a new file - just an old file moved (hence renamed)

PR #111, opened 2024-01-21
VFB-31 Parcel table filters and related changes

stonelink:
[Please replace this line with a brief summary of what you've changed, and tick the boxes below once complete]

- [X] Make sure you've verified it works on dev via `npm run dev`
- [X] Make sure you've verified it works on prod via `npm run build` and `npm run start`


src/app/lists/ListsPage.tsx
```diff
@@ -0,0 +1,37 @@
+"use client";
```
stonelink, line 1: The logic in this file was moved out of lists/page.tsx and adjusted so that it stops hydration errors

PR #115, opened 2024-02-23
VFB-86: Generate seed data

rickaw:
## What's changed
- Use Snaplet to generate a large amount of data for local development
- Rename some database field names `1_quantity`, `1_notes` etc to `quantity_for_1`, `notes_for_1` etc.
- Add scripts to do the following when resetting the local supabase db
  - create an admin user & a caller user 
  - upload the congestion charge postcodes to the local supabase storage

## Checklist
- [x] Make sure you've verified it works on dev via `npm run dev`
- [x] Make sure you've verified it works on prod via `npm run build` and `npm run start`


seed.mts
```diff
@@ -0,0 +1,197 @@
...
+
+    await seed.collectionCentres([
+        {
+            name: "Brixton Hill - Methodist Church",
+            acronym: "BH-MC",
+        },
+        {
+            name: "Clapham - St Stephens Church",
+            acronym: "CLP-SC",
+        },
```
stonelink, line 104: Could you also please put "Delivery"/"DLVR" as a collection centre?

It's really not ideal (see https://softwiretech.atlassian.net/browse/VFB-55) that delivery is represented like this, because it's the most common real-world situation.

For test data, it would be great if about half of the parcels were Delivery and the other half were one of the (other) collection centres - is that possible?

rickaw: Added delivery as a collection centre now. Choosing delivery 50% of the time turned out to be non-trivial. The current implementation should pick a collection centre randomly for each parcel (33% chance) so I'll leave this as is for now.

rickaw: For context, this is how the parcels page looks currently
image

rickaw: (The postcode could do with some more refining - do pick from a set of 10 of postcodes, some of which would be in the congestion charge list? But the new starters are itching to start on the project so I'm keen to get this merged first and do some refining at a later point)

stonelink: This is looking great!
I understand about wanting to get things going.

I think your suggestion about randomly picking from a predefined list of postcodes is good, especially as the client will want to see that the Congestion Charge functionality works as they expect; also so the local devs don't forget that CC functionality exists.

Will it be easy to have everyone replace their data once that's done?

README.md
```diff
@@ -103,8 +103,12 @@ You can either
 #### Apply migrations to local database
```
stonelink, line 103: Thank you for getting the local database solution done!

supabase/migrations/20240222101904_update_hotel_lists_field_names.sql
stonelink, line 1: An actual real needed migration! Exciting!

supabase/tools/getLocalSupabaseClient.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (!supabaseUrl) {
+        throw new Error("Supabse URL it not specified");
+    }
+
+    if (!supabaseUrl.includes("localhost") && !supabaseUrl.includes("127.0.0.1")) {
+        throw new Error("Supabse URL needs to be a localhost port");
+    }
+
+    if (!supabaseKey) {
+        throw new Error("Supabase key is not ");
```
stonelink, line 16: ...not set?
Also a couple of small typos in "Supabse" above

README.md
```diff
@@ -117,6 +121,17 @@ To check they've been applied correctly, either:
...
+The local database is generated based on the code in `seed.mts`. This uses Snaplet to generate a large amount of data.
+To generate te SQL queries needed to populate database, run
+```shell
+npx tsx seed.mts > supabase/seed.sql
+```
+To rebuild the database from the migration files and `seed.sql`, run
+```shell
+npm run dev:reset_supabase
+```
+
```
stonelink, line 134: Should these local setup steps include the createStorage and create-admin-user tools, or have I missed where they are run?

seed.mts
rickaw: While I was waiting, I thought I might as well implement this now:
image

README.md
rickaw: `dev:reset_supabase` script actually resets the database using the seed, and calls `createStorage` and `createAdminAndCallerUsers`, since we want do go through those steps whenever the database is reset.

seed.mts
stonelink: Superhero!

supabase/tools/getLocalSupabaseClient.ts
rickaw: Thanks for catching these, fixed now.

README.md
stonelink: Oh yeah, sorry I didn't read package.json properly

PR #116, opened 2024-02-25
VFB-100: Simplify styling in Client and Parcel modals, and review fields

stonelink:
This is mostly tidying & simplifying the appearance of the two modals.

- [X] Make sure you've verified it works on dev via `npm run dev`
- [X] Make sure you've verified it works on prod via `npm run build` and `npm run start`

### DIFF
--- a/src/app/parcels/ExpandedParcelDetailsFallback.tsx
+++ b/src/app/parcels/ExpandedParcelDetailsFallback.tsx
@@ -4,19 +4,14 @@ import DataViewerFallback from "@/components/DataViewer/DataViewerFallback";
 const clientDetailFields = [
     "VOUCHER #",
     "FULL NAME",
+    "ADDRESS",
     "PHONE NUMBER",
+    "DELIVERY_INSTRUCTIONS",
+    "HOUSEHOLD",
+    "CHILDREN",
     "PACKING DATE",
     "PACKING TIME",
-    "DELIVERY_INSTRUCTIONS",
-    "ADDRESS",
-    "HOUSEHOLD",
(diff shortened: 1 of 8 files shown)


src/app/parcels/ExpandedParcelDetailsFallback.tsx
stonelink, line 1: This highlights an inconsistency between Parcel and Client modals - only Parcel has a Suspense fallback. I'm not sure if there's a reason for that, it's probably an oversight and maybe I should give the Clients modal the same fallback,

src/components/Modal/ModalFormStyles.ts
stonelink, line 1: Btw I don't think any of the styling is responsive - it's just desktop. That's fine for the main use case.

I've added a ticket for looking at responsiveness after the functionality is added (probably after go-live); it might be tricky because the site is quite information dense.

src/app/parcels/ExpandedParcelDetailsFallback.tsx
rickaw: Yeah having a similar fallback would be nice - happy to pick this up later on :)

src/components/Modal/ModalFormStyles.ts
rickaw: The responsiveness being a low priority makes sense :)

rickaw (approved):
image

Love the new look! Thank you!

src/app/parcels/ExpandedParcelDetailsFallback.tsx
stonelink: Thank you :)

PR #118, opened 2024-02-27
VFB-98: Automatic deployment to deployed dev database

rickaw:
- [x] Check that the deployment runs correctly if we run it on every PR
- [x] Remove the rule to run it on every PR, and only run on push / merge to main 

N/A
- [ ] Make sure you've verified it works on dev via `npm run dev`
- [ ] Make sure you've verified it works on prod via `npm run build` and `npm run start`

### DIFF
--- a/.github/workflows/deploy-backend.yml
+++ b/.github/workflows/deploy-backend.yml
@@ -24,6 +24,3 @@ jobs:
         run: npx supabase db push --dry-run
       - name: Apply the migration
         run: npx supabase db push
-
-
-
--- a/.github/workflows/deploy-backend.yml
+++ b/.github/workflows/deploy-backend.yml
@@ -1,6 +1,5 @@
 name: Deploy backend
 on:
-  pull_request:
   push:
     paths:
       supabase/migrations/**.sql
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-backend.yml
```diff
@@ -0,0 +1,26 @@
+name: Deploy backend
+on:
+  push:
+    paths:
+      supabase/migrations/**.sql
```
stonelink, line 5: I wonder about the name of this workflow - it responds to changes in migrations but not other potential backend deployments, like deploying Edge Functions on changes in `supabase/functions`. I'm pretty sure that I had to [deploy Function](https://supabase.com/docs/guides/functions/deploy#deploy-your-edge-functions) changes manually, but I forgot to document in the README, sorry.

So I guess I'm saying: Should we change the name of this workflow to be about migrations, and raise a ticket for adding a workflow to deploy functions?

stonelink: Btw it's great to see that you're adding this, thank you!

rickaw: Ooh yeah that makes sense, I've made this specific to migrations now and raised a [ticket for function deployment](https://softwiretech.atlassian.net/browse/VFB-109!

stonelink (approved):

PR #119, opened 2024-02-27
VFB-107: fix lists page error

rickaw:
## What's changed
- Seed the website_data table in the local database to show comments at the top
- Use Supabase-native way to get one row out of the database
- Fix the lists table not showing the quantities and notes

| Before | After |
| --- | --- |
| image | image |

- [x] Make sure you've verified it works on dev via `npm run dev`
- [x] Make sure you've verified it works on prod via `npm run build` and `npm run start`


src/common/fetch.ts
```diff
@@ -107,9 +107,16 @@ export const fetchLists = async (supabase: Supabase): Promise<Schema["lists"][]>
 };
 
 export const fetchComment = async (supabase: Supabase): Promise<string> => {
-    const { data, error } = await supabase.from("website_data").select().eq("name", "lists_text");
+    const { data, error } = await supabase
+        .from("website_data")
+        .select()
+        .eq("name", "lists_text")
+        .limit(1)
+        .single();
```
stonelink, line 115: Ah great, so it will actually recognise the error now

src/app/lists/ListDataview.tsx
```diff
@@ -102,8 +102,8 @@ const ListsDataView: React.FC<Props> = (props) => {
                     .map(([key]) => [
                         key,
                         {
-                            quantity: row[`${key}_quantity` as keyof Schema["lists"]],
-                            notes: row[`${key}_notes` as keyof Schema["lists"]],
+                            quantity: row[`quantity_for_${key}` as keyof Schema["lists"]],
```
stonelink, line 105: Woops. I should have clocked this in the last review

src/common/fetch.ts
stonelink: Btw [VFB-57](https://softwiretech.atlassian.net/browse/VFB-57) is about the fact that nothing useful happens with the subsequent DatabaseError

src/app/lists/ListDataview.tsx
rickaw: Completely my fault for not checking that table on the page :)

src/common/fetch.ts
rickaw: Thanks for the info!

PR #120, opened 2024-02-27
change today date to start of today and end of today for packing date…

minalui:
… range

## What's changed
Replaced const today with two variables todayStart (start of the day) and todayEnd (end of the day) and used in initial packingDateRange.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image |
image
|

## Checklist
- [x] Make sure you've verified it works via `npm run build && npm run start`

### DIFF
--- a/src/app/parcels/ParcelsPage.tsx
+++ b/src/app/parcels/ParcelsPage.tsx
@@ -116,12 +116,13 @@ const fetchAndFormatParcelTablesData = async (
 };
 
 const ParcelsPage: React.FC<{}> = () => {
-    const today = dayjs();
+    const todayStart = dayjs().startOf("day");
+    const todayEnd = dayjs().endOf("day");
 
     const [isLoading, setIsLoading] = useState(true);
     const [packingDateRange, setPackingDateRange] = useState<DateRangeState>({
-        from: today,
-        to: today,
+        from: todayStart,
+        to: todayEnd,
     });
     const [tableData, setTableData] = useState<ParcelsTableRow[]>([]);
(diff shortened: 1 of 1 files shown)


src/app/parcels/ParcelsPage.tsx
```diff
@@ -116,12 +116,13 @@ const fetchAndFormatParcelTablesData = async (
 };
 
 const ParcelsPage: React.FC<{}> = () => {
-    const today = dayjs();
+    const todayStart = dayjs().startOf("day");
+    const todayEnd = dayjs().endOf("day");
```
rickaw, line 120: nitpick: would `startOfToday` and `endOfToday` be nicer to read?

src/app/parcels/ParcelsPage.tsx
```diff
@@ -116,12 +116,13 @@ const fetchAndFormatParcelTablesData = async (
...
 
     const [isLoading, setIsLoading] = useState(true);
     const [packingDateRange, setPackingDateRange] = useState<DateRangeState>({
-        from: today,
-        to: today,
+        from: todayStart,
+        to: todayEnd,
```
rickaw, line 125: Thanks for these changes! 

What happens when you change the "from" / "to" dates on the page? Would the dates be set to 
- the start of the "from" date and 
- the end of the "to" date
correctly?

rickaw (commented):
First PR woohoo

rickaw (approved):
nice! thank youu

src/app/parcels/ParcelsPage.tsx
rickaw: seems like this is already handled :)

PR #121, opened 2024-02-27
VFB-108: Add job to reset dev database data

rickaw:
## Overview
I've tried using `pg_dump` and `pg_restore` but the tools didn't quite work as expected - the connection was somehow refused even though the credentials were definitely correct. The somewhat easier / safer option (since Supbase db has a bunch of hidden tables for their housekeeping) would be to use Supabase CLI to reset the dev database, including the seed data. This is easily extendable to the prod pipeline. The only difference for prod is that we want both the "deploy" job and the "reset" job to be manual steps.

## What's changed
- Add a job to reset the dev database data

I have checked that:
- [x] the resetting of the data works with the pipeline (by automatically running it on the PR)
- [x] I have changed the config so that it only runs as a manual step on main 

Once this is merged, I still need to check that we can run this individual job manually without triggering the deploy job. If this works as expected, we can create similar jobs to deploy / reset database on prod. I expect the "reset database on prod" to be a one-off, but it would be good to have that up our sleeves.

### DIFF
--- a/.github/workflows/deploy-database.yml
+++ b/.github/workflows/deploy-database.yml
@@ -1,5 +1,7 @@
-name: Deploy backend
+name: Deploy database
 on:
+  pull_request:
+  workflow_dispatch:
   push:
     paths:
       supabase/migrations/**.sql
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-database.yml
```diff
@@ -24,3 +25,20 @@ jobs:
         run: npx supabase db push --dry-run
       - name: Apply the migration
         run: npx supabase db push
+
+  reset-dev-database-data:
+    name: Reset data in dev database
+    if: github.event_name == 'workflow_dispatch'
```
rickaw, line 31: This makes sure that we can only run this as a manual step on the default branch (main)

PR #122, opened 2024-02-28
VFB-110: Test starting supabase locally

rickaw:
## What's changed
- Add SUPABASE_LOCAL_SERVICE_ROLE_KEY and SUPABASE_LOCAL_ANON_KEY to secrets to be used in the pipeline
- When running the e2e tests in the pipeline, start a local Supabase instance and create necessary users / bucket.
- Modify the e2e tests to target the local Supabase instance
- Split the `Check linting and run component tests` job into `Check linting` and `Run component tests`


stonelink (approved):
Nice. Yay Checkout V4!

PR #124, opened 2024-02-28
VFB-35: parcel modal statuses

AmeliaIsCoding:
## What's changed
Added paginated table of events to the bottom of the parcel details modal. Columns are "event", which displays the event name following by paranthesised event data, if any, and "timestamp", which displays the timestamp in DD/MM/YYYY, HH:MM:SS format.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist
- [x] Make sure you've verified it works via `npm run build && npm run start`

### DIFF
--- a/src/app/parcels/getExpandedParcelDetails.ts
+++ b/src/app/parcels/getExpandedParcelDetails.ts
@@ -87,8 +87,8 @@ export interface ExpandedParcelData extends Data {
 }
 
 export interface ExpandedParcelDetails {
-    expandedParcelData: ExpandedParcelData,
-    events: EventTableRow[]
+    expandedParcelData: ExpandedParcelData;
+    events: EventTableRow[];
 }
 
 export const rawDataToExpandedParcelData = (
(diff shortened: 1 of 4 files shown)


src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -190,9 +199,27 @@ export const formatBreakdownOfChildrenFromFamilyDetails = (
...
 };
 
+export const processEventsDetails = (
+    events: Pick<Schema["events"], "event_data" | "event_name" | "timestamp">[]
+): EventTableRow[] => {
+    const eventTableRows: EventTableRow[] = [];
+
+    for (const event of events) {
+        eventTableRows.push({
+            eventInfo: `${event.event_name}` + (event.event_data ? ` (${event.event_data})` : ""),
```
rickaw, line 209:
```suggestion
            eventInfo: event.event_name + (event.event_data ? ` (${event.event_data})` : ""),
```
Is this equivalent?

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -190,9 +199,27 @@ export const formatBreakdownOfChildrenFromFamilyDetails = (
...
+            timestamp: new Date(event.timestamp),
+        });
+    }
```
rickaw, line 212:
```suggestion
    return events.map((event) => ({
        eventInfo: event.event_name + (event.event_data ? ` (${event.event_data})` : ""),
        timestamp: new Date(event.timestamp),
    }));
```

We can use `map` to do the exact same thing here :)

These are some of the most useful methods on JS arrays that might be interesting to look at:
- `.map`
- `.forEach`
- `.filter`
- `.reduce`

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -82,9 +86,14 @@ export interface ExpandedParcelDetails extends Data {
     collection: string;
 }
 
-export const rawDataToExpandedParcelDetails = (
+export interface ExpandedParcelDetails {
+    expandedParcelData: ExpandedParcelData;
+    events: EventTableRow[];
```
rickaw, line 91: Was there any reason not to add `events` directly to the original `ExpandedParcelData`?

rickaw (commented):
The table looks great! Left a few minor comments :)

src/app/parcels/getExpandedParcelDetails.ts
AmeliaIsCoding: Yup, have changed

src/app/parcels/getExpandedParcelDetails.ts
AmeliaIsCoding: Thanks!

src/app/parcels/getExpandedParcelDetails.ts
AmeliaIsCoding: Yes, expandedParcelData is passed into the DataViewer component, so I think the ExpandedParcelData type must extend the Data type, which only allows strings, arrays of strings, booleans, numbers and null. events needs to be an array of EventTableRow, so that we can easily pass it into the Table component. I decided it would be easiest to make a new type ExpandedParcelDetails to contain both expandedParcelData and events, so that we can still call the getExpandedParcelDetails function and get everything we need from that - mostly so that the database call for the event data can be made at the same time as the rest of the data. Is this ok?

rickaw: That's completely fair, thanks!

rickaw (approved):

PR #125, opened 2024-02-28
VFB-69: Add client name button to parcel page and opens client modal when clicked

minalui:
## What's changed
- Add button to parcels page which displays client name
- On click of button, modal opens with client details

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image| image
image |

## Checklist


src/app/parcels/form/ParcelForm.tsx
```diff
@@ -178,10 +197,25 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
                 <Title>Parcel Form</Title>
+                <Button
+                    variant="contained"
+                    type="button"
+                    onClick={() => {
+                        if (clientId) {
+                            getExpandedClientDetails(clientId).then((data) => {
+                                setActiveData(data);
+                                setModalIsOpen(true);
+                            });
```
rickaw, line 213: Sorry I wasn't clear before:

Seems like we call `getExpandedClientDetails` which gets all the client details twice. My original suggestion was to create a function that just gets the full name of the client first, and when needed, we can separately retrieve the rest.

However, I'm not sure how much of a performance gain that is, considering we would be selecting less fields from a row in the table. So I think the best way to tackle this would be to:
- have a state `clientDetails` (which I think should  replace `activeData`)
- call `getExpandedClientDetails` in useEffect (as we currently do) and store the response into `clientDetails`
- when the client details button is clicked, simply do `setModalIsOpen(true)` to open the modal

rickaw (commented):
Thanks! 

We could probably improve how we display the name in the future (Maybe a text for the name and a "i" button to show the details?)

minalui:
Finally passing tests!

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -182,6 +213,21 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
         <CenterComponent>
             <StyledForm>
                 <Title>Parcel Form</Title>
+                {clientDetails && (
+                    <StyledName>
+                        <h2>{clientDetails?.fullName}</h2>
```
rickaw, line 218: I think you can drop the `?` as we are checking `clientDetails` exists already

```suggestion
                        <h2>{clientDetails.fullName}</h2>
```

rickaw (approved):
Nice work!!

PR #130, opened 2024-02-29
Vfb 74 delete parcel action

AmeliaIsCoding:
## What's changed
 Added action to allow parcels to be deleted. The delete action modal shows the parcel details and cancel/delete buttons. On "deletion", an event gets added with name "Request Deleted", but the parcel(s) remain in the database. However, parcels which have a "Request Deleted" event as their most recent event, are filtered out from the parcels table by default. This behaviour can be changed by the user.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| imageimage|imageimageimageimage|

## Checklist

### DIFF
--- a/src/app/parcels/ActionBar/ActionBar.tsx
+++ b/src/app/parcels/ActionBar/ActionBar.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import React, { useState } from "react";
+import React, { SetStateAction, useState } from "react";
 import styled from "styled-components";
 import Button from "@mui/material/Button";
 import { ParcelsTableRow } from "@/app/parcels/getParcelsTableData";
@@ -13,6 +13,7 @@ import { ArrowDropDown } from "@mui/icons-material";
 interface Props {
     selected: number[];
     data: ParcelsTableRow[];
+    setSelected: React.Dispatch<SetStateAction<number[]>>;
 }
 
 const ActionsContainer = styled(ControlContainer)`
--- a/src/app/parcels/ActionBar/ActionBar.tsx
+++ b/src/app/parcels/ActionBar/ActionBar.tsx
@@ -11,10 +11,8 @@ import { ControlContainer } from "@/components/Form/formStyling";
 import { ArrowDropDown } from "@mui/icons-material";
 
 interface Props {
-    selectedRowIndices: number[];
-    parcels: ParcelsTableRow[];
-    setSelectedRowIndices: React.Dispatch<SetStateAction<number[]>>;
-    setSelectedCheckboxes: React.Dispatch<SetStateAction<boolean[]>>;
+    selectedParcels: ParcelsTableRow[];
+    onDeleteParcels: (parcels: ParcelsTableRow[]) => void;
 }
 
 const ActionsContainer = styled(ControlContainer)`
@@ -37,18 +35,15 @@ const ActionBar: React.FC<Props> = (props) => {
         <>
             <ActionsContainer>
                 <Statuses
-                    parcels={props.parcels}
-                    selectedRowIndices={props.selectedRowIndices}
+                    selectedParcels={props.selectedParcels}
                     statusAnchorElement={statusAnchorElement}
                     setStatusAnchorElement={setStatusAnchorElement}
                     modalError={modalError}
                     setModalError={setModalError}
                 />
                 <Actions
-                    parcels={props.parcels}
-                    setSelectedRowIndices={props.setSelectedRowIndices}
-                    selectedRowIndices={props.selectedRowIndices}
-                    setSelectedCheckboxes={props.setSelectedCheckboxes}
+                    selectedParcels={props.selectedParcels}
+                    onDeleteParcels={props.onDeleteParcels}
                     actionAnchorElement={actionAnchorElement}
                     setActionAnchorElement={setActionAnchorElement}
                     modalError={modalError}
--- a/src/app/parcels/ActionBar/ActionBar.tsx
+++ b/src/app/parcels/ActionBar/ActionBar.tsx
@@ -11,9 +11,9 @@ import { ControlContainer } from "@/components/Form/formStyling";
 import { ArrowDropDown } from "@mui/icons-material";
 
 interface Props {
-    selected: number[];
+    selectedRowIndices: number[];
     data: ParcelsTableRow[];
-    setSelected: React.Dispatch<SetStateAction<number[]>>;
+    setSelectedRowIndices: React.Dispatch<SetStateAction<number[]>>;
 }
 
 const ActionsContainer = styled(ControlContainer)`
--- a/src/app/parcels/ActionBar/ActionBar.tsx
+++ b/src/app/parcels/ActionBar/ActionBar.tsx
@@ -37,14 +37,18 @@ const ActionBar: React.FC<Props> = (props) => {
         <>
             <ActionsContainer>
                 <Statuses
-                    {...props}
+                    parcels={props.parcels}
+                    selectedRowIndices={props.selectedRowIndices}
                     statusAnchorElement={statusAnchorElement}
                     setStatusAnchorElement={setStatusAnchorElement}
                     modalError={modalError}
                     setModalError={setModalError}
                 />
                 <Actions
-                    {...props}
+                    parcels={props.parcels}
+                    setSelectedRowIndices={props.setSelectedRowIndices}
+                    selectedRowIndices={props.selectedRowIndices}
+                    setSelectedCheckboxes={props.setSelectedCheckboxes}
                     actionAnchorElement={actionAnchorElement}
                     setActionAnchorElement={setActionAnchorElement}
                     modalError={modalError}
--- a/src/app/parcels/ActionBar/ActionBar.tsx
+++ b/src/app/parcels/ActionBar/ActionBar.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import React, { SetStateAction, useState } from "react";
+import React, { useState } from "react";
 import styled from "styled-components";
 import Button from "@mui/material/Button";
 import { ParcelsTableRow } from "@/app/parcels/getParcelsTableData";
@@ -10,7 +10,7 @@ import Actions from "@/app/parcels/ActionBar/Actions";
 import { ControlContainer } from "@/components/Form/formStyling";
 import { ArrowDropDown } from "@mui/icons-material";
 
-interface Props {
+export interface ActionBarProps {
     selectedParcels: ParcelsTableRow[];
     onDeleteParcels: (parcels: ParcelsTableRow[]) => void;
 }
@@ -25,7 +25,7 @@ const AlertBox = styled.div`
     border-radius: 0.5rem;
 `;
 
-const ActionBar: React.FC<Props> = (props) => {
+const ActionBar: React.FC<ActionBarProps> = (props) => {
     const [statusAnchorElement, setStatusAnchorElement] = useState<HTMLElement | null>(null);
     const [actionAnchorElement, setActionAnchorElement] = useState<HTMLElement | null>(null);
 
--- a/src/app/parcels/ActionBar/ActionBar.tsx
+++ b/src/app/parcels/ActionBar/ActionBar.tsx
@@ -12,7 +12,7 @@ import { ArrowDropDown } from "@mui/icons-material";
 
 interface Props {
     selectedRowIndices: number[];
-    data: ParcelsTableRow[];
+    parcels: ParcelsTableRow[];
     setSelectedRowIndices: React.Dispatch<SetStateAction<number[]>>;
 }
 
--- a/src/app/parcels/ActionBar/ActionBar.tsx
+++ b/src/app/parcels/ActionBar/ActionBar.tsx
@@ -14,6 +14,7 @@ interface Props {
     selectedRowIndices: number[];
     parcels: ParcelsTableRow[];
     setSelectedRowIndices: React.Dispatch<SetStateAction<number[]>>;
+    setSelectedCheckboxes: React.Dispatch<SetStateAction<boolean[]>>;
 }
 
 const ActionsContainer = styled(ControlContainer)`
--- a/src/app/parcels/ActionBar/ActionBar.tsx
+++ b/src/app/parcels/ActionBar/ActionBar.tsx
@@ -13,6 +13,7 @@ import { ArrowDropDown } from "@mui/icons-material";
 export interface ActionBarProps {
     selectedParcels: ParcelsTableRow[];
     onDeleteParcels: (parcels: ParcelsTableRow[]) => void;
+    setIsLoading: (isLoading: boolean) => void;
 }
 
 const ActionsContainer = styled(ControlContainer)`
@@ -40,6 +41,7 @@ const ActionBar: React.FC<ActionBarProps> = (props) => {
                     setStatusAnchorElement={setStatusAnchorElement}
                     modalError={modalError}
                     setModalError={setModalError}
+                    setIsLoading={props.setIsLoading}
                 />
                 <Actions
                     selectedParcels={props.selectedParcels}
--- a/src/app/parcels/ActionBar/ActionBar.tsx
+++ b/src/app/parcels/ActionBar/ActionBar.tsx
@@ -13,7 +13,8 @@ import { ArrowDropDown } from "@mui/icons-material";
 export interface ActionBarProps {
     selectedParcels: ParcelsTableRow[];
     onDeleteParcels: (parcels: ParcelsTableRow[]) => void;
-    setIsLoading: (isLoading: boolean) => void;
+    willSaveParcelStatus: () => void;
+    hasSavedParcelStatus: () => void;
 }
 
 const ActionsContainer = styled(ControlContainer)`
@@ -41,7 +42,8 @@ const ActionBar: React.FC<ActionBarProps> = (props) => {
                     setStatusAnchorElement={setStatusAnchorElement}
                     modalError={modalError}
                     setModalError={setModalError}
-                    setIsLoading={props.setIsLoading}
+                    willSaveParcelStatus={props.willSaveParcelStatus}
+                    hasSavedParcelStatus={props.hasSavedParcelStatus}
                 />
                 <Actions
                     selectedParcels={props.selectedParcels}
--- a/src/app/parcels/ActionBar/Actions.tsx
+++ b/src/app/parcels/ActionBar/Actions.tsx
@@ -170,6 +170,7 @@ interface Props {
     setActionAnchorElement: React.Dispatch<React.SetStateAction<HTMLElement | null>>;
     modalError: string | null;
     setModalError: React.Dispatch<React.SetStateAction<string | null>>;
+    setSelectedCheckboxes: React.Dispatch<React.SetStateAction<boolean[]>>;
 }
 
 const Actions: React.FC<Props> = ({
@@ -180,6 +181,7 @@ const Actions: React.FC<Props> = ({
     setActionAnchorElement,
     modalError,
     setModalError,
+    setSelectedCheckboxes,
 }) => {
     const [selectedAction, setSelectedAction] = useState<PdfType | null>(null);
     const [labelQuantity, setLabelQuantity] = useState<number>(0);
@@ -251,6 +253,7 @@ const Actions: React.FC<Props> = ({
                             headerId="action-modal-header"
                             errorText={modalError}
                             actionType={value.actionType}
+                            setSelectedCheckboxes={setSelectedCheckboxes}
                             inputComponent={
                                 <ActionsInputComponent
                                     pdfType={key}
--- a/src/app/parcels/ActionBar/Actions.tsx
+++ b/src/app/parcels/ActionBar/Actions.tsx
@@ -44,7 +44,6 @@ type AvailableActionsType = {
         showSelectedParcelsInModal: boolean;
         errorCondition: (value: number) => boolean;
         errorMessage: string;
-        message?: string;
         downloadable: boolean;
     };
 };
@@ -79,7 +78,6 @@ const availableActions: AvailableActionsType = {
         showSelectedParcelsInModal: true,
         errorCondition: isNotAtLeastOne,
         errorMessage: "Please select at least one parcel.",
-        message: "Are you sure you want to delete the selected parcel request?",
         downloadable: false,
     },
 };
@@ -249,7 +247,6 @@ const Actions: React.FC<Props> = ({
                             header={key}
                             headerId="action-modal-header"
                             errorText={modalError}
-                            message={value.message}
                             downloadable={value.downloadable}
                             inputComponent={
                                 <ActionsInputComponent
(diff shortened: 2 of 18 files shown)


src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -149,6 +161,7 @@ const ActionsButton: React.FC<ActionsButtonProps> = ({
 
 interface Props {
     selected: number[];
+    setSelected: React.Dispatch<React.SetStateAction<number[]>>;
```
rickaw, line 164: shall we rename these to `selectedRowIndices`  / `setSelectedRowIndices` or something? It's quite hard to understand what this array of numbers is.

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -170,7 +184,13 @@ const Actions: React.FC<Props> = ({
     const [driverName, setDriverName] = useState("");
     const [collectionCentre, setCollectionCentre] = useState("");
 
-    const selectedData = Array.from(selected.map((index) => data[index]));
+    const [selectedData, setSelectedData] = useState(
+        Array.from(selected.map((index) => data[index]))
+    );
+
+    React.useEffect(() => {
```
rickaw, line 191: It's conventional to import `useEffect` directly like we do `useState` in this file so that we can write this without `React.`

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -193,6 +213,7 @@ const Actions: React.FC<Props> = ({
         setModalError(null);
         setDate(dayjs());
         setDriverName("");
+        setSelected([]);
```
rickaw, line 216: So when the modal closes, we unselect all the parcels, is that right?

If I select a parcel, click Delete Parcel Request and close the modal, the parcels I'd selected are still selected. If I then click Delete Parcel Request again, it gives me an error saying I need to select at least one parcel.

_Should_ we really be deselecting all parcels whenever the modal closes? Or only when the delete request is successful? I don't think we should deselect them when we download shipping labels etc.

src/app/parcels/ActionBar/ActionsModal.tsx
```diff
@@ -182,11 +198,24 @@ const ActionsModal: React.FC<ActionsModalProps> = (props) => {
                 {props.errorText && <small>{props.errorText}</small>}
                 {loadPdf ? (
                     <>
-                        <Heading> The PDF is ready to be downloaded. </Heading>
+                        {props.downloadable ? (
```
rickaw, line 201: instead of having a boolean `downloadable`, should we have a enum of some sort like
```
type ActionType = 'pdfDownload' | 'deleteParcel'
```
to be clearer?

src/app/parcels/ActionBar/ActionsModal.tsx
```diff
@@ -163,9 +171,17 @@ export const DayOverviewInput: React.FC<DayOverviewInputProps> = ({
...
+        parcels.map((parcel) => parcel.parcelId),
+        "Request Deleted"
+    );
+    return;
+}; //assume successful for now
+
 const ActionsModal: React.FC<ActionsModalProps> = (props) => {
-    const [loadPdf, setLoadPdf] = useState(false);
     const [loading, setLoading] = useState(false);
+    const [loadPdf, setLoadPdf] = useState(false);
```
rickaw, line 184: I'm not sure I fully understand the intention of these two states. It seems like
- we set both to false
- we set loading to true when we click on the action button
- then when loading is true and loadPdf is false (which is guaranteed to happen at this point), then we toggle both of them.

It seems like in essence this captures two states - (1) we haven't started the action (2) we have finished the action - in which case we should have just one boolean state. 

Am I missing something?

src/app/parcels/ActionBar/Statuses.tsx
```diff
@@ -83,7 +83,13 @@ const Statuses: React.FC<Props> = ({
     const [selectedStatus, setSelectedStatus] = useState<statusType | null>(null);
     const [statusModal, setStatusModal] = useState(false);
 
-    const selectedData = Array.from(selected.map((index) => data[index]));
+    const [selectedData, setSelectedData] = useState(
+        Array.from(selected.map((index) => data[index]))
+    );
+
+    React.useEffect(() => {
```
rickaw, line 90: ditto about importing useEffect

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -149,6 +161,7 @@ const ActionsButton: React.FC<ActionsButtonProps> = ({
 
 interface Props {
     selected: number[];
+    setSelected: React.Dispatch<React.SetStateAction<number[]>>;
     data: ParcelsTableRow[];
```
rickaw, line 165: Shall we rename this to `parcels` or something instead of a very generic `data`?

src/app/parcels/ActionBar/Statuses.tsx
```diff
@@ -83,7 +83,13 @@ const Statuses: React.FC<Props> = ({
     const [selectedStatus, setSelectedStatus] = useState<statusType | null>(null);
     const [statusModal, setStatusModal] = useState(false);
 
-    const selectedData = Array.from(selected.map((index) => data[index]));
+    const [selectedData, setSelectedData] = useState(
+        Array.from(selected.map((index) => data[index]))
+    );
```
rickaw, line 88: What's the reason behind having this as a state and updating it with useEffect? `selected` and `data` (we should've had a better name but alas) are both props so I didn't think this would be necessary.

src/app/parcels/ParcelsPage.tsx
```diff
@@ -147,12 +147,25 @@ const ParcelsPage: React.FC<{}> = () => {
...
-            .on("postgres_changes", { event: "*", schema: "public", table: "parcels" }, async () =>
-                setTableData(await fetchAndFormatParcelTablesData(packingDateRange))
+            .on(
+                "postgres_changes",
+                { event: "*", schema: "public", table: "parcels" },
+                async () => {
+                    setIsLoading(true);
+                    setTableData(await fetchAndFormatParcelTablesData(packingDateRange));
+                    setIsLoading(false);
+                }
```
rickaw, line 160: Shall we pull this out to a method elsewhere?

src/app/parcels/ActionBar/Actions.tsx
AmeliaIsCoding: Have done

src/app/parcels/ActionBar/Actions.tsx
AmeliaIsCoding: So when the modal closes we do unselect all parcels in the selected state, but the corresponding checkboxes still appear ticked in the table (as I couldn't figure out how to untick them), which is why you experienced the above behaviour
Ok I think I agree that it would be best to only deselect if the delete request is successful. -
The reason I did it like this was because if we successfully delete the parcel, it becomes hidden by default, so the user would have to find it again in order to manually unselect it, otherwise that deleted parcel will be passed into any other actions as one of the selected parcels.
I will change the behaviour to only deselect if deletion successful. But I think we should also find a way to make the corresponding checkboxes appear unticked, as they should do, but I couldn't figure this out - maybe another ticket ?

AmeliaIsCoding: Now resolved aside from unselected rows still showing as ticked checkboxes

src/app/parcels/ActionBar/ActionsModal.tsx
AmeliaIsCoding: Yes that's much better, thanks

src/app/parcels/ActionBar/Statuses.tsx
AmeliaIsCoding: Ok I've been unable to recreate the error that caused me to implement this, so I think I will change this back to how it was before, but have described the error below for the record:

since adding the status table in the database as a subscription to the parcels table in the website, on deleting parcels, the data in the parcel table changes. Therefore selectedData, because it is defined in terms of data, changes too. This was causing the below error (which luckily was in my google search history)

`Cannot update a component (`HotReload`) while rendering a different component (`StatusesBarModal`). To locate the bad setState() call inside `StatusesBarModal, follow the stack trace as described]`

And I followed the solution in this Stack Exchange post https://stackoverflow.com/questions/62336340/cannot-update-a-component-while-rendering-a-different-component-warning

Not sure why the error is no longer showing... slightly concerned but at least we have the commit history to look back on if we need ?

src/app/parcels/ActionBar/Statuses.tsx
AmeliaIsCoding: have lost this useEffect because removed this as a state

src/app/parcels/ActionBar/Actions.tsx
AmeliaIsCoding: Have done (and similar occurances)

src/app/parcels/ActionBar/ActionsModal.tsx
AmeliaIsCoding: For record
the 2 states are somehow needed for the "Create PDF" button to change to "Loading..." in the time that it takes to render the download PDF secondary page to the modal. In another ticket will look at combining these pages thus removing need for 2 states

src/components/Tables/Table.tsx
```diff
@@ -126,7 +128,16 @@ const Table = <Data,>({
...
     autoFilter = true,
+    selectedCheckboxes,
+    setSelectedCheckboxes,
 }: Props<Data>): React.ReactElement => {
+    const [selectedCheckboxesDefault, setSelectedCheckboxesDefault] = useState(
+        new Array<boolean>(inputData.length).fill(false)
+    );
+
+    selectedCheckboxes = selectedCheckboxes ?? selectedCheckboxesDefault;
+    setSelectedCheckboxes = setSelectedCheckboxes ?? setSelectedCheckboxesDefault;
```
rickaw, line 139: I think it's an anti-pattern to override the props. 

Another thing we should consider - do we really want the "default" stuff? It might be better if the table is clear about who should manage the checkbox state instead. What I would've expected is something like below, where we don't even need to pass the `setSelectedRowIndices` to the table. The table then uses `selectedRowIndices` to simply display the checkboxes in the correct states, and lets the parent do whatever it needs do to the checkbox states.

```tsx
// in Parcels page

const [selectedRowIndices, setSelectedRowIndices] = useState<number[]>([])

const selectOrDeselectRow = (rowIndex: number) => {
  setSelectedRowIndices(currentIndices => {
    if (currentIndices.includes(rowIndex)) {
      return currentIndices.filter(index => index !== rowIndex)
    }
    return currentIndices.concat([rowIndex])
  })

  // do whatever else we may want to do
}

return (
  ...
  <Table selectedRowIndices={selectedRowIndices} onCheckboxClicked={selectOrDeselectRow} ... />
)
```

src/components/Tables/Table.tsx
```diff
@@ -153,12 +164,8 @@ const Table = <Data,>({
 
     const [data, setData] = useState(inputData);
 
-    const [selectedCheckboxes, setSelectedCheckboxes] = useState(
-        new Array<boolean>(data.length).fill(false)
-    );
-
     const updateCheckboxes = (newSelection: boolean[]): void => {
-        setSelectedCheckboxes(newSelection);
+        setSelectedCheckboxes!(newSelection);
```
rickaw, line 168: Having a non-null assertion is generally an anti-pattern unless it's absolutely necessary.

src/app/parcels/ActionBar/ActionBar.tsx
```diff
@@ -11,8 +11,10 @@ import { ControlContainer } from "@/components/Form/formStyling";
 import { ArrowDropDown } from "@mui/icons-material";
 
 interface Props {
-    selected: number[];
-    data: ParcelsTableRow[];
+    selectedRowIndices: number[];
+    parcels: ParcelsTableRow[];
+    setSelectedRowIndices: React.Dispatch<SetStateAction<number[]>>;
+    setSelectedCheckboxes: React.Dispatch<SetStateAction<boolean[]>>;
```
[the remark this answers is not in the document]
AmeliaIsCoding: Ok I agree, have done

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -36,13 +37,15 @@ type PdfType =
     | "Download Shipping Labels"
     | "Download Shopping Lists"
     | "Download Driver Overview"
-    | "Download Day Overview";
+    | "Download Day Overview"
+    | "Delete Parcel Request";
```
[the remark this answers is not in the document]
AmeliaIsCoding: I changed the name of this to `ActionName`

src/app/parcels/ActionBar/Statuses.tsx
```diff
@@ -59,31 +59,31 @@ export const saveParcelStatus = async (
...
     statusAnchorElement,
     setStatusAnchorElement,
     modalError,
     setModalError,
 }) => {
     const [selectedStatus, setSelectedStatus] = useState<statusType | null>(null);
     const [statusModal, setStatusModal] = useState(false);
-
-    const selectedData = Array.from(selected.map((index) => data[index]));
+    const selectedData = Array.from(selectedRowIndices.map((index) => parcels[index]));
```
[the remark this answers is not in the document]
AmeliaIsCoding: No reason, have changed

src/app/parcels/ActionBar/ActionsModal.tsx
```diff
@@ -163,9 +175,17 @@ export const DayOverviewInput: React.FC<DayOverviewInputProps> = ({
     );
 };
 
+const deleteParcels = async (parcels: ParcelsTableRow[]): Promise<void> => {
+    await saveParcelStatus(
+        parcels.map((parcel) => parcel.parcelId),
+        "Request Deleted"
+    );
+    return;
+}; //assume successful for now
```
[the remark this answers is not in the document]
AmeliaIsCoding: It could be. I'm not sure what we should do if the request isn't successful. Maybe an error message in a pop up or in a snack bar thing?

src/components/Tables/Table.tsx
AmeliaIsCoding: Thank you, have implemented this, and similarly for ActionBar + its children

src/components/Tables/Table.tsx
AmeliaIsCoding: Have removed as part of checkbox refactoring

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -74,16 +74,18 @@ const StatusesBarModal: React.FC<StatusesBarModalProps> = (props) => {
...
-                        <StatusText key={index}>
-                            {parcel.addressPostcode}
-                            {parcel.fullName && ` - ${parcel.fullName}`}
-                            {parcel.collectionDatetime &&
-                                `\n @ ${dayjs(parcel.collectionDatetime!).format(
-                                    "DD/MM/YYYY HH:mm"
-                                )}`}
-                        </StatusText>
-                    );
+                    if (parcel) {
```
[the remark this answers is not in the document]
AmeliaIsCoding: Was able to remove as part of a quick refactor

src/app/parcels/ActionBar/Statuses.tsx
AmeliaIsCoding: Think this error is unlikely to happen now as this modal no longer deals with this state

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -73,18 +67,7 @@ const StatusesBarModal: React.FC<StatusesBarModalProps> = (props) => {
                     <TimePicker value={date} onChange={onTimeChange} disableFuture />
                 </Row>
                 Applying To:
```
rickaw, line 69: We have "_ Parcels selected" as part of ShowParcels. Do we still want "Applying to:"?

src/app/parcels/ParcelsPage.tsx
```diff
@@ -300,10 +366,14 @@ const ParcelsPage: React.FC<{}> = () => {
...
-                <ActionBar data={tableData} selected={selected} />
+                <ActionBar
+                    selectedParcels={selectedRowIndices.map((index) => tableData[index])}
+                    onDeleteParcels={deleteParcels}
+                    setIsLoading={setIsLoading}
+                />
             </PreTableControls>
             {isLoading ? (
-                <></>
+                <Loading />
```
rickaw, line 376: The parcels page is already wrapped in `<main>` and Loading has `<main>` too so we would be showing
```
<main>
  ...
  <main>
    ...
  </main>
  ...
</main>
```

I think we should just use
```
<Centerer>
  <CircularProgress />
</Centerer>
```
or 
```
<CircularProgress />
```

src/app/parcels/ParcelsPage.tsx
```diff
@@ -147,18 +153,68 @@ const ParcelsPage: React.FC<{}> = () => {
...
+    const selectOrDeselectRow = (rowIndex: number): void => {
+        setSelectedRowIndices((currentIndices) => {
+            if (currentIndices.includes(rowIndex)) {
+                return currentIndices.filter((index) => index !== rowIndex);
+            }
+            return currentIndices.concat([rowIndex]);
+        });
+    };
+
+    const toggleAllCheckBox = (isAllCheckBoxSelected: boolean): void => {
```
rickaw, line 202: Do you need this parameter at all? Could we just use the `isAllCheckBoxSelected` state?
```suggestion
    const toggleAllCheckBox = (): void => {
```

src/components/Tables/Table.cy.tsx
```diff
@@ -78,60 +78,83 @@ const headers = [
...
+    const selectOrDeselectRow = (rowIndex: number): void => {
+        setSelectedRowIndices((currentIndices) => {
+            if (currentIndices.includes(rowIndex)) {
+                return currentIndices.filter((index) => index !== rowIndex);
+            }
+            return currentIndices.concat([rowIndex]);
+        });
+    };
+
+    const toggleAllCheckBox = (isAllCheckBoxSelected: boolean): void => {
```
rickaw, line 113: ditto about the parameter

src/components/Tables/Table.cy.tsx
```diff
@@ -78,60 +78,83 @@ const headers = [
...
+        displayed: true,
+        selectedRowIndices: selectedRowIndices,
+        isAllCheckboxChecked: isAllCheckBoxSelected,
+        onCheckboxClicked: (rowIndex: number) => selectOrDeselectRow(rowIndex),
+        onAllCheckboxClicked: (isAllCheckBoxSelected: boolean) =>
+            toggleAllCheckBox(isAllCheckBoxSelected),
+    };
+    const falseCheckboxConfig: CheckboxConfig = {
+        displayed: false,
+    };
```
rickaw, line 139: nitpick: let's add an empty line between these variables / return

rickaw (commented):
Love the changes in state management!

src/app/parcels/ActionBar/ShowParcels.tsx
```diff
@@ -0,0 +1,49 @@
...
+    return (
+        <>
+            <Heading>{props.parcels.length === 1 ? "Parcel" : "Parcels"} selected:</Heading>
+            {props.parcels.slice(0, props.maxParcelsToShow).map((parcel, index) => {
+                return (
+                    <ListItem key={index}>
+                        {parcel.addressPostcode}
+                        {parcel.fullName && ` - ${parcel.fullName}`}
+                        {parcel.collectionDatetime &&
+                            `\n @ ${dayjs(parcel.collectionDatetime!).format("DD/MM/YYYY HH:mm")}`}
```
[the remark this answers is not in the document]
AmeliaIsCoding: no, have removed

src/app/parcels/ActionBar/ShowParcels.tsx
```diff
@@ -0,0 +1,49 @@
...
+            font-weight: 800;
+        `}
+`;
+
+interface ShowParcelsProps {
+    parcels: ParcelsTableRow[];
+    maxParcelsToShow: number;
+}
+
+const ShowParcels: React.FC<ShowParcelsProps> = (props) => {
```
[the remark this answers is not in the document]
AmeliaIsCoding: yes, have changed

src/app/parcels/ActionBar/StatusesModal.tsx
AmeliaIsCoding: Good spot!!

src/app/parcels/ParcelsPage.tsx
AmeliaIsCoding: Have done (with centerer)

rickaw (approved):
Nice one! A couple of minor comments but feel free to merge once they're addressed :)

src/app/parcels/ActionBar/ShowParcels.tsx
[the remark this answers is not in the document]
AmeliaIsCoding: have done!

src/app/parcels/ActionBar/ShowParcels.tsx
```diff
@@ -0,0 +1,45 @@
...
+            {props.parcels.slice(0, props.maxParcelsToShow).map((parcel) => {
+                return (
+                    <ListItem key={parcel.parcelId}>
+                        {parcel.addressPostcode}
+                        {parcel.fullName && ` - ${parcel.fullName}`}
+                        {parcel.collectionDatetime &&
+                            ` @ ${dayjs(parcel.collectionDatetime!).format("DD/MM/YYYY HH:mm")}`}
+                    </ListItem>
+                );
+            })}
```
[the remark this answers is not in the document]
AmeliaIsCoding: it works, have changed

src/app/parcels/ParcelsPage.tsx
```diff
@@ -198,18 +203,68 @@ const ParcelsPage: React.FC<{}> = () => {
...
         const subscriptionChannel = supabase
             .channel("parcels-table-changes")
-            .on("postgres_changes", { event: "*", schema: "public", table: "parcels" }, async () =>
-                setTableData(await fetchAndFormatParcelTablesData(packingDateRange))
+            .on(
+                "postgres_changes",
+                { event: "*", schema: "public", table: "parcels" },
+                async () => {
+                    await loadParcelTableData();
+                }
```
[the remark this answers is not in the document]
AmeliaIsCoding: Yes that works, ty

src/app/parcels/ParcelsPage.tsx
```diff
@@ -168,7 +169,11 @@ const ParcelsPage: React.FC<{}> = () => {
     });
     const [tableData, setTableData] = useState<ParcelsTableRow[]>([]);
     const [selectedParcelId, setSelectedParcelId] = useState<string | null>(null);
-    const [selected, setSelected] = useState<number[]>([]);
+
+    const [selectedRowIndices, setSelectedRowIndices] = useState<number[]>([]);
+    const [isAllCheckBoxSelected, setAllCheckBoxSelected] = useState(false);
+    const timer = useRef<ReturnType<typeof setTimeout> | null>(null);
```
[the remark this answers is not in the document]
AmeliaIsCoding: ok have changed!

PR #131, opened 2024-02-29
VFB-96: Fix empty clients table

AmeliaIsCoding:
## What's changed
Client and parcel tables both now consistently load with data on first render (and others). Removed unnecessary "staleFetch" logic in client page, and fixed "staleFetch" logic in parcels page.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|imageimage|imageimage|

## Checklist


src/app/parcels/ParcelsPage.tsx
```diff
@@ -130,20 +130,17 @@ const ParcelsPage: React.FC<{}> = () => {
...
-
         (async () => {
             setIsLoading(true);
-            const formattedData = await fetchAndFormatParcelTablesData(packingDateRange);
-            if (!staleFetch) {
-                setTableData(formattedData);
+            const dateRangeToFetch = packingDateRange;
+            const tableData = await fetchAndFormatParcelTablesData(dateRangeToFetch);
+
+            if (dateRangeToFetch === packingDateRange) {
```
rickaw, line 138: If you follow the above comment, you'll see that this condition becomes false always.

`===` checks that either
- the reference to the object is the same (if the things you're comparing are reference types e.g. object)
- the value is the same (if the things you're comparing are value types e.g. number, string)

We should use explictly check that the From and To dates are identical instead - see [DayJs docs](https://day.js.org/docs/en/query/is-same)

AmeliaIsCoding: Thank you, makes sense

src/app/parcels/ParcelsPage.tsx
```diff
@@ -130,20 +130,17 @@ const ParcelsPage: React.FC<{}> = () => {
...
 
     useEffect(() => {
-        let staleFetch = false;
-
         (async () => {
             setIsLoading(true);
-            const formattedData = await fetchAndFormatParcelTablesData(packingDateRange);
-            if (!staleFetch) {
-                setTableData(formattedData);
+            const dateRangeToFetch = packingDateRange;
```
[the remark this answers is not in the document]
AmeliaIsCoding: Have changed

src/app/parcels/ParcelsPage.tsx
```diff
@@ -115,6 +115,15 @@ const fetchAndFormatParcelTablesData = async (
     return formattedData;
 };
 
+const areDateRangesIdentical = (
+    dateRangeA: DateRangeState,
+    dateRangeB: DateRangeState
+): boolean => {
+    return dateRangeA.from && dateRangeB.from && dateRangeA.to && dateRangeB.to
+        ? dateRangeA.from?.isSame(dateRangeB.from) && dateRangeA.to?.isSame(dateRangeB.to)
+        : false;
```
rickaw, line 124: Shouldn't we return true if From is `null` for both, and To is the same for both for example?

AmeliaIsCoding: ok, I've changed the logic accordingly so the following are the 4 possibilities for returning true:

From is the same for both and To is the same for both => true
From is `null` for both and To is the same for both => true
From is the same for both and To is `null` for both => true
From is `null` for both and To is `null` for both => true

any other possible combination will return false

rickaw (approved):
Looks great!!

rickaw:
could you change the PR title to `VFB-96: ___`? I think this is what would show up as the commit title when you merge to main

PR #133, opened 2024-03-01
VFB-99: Add migration and reset steps for prod

rickaw:
## What's changed
- Add migration and reset steps for prod
  - This is triggered by pushes to the main branch
- We should now target the `dev` branch when creating pull requests. When we're ready to promote the recent changes to prod, we can move the `main` branch up to the desired commit (or the latest `dev` commit), and that deploys the website and triggers the migration step.
- Add `prod` and `dev` GitHub environments. Workflow jobs for `prod` require approval from myself or @stonelink (config is found [here](https://github.com/VauxhallFoodbank/vauxhall-foodbank/settings/environments))

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -5,6 +5,7 @@ on:
       - .github/**
   push:
     branches:
+      dev
       main
 jobs:
   lint:
--- a/.github/workflows/deploy-database.yml
+++ b/.github/workflows/deploy-database.yml
@@ -1,6 +1,5 @@
 name: Deploy database
 on:
-  workflow_dispatch:
   push:
     paths:
       supabase/migrations/**.sql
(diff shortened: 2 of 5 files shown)


.github/workflows/checks.yml
```diff
@@ -5,6 +5,7 @@ on:
       - .github/**
   push:
     branches:
+      dev
       main
```
rickaw, line 9: I've initially considered using `prod` branch for prod and `main` for development, but this meant that the live site URL would change from `main.___.amplifyapp.com` to `prod.___.amplifyapp.com` so opted to use `dev` for dev and `main` for prod.

stonelink: Thanks for clarifying that, because I don't know how the AWS Amplify stuff was setup.

stonelink (approved):

PR #135, opened 2024-03-05
VFB 76-Controls on Lists comments are a long way from the text

mtharkins:
## What's changed

- Styling amended to reduce the gap between the buttons & comments box. 
- Comment box styling amended to fill more empty space of the comments container to the right
- Adjusted styling to fix uneven spacing between 800px-980px screen size range. 
- Button labels changed from "Reset" to "Cancel" and "Submit" to "Save"
- Soft error handling added to show when unsaved changes have been made in the comment box

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist


node-v20.11.1-x64.msi
rickaw, line 1: I don't think this file was intentional

python-3.12.2-amd64 (1).exe
rickaw, line 1: Think this was also unintentional

python-3.12.2-amd64.exe
rickaw, line 1: We should be able to remove this too. If your computer somehow really needs this exe file, we could add it to the list of files to ignore, but I don't suppose that's the case?

vs_buildtools__a1561ceeec274ccb9195e1680fd80985.exe
rickaw, line 1: This too :)

src/app/lists/CommentBox.tsx
```diff
@@ -73,10 +74,22 @@ const SuccessText = styled.p`
...
+    color: ${(props) => props.theme.primary.background[3]};
+    margin: 1rem 0 0;
+    font-size: 0.8rem;
+`;
+
 const CommentBox: React.FC<CommentProps> = ({ originalComment }) => {
     const [value, setValue] = useState(originalComment);
     const [errorMessage, setErrorMessage] = useState("");
     const [successMessage, setSuccessMessage] = useState("");
+    const [unsavedChanges, setUnsavedChanges] = useState(false);
```
rickaw, line 87: the variable name `unsavedChanges` would imply a list of unsaved changes. We could rename it to something like `hasUnsavedChanges`?

I also don't think we actually need a state here. Since `hasUnsavedChanges` is just depends on `value` and `originalComment`, we could do
```ts
const hasUnsavedChanges = value !== originalComment
```
and this would get updated whenever one of them changes.

rickaw (commented):
Love the warning text!

rickaw (approved):
Looks good to me - feel free to merge once you sort out the linting issue! First merge woohoo

src/app/lists/CommentBox.tsx
```diff
@@ -92,16 +101,17 @@ const CommentBox: React.FC<CommentProps> = ({ originalComment }) => {
             window.location.reload();
         }
     };
+
     return (
         <Wrapper>
             <HeaderAndButtonContainer>
-                <h2>Comments</h2>
```
rickaw, line 98: Ooh i think this might be causing the problem? Maybe the test sees an h3 after seeing an h1 without seeing an h2?

PR #137, opened 2024-03-05
Vfb 112 reset password link points to localhost 3000

minalui:
## What's changed
- Replaced hardcoded route for redirect at login page to {baseUrl}/auth/callback

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| paste_here | paste_here |

## Checklist

This will need to be checked through the deployed website***

### DIFF
--- a/src/components/LoginPanel.tsx
+++ b/src/components/LoginPanel.tsx
@@ -59,7 +59,7 @@ const LoginPanel: React.FC<{}> = () => {
     useEffect(() => {
         setBaseUrl(window.location.origin);
         setLoaded(true);
-    }, [baseUrl]);
+    }, []);
 
     return (
         <MiddleDiv elevation={5} data-loaded={loaded} id="login-panel">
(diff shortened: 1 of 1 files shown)


src/components/LoginPanel.tsx
```diff
@@ -54,10 +54,12 @@ const LoginPanel: React.FC<{}> = () => {
     const theme = useTheme();
 
     const [loaded, setLoaded] = React.useState(false);
+    const [baseUrl, setBaseUrl] = useState<string | null>(null);
 
     useEffect(() => {
+        setBaseUrl(window.location.origin);
         setLoaded(true);
-    }, []);
```
rickaw, line 60: is this intentional?

src/components/LoginPanel.tsx
```diff
@@ -54,10 +54,12 @@ const LoginPanel: React.FC<{}> = () => {
     const theme = useTheme();
 
     const [loaded, setLoaded] = React.useState(false);
```
rickaw, line 56: let's keep this consistent

```suggestion
    const [loaded, setLoaded] = useState(false);
```

rickaw (approved):
Thanks! A few comments but feel free to merge once you're happy

src/components/LoginPanel.tsx
minalui: I've put baseUrl in the dependency - not sure why it thinks I've removed the full line.

PR #139, opened 2024-03-11
VFB 66 - onSwapRows amended to directly update the row order

mtharkins:
## What's changed
The login of onSwapRows has been updated to use the listOfIngridients function as props to update the row_order value of listsData

## Checklist

### DIFF
--- a/src/app/lists/ListDataview.tsx
+++ b/src/app/lists/ListDataview.tsx
@@ -38,7 +38,7 @@ interface QuantityAndNotes {
 
 interface ListDataViewProps {
     listOfIngridients: Schema["lists"][];
-    setData: React.Dispatch<React.SetStateAction<Schema["lists"][]>>;
+    setListOfIngridients: React.Dispatch<React.SetStateAction<Schema["lists"][]>>;
     comment: string;
 }
 
@@ -82,8 +82,8 @@ const listsColumnStyleOptions: ColumnStyles<ListRow> = {
 };
 
 const ListsDataView: React.FC<ListDataViewProps> = ({
-    listOfIngridients: setListOfIngridients,
-    setData,
+    listOfIngridients,
+    setListOfIngridients: setData,
     comment,
 }) => {
     const [modal, setModal] = useState<EditModalState>();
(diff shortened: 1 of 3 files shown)


src/app/lists/ListDataview.tsx
```diff
@@ -36,8 +36,9 @@ interface QuantityAndNotes {
     notes: string | null;
 }
 
-interface Props {
-    data: Schema["lists"][];
+interface ListDataViewProps {
+    listOfIngridients: Schema["lists"][];
+    setData: React.Dispatch<React.SetStateAction<Schema["lists"][]>>;
```
rickaw, line 41: should we also rename this, like `setListOfIngredients`?

src/app/lists/ListDataview.tsx
```diff
@@ -133,9 +134,28 @@ const ListsDataView: React.FC<Props> = (props) => {
...
+
+        item1.row_order = order2;
+        item2.row_order = order1;
+
+        const newData = [...listOfIngridients];
+
+        newData[index1] = item2;
+        newData[index2] = item1;
+
+        setData(newData);
```
rickaw, line 158: maybe worth putting this in a separate function as discussed :)

rickaw (commented):
Nice work! a few comments

src/app/lists/ListDataview.tsx
mtharkins: Renamed

src/app/lists/ListDataview.tsx
mtharkins: Moved this into a separate function

src/app/lists/ListDataview.tsx
```diff
@@ -80,18 +81,22 @@ const listsColumnStyleOptions: ColumnStyles<ListRow> = {
     ),
 };
 
-const ListsDataView: React.FC<Props> = (props) => {
+const ListsDataView: React.FC<ListDataViewProps> = ({
+    listOfIngridients,
+    setListOfIngridients: setData,
```
rickaw, line 86: oops, seems like the same thing has happened here too. If we could just use `setListOfIngridients` throughout, that'd be fab

src/app/lists/ListDataview.tsx
```diff
@@ -114,9 +119,33 @@ const ListsDataView: React.FC<Props> = (props) => {
...
+        const row1Item = listOfIngridients[row1Index];
+        const row1Order = row1Item.row_order;
+
+        const row2Item = listOfIngridients[row2Index];
+        const row2Order = row2Item.row_order;
+
+        row1Item.row_order = row2Order;
+        row2Item.row_order = row1Order;
+
+        const newRowOrder = [...listOfIngridients];
```
rickaw, line 142: shall we rename this to `newListOfIngredients`?

rickaw (approved):
Looks great! A few minor comments but feel free to merge once they're addressed

PR #142, opened 2024-03-11
VFB-39-Specify visible time period for calendar

mtharkins:
## What's changed
Start and end time has been added to the weekly & daily calendar views 

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image 
image
  | image 
image
|

## Checklist

### DIFF
--- a/src/components/Calendar/Calendar.tsx
+++ b/src/components/Calendar/Calendar.tsx
@@ -210,6 +210,9 @@ const Calendar: React.FC<CalendarProps> = ({
 }) => {
     const [eventClick, setEventClick] = useState<CalendarEvent | null>(null);
     const calendarRef = useRef<FullCalendar>(null);
+    const [calendarStartTime] = useState("10:00:00");
+    const [calendarEndTime] = useState("15:00:00");
+    const [locations, setLocations] = useState<string[]>([]);
 
     const handleEventClick = (info: EventClickArg): void => {
         const id = info.event.id;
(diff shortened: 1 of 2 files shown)


src/components/Calendar/Calendar.tsx
```diff
@@ -210,6 +210,9 @@ const Calendar: React.FC<CalendarProps> = ({
 }) => {
     const [eventClick, setEventClick] = useState<CalendarEvent | null>(null);
     const calendarRef = useRef<FullCalendar>(null);
+    const [calendarStartTime] = useState("10:00:00");
+    const [calendarEndTime] = useState("15:00:00");
```
rickaw, line 214: If we're not changing these values then no need to have states for them. We can instead do
```
const calendarStartTime = "10:00:00";
```

src/components/Calendar/Calendar.tsx
```diff
@@ -270,6 +273,8 @@ const Calendar: React.FC<CalendarProps> = ({
                                 meridiem: "short",
                             },
                             titleFormat: { year: "numeric", month: "short", day: "numeric" },
+                            calendarStartTime,
+                            calendarEndTime,
```
rickaw, line 277: are there any docs for these props? I'm struggling to find any

src/components/Calendar/Calendar.tsx
mtharkins: Updated

src/components/Calendar/Calendar.tsx
mtharkins: slotMinTime/slotMaxTime are now being used instead to set the calendar hours.

rickaw (approved):
great thanks!

PR #144, opened 2024-03-12
VFB-52: Add generate maps action in parcels page

minalui:
## What's changed
- Add generate map under actions on parcels page
- Opens google maps in new tab with the selected parcels postcode in a circular pin
- Shows error message if no parcels are selected

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

### DIFF
--- a/src/app/parcels/ActionBar/Actions.tsx
+++ b/src/app/parcels/ActionBar/Actions.tsx
@@ -38,7 +38,8 @@ type ActionName =
     | "Download Shopping Lists"
     | "Download Driver Overview"
     | "Download Day Overview"
-    | "Delete Parcel Request";
+    | "Delete Parcel Request"
+    | "Generate Map";
 
 type AvailableActionsType = {
     [actionKey in ActionName]: {
@@ -81,14 +82,14 @@ const availableActions: AvailableActionsType = {
         errorMessage: "Please select at least one parcel.",
         actionType: "deleteParcel",
     },
-};
-
-const generateMapAction = {
     "Generate Map": {
+        showSelectedParcelsInModal: false,
         errorCondition: isNotAtLeastOne,
         errorMessage: "Please select at least one parcel.",
+        actionType: "generateMap",
     },
 };
+
 interface ActionsInputComponentProps {
     actionName: ActionName;
     selectedParcels: ParcelsTableRow[];
@@ -319,27 +320,33 @@ const Actions: React.FC<Props> = ({
                 >
                     <MenuList id="action-menu">
                         {Object.entries(availableActions).map(([key, value]) => {
-                            return (
-                                <MenuItem
-                                    key={key}
-                                    onClick={onMenuItemClick(
-                                        key as ActionName,
-                                        value.errorCondition,
-                                        value.errorMessage
-                                    )}
-                                >
-                                    {key}
-                                </MenuItem>
-                            );
+                            if (key !== "Generate Map") {
+                                return (
+                                    <MenuItem
+                                        key={key}
+                                        onClick={onMenuItemClick(
+                                            key as ActionName,
+                                            value.errorCondition,
+                                            value.errorMessage
+                                        )}
+                                    >
+                                        {key}
+                                    </MenuItem>
+                                );
+                            } else {
+                                return (
+                                    <MenuItem
+                                        key={key}
+                                        onClick={onMapsClick(
+                                            value.errorCondition,
+                                            value.errorMessage
+                                        )}
+                                    >
+                                        {key}
+                                    </MenuItem>
+                                );
+                            }
                         })}
-                        <MenuItem
-                            onClick={onMapsClick(
-                                generateMapAction["Generate Map"].errorCondition,
-                                generateMapAction["Generate Map"].errorMessage
-                            )}
-                        >
-                            Generate Map
-                        </MenuItem>
                     </MenuList>
                 </Menu>
             )}
--- a/src/app/parcels/ActionBar/Actions.tsx
+++ b/src/app/parcels/ActionBar/Actions.tsx
@@ -254,6 +254,7 @@ const Actions: React.FC<Props> = ({
         };
     };
 
+
     const mapsLinkForSelectedParcels = (): string => {
         return (
             "https://www.google.com/maps/dir/" +
--- a/src/app/parcels/ActionBar/Actions.tsx
+++ b/src/app/parcels/ActionBar/Actions.tsx
@@ -254,7 +254,6 @@ const Actions: React.FC<Props> = ({
         };
     };
 
-
     const mapsLinkForSelectedParcels = (): string => {
         return (
             "https://www.google.com/maps/dir/" +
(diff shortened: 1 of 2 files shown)


src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -234,6 +240,38 @@ const Actions: React.FC<Props> = ({
...
     };
 
+    const mapsLinkForParcels = (): string => {
+        const mapsLinkList = ["https://www.google.com/maps/dir/"];
+        selectedParcels.map((parcel) => {
+            mapsLinkList.push(parcel.addressPostcode + "/");
+        });
+        mapsLinkList.push("/");
+        const mapsLinkString = mapsLinkList.join("");
+        return mapsLinkString;
```
AmeliaIsCoding, line 250: I think there is no need for the mapsLinkList, we could have this as a string and use string.concat rather than list.push, perhaps? Then there is no need to join the list elements :)

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -234,6 +240,38 @@ const Actions: React.FC<Props> = ({
...
+    };
+
+    const openInNewTab = (url: string): void => {
+        const newWindow = window.open(url, "_blank", "noopener, noreferrer");
+        if (newWindow) {
+            newWindow.opener = null;
+        }
```
AmeliaIsCoding, line 257: I'm not sure I understand this if statement, could you please explain? :)

AmeliaIsCoding (commented):
Looks good! I like that you treated the maps action separately to the other actions

src/app/parcels/ActionBar/Actions.tsx
minalui: newWindow is opening a page with the specified url in a new tab(_blank).
noopener:  to prevent an external link from taking control of your user’s originating browser window (this is handled by the major browsers but not android web view - so unsure if we need this anymore)
noreferrer:  keeps external sites from knowing that you have included links to their material on your website
newWindow.opener = null: To set the opener to null again (again to stop the passage into our site - but I'm not sure if its outdated.

src/app/parcels/ActionBar/Actions.tsx
minalui: Yep good plan - I think I'm just more comfortable with array functions than string functions

src/app/parcels/ActionBar/Actions.tsx
AmeliaIsCoding: Ok cool, if you think its needed, then looks good :)

src/app/parcels/ActionBar/Actions.tsx
AmeliaIsCoding: looks good!

AmeliaIsCoding (approved):
looks good!

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -83,6 +83,12 @@ const availableActions: AvailableActionsType = {
     },
 };
 
+const generateMapAction = {
```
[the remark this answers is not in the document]
minalui: > is there a reason why this couldn't be configured like the other actions above?

The other actions open a modal but I think I will try and see if I can make the a conditional statement for the MenuItem so that on click it does a different action.

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -282,18 +320,32 @@ const Actions: React.FC<Props> = ({
...
-                                    onClick={onMenuItemClick(
-                                        key as ActionName,
-                                        value.errorCondition,
-                                        value.errorMessage
-                                    )}
-                                >
-                                    {key}
-                                </MenuItem>
-                            );
+                            if (key !== "Generate Map") {
```
rickaw, line 323: Could we handle this if statement withiin onMenuItemClick for clarity / consistency please? And I think AvailableActionsType could have another field like `shouldOpenModal` or something?

Now that opening the modal isn't the only type of actions, many of the existing variable names should be reviewed to avoid confusion. e.g. `selectedAction` should really be `modalTypeToDisplay` cause that's what decides whether to show modal / which modal to show

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -234,6 +241,37 @@ const Actions: React.FC<Props> = ({
         };
     };
 
+    const mapsLinkForParcels = (): string => {
+        let mapsLinkString = "https://www.google.com/maps/dir";
+        selectedParcels.forEach((parcel) => {
+            mapsLinkString = mapsLinkString.concat("/", parcel.addressPostcode);
+        });
+        mapsLinkString = mapsLinkString.concat("//");
+        return mapsLinkString;
```
rickaw, line 250: Instead of concatnating strings, we can simply that using `map` and `join`.
```suggestion
        return "https://www.google.com/maps/dir/" + selectedParcels.map(parcel => parcel.addressPostcode).join("/") + "//"
```

- Also we should deal with the cases where `parcel.addressPostcode` has a whitespace(s)
- Would this function be better named as `mapLinkForSelectedParcels`? `mapLinkForParcels` kind of implies we need to pass a list of parcels as a parameter, or we have a defined set of parcels that doesn't change.

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -234,6 +241,37 @@ const Actions: React.FC<Props> = ({
...
+    };
+
+    const openInNewTab = (url: string): void => {
+        const newWindow = window.open(url, "_blank", "noopener, noreferrer");
+        if (newWindow) {
+            newWindow.opener = null;
```
rickaw, line 256: What does this do? Looks like the line above guarantees the opener is set to null anyway? (though i'm not sure how this window.open business works)

rickaw (approved):
Nice one! One minor comment but feel free to merge once that's addressed

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -285,11 +326,15 @@ const Actions: React.FC<Props> = ({
...
-                                        value.errorMessage
-                                    )}
+                                    onClick={
+                                        value.shouldOpenModal
+                                            ? onMenuItemClick(
+                                                  key as ActionName,
+                                                  value.errorCondition,
+                                                  value.errorMessage
+                                              ) // eslint-disable-line indent
+                                            : onMapsClick(value.errorCondition, value.errorMessage)
```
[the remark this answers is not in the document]
minalui: @rickaw Can you please check the new commit before I merge? Thanks!

rickaw (approved):
LGTM  - feel free to merge

src/app/parcels/ActionBar/Actions.tsx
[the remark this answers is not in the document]
rickaw: looks sensible!

PR #145, opened 2024-03-12
VFB-114: Table row styling

minalui:
## What's changed
- Add highlight on hover
- Add pointer on hover for parcels, clients, client detail modal tables
- Parcels table checkbox cell on click does not trigger opening of modal

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

### DIFF
--- a/src/app/parcels/ParcelsPage.tsx
+++ b/src/app/parcels/ParcelsPage.tsx
@@ -324,14 +324,11 @@ const ParcelsPage: React.FC<{}> = () => {
 
     const onParcelTableRowClick = (row: Row<ParcelsTableRow>, event: React.MouseEvent): void => {
         const target = event.target as HTMLDivElement;
-        if (target.children.length === 0) {
-            setSelectedParcelId(row.data.parcelId);
-            router.push(`/parcels?${parcelIdParam}=${row.data.parcelId}`);
-        } else if (target.children[0].tagName !== "INPUT") {
-            setSelectedParcelId(row.data.parcelId);
-            router.push(`/parcels?${parcelIdParam}=${row.data.parcelId}`);
-        } else {
+        if (target.children.length > 0 && target.children[0].tagName === "INPUT") {
             return;
+        } else {
+            setSelectedParcelId(row.data.parcelId);
+            router.push(`/parcels?${parcelIdParam}=${row.data.parcelId}`);
         }
     };
 
--- a/src/app/parcels/ParcelsPage.tsx
+++ b/src/app/parcels/ParcelsPage.tsx
@@ -323,7 +323,7 @@ const ParcelsPage: React.FC<{}> = () => {
     };
 
     const onParcelTableRowClick = (row: Row<ParcelsTableRow>, event: React.MouseEvent): void => {
-        const target = event.target as HTMLDivElement;
+        const target = event.target as HTMLElement;
         if (target.children.length > 0 && target.children[0].tagName === "INPUT") {
             return;
         } else {
--- a/src/app/parcels/ParcelsPage.tsx
+++ b/src/app/parcels/ParcelsPage.tsx
@@ -322,7 +322,7 @@ const ParcelsPage: React.FC<{}> = () => {
         lastStatus: rowToLastStatusColumn,
     };
 
-    const onParcelTableRowClick = (row: Row<ParcelsTableRow>, event: MouseEvent): void => {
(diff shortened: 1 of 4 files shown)


src/app/parcels/ParcelsPage.tsx
```diff
@@ -322,9 +322,14 @@ const ParcelsPage: React.FC<{}> = () => {
         lastStatus: rowToLastStatusColumn,
     };
 
-    const onParcelTableRowClick = (row: Row<ParcelsTableRow>): void => {
-        setSelectedParcelId(row.data.parcelId);
-        router.push(`/parcels?${parcelIdParam}=${row.data.parcelId}`);
+    const onParcelTableRowClick = (row: Row<ParcelsTableRow>, event: React.MouseEvent): void => {
+        const target = event.target as HTMLDivElement;
```
rickaw, line 326: I think it's quite hard to ensure this is indeed HTMLDivElement (in fact i'm not sure now). Let's keep it general for now
```suggestion
        const target = event.target as HTMLElement;
```

src/app/parcels/ParcelsPage.tsx
```diff
@@ -322,9 +322,14 @@ const ParcelsPage: React.FC<{}> = () => {
...
+        if (target.children.length > 0 && target.children[0].tagName === "INPUT") {
```
rickaw, line 327: What are we really checking here?

src/components/Tables/Table.tsx
```diff
@@ -368,17 +371,20 @@ const StyledIcon = styled(Icon)`
 
 const TableStyling = styled.div`
     // the component with the filter bars
+
```
rickaw, line 374: nitpick: Are these empty lines intentional?

rickaw (commented):
Great! Though I don't think the images on the PR are showing any diff? Am i being blind?

minalui:
> Great! Though I don't think the images on the PR are showing any diff? Am i being blind?

The first row is highlighted! Printscreen doesn't show the mouse though unfortunately.

src/app/parcels/ParcelsPage.tsx
minalui: We are checking that the table cell that was clicked has children and the first child is an input. The row has two cells that have children elements (cell 1 and cell 5), the others do not have children.

image
* I really want to check if data-column-id is equal to 1 but not sure how to achieve this in an elegant way. Any suggestions?

minalui: Maybe checking id begins with "cell-1"?

src/app/parcels/ParcelsPage.tsx
```diff
@@ -322,9 +322,14 @@ const ParcelsPage: React.FC<{}> = () => {
         lastStatus: rowToLastStatusColumn,
     };
 
-    const onParcelTableRowClick = (row: Row<ParcelsTableRow>): void => {
-        setSelectedParcelId(row.data.parcelId);
-        router.push(`/parcels?${parcelIdParam}=${row.data.parcelId}`);
+    const onParcelTableRowClick = (row: Row<ParcelsTableRow>, event: React.MouseEvent): void => {
+        const target = event.target as HTMLElement;
+        if (target.id.startsWith("cell-1")) {
```
rickaw, line 327: I don't think this is specific enough to target the checkbox. It wouldn't be easy to spot this if we decide to show an icon as the first column for example.

Seems like there's a `ignoreRowClick` flag in the DataTable column. Could we use this, and define the checkbox column to ignore row clicks?
https://react-data-table-component.netlify.app/?path=/docs/api-columns--docs

src/components/Tables/Table.tsx
```diff
@@ -379,6 +382,7 @@ const TableStyling = styled.div`
         background-color: transparent;
 
         // the pagination bar
+
```
rickaw, line 385: there's an extra empty line here too :)

src/app/parcels/ParcelsPage.tsx
minalui: Think I need some help on this

minalui: I realised I can set ignoreRowClick on checkboxconfig

minalui:
I found that there is one column called iconsColumn (flagged for attention/requires followup call) that is not being displayed - just wondered if there was some other uses for it?

rickaw (approved):
Nice one thank you!

PR #163, opened 2024-03-19
VFB-29: Request pagination

AmeliaIsCoding:
## What's changed
Request pagination in clients and parcels tables.
The request to supabase only returns the data that you actually see at any given point, dependant on:
-current page in the table
-number of items per page in the table
-the combination of filters applied and the values of them
-the current sort field and direction, or lack thereof, including if this sort field requires any custom sorting rules 

Small side effect changes:
-Date range selector has moved into the top of the parcels table, alongside the other primary filters, instead of above the table
-A circular progress indicator is shown in clients and parcels pages before table loaded for the first time
-A circular progress indicator is also shown within each table whenever the data is loading.
-The filter options for "last status" and "delivery collection" are no longer dependent on the date range selected. Instead all options are possible at any point.

## Bugs
- Attempting to check lots of parcels (eg 600+) in parcels table throws a database error. Another ticket will deal with this.
- Currently cannot sort or filter by packing time slot in parcels table. Another ticket will deal with this, once unblocked.

## Screenshots / Videos
Not really relevant for this.

## Checklist

  - [x] I have modified the seed in `seed.mts` if appropriate


src/app/parcels/ParcelsPage.tsx
```diff
@@ -181,90 +358,221 @@ const ParcelsPage: React.FC<{}> = () => {
     const searchParams = useSearchParams();
     const parcelId = searchParams.get(parcelIdParam);
 
+    const [sortState, setSortState] = useState<SortState<ParcelsTableRow>>({ sort: false });
```
AmeliaIsCoding, line 361: The component react-data-table-component does not allow us to set a default sort state, even though we want to sort by packing date descending by default. Therefore, in the fetch request, if the sortState is set to {sort: false}, we employ the default sorting behaviour. (see fetchParcelTableData.ts). 
Because of this limitation, the table thinks it is unsorted initially. This means that there is the annoying side effect of, if we were to click the "Packing date" column header to sort by it, the table would think "Ah yes, I need to sort by packing date ascending" so this is what the sortState is updated to. As we were already sorting by this sort field and direction, thus there is no change in the sorting that occurs in fetchParcelTableData.ts (because this depends on sortState). Then when we click it again, the table will think "Ah yes, I need to sort by packing date descending", so the sortState updates to that, and the sort behaviour is correct from now on.

src/app/parcels/ParcelsPage.tsx
```diff
@@ -181,90 +358,221 @@ const ParcelsPage: React.FC<{}> = () => {
...
-            if (areDateRangesIdentical(dateRangeToFetch, packingDateRange)) {
-                setTableData(tableData);
-            }
+    const [primaryFilters, setPrimaryFilters] = useState<Filter<ParcelsTableRow, any>[]>([]);
+    const [additionalFilters, setAdditionalFilters] = useState<Filter<ParcelsTableRow, any>[]>([]);
 
-            setIsLoading(false);
-        })();
-    }, [packingDateRange]);
+    const [areFiltersLoadingForFirstTime, setAreFiltersLoadingForFirstTime] =
```
AmeliaIsCoding, line 378: we need this because the filters are now built asynchronously, because the checkbox filters rely on requests to the database. We don't want to build the table until the filters are ready.

src/app/parcels/ParcelsPage.tsx
```diff
@@ -181,90 +358,221 @@ const ParcelsPage: React.FC<{}> = () => {
...
-                setTableData(await fetchAndFormatParcelTablesData(packingDateRange));
-                setIsLoading(false);
-            }, 500);
+        const buildFilters = async (): Promise<{
+            primaryFilters: Filter<ParcelsTableRow, any>[];
+            additionalFilters: Filter<ParcelsTableRow, any>[];
+        }> => {
+            const startOfToday = dayjs().startOf("day");
+            const endOfToday = dayjs().endOf("day");
+            const dateFilter = buildDateFilter({
```
AmeliaIsCoding, line 388: note that dateFilter is now treated as a regular ol' filter

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -30,52 +35,158 @@ export const getCongestionChargeDetailsForParcels = async (
...
+        if (filter.methodConfig.methodType === PaginationType.Server) {
+            query = filter.methodConfig.method(query, filter.state);
+        }
+    });
+
+    if (sortState.sort && sortState.column.sortMethodConfig?.methodType === PaginationType.Server) {
+        query = sortState.column.sortMethodConfig.method(query, sortState.sortDirection);
+    } else {
+        query = query.order("packing_datetime", { ascending: false });
+    }
```
AmeliaIsCoding, line 58: this is the default sorting behaviour for the parcels table that is shown on initial render, when sort state is set to {sort: false}. see my review comments on ParcelsPage.tsx

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -30,52 +35,158 @@ export const getCongestionChargeDetailsForParcels = async (
...
+    const { data, error } = await query;
+    if (error) {
+        console.error(error);
+        throw new DatabaseError("fetch", "parcels");
+    }
+    return data.map((parcel) => parcel.parcel_id ?? "") ?? [];
```
AmeliaIsCoding, line 140: this function returns parcel ids for all the parcels, after filters are applied, ie, no pagination. This is used when we toggle the all checkbox to true, in order to populate checkedParcelIds with all the parcel ids, after filters are applied.

src/app/parcels/getParcelsTableData.ts
```diff
@@ -43,30 +44,29 @@ export const processingDataToParcelsTableData = (
...
             packingTimeLabel: datetimeToPackingTimeLabel(parcel.packing_datetime),
-            lastStatus: eventToLastStatus(parcel.events[0] ?? null),
+            lastStatus: processLastStatus(parcel),
             voucherNumber: parcel.voucher_number,
             packingDatetime: parcel.packing_datetime ? new Date(parcel.packing_datetime) : null,
             iconsColumn: {
-                flaggedForAttention: client.flagged_for_attention,
-                requiresFollowUpPhoneCall: client.signposting_call_required,
+                flaggedForAttention: parcel.client_flagged_for_attention ?? false,
+                requiresFollowUpPhoneCall: parcel.client_signposting_call_required ?? false,
```
AmeliaIsCoding, line 69: the changes in this file are mostly due to an annoying supabase limitation where it thinks any field in a view can be nullable. So when converting the response to be of type ParcelsTableRow[], we have to pass in a default value for every non-optional attribute of ParcelsTableRow.

src/components/Tables/Filters.tsx
```diff
@@ -1,36 +1,35 @@
...
-    shouldFilter: (data: Data, state: State) => boolean;
+    key: keyof Data;
     filterComponent: (state: State, setState: (state: State) => void) => React.ReactElement;
     state: State;
     initialState: State;
+    methodConfig: MethodConfig<Data, State>;
+    areStatesIdentical: (stateA: State, stateB: State) => boolean;
 }
 
-export interface KeyedFilter<Data, Key extends keyof Data, State> extends Filter<Data, State> {
```
AmeliaIsCoding, line 11: we never really needed the distinction between a Filter and a KeyedFilter, we should just assign a key to every filter, because filters are components that end up in an array, so we need to give them keys.

src/components/Tables/Filters.tsx
```diff
@@ -1,36 +1,35 @@
...
+    | {
+          method: (
+              query: PostgrestFilterBuilder<Database["public"], any, any>,
+              state: State
+          ) => PostgrestFilterBuilder<Database["public"], any, any>;
+          methodType: PaginationType.Server;
+      }
+    | {
+          method: (row: Data, state: State, key: keyof Data) => boolean;
+          methodType: PaginationType.Client;
```
AmeliaIsCoding, line 21: some tables require client side filtering (admin tables, and the component tests).

src/components/Tables/Table.tsx
```diff
@@ -342,12 +420,33 @@ const Table = <Data,>({
...
+                                : () => {}
+                        }
+                        sortServer={sortConfig.sortPossible}
+                        onSort={handleSort}
+                        progressComponent={
+                            <Centerer>
+                                <CircularProgress aria-label="table-progress-bar" />
+                            </Centerer>
+                        }
+                        progressPending={isLoading}
```
AmeliaIsCoding, line 449: the progress component shows within the table in place of the data whenever the data is loading. This is different from rendering a progress component instead of the table in the parent page whenever the data is loading, because this way the table states are maintained.

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -30,52 +35,158 @@ export const getCongestionChargeDetailsForParcels = async (
...
+    }
+
+    if (parcelIds) {
+        query = query.in("parcel_id", parcelIds);
+    }
+
+    const { data, error } = await query;
 
     if (error) {
+        console.error(error);
```
rickaw, line 71: can we replace this with logError?

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -30,52 +35,158 @@ export const getCongestionChargeDetailsForParcels = async (
...
-            timestamp
-        )
-    `
-        )
-        .gte("packing_datetime", dateRange.from)
-        .lte("packing_datetime", dateRange.to)
-        .order("packing_datetime", { ascending: false })
-        .order("timestamp", { ascending: false, foreignTable: "events" })
-        .limit(1, { foreignTable: "events" });
+export const getParcelProcessingData = async (
```
rickaw, line 38:
```suggestion
const getParcelProcessingData = async (
```

src/app/parcels/fetchParcelTableData.ts
rickaw, line 140: nitpick: can we add some empty lines to separate the logic here please?

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -30,52 +35,158 @@ export const getCongestionChargeDetailsForParcels = async (
...
+    ) => PostgrestFilterBuilder<Database["public"], any, any>
+): Promise<Return> => {
+    const query = selectMethod(supabase.from(table));
+    const { data, error } = await query;
+    if (error) {
+        console.error(error);
+        throw new DatabaseError("fetch", table);
+    }
     return data ?? [];
 };
```
rickaw, line 157: I'm not sure I understand what this function is trying to achieve - the caller needs to define `selectMethod` anyway so is it easier to read if you just did this without a function? Maybe it's a naming problem? `getAllValuesForKeys` feels like it's related to key-value pairs but apparently this isn't doing that?

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -30,52 +35,158 @@ export const getCongestionChargeDetailsForParcels = async (
...
+    sortState: SortState<Data>;
+    startPoint: number;
+    endPoint: number;
+}
+
+export const isFreshRequest = <Data>(
+    requestParams: RequestParams<Data>,
+    initialRequestParams: RequestParams<Data>
+): boolean => {
+    const filtersSame = Array.from(requestParams.allFilters).every((filter, index) =>
```
rickaw, line 178: is this equivalent?
```suggestion
    const filtersSame = requestParams.allFilters.every((filter, index) =>
```

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -30,52 +35,158 @@ export const getCongestionChargeDetailsForParcels = async (
...
+}
+
+export interface RequestParams<Data> {
+    allFilters: Filter<Data, any>[];
+    sortState: SortState<Data>;
+    startPoint: number;
+    endPoint: number;
+}
+
+export const isFreshRequest = <Data>(
```
rickaw, line 174: is this function returning whether the two given requests are identical? What do you mean by isFreshRequest?

src/app/parcels/ParcelsPage.tsx
```diff
@@ -132,45 +183,171 @@ const PreTableControls = styled.div`
...
-    const endOfToday = dayjs().endOf("day");
-
     const [isLoading, setIsLoading] = useState(true);
-    const [packingDateRange, setPackingDateRange] = useState<DateRangeState>({
-        from: startOfToday,
-        to: endOfToday,
-    });
-    const [tableData, setTableData] = useState<ParcelsTableRow[]>([]);
+    const [parcelsDataPortion, setParcelsDataPortion] = useState<ParcelsTableRow[]>([]);
+    const [totalRows, setTotalRows] = useState<number>(0);
```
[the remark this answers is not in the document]
rickaw: maybe we should avoid using the word `rows` in that case
```suggestion
    const [filteredParcelCount, setFilteredParcelCount] = useState<number>(0);
```

src/app/parcels/ParcelsPage.tsx
rickaw: think that's good enough for now, thanks!

rickaw (commented):
Amazing! Lots of comments for a big PR but most of them are very minor.

src/app/parcels/ActionBar/Statuses.tsx
```diff
@@ -10,7 +10,7 @@ import { ParcelsTableRow } from "@/app/parcels/getParcelsTableData";
 import StatusesBarModal from "@/app/parcels/ActionBar/StatusesModal";
 import { DatabaseError } from "@/app/errorClasses";
 
-export const statusNamesInWorkflowOrder = [
```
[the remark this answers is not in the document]
AmeliaIsCoding: Cool agreed should I make a ticket for this?

src/app/parcels/ParcelsPage.tsx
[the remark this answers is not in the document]
AmeliaIsCoding: Sounds good. Have updated in the parcels page, clients page, and the table component prop name too

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -36,10 +44,13 @@ const collectionCentresTableHeaderKeysAndLabels: TableHeaders<Schema["collection
 ];
 
 interface Props {
-    collectionCentreData: Schema["collection_centres"][];
+    collectionCentreData: CollectionCentresTableRow[];
 }
 
 const CollectionCentresTables: React.FC<Props> = (props) => {
+    const [collectionCentreDataPortion, setCollectionCentreDataPortion] = useState<
```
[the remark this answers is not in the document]
AmeliaIsCoding: Have changed, also changed usersDataPortion to users etc

seed.mts
```diff
@@ -232,6 +232,28 @@ async function generateSeed(): Promise<void> {
...
+            "Collection Failed",
+            "Parcel Collected",
+            "Shipping Labels Downloaded",
+            "Out for Delivery",
+            "Delivered",
+            "Delivery Failed",
+            "Delivery Cancelled",
+            "Fulfilled with Trussell Trust",
+            "Request Deleted",
+        ].map((event_name, index) => ({ eventName: event_name, workflowOrder: index }))
```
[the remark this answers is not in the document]
AmeliaIsCoding: We could, but I think it would make more sense to keep the list here and get the statuses from the backend in Statuses.tsx
Could change it for now though if you want?

src/app/clients/getClientsData.ts
```diff
@@ -2,38 +2,68 @@ import { ClientsTableRow } from "@/app/clients/ClientsPage";
...
 import { DatabaseError } from "@/app/errorClasses";
 import { Supabase } from "@/supabaseUtils";
+import { Filter, PaginationType } from "@/components/Tables/Filters";
+import { SortState } from "@/components/Tables/Table";
 
-const getClientsData = async (supabase: Supabase): Promise<ClientsTableRow[]> => {
+const getClientsData = async (
+    supabase: Supabase,
+    start: number,
+    end: number,
```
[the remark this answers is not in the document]
AmeliaIsCoding: have done for get parcels data functions too

src/app/clients/getClientsData.ts
```diff
@@ -2,38 +2,68 @@ import { ClientsTableRow } from "@/app/clients/ClientsPage";
...
-        const { count, error: familyError } = await supabase
-            .from("families")
-            .select("*", { count: "exact", head: true })
-            .eq("family_id", client.family_id);
-
-        if (familyError || count === null) {
-            throw new DatabaseError("fetch", "families");
-        }
-
         data.push({
```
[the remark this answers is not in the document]
AmeliaIsCoding: sure sounds good, can you check if what I did is right?

src/app/parcels/ActionBar/Statuses.tsx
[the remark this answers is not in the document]
rickaw: yes please!

seed.mts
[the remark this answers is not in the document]
rickaw: yeah sounds good to me - no actions needed in this file then :)

src/app/lists/ListsPage.tsx
```diff
@@ -18,14 +18,35 @@ const fetchData = async (): Promise<FetchedListsData> => {
...
-            setListsData(fetchedData.data);
+            setListsData(
+                fetchedData.data.map((row) => {
+                    const newRow = {
+                        primaryKey: row.primary_key,
+                        rowOrder: row.row_order,
+                        itemName: row.item_name,
+                        ...Object.fromEntries(
+                            listsHeaderKeysAndLabels
+                                .filter(([key]) => /^\d+$/.test(key))
```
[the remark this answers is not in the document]
AmeliaIsCoding: This is code I moved over from ListsDataView. Do you want me to change this?

src/app/lists/ListDataview.tsx
```diff
@@ -56,13 +58,39 @@ export const headerKeysAndLabels = [
     ["10", "Family of 10"],
 ] satisfies [keyof ListRow, string][];
 
+export const listRowToListDB = (listRow: Partial<ListRow>): Partial<Schema["lists"]> => ({
```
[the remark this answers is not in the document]
AmeliaIsCoding: Great question... I can't remember why I put that there, it works fine without so have removed

src/app/lists/ListDataview.tsx
```diff
@@ -186,11 +194,36 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
...
+    const [listData, setListData] = useState<ListRow[]>(listOfIngridients);
+
+    const filters: Filter<ListRow, string>[] = [
+        buildTextFilter({
+            key: "itemName",
+            label: "Item",
+            headers: listsHeaderKeysAndLabels,
+            methodConfig: { methodType: PaginationType.Client, method: filterRowByText },
+        }),
+    ];
```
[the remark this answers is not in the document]
AmeliaIsCoding: yup

src/app/parcels/ActionBar/ActionBar.tsx
```diff
@@ -11,10 +11,11 @@ import { ControlContainer } from "@/components/Form/formStyling";
 import { ArrowDropDown } from "@mui/icons-material";
 
 export interface ActionBarProps {
-    selectedParcels: ParcelsTableRow[];
+    fetchSelectedParcels: (checkedParcelIds: string[]) => Promise<ParcelsTableRow[]>;
```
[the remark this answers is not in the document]
AmeliaIsCoding: sure, maybe fetchParcelsByIds would be better? lmk

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -179,22 +179,32 @@ const ActionsButton: React.FC<ActionsButtonProps> = ({
...
     setActionAnchorElement,
     modalError,
     setModalError,
+    parcelIds,
 }) => {
+    const [selectedParcels, setSelectedParcels] = useState<ParcelsTableRow[]>([]);
+
+    useEffect(() => {
+        (async () => {
+            setSelectedParcels(await fetchSelectedParcels(parcelIds));
```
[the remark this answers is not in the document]
AmeliaIsCoding: Okay I have added an error message that is shown to the user in this case. Like how "please select at least 1 parcel" is shown to the user. I've changed the code so that the fetchSelectedParcels call is made only when a status/action is selected from their dropdown. This is what I was aiming to do, but it wasn't working - the call was being made whenever parcels were selected/unselected -  now its working as I wanted.

src/app/admin/usersTable/UsersTable.tsx
```diff
@@ -38,6 +40,7 @@ interface Props {
 }
 
 const UsersTable: React.FC<Props> = (props) => {
+    const [userDataPortion, setUserDataPortion] = useState<UserRow[]>(props.userData);
```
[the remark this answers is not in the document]
AmeliaIsCoding: is users, setUsers ok instead? more consistent with the other unpaginated tables

src/app/lists/EditModal.tsx
```diff
@@ -17,7 +17,7 @@ interface Props {
 }
 
 // null => add, undefined => modal closed
-export type EditModalState = Schema["lists"] | null | undefined;
+export type EditModalState = Partial<Schema["lists"]> | null | undefined;
```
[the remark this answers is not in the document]
AmeliaIsCoding: Ok so I think I've figured out that I changed this to Partial to be consistent with the state "toSubmit". But there's no actual need for consistency and would probs be better to not have it partial. Have removed the word partial

src/components/Tables/DateFilter.tsx
```diff
@@ -0,0 +1,46 @@
+"use client";
```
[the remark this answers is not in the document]
AmeliaIsCoding: nope

src/components/Tables/DateFilter.tsx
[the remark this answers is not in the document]
AmeliaIsCoding: I think it needs to be .tsx  -  on line 41 we use html / React

src/app/parcels/fetchParcelTableData.ts
AmeliaIsCoding: image

src/app/parcels/fetchParcelTableData.ts
rickaw: ah weird ok

src/app/parcels/fetchParcelTableData.ts
AmeliaIsCoding: good point, I guess this function doesn't care about the concept of fresh requests or whatever, it just takes in 2 request param objects and compares them, so I will rename it to areRequestsIdentical and rename the objects to requestParamsA and requestParamsB

src/app/parcels/ActionBar/ActionBar.tsx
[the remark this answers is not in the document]
rickaw: fetchParcelsByIds sounds good!

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -179,22 +179,32 @@ const ActionsButton: React.FC<ActionsButtonProps> = ({
 };
 
 interface Props {
-    selectedParcels: ParcelsTableRow[];
+    fetchSelectedParcels: (checkedParcelIds: string[]) => Promise<ParcelsTableRow[]>;
```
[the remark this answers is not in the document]
AmeliaIsCoding: as above, can we use "fetchParcelsByIds"

src/app/parcels/fetchParcelTableData.ts
rickaw: Should we additionally filter the list so it doens't contain empty strings
```suggestion
    return data.map((parcel) => parcel.parcel_id ?? "").filter(parcelId => parcelId) ?? [];
```
or even
```suggestion
    return data.map((parcel) => parcel.parcel_id).filter(parcelId => parcelId) ?? [];
```

AmeliaIsCoding: what about
`return data.reduce((reducedData, parcel) => {parcel.parcel_id && reducedData.push(parcel.parcel_id);
 return reducedData})`

src/app/clients/getClientsData.ts
```diff
@@ -2,38 +2,68 @@ import { ClientsTableRow } from "@/app/clients/ClientsPage";
...
+    let query = supabase.from("clients_plus").select("*", { count: "exact", head: true });
+
+    filters.forEach((filter) => {
+        if (filter.methodConfig.methodType === PaginationType.Server) {
+            query = filter.methodConfig.method(query, filter.state);
+        }
+    });
+    const { count, error: clientError } = await query;
+    if (clientError || count === null) {
+        throw new DatabaseError("fetch", "clients");
```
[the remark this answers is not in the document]
rickaw: It's fine - it just needs to be unique enough that when we find a particular log message on CloudWatch we can find where in the code it came from

src/app/lists/ListsPage.tsx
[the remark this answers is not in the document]
rickaw: hm okay maybe not worth changing it in this PR then 😄

src/components/Tables/DateFilter.tsx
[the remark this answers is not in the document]
rickaw: ok!

src/app/parcels/fetchParcelTableData.ts
rickaw: I think there's definitely a argument for moving the Supabase stuff to one sensible place - I've been thinking about this recently. We should probably have a systematic approach though if we do this - for this PR I feel like they're better defined directly in ParcelsPage.

src/app/parcels/ActionBar/Statuses.tsx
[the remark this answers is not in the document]
AmeliaIsCoding: VFB-158 in backlog

src/app/lists/ListsPage.tsx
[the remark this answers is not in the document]
AmeliaIsCoding: agreed

src/app/parcels/fetchParcelTableData.ts
AmeliaIsCoding: Cool, think I agree

AmeliaIsCoding: have done

src/app/parcels/fetchParcelTableData.ts
rickaw: `return data.reduce((reducedData, parcel) => {parcel.parcel_id && reducedData.push(parcel.parcel_id); ` looks good!

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -66,16 +77,57 @@ const CollectionCentresTables: React.FC<Props> = (props) => {
...
+        }),
+        buildTextFilter({
+            key: "acronym",
+            label: "Acronym",
+            headers: collectionCentresTableHeaderKeysAndLabels,
+            methodConfig: { paginationType: PaginationType.Client, method: filterRowByText },
+        }),
+    ];
+    const [primaryFilters, setPrimaryFilters] =
+        useState<Filter<CollectionCentresTableRow, string>[]>(filters);
```
rickaw, line 95: Can we 
- move `filters` outside the component
- move the `primaryFilters` state to be next to all other states at the top
?

src/app/clients/getClientsData.ts
```diff
@@ -2,38 +2,75 @@ import { ClientsTableRow } from "@/app/clients/ClientsPage";
...
-            .from("families")
-            .select("*", { count: "exact", head: true })
-            .eq("family_id", client.family_id);
-
-        if (familyError || count === null) {
-            throw new DatabaseError("fetch", "families");
-        }
-
+        client.client_id ?? void logError("empty client id");
+        client.full_name ?? void logError("empty client name");
```
rickaw, line 45: I think this would be more conventional but not fussed at all if this works
```
if (!client.client_id) void logError("..,")
```

src/app/lists/ListDataview.tsx
```diff
@@ -56,13 +58,39 @@ export const headerKeysAndLabels = [
     ["10", "Family of 10"],
 ] satisfies [keyof ListRow, string][];
 
+export const listRowToListDB = (listRow: ListRow): Schema["lists"] => ({
+    item_name: listRow.itemName,
+    notes_for_1: listRow[1]?.notes,
```
rickaw, line 63: i don't think you need the ? anymore
```suggestion
    notes_for_1: listRow[1].notes,
```

src/components/Tables/DateFilter.tsx
```diff
@@ -0,0 +1,44 @@
...
+    ): boolean => {
+        return (
+            areDaysIdentical(dateRangeA.from, dateRangeB.from) &&
+            areDaysIdentical(dateRangeA.to, dateRangeB.to)
+        );
+    };
+
+    const areDaysIdentical = (dayA: dayjs.Dayjs | null, dayB: dayjs.Dayjs | null): boolean => {
+        return dayA && dayB ? dayA.isSame(dayB) : dayA === dayB;
+    };
```
rickaw, line 30: I think we might as well move these inner methods out fo `dateFilter`?

rickaw (commented):
some tiny tiny comments!

rickaw (approved):

PR #170, opened 2024-03-21
Vfb 57 add error logging

minalui:
## What's changed
Add error logging to aws cloudwatch

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| paste_here | paste_here |

## Checklist
- [ ] I have documented the testing steps for QA - difficult to achieve errors
- [ ] Make sure you've tested via `npm run test` - component tests failing

If you have made any changes to the database... No changes

### DIFF
--- a/src/app/lists/ListDataview.tsx
+++ b/src/app/lists/ListDataview.tsx
@@ -14,7 +14,6 @@ import Button from "@mui/material/Button";
 import TooltipCell from "@/app/lists/TooltipCell";
 import TableSurface from "@/components/Tables/TableSurface";
 import CommentBox from "@/app/lists/CommentBox";
-import { v4 as uuidv4 } from "uuid";
 import { logError, logInfo } from "@/logger/logger";
 
 interface ListRow {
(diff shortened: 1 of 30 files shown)


package.json
```diff
@@ -66,6 +66,7 @@
     "@types/react-fontawesome": "^1.6.5",
     "@types/react-modal": "^3.16.0",
     "@types/styled-components": "^5.1.26",
+    "@types/uuid": "^9.0.8",
```
rickaw, line 69: can we install @types/uuid as a dev dependency please rather than a dependency? I know the interns did it this way but I don't think this is right. You'll probs have to uninstall @types/uuid and then do `npm install -D @types/uuid`

(unless this leads to next.js being unable to build, in which case find as a dependency)

src/app/lists/ListDataview.tsx
```diff
@@ -82,8 +83,8 @@ const listsColumnStyleOptions: ColumnStyles<ListRow> = {
 };
 
 const ListsDataView: React.FC<ListDataViewProps> = ({
-    listOfIngridients,
-    setListOfIngridients: setListOfIngridients,
+    listOfIngredients,
+    setListOfIngredients: setListOfIngredients,
```
rickaw, line 87:
```suggestion
    setListOfIngredients,
```

rickaw (approved):

PR #174, opened 2024-03-22
VFB-145 fix open client modal after editing client

AmeliaIsCoding:
## What's changed
Fix for VFB-145
Problem was:

open the modal
click Edit
change details (e.g. full name)
get taken back to the modal
click Edit again
then the old detail is still in the form.

This has been fixed.

## Checklist


src/app/clients/edit/[id]/page.tsx
```diff
@@ -1,23 +1,40 @@
...
+                setClientData(await fetchClient(params.id, supabase));
+            }
+        })();
+    }, [params.id]);
+    useEffect(() => {
+        (async () => {
+            if (clientData?.family_id) {
+                setFamilyData(await fetchFamily(clientData.family_id, supabase));
+            }
+        })();
```
Comment (rickaw) [inline src/app/clients/edit/[id]/page.tsx:33]: should we log an error when fetchClient or fetchFamily fails? In which case it might be easier to read with `.then` and `.catch` so we can get rid of the `(async () => { ... })()` structure

i.e.
```ts
if (params.id) {
  fetchClient(params.id, supabase)
    .then(client => setClientData(client))
    .catch(error => {
      logError(...)
      // maybe even set an error message to display on the page, rather than showing an empty page
    })
}
```

src/app/clients/edit/[id]/page.tsx
```diff
@@ -1,23 +1,40 @@
...
-const EditClients: ({ params }: EditClientsParameters) => Promise<React.ReactElement> = async ({
+const EditClients: ({ params }: EditClientsParameters) => React.ReactElement = ({
     params,
 }: EditClientsParameters) => {
-    const supabase = getSupabaseServerComponentClient();
-    const clientData = await fetchClient(params.id, supabase);
-    const familyData = await fetchFamily(clientData.family_id, supabase);
-    const initialFields = autofill(clientData, familyData);
+    const [clientData, setClientData] = useState<Schema["clients"]>();
+    const [familyData, setFamilyData] = useState<Schema["families"][]>();
```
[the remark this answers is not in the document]
Comment (AmeliaIsCoding) [reply src/app/clients/edit/[id]/page.tsx:19]: sure

src/app/clients/edit/[id]/page.tsx
Comment (AmeliaIsCoding) [reply src/app/clients/edit/[id]/page.tsx]: Cool, have done

src/app/clients/edit/[id]/page.tsx
```diff
@@ -1,23 +1,57 @@
...
+        if (params.id) {
+            fetchClient(params.id, supabase)
+                .then((client) => {
+                    setError(null);
+                    setClientData(client);
+                })
+                .catch((error) => {
+                    setError({ name: error, message: "Unable to fetch client data" });
+                    logErrorReturnLogId("error fetching client data in client edit modal");
+                });
```
Comment (rickaw) [inline src/app/clients/edit/[id]/page.tsx:36]: Should we set the error to null first when making a fetch? (In my experience better to do this because if the same request is made multiple times and they all failed, then clearing the error when making a new request makes it clear that the new request is indeed failing to the user. If we displayed the error once and never clear, then it's hard to see if the request is failing or we're not making any new attempts at all, if you know what i mean)

(same in the other use effect)
```suggestion
            setError(null);
            fetchClient(params.id, supabase)
                .then((client) => {
                    setClientData(client);
                })
                .catch((error) => {
                    setError({ name: error, message: "Unable to fetch client data" });
                    void logErrorReturnLogId("error fetching client data in client edit modal");
                });
```

src/app/clients/edit/[id]/page.tsx
```diff
@@ -1,23 +1,57 @@
...
+        }
+    }, [params.id]);
```
Comment (rickaw) [inline src/app/clients/edit/[id]/page.tsx:38]: nitpick: empty line for separation
```suggestion
    }, [params.id]);

```

rickaw (approved):
LGTM! left tiny comments but feel free to merge once you're happy with them

src/app/clients/edit/[id]/page.tsx
Comment (AmeliaIsCoding) [reply src/app/clients/edit/[id]/page.tsx:36]: Ok sure!

Issue #157, opened 2024-03-18
Driver Overview PDF - Editable Message by Admin


dan-langridge:
### Describe the solution / feature you'd like

On the Driver Overview PDF, the message "at the end of your shift" would be great to be editable by admin, to update phone numbers etc as staff change.

rickaw:
## Notes for developers
This would involve 
- adding an entry in the `website_data` table with key `driver_overview_message`
- adding a section on the Admins panel, displaying the editable `website_data` table
  - `item` - human-readable representation of the `key` e.g. "Driver Overview PDF Message"
  - `value` - the current value of the item

AmeliaIsCoding:
@rickaw do you want the website_data table in admins panel to contain just the Driver Overview PDF Message or all fields in the website_data table? (currently the only thing in there is the message at the top of lists page, which is also editable on the lists page itself)

rickaw:
> @rickaw do you want the website_data table in admins panel to contain just the Driver Overview PDF Message or all fields in the website_data table? (currently the only thing in there is the message at the top of lists page, which is also editable on the lists page itself)

@AmeliaIsCoding Good question, I think displaying all fields makes sense and improves visibility for admins :)

dan-langridge:
Just Driver Overview message please, thank youOn 20 Mar 2024, at 15:06, AmeliaIsCoding ***@***.***> wrote:﻿
@rickaw do you want the website_data table in admins panel to contain just the Driver Overview PDF Message or all fields in the website_data table? (currently the only thing in there is the message at the top of lists page, which is also editable on the lists page itself)

dan-langridge:
That’s fine with me too, as long as we keep the functionality for the message on Lists to be edited on the lists page (‘Manager’s’ are allowed to edit this message. Only ‘Admin’ allowed to edit Driver Overview message)On 20 Mar 2024, at 16:15, Ricky Kawagishi ***@***.***> wrote:﻿

@rickaw do you want the website_data table in admins panel to contain just the Driver Overview PDF Message or all fields in the website_data table? (currently the only thing in there is the message at the top of lists page, which is also editable on the lists page itself)

@AmeliaIsCoding Good question, I think displaying all fields makes sense and improves visibility for admins :)

rickaw:
@dan-langridge Sure, we'll keep the editing functionality on the lists page and display all editable data in the admins panel too! Thanks

AmeliaIsCoding:
[...]

PR #175, opened 2024-03-22
#157 Edit website data

AmeliaIsCoding:
## What's changed
Website data is now editable in the admin panel.
driver_overview_message has been added to website data, which corresponds to the message shown at the top of the driver overview PDF. 
lists_text is still part of website data; it is editable both in the admin panel, and in the lists page by any user, as before. 

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image|imageimage|

## Checklist
- [x] Make sure you've tested via `npm run test:e2e` - **accessibility tests failing**

  - [x] I have modified the seed in `seed.mts` if appropriate
  - [x] With my final set up, I can run `npm run dev:reset_supabase` without any errors. --> **Although sometimes the bucket isn't made properly? Think this is unrelated to this ticket**


src/app/admin/websiteDataTable/CustomEditComponent.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const CustomComponent: React.FC<CustomCellProps> = (props: CustomCellProps) => {
+    const { id, field, hasFocus } = props;
+    const apiRef = useGridApiContext();
+    const ref = useRef<HTMLTextAreaElement>(null);
+    const initialValue = useRef(props.value);
+    const [value, setValue] = useState<string>(props.value);
+
+    console.log(initialValue, value);
```
AmeliaIsCoding, line 15: remove

supabase/functions/check-congestion-charge/index.ts
```diff
@@ -29,6 +29,7 @@ serve(async (req: Handler): Promise<Response> => {
             .download("public/congestion-zone-postcodes.txt");
 
         if (error) {
+            console.error(error);
```
AmeliaIsCoding, line 32: remove

supabase/functions/check-congestion-charge/index.ts
```diff
@@ -44,6 +45,7 @@ serve(async (req: Handler): Promise<Response> => {
 
         return new Response(JSON.stringify(result), generateCorsOptionsForJsonResponse(200));
     } catch (error) {
+        console.error(error);
```
AmeliaIsCoding, line 48: remove

AmeliaIsCoding (commented):
self-review

rickaw:
(haven't really reviewed this yet as there are some issues in the code highlighted by the pipeline)

src/app/admin/AdminPage.tsx
```diff
@@ -22,6 +23,7 @@ import CollectionCentresTable from "@/app/admin/collectionCentresTable/Collectio
 import { Schema } from "@/databaseUtils";
 import CreateCollectionCentreForm from "@/app/admin/createCollectionCentre/CreateCollectionCentreForm";
 import PackingSlotsTable from "@/app/admin/packingSlotsTable/PackingSlotsTable";
+import WebsiteDataTable from "./websiteDataTable/WebsiteDataTable";
```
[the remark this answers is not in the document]
AmeliaIsCoding: Yeah, it is the default import. should I change?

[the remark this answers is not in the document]
rickaw: No it's fine in that case.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+    readableName: string;
+    id: string;
+    value: string;
+}
+
+const WebsiteDataTable: React.FC = () => {
+    const [rows, setRows] = useState<WebsiteDataRow[]>([]);
+    const [rowModesModel, setRowModesModel] = useState<GridRowModesModel>({});
+    const [isLoading, setIsLoading] = useState<boolean>(true);
+    const apiRef = useGridApiRef();
```
rickaw, line 35: Could you explain why we need this?

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    useEffect(() => {
+        fetchWebsiteData()
+            .then((response) => setRows(response))
+            .catch(async (error) => {
+                const logId = await logErrorReturnLogId("Error with fetch: website data", error);
+                throw new DatabaseError("fetch", "website data table", logId);
```
rickaw, line 42: There might be other places in the code I've let this past, but I don't think we can catch this error if we throw here? Maybe we should instead just do
- log the error as we already do, and
- show an error message on the table to prompt the user to refresh?

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+            })
+            .finally(() => setIsLoading(false));
```
rickaw, line 44: do we also want to set isloading to true just before fetching?

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+                        const websiteData = await fetchWebsiteData();
+                        setRows(websiteData);
+                    } catch (error) {
+                        if (error) {
+                            const logId = await logErrorReturnLogId(
+                                "Error with fetch: website data subscription",
+                                error
+                            );
+                            setRows([]);
+                            throw new DatabaseError("fetch", "website data table", logId);
```
rickaw, line 65: I don't think we'll be able to catch this error - I think replacing this with displaying an error message will be fine

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    }
+                }
+            )
+            .subscribe(async (status, err) => {
+                if (status === "TIMED_OUT") {
+                    const logId = await logErrorReturnLogId(
+                        "Channel Timed Out: Subscribe to website_data table",
+                        err
+                    );
+                    throw new DatabaseError("fetch", "website data table", logId);
```
rickaw, line 76: ditto about throwing these errors too

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    const logId = await logErrorReturnLogId(
+                        "Channel Error: Subscribe to website_data table",
+                        err
+                    );
+                    throw new DatabaseError("fetch", "website data table", logId);
+                } else if (status === "CLOSED") {
+                    logInfoReturnLogId("Subscription to website_data table closed");
+                } else {
+                    logInfoReturnLogId("Subscribed to website_data table");
+                }
```
rickaw, line 87: is this copied over from somewhere? We should commonise the logic if so.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+            [id]: { mode: GridRowModes.View },
+        }));
+    };
+
+    const processRowUpdate = (newRow: WebsiteDataRow): WebsiteDataRow => {
+        setIsLoading(true);
+        updateDbWebsiteData(newRow)
+            .catch(async (error) => {
+                const logId = await logErrorReturnLogId("Error with update: website data", error);
+                throw new DatabaseError("update", "website data table", logId);
```
rickaw, line 107: ditto about throwing

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    const handleEditClick = (id: GridRowId) => () => {
+        setRowModesModel((currentValue) => ({
+            ...currentValue,
+            [id]: { mode: GridRowModes.Edit },
+        }));
+    };
+
+    const handleCancelClick = (id: GridRowId) => async () => {
+        console.log(rowModesModel[id]);
```
rickaw, line 131: Should remove this console.log

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+            [id]: { mode: GridRowModes.View, ignoreModifications: true },
+        }));
+
+        const editedRow = rows.find((row) => row.id === id);
+        if (editedRow === undefined) {
+            {
+                const logId = await logErrorReturnLogId(
+                    "Edited row in website data admin table is undefined onCancelClick"
+                );
+                throw new DatabaseError("update", "website data table", logId);
```
rickaw, line 143: ditto about throwing an error - displaying an error is probably better?

src/app/admin/websiteDataTable/FetchWebsiteData.tsx
rickaw, line 1: This probably ought to be `fetchWebsiteData.ts` (lowercase first letter + `.ts` extension)

src/app/admin/websiteDataTable/FetchWebsiteData.tsx
```diff
@@ -0,0 +1,51 @@
...
+    name
+        .split("_")
+        .map((word) => `${word[0].toUpperCase()}${word.slice(1)}`)
+        .join(" ");
+
+const getUnreadableName = (name: string): string =>
+    name
+        .split(" ")
+        .map((word) => word.toLowerCase())
+        .join("_");
```
rickaw, line 19: I don't think these are strictly reverse of each other.

If we have a column called `A`, this becomes `A` through `getReadableName` and `a` through `getUnreadableName`.

All this to say, maybe `WebsiteDataRow` should just have an additional field `dbName` or something to just keep track of the original db value so it would always always work.

src/app/admin/websiteDataTable/CustomEditComponent.tsx
```diff
@@ -0,0 +1,57 @@
+import { GridRenderEditCellParams, useGridApiContext } from "@mui/x-data-grid";
+import React, { useEffect, useLayoutEffect, useRef, useState } from "react";
+
+interface CustomCellProps extends GridRenderEditCellParams {
+    editMode: boolean;
+}
+
+const CustomComponent: React.FC<CustomCellProps> = (props: CustomCellProps) => {
```
rickaw, line 8: inconsistency between filename and the component name: though I'm not sure what CustomEditComponent would imply. Maybe `EditableTextAreaForDataGrid` or something? (since this component uses useGridApiContext which requires it to be inside the data grid)

src/app/admin/websiteDataTable/CustomEditComponent.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const { id, field, hasFocus } = props;
```
rickaw, line 9: nitpick: you could just do this within the parameter
```
const CustomComponent: React.FC<CustomCellProps> = ({ id, field, hasFocus }: CustomCellProps) => {
```

src/app/admin/websiteDataTable/CustomEditComponent.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const initialValue = useRef(props.value);
+    const [value, setValue] = useState<string>(props.value);
+
+    useEffect(() => {
+        initialValue.current = props.value;
+        setValue(props.value);
+    }, [props.value]);
+
+    useLayoutEffect(() => {
+        if (hasFocus) {
```
rickaw, line 21: We could additionally check for ref.current since everything inside the if is to do with ref.current anyway
```suggestion
        if (hasFocus && ref.current) {
```

src/app/admin/websiteDataTable/CustomEditComponent.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const { id, field, hasFocus } = props;
+    const apiRef = useGridApiContext();
+    const ref = useRef<HTMLTextAreaElement>(null);
+    const initialValue = useRef(props.value);
+    const [value, setValue] = useState<string>(props.value);
+
+    useEffect(() => {
+        initialValue.current = props.value;
+        setValue(props.value);
+    }, [props.value]);
```
rickaw, line 18: so whenever the prop.value changes, this changes `initialValue` and `value`? Why can't the parent just manage the value and pass setValue down to this component?

src/pdf/DriverOverview/DriverOverview.tsx
```diff
@@ -82,10 +82,17 @@ const DriverOverview = async ({
     date,
 }: Props): Promise<React.ReactElement> => {
     const requiredData = await getRequiredData(parcelIds);
+    const { data, error } = await supabase
+        .from("website_data")
+        .select("name, value")
+        .eq("name", "driver_overview_message")
+        .single();
+    const message = error ? "" : data.value;
```
rickaw, line 90: If there's an error, should we say that in the message?

rickaw (changes requested):

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
AmeliaIsCoding: We use it as a context which allows us to update the value of the edited cell within the custom component by using apiRef.current.setEditCellValue(....)
I think this is the conventional way to deal with custom edit components in the mui data grid
https://mui.com/x/react-data-grid/editing/#create-your-own-edit-component

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
AmeliaIsCoding: Sounds good

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
AmeliaIsCoding: good catch

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
AmeliaIsCoding: Yeah I copied it from packing slots table. I'll commonise the logic, and update WebsiteDataTable.tsx and PackingSlotsTable.tsx to use it. Should I then put this logic in the other pages where we have realtime subscriptions?

src/app/admin/websiteDataTable/FetchWebsiteData.tsx
AmeliaIsCoding: Ah I see. Good solution

src/app/admin/websiteDataTable/CustomEditComponent.tsx
AmeliaIsCoding: Yep sure. I was lazy here

src/app/admin/websiteDataTable/CustomEditComponent.tsx
AmeliaIsCoding: I thought about that, but I wasn't sure how we could keep track of all the states in the parent, given that we have an arbitrary number of rows in the grid, and we would need a state for each row. I'm sure that's possible but I don't know how to write that code

src/pdf/DriverOverview/DriverOverview.tsx
AmeliaIsCoding: yep makes sense

src/app/admin/websiteDataTable/CustomEditComponent.tsx
rickaw: Could the parent store some kind of array as a state? Just had a quick look around the code and the data grid docs. Could we do something like the following using generics so that the data grid is aware that `params.row` in each row is `WebsiteDataRow`?

```
// in WebsiteDataTable.tsx

export interface WebsiteDataRow {
    dbName: string;
    readableName: string;
    id: string;
    value: string;
}

...

const [rows, setRows] = useState<GridRowsProp<WebsiteDataRow>>([]);
...

const websiteDataColumns: GridColDef<WebsiteDataRow>[] = [
  ...,
  {
    renderCell: (params) => <EditableTextAreaForDataGrid {...params} editMode={false} value={params.row.value} />,
  },
  ...
]

...
```

AmeliaIsCoding: Ok. the types are working fine. What I've done is move the apiRef.current.setEditCellValue(...) call into the parent. I think this ref is holding all the values that we want, so I think this is the better way to do it.

No need for the value state in the child anymore, but we do still need the initialValue ref - this is purely to automatically set the cursor to the right position (the end) on clicking edit.

src/common/logSubscriptionStatus.ts
```diff
@@ -0,0 +1,20 @@
...
+        return true;
+    } else if (status === "CHANNEL_ERROR") {
+        await logErrorReturnLogId(`Channel Error: Subscribe to ${tableName} table`, err);
+        return true;
+    } else if (status === "CLOSED") {
+        logInfoReturnLogId(`Subscription to ${tableName} table closed`);
+    } else {
+        logInfoReturnLogId(`Subscribed to ${tableName} table`);
+    }
+    return false;
```
rickaw, line 19: If we void the logging, we can even make this a synchronous function.
```suggestion
export const logSubscriptionStatus = (
    status: "TIMED_OUT" | "CHANNEL_ERROR" | "CLOSED" | "SUBSCRIBED",
    err: Error | undefined,
    tableName: string
): boolean => {
    if (status === "TIMED_OUT") {
        void logErrorReturnLogId(`Channel Timed Out: Subscribe to ${tableName} table`, err);
        return true;
    } else if (status === "CHANNEL_ERROR") {
        void logErrorReturnLogId(`Channel Error: Subscribe to ${tableName} table`, err);
        return true;
    } else if (status === "CLOSED") {
        void logInfoReturnLogId(`Subscription to ${tableName} table closed`);
    } else {
        void logInfoReturnLogId(`Subscribed to ${tableName} table`);
    }
    return false;
```

src/common/logSubscriptionStatus.ts
```diff
@@ -0,0 +1,20 @@
+import { logErrorReturnLogId, logInfoReturnLogId } from "@/logger/logger";
+
+export const logSubscriptionStatus = async (
+    status: "TIMED_OUT" | "CHANNEL_ERROR" | "CLOSED" | "SUBSCRIBED",
+    err: Error | undefined,
+    tableName: string
+): Promise<boolean> => {
+    if (status === "TIMED_OUT") {
+        await logErrorReturnLogId(`Channel Timed Out: Subscribe to ${tableName} table`, err);
+        return true;
```
rickaw, line 10: What does this function return? Can we make that clear in the function name?

rickaw (changes requested):

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
AmeliaIsCoding: makes sense!

src/app/admin/websiteDataTable/EditableTextAreaForDataGrid .tsx
```diff
@@ -0,0 +1,52 @@
...
+    editMode: boolean;
+    handleValueChange: (value: string, id: GridRowId, field: string) => void;
+}
+
+const EditableTextAreaForDataGrid: React.FC<EditableTextAreaForDataGridProps> = ({
+    id,
+    field,
+    hasFocus,
+    value,
+    editMode,
```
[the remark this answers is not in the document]
AmeliaIsCoding: good point, theres no need for hasFocus

src/app/admin/websiteDataTable/EditableTextAreaForDataGrid .tsx
```diff
@@ -0,0 +1,52 @@
...
+    handleValueChange,
+}: EditableTextAreaForDataGridProps) => {
+    const ref = useRef<HTMLTextAreaElement>(null);
+    const initialValue = useRef(value);
```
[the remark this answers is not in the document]
AmeliaIsCoding: discussed in person and decided to remove initialValue and disable the lint rule that necessitated its existence

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -0,0 +1,240 @@
...
+    const [errorMessage, setErrorMessage] = useState<string | null>(null);
+    const apiRef = useGridApiRef();
+
+    useEffect(() => {
+        setIsLoading(true);
+        setErrorMessage(null);
+        fetchWebsiteData()
+            .then((response) => setRows(response))
+            .catch(async (error) => {
+                await logErrorReturnLogId("Error with fetch: website data", error);
```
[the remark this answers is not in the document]
AmeliaIsCoding: Cool, i'll change all of these in the file, where possible

src/common/logSubscriptionStatus.ts
AmeliaIsCoding: Ok I've renamed to checkAndLogSubscriptionStatus because I think that's easier to read than logSubscriptionStatusAndReturnError or something. so have flipped the booleans, so it returns false if there's an error and true if there isn't.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
AmeliaIsCoding: @rickaw do you want me to update the other pages with realtime subscriptions?

rickaw: That'd be great if it's not too much trouble!

src/app/admin/websiteDataTable/EditableTextAreaForDataGrid .tsx
```diff
@@ -0,0 +1,46 @@
...
+    handleValueChange,
+}: EditableTextAreaForDataGridProps) => {
+    const ref = useRef<HTMLTextAreaElement>(null);
+
+    useLayoutEffect(() => {
+        if (editMode && ref.current) {
+            ref.current.focus();
+            ref.current.setSelectionRange(value.length, value.length);
+        }
+    }, [editMode]); //eslint-disable-line react-hooks/exhaustive-deps
```
rickaw, line 23: Shall we add a comment here to explain why we had to do this?

src/common/logSubscriptionStatus.ts
```diff
@@ -0,0 +1,20 @@
...
+        return false;
+    } else if (status === "CHANNEL_ERROR") {
+        void logErrorReturnLogId(`Channel Error: Subscribe to ${tableName} table`, err);
+        return false;
+    } else if (status === "CLOSED") {
+        void logInfoReturnLogId(`Subscription to ${tableName} table closed`);
+    } else {
+        void logInfoReturnLogId(`Subscribed to ${tableName} table`);
+    }
+    return true;
```
rickaw, line 19: nitpick: If we're simply going through every case of `status` then it feels nicer to use swtich case - but it might just be me 
```suggestion
    switch (status) {
        case "TIMED_OUT":
            void logErrorReturnLogId(`Channel Timed Out: Subscribe to ${tableName} table`, err);
            return false;
        case "CHANNEL_ERROR":
            void logErrorReturnLogId(`Channel Error: Subscribe to ${tableName} table`, err);
            return false;
        case "CLOSED":
            void logInfoReturnLogId(`Subscription to ${tableName} table closed`);
            return true;
        case "SUBSCRIBED":
            void logInfoReturnLogId(`Subscribed to ${tableName} table`);
            return true;
    }
```

rickaw (approved):
A few minor comments but feel free to merge after they're addressed!

src/common/logSubscriptionStatus.ts
AmeliaIsCoding: Agreed

src/common/logSubscriptionStatus.ts
```diff
@@ -0,0 +1,20 @@
+import { logErrorReturnLogId, logInfoReturnLogId } from "@/logger/logger";
+
+export const checkAndLogSubscriptionStatus = (
```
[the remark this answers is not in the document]
AmeliaIsCoding: Okay, sure

PR #189, opened 2024-03-25
VFB-29: FIX request pagination

AmeliaIsCoding:
## What's changed
Fixed stale fetch logic, that was supposed to ensure that the most recently sent request is the one which the data is set to, no matter if another request returns sooner. It wasn't working, now it is. Both parcels and clients page.

## Checklist

If you have made any changes to the database... nope

### DIFF
--- a/src/app/clients/ClientsPage.tsx
+++ b/src/app/clients/ClientsPage.tsx
@@ -84,7 +84,7 @@ const ClientsPage: React.FC<{}> = () => {
     const [filteredClientCount, setFilteredClientCount] = useState<number>(0);
     const [sortState, setSortState] = useState<SortState<ClientsTableRow>>({ sortEnabled: false });
     const [primaryFilters, setPrimaryFilters] = useState<Filter<ClientsTableRow, any>[]>(filters);
-    const abortController = useRef<AbortController>();
+    const clientTableFetchAbortController = useRef<AbortController | null>(null);
 
     const [perPage, setPerPage] = useState(10);
     const [currentPage, setCurrentPage] = useState(1);
@@ -94,15 +94,15 @@ const ClientsPage: React.FC<{}> = () => {
     useEffect(() => {
         (async () => {
             setIsLoading(true);
-            if (abortController.current) {
-                abortController.current.abort("stale request");
+            if (clientTableFetchAbortController.current) {
+                clientTableFetchAbortController.current.abort("stale request");
             }
-            abortController.current = new AbortController();
-            if (abortController.current) {
+            clientTableFetchAbortController.current = new AbortController();
+            if (clientTableFetchAbortController.current) {
                 const filteredClientCount = await getClientsCount(
                     supabase,
                     primaryFilters,
-                    abortController.current.signal
+                    clientTableFetchAbortController.current.signal
                 );
                 const fetchedData = await getClientsData(
                     supabase,
@@ -110,9 +110,9 @@ const ClientsPage: React.FC<{}> = () => {
                     endPoint,
                     primaryFilters,
                     sortState,
-                    abortController.current.signal
+                    clientTableFetchAbortController.current.signal
                 );
-                abortController.current = undefined;
+                clientTableFetchAbortController.current = null;
                 !filteredClientCount.abortSignalResponse.aborted &&
                     setFilteredClientCount(filteredClientCount.count);
                 !fetchedData.abortSignalResponse.aborted && setClientsDataPortion(fetchedData.data);
@@ -125,15 +125,15 @@ const ClientsPage: React.FC<{}> = () => {
     useEffect(() => {
         const loadCountAndData = async (): Promise<void> => {
             setIsLoading(true);
-            if (abortController.current) {
-                abortController.current.abort("stale request");
+            if (clientTableFetchAbortController.current) {
+                clientTableFetchAbortController.current.abort("stale request");
             }
-            abortController.current = new AbortController();
-            if (abortController.current) {
+            clientTableFetchAbortController.current = new AbortController();
+            if (clientTableFetchAbortController.current) {
                 const filteredParcelCount = await getClientsCount(
                     supabase,
                     primaryFilters,
-                    abortController.current.signal
+                    clientTableFetchAbortController.current.signal
                 );
                 const fetchedData = await getClientsData(
                     supabase,
@@ -141,9 +141,9 @@ const ClientsPage: React.FC<{}> = () => {
                     endPoint,
                     primaryFilters,
                     sortState,
-                    abortController.current.signal
+                    clientTableFetchAbortController.current.signal
                 );
-                abortController.current = undefined;
+                clientTableFetchAbortController.current = null;
                 !filteredParcelCount.abortSignalResponse.aborted &&
                     setFilteredClientCount(filteredParcelCount.count);
                 !fetchedData.abortSignalResponse.aborted && setClientsDataPortion(fetchedData.data);
--- a/src/app/clients/ClientsPage.tsx
+++ b/src/app/clients/ClientsPage.tsx
@@ -8,7 +8,7 @@ import Table, { SortOptions, TableHeaders, SortState } from "@/components/Tables
 import TableSurface from "@/components/Tables/TableSurface";
 import supabase from "@/supabaseClient";
 import { faUser } from "@fortawesome/free-solid-svg-icons";
-import React, { useEffect, useState, Suspense } from "react";
+import React, { useEffect, useState, Suspense, useRef } from "react";
 import { useTheme } from "styled-components";
 import getClientsData, { getClientsCount } from "./getClientsData";
 import { useSearchParams, useRouter } from "next/navigation";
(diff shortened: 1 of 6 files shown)


AmeliaIsCoding (commented):
self -review

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -159,37 +168,49 @@ export const getParcelIds = async (
...
+    });
+
+    if (
+        sortState.sortEnabled &&
+        sortState.column.sortMethodConfig?.paginationType === PaginationType.Server
+    ) {
+        query = sortState.column.sortMethodConfig.method(query, sortState.sortDirection);
+    } else {
+        query = query.order("packing_datetime", { ascending: false });
+    }
```
rickaw, line 192: these lines are exactly the same in 3 places (getParcelsByIds, getParcelProcessingData & getParcelIds). shall we commonise this?

rickaw (changes requested):

src/app/clients/ClientsPage.tsx
```diff
@@ -85,56 +84,70 @@ const ClientsPage: React.FC<{}> = () => {
     const [filteredClientCount, setFilteredClientCount] = useState<number>(0);
     const [sortState, setSortState] = useState<SortState<ClientsTableRow>>({ sortEnabled: false });
     const [primaryFilters, setPrimaryFilters] = useState<Filter<ClientsTableRow, any>[]>(filters);
+    const abortController = useRef<AbortController>();
```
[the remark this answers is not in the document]
AmeliaIsCoding: cool will change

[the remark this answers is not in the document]
AmeliaIsCoding: is it ok if I change to clientTableFetchAbortController, because the same abortController is used for both the count and the data requests (because if we abort one we will abort the other, and then we can just run abortController.abort() and it deals with both)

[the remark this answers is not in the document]
rickaw: that sounds reasonable!

src/app/clients/getClientsData.ts
```diff
@@ -33,8 +34,14 @@ const getClientsData = async (
 
     query = query.range(startIndex, endIndex);
 
+    query = query.abortSignal(abortSignal);
+
     const { data: clients, error: clientError } = await query;
 
+    if (abortSignal.aborted) {
+        return { data: [], abortSignalResponse: abortSignal };
+    }
```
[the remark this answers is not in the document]
AmeliaIsCoding: yes it does, but after this line, we check for errors in the response and throw an error if there is one. This is bad because if we abort, we don't want to throw an error. so we need to return from the fetch function before. 

Another option that I tried was not throwing an error if there is a specific "abort" error, but the abort error returned by supabase was inconsistent. 

Plus, we have to return something from the fetch function, so if we were to return null or [] or something, we'd have to have code in ParcelsPage.tsx checking if the data returned from the fetch function was truthy, and if so, set the data to be that. I thought that was less readable than returning the abortSignal, which allows us to check if the abortSignal.aborted is true

thoughts?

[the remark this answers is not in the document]
rickaw: Interesting - I would've expected the error returned by supabase to have a consistent code or something at least.

I understand your comment about dealing with the error just after this. IMO the real problem lies in the lack of error handling by this function's caller. If this fetching fails somehow, we should handle it gracefully on the frontend (maybe _not doing anything_ is what we want in the `catch`). 

It'd then be fine to throw an error here even when the request is aborted. (obviously it'd be much better if we could log something else when aborted, but if we can't meaningfully distinguish between an aborted request and other errors, that's fine)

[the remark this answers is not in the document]
AmeliaIsCoding: Okay cool I agree we should be dealing with it gracefully in ParcelsPage.tsx makes sense. 

Are you sure its ok to throw an error when the request is aborted? because we're expecting that to happen quite a lot of the time, so it would really clog up the cloudwatch log.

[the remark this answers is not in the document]
AmeliaIsCoding: Ok sure good idea. Have done

src/app/clients/getClientsData.ts
```diff
@@ -33,11 +34,18 @@ const getClientsData = async (
...
 
     if (clientError) {
-        const logId = await logErrorReturnLogId("Error with fetch: Clients", clientError);
-        throw new DatabaseError("fetch", "clients", logId);
+        let logId;
+        if (abortSignal.aborted) {
+            logId = await logInfoReturnLogId("Aborted fetch: client table count", clientError);
+        } else {
+            logId = await logErrorReturnLogId("Error with fetch: client table count", clientError);
+        }
```
rickaw, line 47:
```suggestion
        const logId = abortSignal.aborted
            ? await logInfoReturnLogId("Aborted fetch: client table count", clientError);
            : await logErrorReturnLogId("Error with fetch: client table count", clientError);
```

(same change could be made in getClientsCount and fetchParcelTableData)

src/app/parcels/ParcelsPage.tsx
```diff
@@ -358,6 +357,19 @@ const ParcelsPage: React.FC<{}> = () => {
 
     const [sortState, setSortState] = useState<SortState<ParcelsTableRow>>({ sortEnabled: false });
 
+    const [perPage, setPerPage] = useState(10);
```
rickaw, line 360: shall we rename this to parcelCountPerPage while we're at it?

src/app/parcels/ParcelsPage.tsx
```diff
@@ -458,35 +459,33 @@ const ParcelsPage: React.FC<{}> = () => {
...
-                allFilters: { ...allFilters },
-                sortState: { ...sortState },
-                startPoint: startPoint,
-                endPoint: endPoint,
-            };
-            (async () => {
-                setIsLoading(true);
-                const filteredParcelCount = await getParcelsCount(supabase, allFilters);
-                const fetchedData = await getParcelsData(
+            const allFilters = [...primaryFilters.slice(), ...additionalFilters.slice()];
```
rickaw, line 462: Is this another weird array behaviour that required you to do this?
```suggestion
            const allFilters = [...primaryFilters, ...additionalFilters];
```
or even
```suggestion
            const allFilters = primaryFilters.concat(additionalFilters);
```

src/app/parcels/ParcelsPage.tsx
```diff
@@ -458,35 +459,33 @@ const ParcelsPage: React.FC<{}> = () => {
...
+            }
+            parcelsTableFetchAbortController.current = new AbortController();
+            if (parcelsTableFetchAbortController.current) {
+                getParcelsCount(
+                    supabase,
+                    allFilters,
+                    parcelsTableFetchAbortController.current.signal
+                )
+                    .then((count) => setFilteredParcelCount(count))
+                    .catch();
```
rickaw, line 475: should we display set any error message in these catches?

src/app/parcels/ParcelsPage.tsx
```diff
@@ -458,35 +459,33 @@ const ParcelsPage: React.FC<{}> = () => {
...
-            (async () => {
-                setIsLoading(true);
-                const filteredParcelCount = await getParcelsCount(supabase, allFilters);
-                const fetchedData = await getParcelsData(
+            const allFilters = [...primaryFilters.slice(), ...additionalFilters.slice()];
+            setIsLoading(true);
+            if (parcelsTableFetchAbortController.current) {
+                parcelsTableFetchAbortController.current.abort("stale request");
+            }
+            parcelsTableFetchAbortController.current = new AbortController();
```
rickaw, line 467: nitpick: could probably do with some empty lines to separate logic around here

src/app/clients/ClientsPage.tsx
```diff
@@ -85,56 +84,70 @@ const ClientsPage: React.FC<{}> = () => {
...
+        }
+        clientTableFetchAbortController.current = new AbortController();
+        if (clientTableFetchAbortController.current) {
+            getClientsCount(
+                supabase,
+                primaryFilters,
+                clientTableFetchAbortController.current.signal
+            )
+                .then((count) => setFilteredClientCount(count))
+                .catch();
```
rickaw, line 107: should we set error in these catches?

rickaw (changes requested):

src/app/clients/getClientsData.ts
AmeliaIsCoding: nice

src/app/parcels/ParcelsPage.tsx
AmeliaIsCoding: image

src/app/parcels/ParcelsPage.tsx
AmeliaIsCoding: I don't think so because it could be an aborted request, in which case we don't want to set an error

src/app/clients/ClientsPage.tsx
AmeliaIsCoding: I don't think so because it could be an aborted request, in which case we don't want to set an error

rickaw: What's stopping us from throwing a different custom error when the request is aborted, not the DatabaseError? Then in the catch, we can check if the thrown error is an instance of the DatabaseError or not

src/app/parcels/ParcelsPage.tsx
AmeliaIsCoding: Nope, this was leftover from when I was trying to fix the weird array behaviour, but we didn't actually need this, it had no effect. have removed

src/app/clients/ClientsPage.tsx
AmeliaIsCoding: sounds good! have done

rickaw (approved):

PR #190, opened 2024-03-25
NONE: Fix accessibility issues in parcels page

AmeliaIsCoding:
## What's changed
Fixed breaking accessibility test on parcels page:
- table didn't have the correct roles for its children when the loading symbol was displayed within the table
- there was a duplicate id "status-button"

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist
- [ ] The ticket is up-to-date - Please document any deviations from the original approach if there is any. 

### DIFF
--- a/src/components/DataInput/CheckboxGroupPopup.tsx
+++ b/src/components/DataInput/CheckboxGroupPopup.tsx
@@ -43,7 +43,7 @@ const CheckboxGroupPopup: React.FC<Props> = (props) => {
                 onClick={(event) => setPopoverAnchorElement(event.currentTarget)}
                 disabled={!props.labelsAndKeys.length}
                 type="button"
-                id="status-button"
+                id="checkbox-status-button"
                 endIcon={<ArrowDropDown />}
             >
                 {props.groupLabel}
--- a/src/components/DataInput/CheckboxGroupPopup.tsx
+++ b/src/components/DataInput/CheckboxGroupPopup.tsx
@@ -43,7 +43,6 @@ const CheckboxGroupPopup: React.FC<Props> = (props) => {
                 onClick={(event) => setPopoverAnchorElement(event.currentTarget)}
                 disabled={!props.labelsAndKeys.length}
                 type="button"
-                id="checkbox-status-button"
                 endIcon={<ArrowDropDown />}
             >
                 {props.groupLabel}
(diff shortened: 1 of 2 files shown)


src/components/DataInput/CheckboxGroupPopup.tsx
```diff
@@ -43,7 +43,7 @@ const CheckboxGroupPopup: React.FC<Props> = (props) => {
                 onClick={(event) => setPopoverAnchorElement(event.currentTarget)}
                 disabled={!props.labelsAndKeys.length}
                 type="button"
-                id="status-button"
+                id="checkbox-status-button"
```
rickaw, line 46: This component is used in multiple places so I don't think setting this to a hard-coded id is a good idea.

If we don't actually need the id, can we remove it? (this seems to be the case?)

If we do need the ID for this to work, can we change this so that we have some sort of dynamic id name?

src/components/Tables/Table.tsx
```diff
@@ -444,8 +444,12 @@ const Table = <Data,>({
                         sortServer={sortConfig.sortPossible}
                         onSort={handleSort}
                         progressComponent={
-                            <Centerer>
-                                <CircularProgress aria-label="table-progress-bar" />
+                            <Centerer role="rowgroup">
+                                <CircularProgress
+                                    role="row"
```
rickaw, line 449: can't we use the `progressbar` role?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role

rickaw (changes requested):

src/components/Tables/Table.tsx
AmeliaIsCoding: No, the table specifically needs children with roles "row" and "rowgroup". I tried to override this requirement by setting aria-busy={isLoading} on the table component and various divs above of it in the hierarchy, which supposedly should work, but unfortunately it wouldn't. I think aria-busy needs to be set to true on the underlying table component, like, the library's component, but there's no way to do this unfortunately, think it's an oversight on their behalf. I thought this was the best fix.

src/components/DataInput/CheckboxGroupPopup.tsx
AmeliaIsCoding: sure, I think you're right, we don't need it

src/components/Tables/Table.tsx
AmeliaIsCoding: let me know your thoughts

AmeliaIsCoding: I don't think I can merge because the requested changes haven't been addressed

rickaw (approved):

Issue #199, opened 2024-03-26
Improve the password reset flow


rickaw:
### Describe the solution / feature you'd like

- Edit the password reset email to say that the link needs to be opened using the same machine and browser
- If the password reset link did not work as expected, we should display an error and explain that it did not succeed, and a likely reason why.
- If password reset is successful, direct user to / which should redirect to parcels page automatically (as the user is already signed in at this point)

Raised by #177

rickaw:
QA Steps
- [ ] Password reset email mentions that the link needs to be opened using the same machine and browser
- [ ] When password reset is successful, direct user to / which should redirect to parcels page automatically (as the user is already signed in at this point)
- [ ] If you click the same link twice, an error is displayed and explains that it did not succeed and a likely reason why.

minalui:
QA passed

PR #216, opened 2024-03-28
#199 improve password reset flow

rickaw:
## What's changed
- Edit the password reset email to say that the link needs to be opened using the same machine and browser (change in Supabase)
- If the password reset link did not work as expected, we should display an error and explain that it did not succeed, and a likely reason why.
- If password reset is successful, direct user to / which should redirect to parcels page automatically (as the user is already signed in at this point)

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| N/A | Example of how the error message is displayed image |

## Checklist
- [x] Make sure you've tested via `npm run test:e2e` (all good apart from the one Mina's fixing)

If you have made any changes to the database... NA

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,6 +1,7 @@
 name: Checks
 on:
   pull_request:
+  merge_group:
   push:
     branches:
       - dev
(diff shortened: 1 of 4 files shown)


AmeliaIsCoding (approved):
Looks good to me

PR #217, opened 2024-03-28
NONE: Enable realtime through migration

rickaw:
## What's changed
Enable realtime through migration.

I have checked
- [x] DB reset works successfully with realtime turned on

### DIFF
--- a/supabase/migrations/20240328103536_enable_realtime.sql
+++ b/supabase/migrations/20240328103536_enable_realtime.sql
@@ -1,0 +1,1 @@
+alter publication supabase_realtime add table clients, collection_centres, events, families, packing_slots, parcels;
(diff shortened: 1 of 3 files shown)


supabase/migrations/20240328103536_enable_realtime.sql
```diff
@@ -0,0 +1 @@
+alter publication supabase_realtime add table clients, collection_centres, events, families, packing_slots, parcels;
```
AmeliaIsCoding, line 1: I think we need website_data here too

AmeliaIsCoding (commented):
Yay!! one table missing

supabase/migrations/20240328103536_enable_realtime.sql
rickaw: oh you're right!

PR #238, opened 2024-04-05
VFB-173: Set up audit logging and log changes when parcel is added or edited

minalui:
## What's changed
Add audit logging when parcels are added or edited

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| paste_here | paste_here |

## Checklist

  - [N/A] I have modified the seed in `supabase/seed/seed.mts` if appropriate
  - [N/A ] If I have modified the seed, I have also generated the seed with `npm run db:generate_seed` 
  - Does this require resetting the deployed database? - YES (remove as appropriate)
    - Checkout the dev branch
    - Checkout your branch

### DIFF
--- a/.snaplet/snaplet-client.d.ts
+++ b/.snaplet/snaplet-client.d.ts
@@ -85,15 +85,6 @@ type Override = {
       objects?: string;
     };
   }
-  channels?: {
-    name?: string;
-    fields?: {
-      id?: string;
-      name?: string;
-      inserted_at?: string;
-      updated_at?: string;
-    };
-  }
   clients?: {
     name?: string;
     fields?: {
@@ -167,7 +158,6 @@ type Override = {
       created_at?: string;
       updated_at?: string;
       authentication_method?: string;
-      auth_code_issued_at?: string;
       saml_relay_states?: string;
     };
   }
@@ -457,13 +447,6 @@ type Override = {
       version?: string;
     };
   }
-  realtime_schema_migrations?: {
-    name?: string;
-    fields?: {
-      version?: string;
-      inserted_at?: string;
-    };
-  }
   supabase_migrations_schema_migrations?: {
     name?: string;
     fields?: {
@@ -534,18 +517,6 @@ type Override = {
       event_name?: string;
       workflow_order?: string;
       audit_log?: string;
-    };
-  }
-  subscription?: {
-    name?: string;
-    fields?: {
-      id?: string;
-      subscription_id?: string;
-      entity?: string;
-      filters?: string;
-      claims?: string;
-      claims_role?: string;
-      created_at?: string;
     };
   }
   users?: {
@@ -653,11 +624,6 @@ export interface Fingerprint {
     fileSizeLimit?: FingerprintNumberField;
     objects?: FingerprintRelationField;
   }
-  channels?: {
-    id?: FingerprintNumberField;
-    insertedAt?: FingerprintDateField;
-    updatedAt?: FingerprintDateField;
-  }
   clients?: {
     auditLogsByClientId?: FingerprintRelationField;
     families?: FingerprintRelationField;
@@ -680,7 +646,6 @@ export interface Fingerprint {
   flowStates?: {
     createdAt?: FingerprintDateField;
     updatedAt?: FingerprintDateField;
-    authCodeIssuedAt?: FingerprintDateField;
     samlRelayStates?: FingerprintRelationField;
   }
   hooks?: {
@@ -789,10 +754,6 @@ export interface Fingerprint {
   authSchemaMigrations?: {
 
   }
-  realtimeSchemaMigrations?: {
-    version?: FingerprintNumberField;
-    insertedAt?: FingerprintDateField;
-  }
   supabaseMigrationsSchemaMigrations?: {
 
   }
@@ -825,11 +786,6 @@ export interface Fingerprint {
   statusOrders?: {
     workflowOrder?: FingerprintNumberField;
     auditLogsByStatusOrder?: FingerprintRelationField;
-  }
-  subscriptions?: {
-    id?: FingerprintNumberField;
-    claims?: FingerprintJsonField;
-    createdAt?: FingerprintDateField;
   }
   users?: {
     emailConfirmedAt?: FingerprintDateField;
--- a/.snaplet/snaplet-client.d.ts
+++ b/.snaplet/snaplet-client.d.ts
@@ -27,6 +27,38 @@ type Override = {
       created?: string;
     };
   }
+  audit_log?: {
+    name?: string;
+    fields?: {
+      primary_key?: string;
+      user_id?: string;
+      action?: string;
+      client_id?: string;
+      collection_centre_id?: string;
+      event_id?: string;
+      family_member_id?: string;
+      list_id?: string;
+      list_hotel_id?: string;
+      packing_slot_id?: string;
+      parcel_id?: string;
+      status_order?: string;
+      website_data?: string;
+      content?: string;
+      wasSuccess?: string;
+      log_id?: string;
+      users?: string;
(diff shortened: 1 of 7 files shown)


src/app/parcels/form/clientDatabaseFunctions.ts
```diff
@@ -2,34 +2,71 @@ import supabase from "@/supabaseClient";
...
+    const { data, error } = await supabase
         .from("parcels")
         .insert(parcelRecord)
-        .select("primary_key, client_id");
+        .select("primary_key, client_id")
+        .single();
+
+    const auditLog: AuditLogProps = {
+        action: "add a parcel",
+        content: { parcel: parcelRecord },
```
minalui, line 19: rename parcelDetails

src/app/parcels/form/clientDatabaseFunctions.ts
```diff
@@ -2,34 +2,71 @@ import supabase from "@/supabaseClient";
...
         .from("parcels")
         .update(parcelRecord)
         .eq("primary_key", primaryKey)
-        .select("primary_key, client_id");
+        .select("primary_key, client_id")
+        .single();
+
+    const auditLog: AuditLogProps = {
+        action: "edit a parcel",
+        content: { parcel: parcelRecord },
```
minalui, line 53: rename parcelDetails

src/server/auditLog.ts
```diff
@@ -0,0 +1,55 @@
+import { getCurrentUser } from "@/server/getCurrentUser";
+import supabase from "@/supabaseClient";
+import { InsertSchema } from "@/databaseUtils";
+import { logErrorReturnLogId } from "@/logger/logger";
+import { DatabaseError } from "@/app/errorClasses";
+
+type AuditLogInsertRecord = InsertSchema["audit_log"];
+export interface AuditLogProps {
+    action?: string;
+    content: Record<string, any>;
```
minalui, line 10: I'm not sure about the any - it could potentially be type ParcelDatabaseInsertRecord for now but it won't cover the content that is not a parcel. What do you think?

src/server/getCurrentUser.ts
```diff
@@ -0,0 +1,18 @@
...
+import { User } from "@supabase/gotrue-js";
+
+export async function getCurrentUser(): Promise<User | null> {
+    const serverComponentClient = getSupabaseServerComponentClient();
+
+    const { data, error } = await serverComponentClient.auth.getUser();
+
+    if (error) {
+        const logId = await logErrorReturnLogId("error with auth getUser", error);
+        throw new Error(`Error getting current user. Log ID: ${logId}`);
```
minalui, line 14: Probably don't need to throw as it would stop null from being passed back, error should be handled at the caller of the function

supabase/migrations/20240405104731_add_audit_log_table.sql
```diff
@@ -0,0 +1,118 @@
...
+    "client_id" uuid,
+    "collection_centre_id" uuid,
+    "event_id" uuid,
+    "family_member_id" uuid,
+    "list_id" uuid,
+    "list_hotel_id" uuid,
+    "packing_slot_id" uuid,
+    "parcel_id" uuid,
+    "status_order" text,
+    "website_data" text,
```
minalui, line 14: status_order and website_date do not have primary keys

rickaw: they actually have primary keys! They just happen to be text, not uuid or number! so this is fine!

.snaplet/snaplet-client.d.ts
rickaw, line 1: Can we make sure these generated files are up to date please?

src/server/getCurrentUser.ts
```diff
@@ -0,0 +1,17 @@
+"use server";
+
+import { getSupabaseServerComponentClient } from "@/supabaseServer";
+import { logErrorReturnLogId } from "@/logger/logger";
+import { User } from "@supabase/gotrue-js";
+
+export async function getCurrentUser(): Promise<User | null> {
+    const serverComponentClient = getSupabaseServerComponentClient();
+
+    const { data, error } = await serverComponentClient.auth.getUser();
```
rickaw, line 10: nitpick: just for consistency
```suggestion
    const supabase = getSupabaseServerComponentClient();

    const { data, error } = await supabase.auth.getUser();
```

src/server/auditLog.ts
rickaw: How about using `Json | null` with `import { Json } from "@/databaseTypesFile";`? since that's what's needed by `AuditLogInsertRecord` anyway?

src/server/auditLog.ts
```diff
@@ -0,0 +1,56 @@
+import { getCurrentUser } from "@/server/getCurrentUser";
```
rickaw, line 1:
```suggestion
"use server";

import { getCurrentUser } from "@/server/getCurrentUser";
```

src/server/auditLog.ts
```diff
@@ -0,0 +1,56 @@
...
+        list_id: auditLogProps.listId,
+        list_hotel_id: auditLogProps.listHotelId,
+        packing_slot_id: auditLogProps.packingSlotId,
+        parcel_id: auditLogProps.parcelId,
+        content: auditLogProps.content,
+        wasSuccess: auditLogProps.wasSuccess,
+        log_id: auditLogProps.logId,
+    };
+
+    const { error } = await supabase
```
rickaw, line 47: I'm weary of using this `supabase` since it is a supabase client-side client. Can we just instantiate a supabase server client?

src/app/parcels/form/clientDatabaseFunctions.ts
```diff
@@ -2,34 +2,71 @@ import supabase from "@/supabaseClient";
...
     if (error) {
         const logId = await logErrorReturnLogId("Error with insert: parcel data", error);
+        auditLog.logId = logId;
+        await sendAuditLog(auditLog);
         throw new DatabaseError("insert", "parcel data", logId);
     }
+
+    auditLog.parcelId = data.primary_key;
+    auditLog.wasSuccess = true;
+    await sendAuditLog(auditLog);
```
rickaw, line 37: I'm a little wary of changing an object for readability, so it might be better to use `...` to construct a new object to pass to `sendAuditLog`. (similarly in the updateParcel function)

```suggestion
    const auditLog = {
        action: "add a parcel",
        content: { parcelDetails: parcelRecord },
        clientId: parcelRecord.client_id,
        collectionCentreId: parcelRecord.collection_centre
            ? parcelRecord.collection_centre
            : undefined,
        packingSlotId: parcelRecord.packing_slot ? parcelRecord.packing_slot : undefined,
    } as const satisfies Partial<AuditLogProps>;

    if (error) {
        const logId = await logErrorReturnLogId("Error with insert: parcel data", error);
        await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
        throw new DatabaseError("insert", "parcel data", logId);
    }

    await sendAuditLog({ ...auditLog, wasSuccess: true, parcelId: data.primary_key });
```

rickaw (changes requested):
Nice!

src/server/auditLog.ts
```diff
@@ -0,0 +1,61 @@
...
+    collectionCentreId?: string;
+    eventId?: string;
+    familyMemberId?: string;
+    listId?: string;
+    listHotelId?: string;
+    packingSlotId?: string;
+    parcelId?: string;
+}
+
+export async function sendAuditLog(auditLogProps: AuditLogProps): Promise<void> {
```
rickaw, line 26: nitpick: I don't think the name `prop` is needed here
```suggestion
export async function sendAuditLog(auditLog: AuditLog): Promise<void> {
```

rickaw (approved):

PR #240, opened 2024-04-08
VFB-180: Changing the number of parcels changes the order of rows

AmeliaIsCoding:
## What's changed

- Fixed bug that was showing parcels with packing dates 1 day behind the date filter settings, as well as the selected range
- By default the parcels are now ordered by packing date, packing slot and then client name (previously just packing date)
- If the user explicitly orders the table by packing date, then we also order by packing date, packing slot and client name
- If the user orders the table by packing date but in reverse, then we order by reverse packing date, then regular packing slot and client name order
- ALL requests to the parcels or clients tables are ordered by the primary key (ie parcel id or client id) as the final field to sort by: ie after either the default sorts, or the user input sorts. This ensures there is only one valid sort order for any response.

This last bullet point fixed a bug where, on changing the number of parcels per page, the order of the parcel rows changed. 

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image|image|

## Checklist


AmeliaIsCoding (commented):
self-review

src/app/parcels/parcelsTableFilters.ts
```diff
@@ -78,7 +78,7 @@ export const buildDateFilter = (
         query: PostgrestFilterBuilder<Database["public"], any, any>,
         state: DateRangeState
     ): PostgrestFilterBuilder<Database["public"], any, any> => {
-        return query.gte("packing_date", state.from).lte("packing_date", state.to);
+        return query.gt("packing_date", state.from).lte("packing_date", state.to);
```
rickaw, line 81: Do we perhaps also want to do `lt` instead of `lte`?
```suggestion
        return query.gt("packing_date", state.from).lt("packing_date", state.to);
```

AmeliaIsCoding: This would miss some parcels right - and if I do this no parcels show :/

Because we changed packing_date in the DB to be a date rather than a time, but the datefilter uses dayjs objects which have a time, I think the date filter gets effectively converted into a date without a time at the point of filtering in supabase.

eg if we have the date filter from 08/04/2024 00:00 to 08/04/2024 to 23:23:59, this gets converted into 08/04/2024 to 08/04/2024 at the point of filtering in supabase. so if we were to do gt and lt, we would have no parcels :0

AmeliaIsCoding: aware that if the above was true, the bug that caused me to change this shouldnt have occured, because then lte and gte would result in fine behaviour. I think this could be to do with timezone changes (yikes) which would explain why we only noticed this after the change to BST - i.e., the dayjs objects are in BST, this gets converted to GMT when comparing against the supabase packing_date because this is in UTC *, so an hour is subtracted from both to to date and the from date before truncating the time, thus in the above example we'd filter by the range 07/04/2024 - 08/04/2024 at the point of filtering in supabase. this was the behaviour that is occurring in dev that I fixed by doing this. this explains how using gt and lte gets around this.

suggestion: change date filter to use a date object / something without time, with timezone UTC, so that no conversion occurs when we compare to supabase. and filter by "lte and gte" as this would make most sense. 

* see https://supabase.com/docs/guides/database/postgres/configuration

AmeliaIsCoding: update: looks like there's not a good date object without time. maybe we carry on using dayjs but set the timezone to be UTC.

rickaw: Since the `packing_date` is a date type in the db, can't we use 'YYYY-MM-DD' format to compare the date and use `gte` and `lte`? i.e.

```
function getDbDate(dateTime: Dayjs): string {
  // return 2020-01-15 format
}

...

query.gt("packing_date", getDbDate(state.from)).lt("packing_date", getDbDate(state.to));

```

posgres docs on dates / formats: https://www.postgresql.org/docs/8.3/datatype-datetime.html

AmeliaIsCoding: update: have implemented the above.
TL;DR: changed date filter to use UTC to be consistent with supabase, which is an alternative fix to the problem in dev (that currently yesterdays parcels show by default too). perhaps better than the potentially less intuitive gt/lte fix.

rickaw: I don't think we should be always using UTC. When showing the parcels page for example, we want to show "today's parcels" by default. So we want to be using `dayjs()` to get the user's current date in whatever timezone they're in i.e. UK timezone at the time, and use that date to query from supabase. See my comment above.

rickaw (changes requested):

src/app/parcels/parcelsTableFilters.ts
rickaw: (there's definitely a change to be made with respect to collection date time though - The collection time to be entered by volunteers are on an artbitrary future date. I don't think there's a simple way to store time in BST if that future date is using BST, and store time in UTC otherwise. In which case the solution would be to store the collection date time without timezone. Ultimately we only care about the local collection time.

AmeliaIsCoding: yes this makes sense. have done

src/app/parcels/ParcelsPage.tsx
```diff
@@ -286,11 +292,10 @@ const ParcelsPage: React.FC<{}> = () => {
             primaryFilters: Filter<ParcelsTableRow, any>[];
             additionalFilters: Filter<ParcelsTableRow, any>[];
         }> => {
-            const startOfToday = dayjs().startOf("day");
-            const endOfToday = dayjs().endOf("day");
+            const today = dayjs().utc();
```
rickaw, line 295: I don't think we want to use utc here - we'd probably want to use "today's date" for the user which would be BST or UTC depending on season. `dayjs()` should just provide the current time in the current timezone
```suggestion
            const today = dayjs();
```

src/components/DateRangeInputs/DateRangeInputs.tsx
```diff
@@ -3,8 +3,10 @@
 import { DatePicker } from "@mui/x-date-pickers";
 import dayjs, { Dayjs } from "dayjs";
 import React from "react";
+import utc from "dayjs/plugin/utc";
 
-const reasonableMinDate = dayjs("2020-01-01");
+const reasonableMinDate = dayjs("2020-01-01").utc();
+dayjs.extend(utc);
```
rickaw, line 9: I don't think this is needed

src/app/parcels/ParcelsPage.tsx
AmeliaIsCoding: removed

src/components/DateRangeInputs/DateRangeInputs.tsx
AmeliaIsCoding: removed

src/app/parcels/ParcelsPage.tsx
rickaw, line 1: Can we also make this change in line 292 please? This would signal to the future devs that time has no significance for the date filter.

```
const today = dayjs()
const dateFilter = buildDateFilter({
    from: today,
    to: today,
});
```

src/app/parcels/parcelsTableFilters.ts
```diff
@@ -11,12 +11,15 @@ import { logErrorReturnLogId } from "@/logger/logger";
...
 import { checklistFilter } from "@/components/Tables/ChecklistFilter";
 import { CollectionCentresOptions } from "@/app/parcels/fetchParcelTableData";
+import { Dayjs } from "dayjs";
 
 interface packingSlotOptionsSet {
     key: string;
     value: string;
 }
 
+const getDbDate = (dateTime: Dayjs): string => dateTime.format("YYYY-MM-DD");
```
rickaw, line 21: just a nitpick: can we put this in format.ts please?

rickaw (approved):

PR #249, opened 2024-04-09
VFB-188: Move critial website data from seed to migrations

rickaw:
## What's changed
See ticket for the details.

## Checklist

  - [ ] I have updated the typescript definitions for the database with `db:local:generate_types` - NA
    - Checkout the dev branch
    - Checkout your branch


supabase/seed.sql
rickaw, line 1: website data and status_order seed is removed from this sql.

supabase/migrations/20240409135335_add_website_data_and_status_orders.sql
```diff
@@ -0,0 +1,31 @@
+DELETE FROM public.website_data;
```
rickaw, line 1: We're happy to just delete all items from website_data and status_order since we know the exact values we need them to be for now.

In the future
- Adding to website_data should be done using insert like in any other table
- Adding to status_order is a bit more tricky, as we'd likely want to reshuffle the order too. We'd probably end up deleting all and defining all as that's the easiest thing to do. (We could instead define a logic to add one item at a certain position and shift the order number up by one for some existing items, but might be an overkill.)

PR #251, opened 2024-04-09
VFB-184: Anonymise user's profile on deleting their account

AmeliaIsCoding:
## What's changed

- If deletion fails, close the "danger dialog" so that the error message under the users table is visible
- Changed profiles table so that the primary key is no longer an FK to auth.users, and add another column user_id to be a FK to auth_users
- When deleting a user, the corresponding row in auth.users is deleted, and the user_id, first name, last name and telephone number in profiles are all set to null

## Checklist

  - [ ] I have modified the seed in `supabase/seed/seed.mts` if appropriate -> not needed
  - [ ] If I have modified the seed, I have also generated the seed with `npm run db:generate_seed` -> not needed
    - Checkout the dev branch
    - Checkout your branch


src/server/adminInviteUser.ts
```diff
@@ -42,24 +42,27 @@ export async function adminInviteUser(
...
-            };
-        }
-        void logInfoReturnLogId(
-            `Created a profile for ${userDetails.role} user: ${userDetails.email}`
-        );
+    if (!data.user) {
+        return {
+            data: null,
+            error: { "Failed to create user": "Email may be in use already." },
+        };
```
AmeliaIsCoding, line 49: Not ready for team review, I need to check this behaviour on trying to create a user with an email address that already is in use

AmeliaIsCoding (commented):
Self review - not ready for team review yet
EDIT now ready

src/server/adminInviteUser.ts
AmeliaIsCoding: I have removed this as it was not working as intended. Created a new ticket for the error handling on creating a duplicate user, as this is a little trickier than first thought - [VFB-189](https://softwiretech.atlassian.net/browse/VFB-189)

supabase/migrations/20240409152753_anonymise-profile-on-deleting-user.sql
rickaw, line 1: can we add some sql code to deal with the existing profiles? We should populate `user_id` with what's currently in `primary_key`.

src/server/adminInviteUser.ts
```diff
@@ -42,24 +42,21 @@ export async function adminInviteUser(
...
+    const { error: createRoleError } = await supabase.from("profiles").insert({
+        role: userDetails.role,
+        first_name: userDetails.firstName,
+        last_name: userDetails.lastName,
+        telephone_number: userDetails.telephoneNumber,
+        user_id: data.user.id,
+    });
+
+    if (createRoleError) {
+        return {
```
rickaw, line 54: Should we log an error here?

rickaw (changes requested):
`UpdateUserProfiles` has the following which also needs to be updated

```ts
    const { error } = await supabase
        .from("profiles")
        .update({
            role: userDetails.role,
        })
        .eq("primary_key", userDetails.userId)
        .select()
        .single();
```

src/server/adminInviteUser.ts
AmeliaIsCoding: have done

supabase/migrations/20240409152753_anonymise-profile-on-deleting-user.sql
rickaw: Either, 
- put the additional sql code in the same file (`anonymise-profile-___.sql`)
- You create a new migration file with `npx supabase migration new <name_of_migration>`

I lean towards the first option cause then we can guarantee that the adding `user_id` column can happen in the same operation as populating the data.

### How to test your migration file
- checkout a commit without this migration file (latest dev if you've merged dev into your branch, otherwise use the base of you branch)
- run `npm run post_checkout` to reset the database in that version
- checkout your branch
- run `npx supabase migrations up --include-all --local` to apply outstanding migrations

src/server/adminInviteUser.ts
```diff
@@ -5,7 +5,12 @@ import { getSupabaseAdminAuthClient } from "@/supabaseAdminAuthClient";
 import { User } from "@supabase/gotrue-js";
 import { InviteUserDetails } from "@/app/admin/createUser/CreateUserForm";
 import supabase from "@/supabaseClient";
-import { logInfoReturnLogId } from "@/logger/logger";
+import { logErrorReturnLogId, logInfoReturnLogId } from "@/logger/logger";
+
+export type InviteUserErrorType =
+    | "Admin Authentication Failure"
```
rickaw, line 11: These string union types tend to use camelCase or just lower case. One reason is to signal that this isn't supposed to be used for a message for end-users. It's just for the code to identify which error has occurred. I gravitate towards camelCase though cause it's a little easier to refactor / rename with IDEs if we needed to.

```suggestion
    | "adminAuthenticationFailure" // (or "admin authentication failure")
```

src/server/adminInviteUser.ts
```diff
@@ -36,30 +47,31 @@ export async function adminInviteUser(
...
+        first_name: userDetails.firstName,
+        last_name: userDetails.lastName,
+        telephone_number: userDetails.telephoneNumber,
+        user_id: data.user.id,
+    });
+
+    if (createRoleError) {
+        const logId = await logErrorReturnLogId(
+            `Error with insert profile: ${createRoleError.message}`
         );
```
rickaw, line 68: why not log the error object
```suggestion
        const logId = await logErrorReturnLogId(
            `Error with insert profile`,
            { error: createRoleError }
        );
```

src/app/admin/createUser/CreateUserForm.tsx
```diff
@@ -120,8 +122,8 @@ const CreateUserForm: React.FC<{}> = () => {
                         Invite User
                     </Button>
                 )}
-
-                {submitError && <Alert severity="error">{submitError}</Alert>}
+                {serverError && <Alert severity="error">{serverError}</Alert>}
```
rickaw, line 125: Instead of directly using serverError as a string here, I think we'd want to show custom message based on serverError.

In principle, if we want to change the error message shown on the CreateUserForm when something goes wrong, we want to be editing `CreateUserForm.tsx`, not `adminInviteUser.ts`, right?

Additionally, if we were to have two separate pathways to invite a user on the website, maybe the error messages should be different for the same error (because there's a suspected user error in one case but not the other for example).

All this to say, `adminInviteUser.ts` shouldn't concern itself with error messages displayed on frontend.

rickaw (changes requested):

supabase/migrations/20240409152753_anonymise-profile-on-deleting-user.sql
AmeliaIsCoding: yay it worked

src/server/adminInviteUser.ts
AmeliaIsCoding: sure

src/app/admin/createUser/CreateUserForm.tsx
AmeliaIsCoding: okay sure, makes sense

src/server/adminInviteUser.ts
```diff
@@ -36,30 +47,31 @@ export async function adminInviteUser(
     });
 
     if (error) {
+        const logId = await logErrorReturnLogId(`Error with inviting user: ${error}`);
```
rickaw, line 50: How is this log displayed? I'm worried about this `error` being a class instance.

src/server/adminInviteUser.ts
```diff
@@ -36,30 +47,31 @@ export async function adminInviteUser(
...
+        first_name: userDetails.firstName,
+        last_name: userDetails.lastName,
+        telephone_number: userDetails.telephoneNumber,
+        user_id: data.user.id,
+    });
+
+    if (createRoleError) {
+        const logId = await logErrorReturnLogId(
+            `Error with insert profile: ${createRoleError}`
         );
```
rickaw: Why don't we log the createRoleError in the metadata rather in the message string?
```suggestion
        const logId = await logErrorReturnLogId(
            `Error with insert profile`,
            { error: createRoleError }
        );
```

supabase/migrations/20240409152753_anonymise-profile-on-deleting-user.sql
```diff
@@ -0,0 +1,38 @@
...
+    end if;
+    return new;
+END;$function$
+;
+
+CREATE TRIGGER anonymise_profile_trigger AFTER UPDATE OF user_id ON public.profiles FOR EACH ROW EXECUTE FUNCTION anonymise_profile();
+
+UPDATE profiles
+SET user_id = primary_key
+WHERE user_id IS NULL;
```
rickaw, line 38: nitpick: Let's be more specific like `alter table "public"."profiles" add column "user_id" uuid;`

```suggestion
UPDATE "public"."profiles"
SET "user_id" = "primary_key"
WHERE "user_id" IS NULL;
```

src/app/admin/createUser/CreateUserForm.tsx
```diff
@@ -75,15 +90,13 @@ const CreateUserForm: React.FC<{}> = () => {
...
 
         if (error) {
-            setSubmitError(Errors.external);
+            setServerError(error.type);
             setSubmitDisabled(false);
             setInvitedUser(null);
-
-            const logId = await logErrorReturnLogId("Error with auth: Invite user", error);
-            throw new DatabaseError("insert", "invite user", logId);
+            throw new DatabaseError("insert", "invite user", error.logId);
```
rickaw, line 96: I don't think we catch this error anywhere, do we even need to throw? Displaying an error message is how we deal with these errors right? 

And also, maybe we should display the log ID as part of the server error message, so they can tell us what to look for when something goes wrong?

src/server/adminInviteUser.ts
AmeliaIsCoding: should we do this for the above 2 errors as well?

src/server/adminInviteUser.ts
rickaw: Will it give us more info if we do this? I think we prefer sticking the error into the metadata object rather than trusting the string interpolation anyway.

```
logErrorReturnLogId(`Error with inviting user`, { error });
```

src/server/adminInviteUser.ts
AmeliaIsCoding: (have done)

src/app/admin/createUser/CreateUserForm.tsx
AmeliaIsCoding: yep agreed with both. have done

rickaw (approved):

rickaw:
Thanks for being patient with all my comments!

PR #261, opened 2024-04-11
VFB-194 VFB-195: Add audit logs for adding and editing packing slots and fix RLS policies checking for admin access

rickaw:
## What's changed
- Add audit logs for adding and editing packing slots
- Fix RLS policies checking for admin access
- Add an e2e test for moving packing slots

## Checklist
- [x] The ticket is up-to-date - Please document any deviations from the original approach if there is any.
- [x] I have documented the testing steps for QA
- [x] I have self-reviewed this PR
- [x] Make sure you've verified it works via `npm run dev`
- [x] Make sure you've verified it works via `npm run build` and `npm run start`
- [x] Make sure you've fixed all linting problems with `npm run lint_fix`
- [x] Make sure you've tested via `npm run test:e2e`

If you have made any changes to the database...
  - [x] The migration files are up-to-date with my final set up (`npx supabase db diff -f <name_of_migration>` should create nothing at this point)
  - [x] I have updated the typescript definitions for the database with `db:local:generate_types`
  - [x] I have modified the seed in `supabase/seed/seed.mts` if appropriate
  - [x] If I have modified the seed, I have also generated the seed with `npm run db:generate_seed` 
  - [x] With my final set up, I can run `npm run dev:reset_supabase` without any errors.
  - Does this require resetting the deployed database? - YES / NO (remove as appropriate)
  - [x] I have checked that migration can happen successfully without resetting the database, doing the following.
    - Make sure you have rebased your branch onto dev or merged dev into your branch.
    - Checkout the dev branch
    - Run `npm run post_checkout` to reset the database, including the seed data.
    - Checkout your branch
    - Run `npx supabase migrations up --include-all --local` to run all outstanding migration files
    - Check that the resulting database is what you expect, bar any seed data changes.

PR #262, opened 2024-04-12
VFB-191: Add audit logs when an ingredient is added or edited in the lists page

minalui:
## What's changed
- Add audit log for when list item is added, edited or deleted
- Change list item id in audit log table to become null if corresponding row in list table is deleted
- Add subscription for list table and allow table to be updated automatically. 
- Handle modal close actions.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| paste_here | paste_here |

## Checklist

    - Checkout the dev branch
    - Checkout your branch

### DIFF
--- a/src/app/lists/EditModal.tsx
+++ b/src/app/lists/EditModal.tsx
@@ -64,59 +64,67 @@ const EditModal: React.FC<Props> = ({ data, onClose }) => {
         setToSubmit({ ...toSubmit, [key]: newValue });
     };
 
+    const addListItem = async (listItem: Partial<Schema["lists"]>): Promise<void> => {
+        const { data: returnedListData, error: insertListItemError } = await supabase
+            .from("lists")
+            .insert(listItem)
+            .select()
+            .single();
+
+        const auditLog = {
+            content: { itemDetails: listItem },
+            action: "add a list item",
+        } as const satisfies Partial<AuditLog>;
+
+        if (insertListItemError) {
+            const logId = await logErrorReturnLogId("failed to insert list item", {
+                error: insertListItemError,
+            });
+            void sendAuditLog({ ...auditLog, wasSuccess: false, logId });
+            setErrorMessage(`Failed to add list item. Log ID: ${logId}`);
+            return;
+        }
+
+        void sendAuditLog({
+            ...auditLog,
+            wasSuccess: true,
+            listId: returnedListData.primary_key,
+        });
+    };
+
+    const editListItem = async (listItem: Partial<Schema["lists"]>): Promise<void> => {
+        const { data: returnedListData, error: updateListItemError } = await supabase
+            .from("lists")
+            .update(listItem)
+            .eq("primary_key", listItem.primary_key)
+            .select()
+            .single();
+
+        const auditLog = {
+            content: listItem ?? {},
+            action: "edit a list item",
+        } as const satisfies Partial<AuditLog>;
+
+        if (updateListItemError) {
+            const logId = await logErrorReturnLogId("failed to update list item", {
+                error: updateListItemError,
+            });
+            void sendAuditLog({ ...auditLog, wasSuccess: false, logId });
+            setErrorMessage(`Failed to update a list item. Log ID: ${logId}`);
+            return;
+        }
+        void sendAuditLog({
+            ...auditLog,
+            wasSuccess: true,
+            listId: returnedListData.primary_key,
+        });
+    };
+
     const onSubmit = async (): Promise<void> => {
         if (data === null) {
-            const { data: returnedListData, error: insertListItemError } = await supabase
-                .from("lists")
-                .insert(toSubmit)
-                .select()
-                .single();
-
-            const auditLog = {
-                content: { itemDetails: toSubmit },
-                action: "add a list item",
-            } as const satisfies Partial<AuditLog>;
-
-            if (insertListItemError) {
-                const logId = await logErrorReturnLogId("failed to insert list item", {
-                    error: insertListItemError,
-                });
-                void sendAuditLog({ ...auditLog, wasSuccess: false, logId });
-                setErrorMessage(`Failed to add list item. Log ID: ${logId}`);
-                return;
-            }
-
-            void sendAuditLog({
-                ...auditLog,
-                wasSuccess: true,
-                listId: returnedListData.primary_key,
-            });
+            void addListItem(toSubmit);
         } else {
-            const { data: returnedListData, error: updateListItemError } = await supabase
-                .from("lists")
-                .update(toSubmit)
-                .eq("primary_key", toSubmit.primary_key)
-                .select()
-                .single();
-
-            const auditLog = {
-                content: toSubmit ?? {},
-                action: "edit a list item",
-            } as const satisfies Partial<AuditLog>;
-
-            if (updateListItemError) {
-                const logId = await logErrorReturnLogId("failed to update list item", {
-                    error: updateListItemError,
-                });
-                void sendAuditLog({ ...auditLog, wasSuccess: false, logId });
-                setErrorMessage(`Failed to update a list item. Log ID: ${logId}`);
-                return;
-            }
-            void sendAuditLog({
-                ...auditLog,
-                wasSuccess: true,
-                listId: returnedListData.primary_key,
-            });
+            void editListItem(toSubmit);
         }
 
         setToSubmit({});
--- a/src/app/lists/EditModal.tsx
+++ b/src/app/lists/EditModal.tsx
@@ -81,12 +81,12 @@ const EditModal: React.FC<Props> = ({ data, onClose }) => {
                 const logId = await logErrorReturnLogId("failed to insert list item", {
                     error: insertListItemError,
                 });
-                await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
+                void sendAuditLog({ ...auditLog, wasSuccess: false, logId });
                 setErrorMessage(`Error: failedToInsertListItem. Log ID: ${logId}`);
                 return;
             }
 
-            await sendAuditLog({
+            void sendAuditLog({
                 ...auditLog,
                 wasSuccess: true,
                 listId: returnedListData.primary_key,
@@ -108,11 +108,11 @@ const EditModal: React.FC<Props> = ({ data, onClose }) => {
                 const logId = await logErrorReturnLogId("failed to update list item", {
                     error: updateListItemError,
                 });
-                await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
+                void sendAuditLog({ ...auditLog, wasSuccess: false, logId });
                 setErrorMessage(`Error: failedToUpdateListItem. Log ID: ${logId}`);
                 return;
             }
-            await sendAuditLog({
+            void sendAuditLog({
                 ...auditLog,
                 wasSuccess: true,
                 listId: returnedListData.primary_key,
--- a/src/app/lists/EditModal.tsx
+++ b/src/app/lists/EditModal.tsx
@@ -55,7 +55,7 @@ const listQuantityNoteAndLabels: [keyof Schema["lists"], keyof Schema["lists"], 
 ];
 
 const EditModal: React.FC<Props> = ({ data, onClose }) => {
-    const [toSubmit, setToSubmit] = useState<Partial<Schema["lists"]>>(data ?? {});
+    const [toSubmit, setToSubmit] = useState<Partial<Schema["lists"]> | null>(data ?? null);
 
     const [errorMessage, setErrorMessage] = useState<string | null>(null);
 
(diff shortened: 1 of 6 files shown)


minalui:
I can't seem to test for delete function - any ideas?

rickaw:
> I can't seem to test for delete function - any ideas?

It might be because the RLS policies are borked at the moment? Or something else? You could either wait for #261 to be merged or change the RLS policies to just `true` to see if it works then

src/app/lists/ListsPage.tsx
```diff
@@ -5,60 +5,92 @@ import { Schema } from "@/databaseUtils";
...
+                            quantity: row[`quantity_for_${key}` as keyof Schema["lists"]],
+                            notes: row[`notes_for_${key}` as keyof Schema["lists"]],
+                        },
+                    ])
+            ),
+        } as ListRow; // this cast is needed here as the type system can't infer what Object.fromEntries will return
+        return newRow;
+    });
+    return formattedListData;
+};
```
rickaw, line 42: nitpick: can be simplified a little
```suggestion
const formatListData = (listData: FetchedListsData): ListRow[] =>
    listData.data.map(
        (row): ListRow =>
            ({
                primaryKey: row.primary_key,
                rowOrder: row.row_order,
                itemName: row.item_name,
                ...Object.fromEntries(
                    listsHeaderKeysAndLabels
                        // .filter(([key]) => /^\d+$/.test(key))
                        .map(([key]) => [
                            key,
                            {
                                quantity: row[`quantity_for_${key}` as keyof Schema["lists"]],
                                notes: row[`notes_for_${key}` as keyof Schema["lists"]],
                            },
                        ])
                ),
            }) as ListRow // this cast is needed here as the type system can't infer what Object.fromEntries will return
    );
```

src/app/lists/ListsPage.tsx
```diff
@@ -5,60 +5,92 @@ import { Schema } from "@/databaseUtils";
...
+                    const logId = logErrorReturnLogId("Error with fetch: list data subscription", {
+                        error: error,
+                    });
+                    setListData([]);
+                    setErrorMessage(`Error fetching data, please reload. Log ID: ${logId}`);
+                }
+            })
+            .subscribe((status, err) => {
+                subscriptionStatusRequiresErrorMessage(status, err, "website_data") &&
+                    setErrorMessage("Error fetching data, please reload");
```
rickaw, line 79: I think this will be more readable than using the logical AND operator
```suggestion
                if (subscriptionStatusRequiresErrorMessage(status, err, "website_data")) {
                    setErrorMessage("Error fetching data, please reload");
                }
```

src/app/lists/ListDataview.tsx
```diff
@@ -123,12 +124,13 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
...
     setListOfIngredients,
     comment,
+    error,
 }) => {
     const [modal, setModal] = useState<EditModalState>();
     const [toDelete, setToDelete] = useState<number | null>(null);
     // need another setState otherwise the modal content changes before the close animation finishes
     const [toDeleteModalOpen, setToDeleteModalOpen] = useState<boolean>(false);
-    const [errorMsg, setErrorMsg] = useState<string | null>(null);
+    const [errorMessage, setErrorMessage] = useState<string | null>(error);
```
rickaw, line 133: So it seems like we manage the state of error messages in ListsPage and ListsDataView? Is this intended?

We should let ListsPage manage the state and pass `errorMessage` and `setErrorMessage` down to `ListsDataView` maybe?

src/app/lists/EditModal.tsx
```diff
@@ -55,25 +57,70 @@ const listQuantityNoteAndLabels: [keyof Schema["lists"], keyof Schema["lists"],
...
-        const table = supabase.from("lists");
-        const { error } =
-            data === null
-                ? await table.insert(toSubmit)
-                : await table.update(toSubmit).eq("primary_key", toSubmit.primary_key);
-
-        if (error) {
-            setErrorMsg(error.message);
+        if (data === null) {
+            const { data: returnedListData, error: insertListItemError } = await supabase
```
rickaw, line 69: Should we pull the insert and update into two separate methods here so that we have something like
```ts
if (data === null) {
    addListItem(listItem)
} else {
    updateListItem(listItem)
}
```

src/app/lists/EditModal.tsx
```diff
@@ -55,25 +57,70 @@ const listQuantityNoteAndLabels: [keyof Schema["lists"], keyof Schema["lists"],
...
+            if (insertListItemError) {
+                const logId = await logErrorReturnLogId("failed to insert list item", {
+                    error: insertListItemError,
+                });
+                await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
+                setErrorMessage(`Error: failedToInsertListItem. Log ID: ${logId}`);
+                return;
+            }
+
+            await sendAuditLog({
```
rickaw, line 89: Since the subsequent operations don't need to wait for this to finish...
```suggestion
            void sendAuditLog({
```

src/app/lists/EditModal.tsx
```diff
@@ -55,25 +57,70 @@ const listQuantityNoteAndLabels: [keyof Schema["lists"], keyof Schema["lists"],
...
+
+            if (updateListItemError) {
+                const logId = await logErrorReturnLogId("failed to update list item", {
+                    error: updateListItemError,
+                });
+                await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
+                setErrorMessage(`Error: failedToUpdateListItem. Log ID: ${logId}`);
+                return;
+            }
+            await sendAuditLog({
```
rickaw, line 115:
```suggestion
            void sendAuditLog({
```

src/app/lists/EditModal.tsx
```diff
@@ -55,25 +57,70 @@ const listQuantityNoteAndLabels: [keyof Schema["lists"], keyof Schema["lists"],
...
+                content: toSubmit ?? {},
+                action: "edit a list item",
+            } as const satisfies Partial<AuditLog>;
+
+            if (updateListItemError) {
+                const logId = await logErrorReturnLogId("failed to update list item", {
+                    error: updateListItemError,
+                });
+                await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
+                setErrorMessage(`Error: failedToUpdateListItem. Log ID: ${logId}`);
```
rickaw, line 112: This is a user-facing message right?
```suggestion
                setErrorMessage(`Failed to update a list item. Log ID: ${logId}`);
```

src/app/lists/EditModal.tsx
```diff
@@ -55,25 +57,70 @@ const listQuantityNoteAndLabels: [keyof Schema["lists"], keyof Schema["lists"],
...
+                content: { itemDetails: toSubmit },
+                action: "add a list item",
+            } as const satisfies Partial<AuditLog>;
+
+            if (insertListItemError) {
+                const logId = await logErrorReturnLogId("failed to insert list item", {
+                    error: insertListItemError,
+                });
+                await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
+                setErrorMessage(`Error: failedToInsertListItem. Log ID: ${logId}`);
```
rickaw, line 85:
```suggestion
                setErrorMessage(`Failed to add a list item. Log ID: ${logId}`);
```

src/app/lists/EditModal.tsx
```diff
@@ -55,25 +57,70 @@ const listQuantityNoteAndLabels: [keyof Schema["lists"], keyof Schema["lists"],
...
+                return;
+            }
+            await sendAuditLog({
+                ...auditLog,
+                wasSuccess: true,
+                listId: returnedListData.primary_key,
+            });
         }
+
+        setToSubmit({});
```
rickaw, line 122: I think we want to set this to null when there's nothing to submit?

rickaw (changes requested):

src/app/lists/EditModal.tsx
```diff
@@ -105,7 +163,7 @@ const EditModal: React.FC<Props> = ({ data, onClose }) => {
             <ModalInner>
                 <h3>Description</h3>
                 <TextInput
-                    defaultValue={toSubmit.item_name ?? ""}
+                    defaultValue={toSubmit ? toSubmit.item_name : ""}
```
rickaw, line 166:
```suggestion
                    defaultValue={toSubmit?.item_name ?? ""}
```

src/app/lists/EditModal.tsx
```diff
@@ -115,12 +173,12 @@ const EditModal: React.FC<Props> = ({ data, onClose }) => {
                             <h3>{label}</h3>
                             <DataWithTooltipDiv>
                                 <TextInput
-                                    defaultValue={toSubmit[quantityKey]?.toString() ?? ""}
+                                    defaultValue={toSubmit ? toSubmit[quantityKey]?.toString() : ""}
```
rickaw, line 176: Are you aware that this expression is either string or undefined since `toSubmit[quantityKey]?.toString()` can be undefined due to optional chaining? I'm guessing this makes no diff in the behaviour but just wanted to make sure you're aware

src/app/lists/ListDataview.tsx
```diff
@@ -123,12 +125,13 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
     listOfIngredients,
     setListOfIngredients,
     comment,
+    error,
+    handleSetError,
```
rickaw, line 129: We should probably make it clear that this is a user-facing error message, not some error object for internal use.
```suggestion
    errorMessage,
    setErrorMessage,
```

src/app/lists/ListDataview.tsx
```diff
@@ -203,15 +206,32 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
...
-                setErrorMsg(error.message + `Error ID: ${logId}`);
-            } else {
-                window.location.reload();
+                await sendAuditLog({
+                    ...auditLog,
+                    wasSuccess: false,
+                    logId,
+                    listId: itemToDelete.primaryKey,
+                });
+                handleSetError(error.message + `Error ID: ${logId}`);
```
rickaw, line 228: We probably won't want to rely on the error message coming out of Supabase here because we don't know ho
```suggestion
                handleSetError(`Failed to delete a list item. Log ID: ${logId}`);
```

src/app/lists/ListDataview.tsx
```diff
@@ -203,15 +206,32 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
...
+                return;
             }
+
+            await sendAuditLog({ ...auditLog, wasSuccess: true });
+            setToDelete(null);
+            setToDeleteModalOpen(false);
```
rickaw, line 234: I'm not sure if this makes any difference, but conceptually it makes more sense to close the modal without changing its content, and then setting the parcel to be deleted to `null`
```suggestion
            setToDeleteModalOpen(false);
            setToDelete(null);
```

src/app/lists/ListsPage.tsx
```diff
@@ -5,60 +5,95 @@ import { Schema } from "@/databaseUtils";
...
 const ListsPage: React.FC<{}> = () => {
     const [isLoading, setIsLoading] = useState(true);
-    const [listsData, setListsData] = useState<ListRow[]>([]);
+    const [listData, setListData] = useState<ListRow[]>([]);
     const [comment, setComment] = useState("");
+    const [errorMessage, setErrorMessage] = useState<string | null>(null);
 
+    function handleSetError(error: string | null): void {
+        error ? setErrorMessage(error) : setErrorMessage(null);
+    }
```
rickaw, line 50: Isn't it just this basically? (the child shouldn't be setting the error message to `""` anyway). And if this is the case, we can just pass `setErrorMessage` as a prop instead of defining a whole other function.
```suggestion
    function handleSetError(error: string | null): void {
        setErrorMessage(error)
    }
```

src/app/lists/ListsPage.tsx
```diff
@@ -5,60 +5,95 @@ import { Schema } from "@/databaseUtils";
...
-            setComment(fetchedData.comment);
+            const listData = await fetchListData();
+            setListData(formatListData(listData));
+            setComment(listData.comment);
             setIsLoading(false);
         })();
     }, []);
 
+    useEffect(() => {
+        // This requires that the DB table has Realtime turned on
```
rickaw, line 62:
```suggestion
```

src/app/lists/ListsPage.tsx
```diff
@@ -5,60 +5,95 @@ import { Schema } from "@/databaseUtils";
...
+        const subscriptionChannel = supabase
+            .channel("lists-table-changes")
+            .on("postgres_changes", { event: "*", schema: "public", table: "lists" }, async () => {
+                setErrorMessage(null);
```
rickaw, line 66: I'm not sure how I feel about clearing the message every time there's a change in db. What do you think?

src/app/lists/ListsPage.tsx
```diff
@@ -5,60 +5,95 @@ import { Schema } from "@/databaseUtils";
...
+                try {
+                    const listData = await fetchListData();
+                    setListData(formatListData(listData));
```
rickaw, line 69: Can we write this using our new style of returning errors rather than throwing errors if possible?

rickaw (changes requested):

src/app/lists/ListsPage.tsx
minalui: Affects shopping lists pdf so parked for now

PR #265, opened 2024-04-15
VFB-197: Add audit log to delete user action

minalui:
## What's changed
- Add audit log when user is deleted

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| paste_here | paste_here |

## Checklist

If you have made any changes to the database... N/A
    - Checkout the dev branch
    - Checkout your branch

### DIFF
--- a/src/app/admin/deleteUser/DeleteUserDialog.tsx
+++ b/src/app/admin/deleteUser/DeleteUserDialog.tsx
@@ -6,10 +6,8 @@ import styled from "styled-components";
 import Modal from "@/components/Modal/Modal";
 import OptionButtonsDiv from "@/app/admin/common/OptionButtonsDiv";
 import { SetAlertOptions } from "@/app/admin/common/SuccessFailureAlert";
-import { logErrorReturnLogId, logInfoReturnLogId } from "@/logger/logger";
+import { logInfoReturnLogId } from "@/logger/logger";
 import { adminDeleteUser } from "@/server/adminDeleteUser";
-import { AuditLog, sendAuditLog } from "@/server/auditLog";
-import supabase from "@/supabaseClient";
 
 const DangerDialog = styled(Modal)`
     .MuiPaper-root > div:first-child {
@@ -38,45 +36,12 @@ const DeleteUserDialog: React.FC<Props> = (props) => {
             return;
         }
 
-        const { data: profileUserId, error: profileUserIdError } = await supabase
-            .from("profiles")
-            .select("primary_key")
-            .eq("user_id", props.userToDelete.id)
-            .single();
-
-        if (profileUserIdError) {
-            const logId = await logErrorReturnLogId(
-                `Failed to fetch user id from profiles. ${props.userToDelete.id}`,
-                { error: profileUserIdError }
-            );
-            props.setAlertOptions({
-                success: false,
-                message: <>{`Failed to get profile user id. Log ID: ${logId}`}</>,
-            });
-            props.setUserToDelete(null);
-            return;
-        }
-
         const { error: deleteUserError } = await adminDeleteUser(props.userToDelete.id);
 
-        const auditLog = {
-            action: "delete a user",
-            content: {
-                userId: props.userToDelete.id,
-                role: props.userToDelete.userRole,
-            },
-            profileId: profileUserId.primary_key ?? "",
-        } as const satisfies Partial<AuditLog>;
-
         if (deleteUserError) {
-            const logId = await logErrorReturnLogId(
-                `Error with delete: User ${props.userToDelete.email}`,
-                { error: deleteUserError }
-            );
-            void sendAuditLog({ ...auditLog, wasSuccess: false, logId });
             props.setAlertOptions({
                 success: false,
-                message: <>{`Delete User Operation Failed. Log ID: ${logId}`}</>,
+                message: <>{`${deleteUserError.type}. Log ID: ${deleteUserError.logId}`}</>,
             });
         } else {
             props.setAlertOptions({
@@ -87,7 +52,6 @@ const DeleteUserDialog: React.FC<Props> = (props) => {
                     </>
                 ),
             });
-            void sendAuditLog({ ...auditLog, wasSuccess: true });
             void logInfoReturnLogId(`${props.userToDelete?.email} deleted successfully.`);
         }
 
--- a/src/app/admin/deleteUser/DeleteUserDialog.tsx
+++ b/src/app/admin/deleteUser/DeleteUserDialog.tsx
@@ -26,6 +26,19 @@ interface Props {
     setAlertOptions: SetAlertOptions;
 }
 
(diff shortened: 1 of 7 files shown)


src/app/admin/deleteUser/DeleteUserDialog.tsx
```diff
@@ -36,29 +38,59 @@ const DeleteUserDialog: React.FC<Props> = (props) => {
             return;
         }
 
-        const { error } = await adminDeleteUser(props.userToDelete.id);
+        const { data: profileUserId, error: profileUserIdError } = await supabase
```
rickaw, line 41: I wonder if we should get the profile id and send the audit log within `adminDeleteUser` instead?

minalui:
I've added the subscription to the profiles table so the user table updates automatically. I've also removed the refresh button. However, the create user form remains populated after the submit button (and the success message is displayed instead of a submit button) - just wondering if the form should be cleared and submit button remain?

rickaw:
> I've added the subscription to the profiles table so the user table updates automatically. I've also removed the refresh button. However, the create user form remains populated after the submit button (and the success message is displayed instead of a submit button) - just wondering if the form should be cleared and submit button remain?
> 
> image

I think it's fine as is for now!

src/app/admin/deleteUser/DeleteUserDialog.tsx
```diff
@@ -36,9 +36,14 @@ const DeleteUserDialog: React.FC<Props> = (props) => {
...
         }
 
-        const { error } = await adminDeleteUser(props.userToDelete.id);
+        const { error: deleteUserError } = await adminDeleteUser(props.userToDelete.id);
 
-        if (!error) {
+        if (deleteUserError) {
+            props.setAlertOptions({
+                success: false,
+                message: <>{`${deleteUserError.type}. Log ID: ${deleteUserError.logId}`}</>,
```
rickaw, line 44: since `deleteUserError.type` is meant to be just an identifier of the error, we should compute the user friendly error message based on that and display it there :)

src/app/admin/deleteUser/DeleteUserDialog.tsx
```diff
@@ -48,17 +53,8 @@ const DeleteUserDialog: React.FC<Props> = (props) => {
                 ),
             });
             void logInfoReturnLogId(`${props.userToDelete?.email} deleted successfully.`);
-        } else {
-            const logId = await logErrorReturnLogId(
-                `Error with delete: User ${props.userToDelete.email}`,
-                { error }
-            );
```
rickaw, line 55: We should still log an error

src/app/admin/usersTable/UsersTable.tsx
```diff
@@ -230,6 +231,26 @@ const UsersTable: React.FC = () => {
         void fetchAndDisplayUserData();
     }, [fetchAndDisplayUserData]);
 
+    useEffect(() => {
+        void fetchAndDisplayUserData();
+        // This requires that the profiles tables have Realtime turned on
```
rickaw, line 236: I don't think this comment is needed anymore as we manage realtime through migrations

src/app/admin/usersTable/UsersTable.tsx
```diff
@@ -230,6 +231,26 @@ const UsersTable: React.FC = () => {
         void fetchAndDisplayUserData();
     }, [fetchAndDisplayUserData]);
 
+    useEffect(() => {
+        void fetchAndDisplayUserData();
```
rickaw, line 235: why do we need to fetch here? I thought this useEffect should just run once at the beginning to try and subscribe to the db changes?

src/app/admin/usersTable/UsersTable.tsx
```diff
@@ -230,6 +231,26 @@ const UsersTable: React.FC = () => {
...
+            )
+            .subscribe((status, err) => {
+                subscriptionStatusRequiresErrorMessage(status, err, "website_data") &&
+                    setErrorMessage("Error fetching users data, please reload");
+            });
+
+        return () => {
+            supabase.removeChannel(subscriptionChannel);
+        };
+    }, [startIndex, endIndex, primaryFilters, sortState]);
```
rickaw, line 252: > I thought this useEffect should just run once at the beginning to try and subscribe to the db changes?
```suggestion
    }, [fetchAndDisplayUserData]);
```

src/server/adminInviteUser.ts
```diff
@@ -88,11 +88,11 @@ export async function adminInviteUser(
     const { data: profileData, error: createProfileError } = await supabase
         .from("profiles")
         .insert({
-            primary_key: newUserData.user.id,
             role: userDetails.role,
             first_name: userDetails.firstName,
             last_name: userDetails.lastName,
             telephone_number: userDetails.telephoneNumber,
+            user_id: newUserData.user.id,
```
rickaw, line 95: Ooh, was this another user_id related bug?

src/server/adminDeleteUser.ts
```diff
@@ -2,30 +2,78 @@
...
+import { AuditLog, sendAuditLog } from "@/server/auditLog";
+import { getSupabaseServerComponentClient } from "@/supabaseServer";
 
-export type DeleteUserErrorType = {
-    error: Record<string, string> | null;
-};
+type DeleteUserErrorType =
+    | "Failed to authenticate as admin"
+    | "Failed to get user id from profiles"
+    | "Failed to delete user";
```
rickaw, line 12: These tend to be camelCase or all lowercase, since they serve as a way for the code to differentiate what kind of error it is returning, not a user friendly error message. (This function shouldn't be concerned about what error message is displayed, but about returning the error data back to the caller)

e.g. `"failedToAuthenticateAsAdmin"`

src/server/adminDeleteUser.ts
```diff
@@ -2,30 +2,78 @@
...
 
-export async function adminDeleteUser(userId: string): Promise<DeleteUserErrorType> {
+type DeleteUserError =
```
rickaw, line 14: nitpick: this whole object is a result, which may or may not have an error inside
```suggestion
type DeleteUserResult =
```

rickaw (commented):
Nice! a few comments

src/app/admin/deleteUser/DeleteUserDialog.tsx
minalui: I've sent a log within adminDeleteUser and returned the logID to be displayed

src/server/adminInviteUser.ts
minalui: Yes - it wasn't making the connection to users table and not displaying the data in the users table

rickaw: oh cool - let's merge this asap so we can include it in today's release

rickaw (approved):

minalui:
Can you please double check the switch because it required to return a default value.

Issue #191, opened 2024-03-25
Add date+time stamp of when parcel added to a client (separate to packing date)


dan-langridge:
### Describe the solution / feature you'd like

When I open the client modal to see the history of the parcels that client has received, it would be helpful to see a date and timestamp of when the parcel was originally added for that client.

This is different to the Packing Date, as the parcel could get moved to a few different dates on the app during the lifecycle of that parcel, but it's helpful to know when it was originally entered.

example: life cycle of a chaotic parcel
10th - Added to App for 11th packing shift
11th - Called, can't receive delivery today, move to 12th
12th - attempted delivery, not home, called, will be home tomorrow, move to 13th
13th - another failed delivery, called, moved to 15th
15th - successful delivery

In the current app, we wouldnt know when that parcel was added originally to the app.

rickaw:
## Developer Notes
- Add `created_at` field in the parcels table, to be set to the current time when a parcel is inserted automatically
- Make necessary chanegs to display the timestamp in the following places
  - on the parcels table as an additional column
  - on the parcel detail modal as an additional field "Created at"
- In the generated seed data, manually set this to noon on 31 Dec 2023 (this is before all parcel events) so it visually makes sense

minalui:
QA passed 👍

PR #269, opened 2024-04-16
#191: Add parcel creation timestamp

rickaw:
## What's changed
Add a timestamp for when the parcel is created.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |
| image | image |

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/parcels/ExpandedParcelDetails.tsx
```diff
@@ -1,7 +1,13 @@
-import React from "react";
+"use client";
```
rickaw, line 1: This fixes the "can't use async component in a client component" issue

src/app/parcels/ParcelsPage.cy.tsx
```diff
@@ -162,25 +126,6 @@ describe("Parcels Page", () => {
     });
 
     describe("Backend Processing for Expanded Parcel Details", () => {
-        it("Fields are set correctly", () => {
-            const expandedClientDetails = rawDataToExpandedParcelData(
```
rickaw, line 166: This test is no longer relevant

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -1,19 +1,44 @@
...
+          parcelDetails: null;
+          error: FetchExpandedParcelDetailsError;
+      };
+
+export interface FetchExpandedParcelDetailsError {
+    type: FetchExpandedParcelDetailsErrorType;
+    logId: string;
+}
+
+export type FetchExpandedParcelDetailsErrorType =
```
rickaw, line 28: no export

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -138,72 +151,6 @@ export const formatDatetimeAsTime = (datetime: string | null): string => {
...
-            childDetails.push(`${age}-year-old ${familyMember.gender}`);
-        }
-    }
-
-    if (childDetails.length === 0) {
-        return "No Children";
-    }
-
-    return childDetails.join(", ");
-};
```
rickaw, line 205: The exact copy of these functions already exist in `getExpandedClientDetails.ts`

src/app/parcels/getParcelsTableData.ts
```diff
@@ -72,6 +73,7 @@ export const processingDataToParcelsTableData = async (
                 flaggedForAttention: parcel.client_flagged_for_attention ?? false,
                 requiresFollowUpPhoneCall: parcel.client_signposting_call_required ?? false,
             },
+            createdAt: parcel.created_at ? new Date(parcel.created_at) : null,
```
rickaw, line 76: use map?

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -47,35 +72,56 @@ export const getRawParcelDetails = async (parcelId: string) => {
...
+        return {
+            parcelDetails: null,
+            error: {
+                type: "clientDetailDoesNotExist",
+                logId,
+            },
+        };
     }
 
-    return new Date(datetime).toLocaleDateString("en-GB");
```
rickaw, line 78: this has been moved to `format.ts`

supabase/migrations/20240415132809_add_creation_timestamp_to_parcels.sql
```diff
@@ -0,0 +1,33 @@
...
+    clients.phone_number AS client_phone_number,
+    family_count.family_count,
+    last_status.event_name AS last_status_event_name,
+    last_status.event_data AS last_status_event_data,
+    last_status."timestamp" AS last_status_timestamp,
+    last_status.workflow_order AS last_status_workflow_order
+   FROM (((((parcels
+     LEFT JOIN collection_centres ON ((parcels.collection_centre = collection_centres.primary_key)))
+     LEFT JOIN clients ON ((parcels.client_id = clients.primary_key)))
+     LEFT JOIN packing_slots ON ((parcels.packing_slot = packing_slots.primary_key)))
```
rickaw, line 30: These three used to be `JOIN` but changed to `LEFT JOIN` as we want all parcels to be in parcels_plus whether they have a corresponding collection centre, client or packing slots or not.

supabase/migrations/20240415132809_add_creation_timestamp_to_parcels.sql
```diff
@@ -0,0 +1,33 @@
+alter table "public"."parcels" add column "created_at" timestamp with time zone not null default (now() AT TIME ZONE 'utc'::text);
+
+update "public"."parcels" set "created_at"='2023-12-31 12:00:00+00';
+
+drop view if exists "public"."parcels_plus";
+
+create or replace view "public"."parcels_plus" as  SELECT parcels.primary_key AS parcel_id,
+    parcels.collection_datetime,
+    parcels.packing_date,
+    parcels.created_at,
```
rickaw, line 10: Added this.

src/app/parcels/ExpandedParcelDetails.tsx
```diff
@@ -11,20 +17,56 @@ const sortByTimestampWithMostRecentFirst = (events: EventTableRow[]): EventTable
...
+            setErrorMessage(newErrorMessage);
+            return;
+        }
+
+        setParcelDetails(expandedParcelDetails);
+    }, [parcelId]);
+
+    useEffect(() => {
+        void fetchAndSetParcelDetails();
+    }, [fetchAndSetParcelDetails]);
```
AmeliaIsCoding, line 54: just for my own educational benefit, can you please explain to me why we would want to do this useCallback thing?

src/app/parcels/ExpandedParcelDetails.tsx
```diff
@@ -11,20 +17,56 @@ const sortByTimestampWithMostRecentFirst = (events: EventTableRow[]): EventTable
...
 
     return (
         <>
-            <DataViewer data={expandedParcelDetails.expandedParcelData} />
-            <EventTable
-                tableData={sortByTimestampWithMostRecentFirst(expandedParcelDetails.events)}
-            />
+            {errorMessage && <ErrorLargeText>{errorMessage}</ErrorLargeText>}
```
AmeliaIsCoding, line 58: just to point out I was using `<ErrorSecondaryText>` in the action modals - should we choose one to be consistent? Don't mind which

src/app/parcels/ParcelsPage.cy.tsx
AmeliaIsCoding: is this because the data is got from the parcels_plus view?

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -1,19 +1,44 @@
...
 import { logErrorReturnLogId } from "@/logger/logger";
-export type RawParcelDetails = Awaited<ReturnType<typeof getRawParcelDetails>>;
+import {
+    formatAddressFromClientDetails,
+    formatBreakdownOfChildrenFromFamilyDetails,
+    formatHouseholdFromFamilyDetails,
+} from "@/app/clients/getExpandedClientDetails";
+import { formatDateTime, formatDatetimeAsDate } from "@/common/format";
+
+type FetchExpandedParcelDetailsResult =
```
AmeliaIsCoding, line 13: I've been using "...Response" instead of "...Result", again should we choose one to be consistent, don't mind which

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -47,35 +72,56 @@ export const getRawParcelDetails = async (parcelId: string) => {
...
+                logId,
+            },
+        };
     }
 
-    return new Date(datetime).toLocaleDateString("en-GB");
+    return {
+        parcelDetails: {
+            expandedParcelData: {
+                voucherNumber: rawParcelDetails.voucher_number ?? "",
```
AmeliaIsCoding, line 109: is it possible for rawParcelDetails to be null? if so should we return a specific error?

supabase/migrations/20240415132809_add_creation_timestamp_to_parcels.sql
```diff
@@ -0,0 +1,33 @@
+alter table "public"."parcels" add column "created_at" timestamp with time zone not null default (now() AT TIME ZONE 'utc'::text);
+
+update "public"."parcels" set "created_at"='2023-12-31 12:00:00+00';
```
AmeliaIsCoding, line 3: Maybe I'm confused as to the point in which this file runs, but should this be  `....WHERE "created_at" IS NULL;`?

supabase/migrations/20240415132809_add_creation_timestamp_to_parcels.sql
AmeliaIsCoding: is the number of entries in parcels_plus still as expected (think it should be 5000?)? I think I might have experienced problems with these different types of joins when originally making the view. But I can't remember the specifics so happy with it if it works / results in right number of parcels

supabase/migrations/20240415132809_add_creation_timestamp_to_parcels.sql
```diff
@@ -0,0 +1,33 @@
+alter table "public"."parcels" add column "created_at" timestamp with time zone not null default (now() AT TIME ZONE 'utc'::text);
```
AmeliaIsCoding, line 1: does this default value result in the right time in terms of time zones? (time zones confuse me)

AmeliaIsCoding (commented):
Looks good, just had a few small questions

src/app/parcels/ExpandedParcelDetails.tsx
rickaw: The main reason is to make the useEffect simple and easy to read. If we instead had the content of `fetchAndSetParcelDetails` inside the use effect, then future devs would have to read the content to understand what happens in the use effect.

src/app/parcels/ExpandedParcelDetails.tsx
rickaw: happy to use ErrorSecondaryText! changed now.

src/app/parcels/ParcelsPage.cy.tsx
rickaw: This test specifically tests `rawDataToExpandedParcelData`, which we don't have anymore. (Though we could spend some effort to stub out the supabase calls to test effectively the same thing, but probably not worth the effort for now)

supabase/migrations/20240415132809_add_creation_timestamp_to_parcels.sql
rickaw: parcels_plus still has 5000 items 👍 

`LEFT_JOIN` means we retain every row from the `parcels` table in this context, so we're guaranteeing that there will be 5000 items this way.

`JOIN` means we retain rows from the `parcels` table where the `ON` condition is matched, so we may end up with less than 5000 items. However, we happen to have 5000 items anyway because every parcel has a collection centre, a client and a packing slot.

src/app/parcels/getExpandedParcelDetails.ts
rickaw: idm which either, I think both exist in the code base now. Slight preference on Result because Response would imply it's the response for some sort of request (the data coming directly out of a supabase call would be a response, but not after we do a potential data manipulation / make multiple calls).

We can try standardise to Result as we go along.

supabase/migrations/20240415132809_add_creation_timestamp_to_parcels.sql
rickaw: This runs when the migration happens. When this is merged to dev for example, the outstanding migrations (just this file) runs against the dev db, which would
- create the created_at column
- set created_at to a certain time.

Since we're creating the column just before, we don't need to worry about `....WHERE "created_at" IS NULL;`

src/app/parcels/ExpandedParcelDetails.tsx
AmeliaIsCoding: Thought so, nice, thank you!

src/app/parcels/getExpandedParcelDetails.ts
AmeliaIsCoding: Okay cool!

supabase/migrations/20240415132809_add_creation_timestamp_to_parcels.sql
AmeliaIsCoding: makes sense, thanks

supabase/migrations/20240415132809_add_creation_timestamp_to_parcels.sql
AmeliaIsCoding: niice

AmeliaIsCoding (approved):

src/app/parcels/getParcelsTableData.ts
```diff
@@ -46,36 +45,31 @@ export const processingDataToParcelsTableData = async (
         );
```
minalui, line 45: Should we return and display the error?

src/app/parcels/ParcelsPage.tsx
minalui, line 1: Also line 226

minalui (commented):
Looks good I don't know if you want to handle some errors that are being thrown as well 🙊

src/app/parcels/getParcelsTableData.ts
rickaw: Good point, done!

src/app/parcels/ParcelsPage.tsx
rickaw: This one would probs conflict with Amelia's big ticket so happy to leave this for now!

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -190,9 +230,14 @@ export const getParcelsByIds = async (
     }
 
     const congestionCharge = await getCongestionChargeDetailsForParcels(data, supabase);
-    const formattedData = processingDataToParcelsTableData(data, congestionCharge);
+    const { parcelTableRows, error: processParcelDataError } =
+        await processingDataToParcelsTableData(data, congestionCharge);
+
+    if (processParcelDataError) {
+        throw new Error("Failed to process parcels.", { cause: processParcelDataError });
```
rickaw, line 237: I've had to put move `processingDataToParcelsTableData`'s throw here to keep the current behaviour without a big conflict with Amelia 😂

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -86,37 +99,107 @@ const getParcelProcessingData = async (
...
+
+    if (getDbParcelsError) {
+        let errorType: GetParcelDataAndCountErrorType;
+        switch (getDbParcelsError.type) {
+            case "abortedFetch":
+                errorType = "abortedFetch";
+                break;
+            case "failedToFetchParcelTable":
+                errorType = "failedToFetchParcels";
+                break;
```
minalui, line 163: I don't know if this feels a little redundant to me 😕

rickaw:
note to self: make sure the parcels_plus is correct after Mina's change

PR #278, opened 2024-04-18
VFB-198: Make it not possible to delete self

AmeliaIsCoding:
## What's changed
It is no longer possible to delete yourself; the bin icon that allows a user to be deleted in the Users Table in the Admin Page is not displayed in the row representing the currently logged in user.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image|image|

## Checklist

No DB changes.


src/app/admin/usersTable/UsersTable.tsx
```diff
@@ -224,10 +226,20 @@ const UsersTable: React.FC = () => {
                     ${error.logId}`);
                         break;
                 }
-            } else {
-                setUsers(data.userData);
-                setFilteredUsersCount(data.count);
+                return;
```
rickaw, line 229: Does this mean if there's any errors in the fetch, we return from fetchAndDisplayUserData without setting isLoading to false etc. (line 244 & 245)?

src/app/admin/usersTable/UsersTable.tsx
```diff
@@ -224,10 +226,20 @@ const UsersTable: React.FC = () => {
...
             }
+            setUsers(data.userData);
+            setFilteredUsersCount(data.count);
+
+            const { data: currentUser, error: currentUserError } = await getCurrentUser();
+            if (currentUserError) {
+                setErrorMessage(
+                    `Error occured when fetching current user: ${currentUserError.type}, Log ID: ${currentUserError.logId}`
+                );
+                return;
```
rickaw, line 239: Same here - if there's any errors in getCurrentUser, we return from fetchAndDisplayUserData without setting isLoading to false etc. (line 244 & 245)?

src/components/Tables/Table.tsx
```diff
@@ -307,44 +308,51 @@ const Table = <Data,>({
...
-                        </StyledIconButton>
-                    )}
-                </EditAndReorderArrowDiv>
-            ),
+            cell: (row: Row<Data>) => {
+                const isRowDeletable =
+                    editableConfig.onDelete &&
+                    ("id" in row.data && typeof row.data.id === "string"
+                        ? !editableConfig.undeletableIds?.includes(row.data.id)
+                        : true);
```
rickaw, line 316: Instead of checking the "id" field in the data internally, why don't we 
- define a prop `isDeletable: (row: Data) => boolean`
- use that prop to determine if the row is deletable
- let the parent define which row is deletable

This gives us the flexibility to define which one is deletable in the parent in any way we want, not just using the "id" field.

src/server/getCurrentUser.ts
```diff
@@ -4,14 +4,35 @@ import { getSupabaseServerComponentClient } from "@/supabaseServer";
...
     if (error) {
-        void logErrorReturnLogId("error with auth getUser", error);
+        const logId = await logErrorReturnLogId("error with auth getUser", error);
+        return { data: null, error: { type: "userFetchFailed", logId: logId } };
+    }
+
+    if (data === null) {
+        const logId = await logErrorReturnLogId("error with auth getUser: no matching users");
+        return { data: null, error: { type: "noMatchingUsers", logId: logId } };
     }
```
rickaw, line 35: Is this really needed? I think data is guaranteed to be not null if error is null?

src/app/admin/usersTable/UsersTable.tsx
AmeliaIsCoding: You're right. That's not what we want to do. Have fixed

src/app/admin/usersTable/UsersTable.tsx
AmeliaIsCoding: Also right, have fixed

src/components/Tables/Table.tsx
AmeliaIsCoding: that makes a lot of sense, thank you! have done

src/server/getCurrentUser.ts
AmeliaIsCoding: You're right, have removed

rickaw (commented):
the security changes have been reverted so pls resolve the conflicts :) sorry!

src/components/Tables/Table.tsx
```diff
@@ -307,44 +308,49 @@ const Table = <Data,>({
...
+                                onClick={() => swapRows(row.rowId, false)}
+                                aria-label="reorder row downwards"
+                                disabled={isSwapping}
+                            >
+                                <StyledIcon icon={faAnglesDown} />
+                            </StyledIconButton>
+                        )}
+                        {isRowDeletable && (
+                            <StyledIconButton
+                                onClick={() => editableConfig.onDelete!(row.rowId)}
```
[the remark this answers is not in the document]
AmeliaIsCoding: Typescript still doesn't realise that it's not undefined
image
image
Should I make the change of doing && editableConfig.onDelete, but keep the null coercion?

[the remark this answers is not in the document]
rickaw: ah weird, let's do this instead to be extra safe

```ts
{editableConfig.onDelete && isRowDeletable && (
    <StyledIconButton
        onClick={() => {
            if (editableConfig.onDelete) {
                editableConfig.onDelete(row.rowId);
            }
        }}
        aria-label="delete"
    >
        <StyledIcon icon={faTrashAlt} />
    </StyledIconButton>
)}
```

rickaw (approved):

PR #284, opened 2024-04-19
VFB-191 fix: Handle errors from QA

minalui:
## What's changed
- Declare audit log with itemName before call to supabase to avoid undefined ('reading itemName') when deleting an item
- Add list_id to edit audit log
- Change logic to avoid closing modal when error occurs

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| paste_here | image |

## Checklist

If you have made any changes to the database... N/A
    - Checkout the dev branch
    - Checkout your branch


src/app/lists/EditModal.tsx
```diff
@@ -126,11 +137,24 @@ const EditModal: React.FC<Props> = ({ data, onClose }) => {
...
+            }
         } else {
-            void editListItem(toSubmit);
+            const { error } = await editListItem(toSubmit);
+            if (error) {
+                switch (error.type) {
+                    case "failedToEditListItem":
+                        setErrorMessage(`Failed to edit list item. Log ID: ${error.logId}`);
+                }
+                return;
```
minalui, line 153: Return from function to prevent calling setToSubmit which closes the modal

src/app/lists/ListDataview.tsx
```diff
@@ -214,6 +210,11 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
                 },
             } as const satisfies Partial<AuditLog>;
 
+            const { error } = await supabase
+                .from("lists")
+                .delete()
+                .eq("primary_key", itemToDelete.primaryKey);
```
minalui, line 216: Moved this to avoid audit log calling itemToDelete.itemName after the item has been deleted

rickaw: I"m not sure if this was causing any problems... but this new version is also fine? haha

src/app/lists/EditModal.tsx
```diff
@@ -21,6 +21,11 @@ interface Props {
 // null => add, undefined => modal closed
 export type EditModalState = Schema["lists"] | null | undefined;
 
+type AddEditListErrorTypes = "failedToAddListItem" | "failedToEditListItem";
+type AddEditListReturn = {
+    error: { type: AddEditListErrorTypes; logId: string } | null;
+};
```
rickaw, line 27: I feel like it'd make sense to have separate types for add and edit operations. The caller of addListItem shouldn't be expecting `"failedToEditListItem"` as a possible error type right?

src/app/lists/ListDataview.tsx
minalui: This works now 👀

rickaw (approved):

Issue #188, opened 2024-03-25
Client Postcode - Amend Response Validation to allow "NFA"


dan-langridge:
### Reproduction steps

When I:
- Add a new client who is homeless and doesn't have an address, I give them the postcode NFA (No Fixed Abode)
- Error : Please enter a valid entry.

### Expected behaviour

A client who is homeless and doesn't have an address, in the address field we would just give them the postcode NFA (No Fixed Abode). This is always for clients who are collecting their parcel from a Collection Centre.

I see 2 solutions:

Easy) Change/Remove the Response Validation for Postcode to allow for free text (NFA)

Less Easy) Add a tick button to say "No Address" which then removes the address fields and automatically adds 'NFA' onto the app, and shopping list, shipping label, driver overview etc.

We are aware that using NFA as a postcode breaks the "Generate Map" feature but it's something we know and know not to do....

### What browsers are you seeing the problem on?

Safari

### What OS is your computer running on?

macOS

rickaw:
## Developer Notes
Seems like the best course of action would be to:
- Make the postcode nullable in the db (if it isn't already)
- Add a "No address" checkbox next to the address field on the form
  - When the checkbox is ticked, the address field should be disabled, and the value set to null
- In pages that use the postcode (shopping list, shipping label, driver overview etc), display "NFA" whenever the postcode is null
- For generating the map, ignore null postcodes.
- Change the generated seed data so that some clients have a null postcode.

AmeliaIsCoding:
Ricky's above notes was the approach I used

Also note:

-when a client has null postcode, their entire address is displayed as "NFA", so the other address fields are ignored even if populated
-in the client add/edit form, ticking the checkbox hides the address field, not just disabling. the address fields are all set to an empty string or null (postcode)
-if a client with null postcode is edited, the other address fields are set to an empty string on submitting the edit form (unless they are edited by the user). They appear empty in the form during the editing process.

AmeliaIsCoding:
### QA

- [ ]  In parcels page, some parcels are shown with postcode equal to NFA
- [ ] On opening the parcels modal, the entire address of such parcels is shown only as NFA
- [ ] Select a parcel with NFA -> create shipping label -> the postcode is shown as NFA -> download PDF -> the entire address is shown only as NFA in the PDF
- [ ] Repeat for shopping list action
- [ ] Repeat for driver overview action
[...]

PR #286, opened 2024-04-23
#188: Allow a client to have no address

AmeliaIsCoding:
## What's changed
- Client's postcode is now nullable
- If a client's postcode is null, their address is displayed as "NFA" everywhere on the front end, including parcels page and modal, clients page and modal, actions modal, and all PDFs
- When generating a map for parcels with NFA, these parcels are ignored in the map; if only NFA parcels are selected, an error message is displayed in the generate map modal
- When creating or editing a client, there is a checkbox for "No Address"; when checked, the address fields disappear and are set to empty strings or null in the case of postcode
- The seeded data includes some clients with null postcode

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image|imageimageimage|

## Checklist

  - Does this require resetting the deployed database? - NO unless we want the seeded data to include some clients/parcels with NFA
    - Checkout the dev branch
    - Checkout your branch


src/app/parcels/parcelsTableFilters.ts
```diff
@@ -29,6 +30,14 @@ export const postcodeSearch = (
...
     state: string
 ): PostgrestFilterBuilder<Database["public"], any, any> => {
+    if (state === "") {
+        return query;
+    }
+    if (nullPostcodeDisplay.toLowerCase().includes(state.toLowerCase())) {
+        return query.or(
+            `client_address_postcode.ilike.%${state}%, client_address_postcode.is.null`
+        );
+    }
```
AmeliaIsCoding, line 40: we need to change the postcode filter to search for null postcodes if "nfa" is searched for

src/components/Form/formFunctions.ts
```diff
@@ -47,25 +50,20 @@ export interface Fields {
...
-): FieldSetter => {
-    return (fieldKey, newFieldValue) => {
-        fieldSetter({ ...fieldValues, [fieldKey]: newFieldValue });
+): Setter<SpecificFields> => {
+    return <Key extends keyof SpecificFields>(keyValuePairs: [Key, any][]): void => {
+        const newFieldValues: SpecificFields = { ...fieldValues };
+        keyValuePairs.forEach((keyValuePair) => {
+            newFieldValues[keyValuePair[0]] = keyValuePair[1];
+        });
+        setFields(newFieldValues);
```
AmeliaIsCoding, line 66: This new createSetter function takes in a state and a setState action, and returns a setter function which takes in key value pairs and updates the corresponding keys in the state. The differences are that now the setter function can update multiple keys simultaneously, and also the createSetter can be called for both the fields and the error fields.

supabase/migrations/20240422113025_allow_null_postcode.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."clients" drop constraint "clients_address_postcode_check";
+
+alter table "public"."clients" alter column "address_postcode" drop not null;
+
```
AmeliaIsCoding, line 4: As well as dropping the "not null" constraint on the postcode, I also needed to drop a custom constraint where we checked that the length of postcode was greater than 0.

supabase/seed/clientsSeed.mts
```diff
@@ -35,4 +35,5 @@ export const possibleParcelPostCodes = [
     "CB2 3JU",
     "CB24 4RT",
     "CB8 9LJ",
+    null,
```
AmeliaIsCoding, line 38: adding null postcode as a possible seeded postcode

AmeliaIsCoding (commented):
Self review `

AmeliaIsCoding (commented):
Self review 2

src/app/admin/createUser/CreateUserForm.tsx
```diff
@@ -22,23 +22,32 @@ import UserDetailsCard from "@/app/admin/createUser/UserDetailsCard";
...
     lastName: string;
     telephoneNumber: string;
 }
+interface InviteUserErrors extends FormErrors<InviteUserFields> {
+    email: Errors;
+    role: Errors;
+    firstName: Errors;
+    lastName: Errors;
+    telephoneNumber: Errors;
+}
```
rickaw, line 38:
```suggestion
interface InviteUserErrors = Required<FormErrors<InviteUserFields>>
```

src/components/Form/formFunctions.ts
```diff
@@ -6,9 +6,9 @@ import {
 import { Database } from "@/databaseTypesFile";
 import dayjs, { Dayjs } from "dayjs";
 
-type Field = Fields[keyof Fields];
-export type ErrorSetter = (errorKey: string, errorType: Errors) => void;
-export type FieldSetter = (fieldKey: string, newFieldValue: Field) => void;
+export type Setter<SpecificFields extends Fields> = <Key extends keyof SpecificFields>(
+    keyValuePairs: [Key, any][]
```
rickaw, line 10: I think a list of { key: Key, value: any } is easier read than a list of [Key, value] because we can name the two fields.
```suggestion
    keyValuePairs: { key: Key, value: any }[]
```


Or an even better way of doing this might be something like this, where the setter expects one object as a parameter, which can then set multiple fields at the same time?
```ts
export type Setter<SpecificFields extends Fields> = (keyValuePairs: {
    [key in keyof SpecificFields]: any;
}) => void;
```

src/components/Form/formFunctions.ts
```diff
@@ -87,89 +85,104 @@ const getErrorType = (
     return Errors.none;
 };
 
-export const onChangeText = (
-    fieldSetter: FieldSetter,
-    errorSetter: ErrorSetter,
+export const onChangeText = <
+    SpecificFields extends Fields,
+    SpecificErrors extends FormErrors<SpecificFields>,
```
rickaw, line 90: Does this second generic parameter real parameter? Can't we just have `errorSetter: Setter<FormErrors<SpecificFields>>` and so have one generic parameter?

(similarly in the other places)

src/app/parcels/ParcelsPage.tsx
```diff
@@ -248,6 +248,7 @@ const parcelIdParam = "parcelId";
 
 const defaultNumberOfParcelsPerPage = 100;
 const numberOfParcelsPerPageOptions = [10, 25, 50, 100];
+export const nullPostcodeDisplay = "NFA";
```
rickaw, line 251: we should probably have this line in a different file as this is used in multiple places, not just the ParcelsPage. Is there anywhere sensible? We could just create a file `src/common/nullPostCodeDisplay.ts` or something

src/app/parcels/ActionBar/ActionAndStatusBar.cy.tsx
```diff
@@ -156,7 +156,7 @@ describe("Parcels - Action Bar", () => {
             });
             cy.get(`input[value="${timeString}"]`).should("exist");
             mockData.forEach((parcel: ParcelsTableRow) => {
-                cy.get(".MuiPaper-root").contains(parcel.addressPostcode);
+                parcel.addressPostcode && cy.get(".MuiPaper-root").contains(parcel.addressPostcode);
```
rickaw, line 159: (same in other places)
```suggestion
                if (parcel.addressPostcode) {
                    cy.get(".MuiPaper-root").contains(parcel.addressPostcode);
                }
```

rickaw (changes requested):

src/app/parcels/ParcelsPage.tsx
AmeliaIsCoding: I think putting it in common/format.ts would be sensible

src/app/admin/createCollectionCentre/CreateCollectionCentreForm.tsx
```diff
@@ -7,26 +7,37 @@ import {
...
-const initialFields: InsertSchema["collection_centres"] = {
+interface CollectionCentreFields {
+    name: string;
+    acronym: string;
+}
+
+interface CollectionCentreErrors extends FormErrors<CollectionCentreFields> {
+    name: Errors;
+    acronym: Errors;
+}
```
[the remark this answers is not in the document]
AmeliaIsCoding: yes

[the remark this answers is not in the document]
AmeliaIsCoding: nevermind, I get an error saying that Required<FormErrors<SpecificFields>> is not assignable to FormErrors<SpecificFields> because the values in the latter can be undefined but in the former they have to be of type Errors. Not sure how to fix :(

src/components/Form/formFunctions.ts
AmeliaIsCoding: Okay - I think we still need SpecificErrors as a generic parameter for checkErrorOnSubmit because it has errors as an actual parameter, but let me know if there's something I could do to get rid of the generic parameter in this case?

src/components/Form/formFunctions.ts
AmeliaIsCoding: I agree that {key: Key, value: any}[] would be better, but I'm not sure I understand your second suggestion. If we were to just pass in one object, how would we then set multiple fields?

AmeliaIsCoding: Pushing the first suggestion :)

rickaw: If we pass in an object like 
```ts
{
  field1: value1,
  field2: value2,
}
```
then field1 should be updated to value1, field2 should be updated to value2 etc. Am I missing something?

src/components/Form/formFunctions.ts
rickaw: Does this work?

```ts
export const checkErrorOnSubmit = <SpecificFields extends Fields>(
    errorType: FormErrors<SpecificFields>,
    errorSetter: (errors: FormErrors<SpecificFields>) => void,
    keysToCheck?: string[]
): boolean => {
```

src/app/admin/createCollectionCentre/CreateCollectionCentreForm.tsx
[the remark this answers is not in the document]
rickaw: hm what was the problem here?

rickaw (commented):
thanks! just a few bits left from before :)

src/app/admin/createCollectionCentre/CreateCollectionCentreForm.tsx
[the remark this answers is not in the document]
AmeliaIsCoding: I fixed it, struggling to remember details - sorry :(

src/components/Form/formFunctions.ts
AmeliaIsCoding: No, because errorType isn't necessarily FormErrors<SpecificFields> it should be extends FormErrors<SpecificFields> I think
I get this error
```
Argument of type 'Dispatch<SetStateAction<Required<FormErrors<CollectionCentreFields>>>>' is not assignable to parameter of type '(errors: FormErrors<CollectionCentreFields>) => void'.
  Types of parameters 'value' and 'errors' are incompatible.
    Type 'FormErrors<CollectionCentreFields>' is not assignable to type 'SetStateAction<Required<FormErrors<CollectionCentreFields>>>'.
      Type 'FormErrors<CollectionCentreFields>' is not assignable to type 'Required<FormErrors<CollectionCentreFields>>'.
        Types of property 'name' are incompatible.
          Type 'Errors | undefined' is not assignable to type 'Errors'.
            Type 'undefined' is not assignable to type 'Errors'.
```

rickaw:
Could you possibly add a few e2e tests to check this feature works correctly?

src/components/Form/formFunctions.ts
AmeliaIsCoding: we got this to work

src/app/admin/createUser/UserRoleCard.tsx
```diff
@@ -3,14 +3,15 @@ import React from "react";
...
 
 const UserRoleCard: React.FC<InviteUserCardProps> = ({ fieldSetter }) => {
     return (
         <GenericFormCard title="User Role" required>
             <UserRoleDropdownInput
-                defaultValue="caller"
+                defaultValue="volunteer"
                 onChange={getDropdownListHandler((role: string) =>
-                    fieldSetter([{ key: "role", value: role }])
+                    fieldSetter({ role: role as UserRole })
```
rickaw, line 14: instead of asserting the value as UserRole here, can we make it so that `UserRoleDropdownInput` has a prop onChange that is `(userRole: UserRole) => void` or something similar?

src/components/Form/formFunctions.ts
```diff
@@ -58,12 +58,8 @@ export const createSetter = <SpecificFields extends Fields>(
...
     fieldValues: SpecificFields
 ): Setter<SpecificFields> => {
-    return <Key extends keyof SpecificFields>(keyValuePairs: { key: Key; value: any }[]): void => {
-        const newFieldValues: SpecificFields = { ...fieldValues };
-        keyValuePairs.forEach((keyValuePair) => {
-            newFieldValues[keyValuePair.key] = keyValuePair.value;
-        });
-        setFields(newFieldValues);
+    return (fieldValuesToUpdate: Partial<SpecificFields>): void => {
+        setFields({ ...fieldValues, ...fieldValuesToUpdate });
```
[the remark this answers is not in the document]
AmeliaIsCoding: I thought the same when I first saw this, but it does retain all updates. I guess even though fieldValues is passed in when creating the setter, it's by reference instead of value or something? 
I've just double checked and it definitely retains the updates :)

src/app/admin/createUser/UserRoleCard.tsx
AmeliaIsCoding: I gave this a go yesterday but I gave up because I got this error and I didn't understand it :'(
image
Please help when you have some time :)

rickaw: I think you meant to use `Value extends string`.

Regardless you'll see an error message.
```
Argument of type string | Value is not assignable to parameter of type Value
string | Value is assignable to the constraint of type Value, but Value could be instantiated with a different subtype of constraint string
```


You'd probably have to delegate the validation of the value to the function caller i.e.
```ts
export const getDropdownListHandler = <Value extends string>(
    setValue: ValueSetter<Value>,
    isValueValid: (value: string) => value is Value
): SelectChangeEventHandler<Value> => {
    return (event: SelectChangeEvent<Value>) => {
        const value = event.target.value;
        if (isValueValid(value)) {
            setValue(value);
        }
    };
};
```

src/components/Form/formFunctions.ts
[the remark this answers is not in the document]
rickaw: ok cool!

src/app/admin/createUser/UserRoleCard.tsx
AmeliaIsCoding: Discussed in person and think is fixed

rickaw (approved):

src/components/DataInput/DataInput.cy.tsx
```diff
@@ -32,7 +32,10 @@ describe("Data Input Components", () => {
                 ]}
                 listTitle="Dropdown List"
                 defaultValue="Default"
-                onChange={getDropdownListHandler(() => console.log("Dropdown Changed"))}
+                onChange={getDropdownListHandler(
+                    () => console.log("Dropdown Changed"),
```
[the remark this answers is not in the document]
AmeliaIsCoding: Have done

src/app/admin/common/UserRoleDropdownInput.tsx
```diff
@@ -12,7 +12,7 @@ const roleLabelsAndValues: [string, UserRole][] = [
 
 interface Props {
     defaultValue: UserRole;
-    onChange: SelectChangeEventHandler;
+    onChange: SelectChangeEventHandler<any>;
```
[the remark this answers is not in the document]
AmeliaIsCoding: Have done

PR #290, opened 2024-04-23
VFB-175: Fix

AmeliaIsCoding:
## What's changed
- Keep status modal open if the status change fails, so that the user can see the error message
- When selecting a date via the calendar view on the Day Overview Modal, a valid date now enables the PDF download button

## Checklist
- [ ] The ticket is up-to-date - Please document any deviations from the original approach if there is any.
- [ ] I have self-reviewed this PR
- [ ] Make sure you've verified it works via `npm run dev`

No DB changes.

### DIFF
--- a/src/app/parcels/ActionBar/Statuses.tsx
+++ b/src/app/parcels/ActionBar/Statuses.tsx
@@ -58,7 +58,7 @@ export const saveParcelStatus = async (
     }));
 
     const { data, error } = await supabase
-        .from("events")
+        .from("ss")
         .insert(eventsToInsert)
         .select("event_id:primary_key, parcel_id");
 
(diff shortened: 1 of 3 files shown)


src/app/parcels/ActionBar/Statuses.tsx
```diff
@@ -152,7 +152,7 @@ const Statuses: React.FC<Props> = ({
             setServerErrorMessage(`${getStatusErrorMessage(error)} Log ID: ${error.logId}`);
         }
         hasAttemptedToSaveParcelStatus();
-        setStatusModal(false);
+        error ?? setStatusModal(false);
```
rickaw, line 155: nitpick: maybe this is just a personal preference, but IMO `??` and `&&` etc form expressions, not a conditional execution of tasks. To me it reads better to use `if` statements, even if at the cost of extra lines.

```suggestion
        if (!error) {
            setStatusModal(false);
        }
```

rickaw (approved):

src/app/parcels/ActionBar/Statuses.tsx
AmeliaIsCoding: have done

PR #291, opened 2024-04-23
VFB-81: Concurrent record editing

minalui:
## What's changed
- Add last_updated column to parcels and clients table, default set to current utc
- Existing rows will have last_update set to utc when migration happens
- Add function to updated last_updated to current utc when an update is made to the row
- Change update functions to check last_updated in client side is the same as database
- Handle and display errors

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| paste_here | paste_here |

## Checklist

  - [ ] I have modified the seed in `supabase/seed/seed.mts` if appropriate N/A
  - [ ] If I have modified the seed, I have also generated the seed with `npm run db:generate_seed`  N/A
  - [ ] With my final set up, I can run `npm run dev:reset_supabase` without any errors. N/A
    - Checkout the dev branch
    - Checkout your branch


src/app/clients/form/submitFormHelpers.ts
```diff
@@ -60,13 +58,19 @@ const formatClientRecord = (
...
+        last_updated: fields.lastUpdated,
     };
 };
 
-export const submitAddClientForm = async (
-    fields: ClientFields,
-    router: AppRouterInstance
-): Promise<void> => {
+type addClientErrors = "failedToInsertClientAndFamily";
+type addClientReturnType =
```
rickaw, line 66: For convention, we'll stick to `...Result` :)
```suggestion
type addClientResult =
```

src/app/clients/form/submitFormHelpers.ts
```diff
@@ -99,21 +103,32 @@ export const submitAddClientForm = async (
         clientId: clientId,
     });
 
-    router.push(`/parcels/add/${clientId}`);
+    return { clientId: clientId, error: null };
 };
 
+type editClientErrors = "failedToUpdateClientAndFamily" | "concurrentUpdateConflict";
+type editClientReturnType =
```
rickaw, line 110:
```suggestion
type editClientResult =
```

rickaw (changes requested):

supabase/migrations/20240423141258_update_client_function_change.sql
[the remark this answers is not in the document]
minalui: The function now only updates a row if primary_key and last_updated matches. It also returns the client ID and number of rows updated in a new database type.

src/app/error.tsx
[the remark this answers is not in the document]
minalui: OK!

src/app/clients/ClientsPage.tsx
```diff
@@ -162,9 +160,10 @@ const ClientsPage: React.FC<{}> = () => {
             });
 
         return () => {
-            supabase.removeChannel(subscriptionChannel);
+            void supabase.removeChannel(subscriptionChannel);
         };
-    }, [startPoint, endPoint, primaryFilters, sortState]);
+    }, [startPoint, endPoint, primaryFilters, sortState, fetchAndDisplayClientsData]);
```
rickaw, line 165: I don't think you need this change anymore since we don't call fetchAndDisplayClientsData?

rickaw (approved):
one last thing but feel free to merge once that's addressed!

src/app/clients/ClientsPage.tsx
rickaw: ah that makes sense!

PR #293, opened 2024-04-24
VFB-177: Create admin view of audit logs

AmeliaIsCoding:
## What's changed

- Audit log table shown in admin page in its own panel
- All columns except primary key displayed, with action, created_at, user_id, content, wasSuccess and log_id shown as the default headers, with the other headers toggleable
- All columns are sortable
- No filters are implemented yet
- The table uses server side pagination
- The table uses realtime subscription
- All columns displayed as simple text, except created_at is formatted as a datetime and wasSuccess formatted as “True” or “False”
- The default sorting of the table is by created_at, with most recent first

## Screenshots / Videos
| After      |
|------------|
|imageimage|

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/components/Tables/TableFilterBar.tsx
AmeliaIsCoding, line 1: The changes in this file amount to 

- only showing the filter icon(s) if filters exist for the table. 
- renaming variables to make it clear that this component isn't just the filter bar, it also serves to display additional columns

AmeliaIsCoding (commented):
Self-review

AmeliaIsCoding (commented):
Self-review 2

src/app/admin/auditLogTable/AuditLogTable.tsx
rickaw, line 1: Shall we move _the bits of this file that are not the component itself_ into different files? 

For readability, we'd ideally have AuditLogTable.tsx that is like
```ts
const AuditLogTable: React.FC = () => {
  ...
}

export default AuditLogTable;
```


Another thing I noticed was that `fetchAuditLogData.ts` imports `AuditLogRow` from this file and this file imports stuff from `fetchAuditLogData.ts`. To me circular dependencies between files are an indication of us not structuring the code clearly enough. For example, should `AuditLogRow` belong to the table element itself or the fetch function, or is it equally relevant to both? Seems like it is more relevant to the table itself because it forces every field to have a string. But should the fetch function care about what the table needs to show? Or can we return some intermediate type that is convenient from the fetch function's perspective, and convert that data within the table to suit its needs?

src/app/admin/auditLogTable/fetchAuditLogData.tsx
rickaw, line 1: This could just be a .ts file, not .tsx?

src/app/admin/auditLogTable/AuditLogTable.tsx
AmeliaIsCoding: I've had a go at doing this. One thing I was unsure about was what to name that file that has the things that were in AuditLogTable.tsx but were not part of the component. I called it auditLogTableConstants.ts but let me know if there is something better / more conventional.

rickaw: is there no way of splitting it up into sensible parts? I feel like `auditLogTableConstants` doesn't tell us much. Maybe a file for all the columns, another file for display functions?

rickaw (approved):

src/app/admin/auditLogTable/AuditLogTable.tsx
AmeliaIsCoding: Can do if you think it's worth the extra files

rickaw: I'd err on the side of more files with independent pieces of code in each, rather than all of them in one giant file.

As a guide, when the file exceeds 200 lines or so, this might be an indication of us doing too many things in that file. (but it depends)

AmeliaIsCoding: Have done, might have gone a bit crazy with the file separation lol tell me if it's too much

rickaw (approved):

src/app/admin/auditLogTable/AuditLogTable.tsx
rickaw: nah this is great! the separation of logic makes it much easier for future devs to read the code!

AmeliaIsCoding: ParcelsPage.tsx might need some refactoring then lol :(

PR #294, opened 2024-04-24
VFB-131: Add manager and staff roles and rename caller to volunteer

AmeliaIsCoding:
## What's changed

- Manager and staff roles added
- Both manager and staff can access parcels, clients, lists and calendars
- Manager and staff can edit/update/delete items on the lists page
- Caller role renamed to volunteer
- On restarting the database, one user of each role is created

## Checklist

    - Checkout the dev branch
    - Checkout your branch


AmeliaIsCoding (commented):
Self-review

supabase/migrations/20240424132005_add_manager_and_staff_roles_and_rename_caller_to_volunteer.sql
```diff
@@ -0,0 +1,65 @@
...
+    users.email,
+    users.created_at,
+    users.updated_at,
+    profiles.first_name,
+    profiles.last_name,
+    profiles.role,
+    profiles.telephone_number
+   FROM (auth.users
+     LEFT JOIN profiles ON ((users.id = profiles.user_id)))
+  ORDER BY profiles.first_name;
```
rickaw, line 26: Is this exactly the same as the existing definition? Did we need this to make the role work just out of curiosity?

src/app/roles.tsx
```diff
@@ -22,14 +22,20 @@ const pathsShownToAllAuthenticatedUsers = [
...
     switch (role) {
         case "admin":
             return [...pathsShownToAllAuthenticatedUsers, ...pathsOnlyShownToAdmin];
-        case "caller":
+        case "volunteer":
             return pathsShownToAllAuthenticatedUsers;
+        case "manager":
+            return [...pathsShownToAllAuthenticatedUsers, ...pathsOnlyShownToManager];
+        case "staff":
+            return [...pathsShownToAllAuthenticatedUsers, ...pathsOnlyShownToManager];
```
rickaw, line 38: since we have a very clear 1-dimentional structure of permissions, why don't we "incrementally add paths as we go up the roles" to avoid repetition? i.e.
```ts
const pathsShownToStaffAndAbove = ["/lists"] as const;
const const pathsOnlyShownToAdmin = ["/admin"] as const;

...

    case "admin:
        return [...pathsShownToAllAuthenticatedUsers, ...pathsShownToStaffAndAbove, ...pathsOnlyShownToAdmin];

File: supabase/migrations/20240424132005_add_manager_and_staff_roles_and_rename_caller_to_volunteer.sql (hunk 1 above)
Comment (AmeliaIsCoding) [reply supabase/migrations/20240424132005_add_manager_and_staff_roles_and_rename_caller_to_volunteer.sql:26]: This was automatically generated when making the migration file via the diff method. I think we need to drop and then recreate the view (yes its exactly the same) just because the role enum changed.

File: src/app/roles.tsx (hunk 2 above)
Comment (AmeliaIsCoding) [reply src/app/roles.tsx:38]: Sure, I wasn't sure whether you would prefer this or what I did but will do

File: supabase/tools/createUserForEachRole.ts [hunk 3]
```diff
@@ -6,7 +6,7 @@ dotenv.config({ path: "./.env.local" });
 
 createAdminAndCallerUsers();
```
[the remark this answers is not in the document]
AmeliaIsCoding: thanks, good catch

supabase/migrations/20240403101727_add_profiles_to_existing_users.sql
```diff
@@ -20,7 +20,7 @@ BEGIN
     WHILE index <= total_count LOOP
         SELECT user_id INTO profile_id FROM user_id_array WHERE array_index = index;
         SELECT "role" INTO profile_role FROM user_id_array WHERE array_index = index;
-        INSERT INTO profiles (primary_key, "role") VALUES (profile_id, COALESCE(profile_role, 'caller'));
+        INSERT INTO profiles (primary_key, "role") VALUES (profile_id, COALESCE(profile_role, 'volunteer'));
```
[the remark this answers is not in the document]
AmeliaIsCoding: Okay thanks. I think at this point it's correct to use 'caller' as the role enum is 'caller' | 'volunteer'.
Changing this to 'caller' then checking the migration works is successful so I think we're okay.

The steps:
checking out dev -> npm run post_checkout -> checkout my branch -> npx supabase migrations up --include-all --local 
has been very very useful to check that the migration works by the way!

supabase/migrations/20240424132005_add_manager_and_staff_roles_and_rename_caller_to_volunteer.sql
rickaw: fair enough!

rickaw (approved):

PR #297, opened 2024-04-25
VFB-206: Change audit log table to use actor profile ID instead of user ID

AmeliaIsCoding:
## What's changed

- delete column user ID in audit log table in supabase
- add column actor profile ID which is a non nullable FK to profiles
- populate the new column with the profile ID of the user from the previous user ID column
- whenever we make a new audit log, populate the new column with the profile ID of the currently logged in user

## Checklist

  - Does this require resetting the deployed database? -  NO
    - Checkout the dev branch
    - Checkout your branch


AmeliaIsCoding (commented):
self-review

supabase/migrations/20240425082036_actor_profile_id_in_audit_log_table.sql
```diff
@@ -0,0 +1,11 @@
+alter table "public"."audit_log" add column "actor_profile_id" uuid;
+
+alter table "public"."audit_log" add constraint "audit_log_actor_profile_id_fkey" FOREIGN KEY ("actor_profile_id") REFERENCES "public"."profiles"("primary_key") ON DELETE CASCADE;
```
rickaw, line 3: let's use ON DELETE SET NULL, just in case. This means `actor_profile_id`  needs to be nullable, but I think that's better than an accidental removal of a profile deleting all the logs for that profile.

rickaw (commented):
looks great, thanks!

supabase/migrations/20240425082036_actor_profile_id_in_audit_log_table.sql
AmeliaIsCoding: Ok makes sense. Do I need to remove the not null constraint? I didn't, and tried resetting the database, and everything worked fine - the column is not nullable but has "ON DELETE SET NULL". Not sure if it will work or not?

AmeliaIsCoding: ignore me i answered this by trying to delete a profile with associated audit logs - supabase errored - so i'll remove the line in the migration file that sets the null constraint

rickaw (approved):

PR #305, opened 2024-04-29
VFB-205: Audit log modal and updates to audit log table

AmeliaIsCoding:
## What's changed

- On clicking a row in the audit log table, a modal opens
- The modal header is the name of the action
- The modal has one row for each foreign key that is populated in the DB audit log table
- When parcel_id or client_id is populated, a link to the corresponding parcel or client modal is shown
- For any other foreign key, text is displayed according to below
- The content column was also moved inside the audit log modal
- The foreign key columns were removed from the table (and the content column)
- Instead of displaying the user ID of the user who carried out the action, their full name is shown, unless their user id is null, in which case "Deleted User" is shown. This column is sortable alphabetically
- Change title of column "Log ID" to "Error Log ID"

For each foreign key that is populated, the following data/link is displayed: 

Parcels → display the client name, postcode and collection time, and on clicking this, link to the parcel details modal
Clients → display the client name and on clicking this, link to the client details modal
Collection centres → display the collection centre name
Events → display the new parcel status name
Profiles → display the first and last name of the profile, or “Deleted User” if the profile has no user ID
Lists → display the ingredient name
Lists hotel → display the ingredient name
Packing slots → display the packing slot name
Status order → display the event name
Website data → display the name

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image|imageimage|

## Checklist

    - Checkout the dev branch
    - Checkout your branch


AmeliaIsCoding (commented):
Self-review

AmeliaIsCoding (commented):
Self-review 2

AmeliaIsCoding (commented):
Self review 3

src/app/admin/auditLogTable/types.ts
```diff
@@ -1,9 +1,12 @@
 import { Json } from "@/databaseTypesFile";
-import { Schema } from "@/databaseUtils";
+import { ViewSchema } from "@/databaseUtils";
+
+export type AuditLogResponseData = ViewSchema["audit_log_plus"][];
```
rickaw, line 4: Shall we put this inside databaseUtils as well?

src/app/admin/auditLogTable/AuditLogTable.tsx
```diff
@@ -26,6 +27,8 @@ const AuditLogTable: React.FC = () => {
     const [currentPage, setCurrentPage] = useState(1);
     const startPoint = (currentPage - 1) * auditLogCountPerPage;
     const endPoint = currentPage * auditLogCountPerPage - 1;
+    const [modalIsOpen, setModalIsOpen] = useState<boolean>(false);
```
rickaw, line 30: Do we really need this boolean state? Can't we use the presence of `selectedAuditLog` as a way to signal whether to show the modal?

src/app/admin/auditLogTable/auditLogModal/AuditLogModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    )}
+                    {selectedAuditLogRow.websiteData && (
+                        <WebsiteDataAuditLogModalRow
+                            websiteDataName={selectedAuditLogRow.websiteData}
+                        />
+                    )}
+                    <AuditLogModalItem>
+                        <Key>ACTION CONTENT: </Key>
+                        <TextValueContainer>
+                            {formatJson(selectedAuditLogRow.content)}
```
rickaw, line 104: could we change formatJson to use something like `JSON.stringify(json, null, 2)` to see if that makes it prettier?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

src/app/admin/auditLogTable/auditLogModal/AuditLogModalRow.tsx
```diff
@@ -0,0 +1,67 @@
...
+    fetchResponse: (foreignKey: string) => Promise<AuditLogModalRowResponse<Data, Error>>;
+    getErrorMessage: (error: Error) => string;
+    RowComponentWhenSuccessful: React.FC<Data>;
+    header: string;
+}
+
+const AuditLogModalRow = <Data, Error>({
+    foreignKey,
+    fetchResponse,
+    getErrorMessage,
```
rickaw, line 36: I don't think we need all three of these. We could instead have one function.

```ts
getDataOrErrorToDisplay: () => { 
  data: Data, 
  errorMessage: null 
} | { 
  data: null, 
  errorMessage: string 
}
```

src/app/admin/auditLogTable/auditLogModal/auditLogModalRows/Client.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+const fetchClientLinkDetails = async (
+    clientId: string
+): Promise<AuditLogModalRowResponse<ClientLinkDetails, ClientLinkDetailsError>> => {
+    const { data: data, error } = await supabase
+        .from("clients")
+        .select("primary_key, full_name")
+        .eq("primary_key", clientId)
+        .limit(1)
```
rickaw, line 29: not fussed, but i don't think we need this `limit(1)` since we're matching by primary key (same in other places)

src/app/admin/auditLogTable/types.ts
```diff
@@ -16,16 +19,20 @@ export interface AuditLogRow {
...
-    auditLogResponse: Schema["audit_log"][]
+    auditLogResponse: AuditLogResponseData
 ): AuditLogRow[] =>
     auditLogResponse.map((auditLogResponseRow) => ({
+        auditLogId: auditLogResponseRow.primary_key ?? "",
         action: auditLogResponseRow.action ?? "",
-        actorProfileId: auditLogResponseRow.actor_profile_id ?? "",
+        actorName:
+            auditLogResponseRow.actor_user_id === null
+                ? "User Deleted"
```
rickaw, line 34: to be consistent with the modal row (or even put this in a common const)
```suggestion
                ? "Deleted User"
```

rickaw (commented):
I like how different modal rows are nicely commonised and yet separated!

src/app/admin/auditLogTable/AuditLogTable.tsx
AmeliaIsCoding: That's much better yes! I'm wondering if, within the modal component, there's a way for typescript to know that selectedAuditLog must be truthy?

src/app/admin/auditLogTable/types.ts
AmeliaIsCoding: Sure

src/app/admin/auditLogTable/AuditLogTable.tsx
AmeliaIsCoding: image
ie remove the need for the {selectedAuditLog && ....} check

rickaw: you could do 
```ts
if (selectedAuditLogRow === null) {
  return <></>
}
```
or something although this is also not super clean / not aligned with the `isOpen` way of showing modals.

Up to you.

src/app/admin/auditLogTable/auditLogModal/AuditLogModal.tsx
AmeliaIsCoding: Yep it's slightly better

src/app/admin/auditLogTable/AuditLogTable.tsx
AmeliaIsCoding: I'll leave it how it is

AmeliaIsCoding (commented):
Self-review

rickaw (approved):

PR #308, opened 2024-04-30
VFB-209: Audit log editing a user

AmeliaIsCoding:
## What's changed

- Audit log editing a user - currently only role changes are possible in the front end but the audit log logic will extend to any changes.
- Also fixed a bug where when you change a users role to one that was not previously held by any user, the user disappears from the users table until refreshing. Fixed this by filtering by all user roles not just those held by users in the DB. 

## Checklist

    - Checkout the dev branch
    - Checkout your branch

### DIFF
--- a/src/app/admin/usersTable/UsersTable.tsx
+++ b/src/app/admin/usersTable/UsersTable.tsx
@@ -22,7 +22,7 @@ import { getCurrentUser } from "@/server/getCurrentUser";
 import { subscriptionStatusRequiresErrorMessage } from "@/common/subscriptionStatusRequiresErrorMessage";
 
 const usersTableHeaderKeysAndLabels: TableHeaders<UserRow> = [
-    ["id", "User ID"],
+    ["userId", "User ID"],
     ["firstName", "First Name"],
     ["lastName", "Last Name"],
     ["email", "Email"],
@@ -273,7 +273,7 @@ const UsersTable: React.FC = () => {
                 dataPortion={users}
                 headerKeysAndLabels={usersTableHeaderKeysAndLabels}
                 columnDisplayFunctions={userTableColumnDisplayFunctions}
-                toggleableHeaders={["id", "createdAt", "updatedAt"]}
+                toggleableHeaders={["userId", "createdAt", "updatedAt"]}
                 defaultShownHeaders={[
                     "firstName",
                     "lastName",
@@ -302,7 +302,7 @@ const UsersTable: React.FC = () => {
                     onDelete: userOnDelete,
                     onEdit: userOnEdit,
                     setDataPortion: setUsers,
-                    isDeletable: (row: UserRow) => row.id !== currentUserId,
+                    isDeletable: (row: UserRow) => row.userId !== currentUserId,
                 }}
                 filterConfig={{
                     primaryFiltersShown: true,
--- a/src/app/admin/usersTable/UsersTable.tsx
+++ b/src/app/admin/usersTable/UsersTable.tsx
@@ -162,18 +162,9 @@ const UsersTable: React.FC = () => {
                     method: emailSearch,
                 },
             }),
+            buildUserRoleFilter(),
         ];
 
-        const { data: userRoleFilter, error } = await buildUserRoleFilter();
-        if (error) {
-            switch (error.type) {
-                case "failedToFetchUserRoleFilterOptions":
-                    setErrorMessage(`Failed to retrieve user role filters. Log ID: ${error.logId}`);
-                    break;
-            }
-        } else {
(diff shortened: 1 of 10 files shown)


src/app/admin/usersTable/usersTableFilters.ts
```diff
@@ -67,20 +67,11 @@ export const buildUserRoleFilter = async (): Promise<BuildUserRoleFilterAndError
...
-
-    uniqueUserRoles.sort();
-
     const userRoleFilter = checklistFilter<UserRow>({
         key: "userRole",
         filterLabel: "User Role",
-        itemLabelsAndKeys: uniqueUserRoles.map((userRole) => [userRole, userRole]),
-        initialCheckedKeys: uniqueUserRoles,
+        itemLabelsAndKeys: userRolesFromDb.map((userRole) => [userRole, userRole]),
+        initialCheckedKeys: userRolesFromDb,
```
AmeliaIsCoding, line 74: This fixes the bug described in the PR

supabase/migrations/20240430155721_rename_profiles_plus_to_users_plus_and_include_profile_id.sql
```diff
@@ -0,0 +1,42 @@
+drop view if exists "public"."profiles_plus";
+
+create or replace view "public"."users_plus" as  SELECT users.id AS user_id,
```
AmeliaIsCoding, line 3: Renamed the users_plus view to be consistent with the other views which are named off of the primary table that they are built from

supabase/migrations/20240430155721_rename_profiles_plus_to_users_plus_and_include_profile_id.sql
```diff
@@ -0,0 +1,42 @@
...
+    users.email,
+    users.created_at,
+    users.updated_at,
+    profiles.primary_key AS profile_id,
```
AmeliaIsCoding, line 7: The view now includes profile ID. This allows us to update a user's by their profile ID instead of by their user ID. This means we can populate the profile ID in the audit log without having to select it from the DB after updating. This is needed because if the update fails, we do not retrieve the profile ID, meaning we can't populate it in the audit log, if it fails.

AmeliaIsCoding (commented):
Self-review

src/app/admin/usersTable/usersTableFilters.ts
AmeliaIsCoding: You're right it didn't actually fix it

I see two options
1. hardcode the roles in the dropdown 
2. update the roles in the dropdown automatically through a realtime channel

I tried to do the second option, but the problem was that, if we rebuild the filter every time the unique roles in the DB change, we lose the current state of the filter. To fix this, could a) change the way the filters work to be stateful we don't have to rebuild it every time, but this would be a lot of effort, or b) store the current state of the filter in the parent and pass that in to the function that build the filter whenever we need to rebuild it. This would work I think, but would add yet another state to the parent, when they shouldn't really be the one controlling the current state of the filter.

I've done option 1 and pushed it, but let me know if you want me to spend time on doing option 2

rickaw: yeah i think option 1 is okay. We'd definitely want to make it clear that it is tied to the database so that if we decide to change roles in the future, it'd be obvious enough to pick up (or even give an error)

src/app/admin/usersTable/usersTableFilters.ts
```diff
@@ -48,32 +34,23 @@ export const emailSearch = (
...
-
-    if (error) {
-        const logId = await logErrorReturnLogId("Error with fetch: User role filter options", {
-            error: error,
-        });
-        return { data: null, error: { type: "failedToFetchUserRoleFilterOptions", logId: logId } };
-    }
-
-    const userRolesFromDb: UserRole[] = data.map((row) => row.role) ?? [];
+    const allUserRoles: UserRole[] = ["volunteer", "manager", "staff", "admin"];
```
rickaw, line 45: Yeah i think typing this as UserRole[] is good enough. It can at least cope with renaming of roles. It just can't cope with adding a role in the database, but it's a compromise i guess

src/app/admin/usersTable/UsersTable.tsx
```diff
@@ -162,18 +162,9 @@ const UsersTable: React.FC = () => {
                     method: emailSearch,
                 },
             }),
+            buildUserRoleFilter(),
```
rickaw, line 165: does this mean `buildFilters` can be non-async anymore?

AmeliaIsCoding: yup

PR #311, opened 2024-05-01
VFB-210: Audit log list order change

AmeliaIsCoding:
## What's changed
Send an audit log when a list item is moved up or down.

Please note: I decided to only send an audit log for the list item that was clicked, so that we have one audit log per list item order change, to be consistent with packing slots. (Opposed to two audit logs, one for each list item that moves - the one that is clicked and the one that it swaps with).

## Checklist

No DB changes.


AmeliaIsCoding (commented):
Self-review

rickaw (approved):

Issue #232, opened 2024-04-04
Calendar Page - Week View - Change Date format to day/month instead of month/day


dan-langridge:
### Describe the solution / feature you'd like

On the Calendar page, on the Week and Day View, the format is in US Month/Day. Please can we change it to UK Day/Month
<img width="1543" alt="Screenshot 2024-04-04 at 17 08 53" src="https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/80202270/f42f37c8-79fb-4adf-99c8-e310d1a4d99f">

maxjohns0n:
### QA

- Navigate to calendar page
- Switch to "week" view
  - [ ] Dates are in UK format (day/month instead of month/day)
- Switch to "day" view
  - [ ] Title is in UK format (e.g. 2 May 2024 instead of May 2, 2024)
- For all three views (month/week/day):
  - [ ] Monday is the first day of the week
  - [ ] Times are displayed as 24h

AmeliaIsCoding:
QA passed! Well done for your first merge! <3

PR #313, opened 2024-05-02
#232: Use en-GB locale for calendar

maxjohns0n:
## What's changed
* Set the calendar to use en-GB locale so that the date is formatted correctly (day/month).
  * This also makes Monday the first day of the week
* Kept the times as 12h (this overrides the locale defaults)
* Also extracted the `eventTimeFormat` specification to apply to all views because this was duplicated

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

No database changes

### DIFF
--- a/src/components/Calendar/Calendar.tsx
+++ b/src/components/Calendar/Calendar.tsx
@@ -252,11 +252,6 @@ const Calendar: React.FC<CalendarProps> = ({
                     initialDate={initialDate}
                     views={{
                         dayGridMonth: {
-                            eventTimeFormat: {
-                                hour: "numeric",
-                                minute: "2-digit",
-                                meridiem: "short",
-                            },
                             titleFormat: { year: "numeric", month: "short" },
                             dayMaxEventRows: 3,
                             moreLinkClick: "day",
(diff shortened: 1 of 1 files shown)


maxjohns0n:
Changing the locale also means Monday is the first day of the week, and all times are in 24h instead of 12h. Is this what we want? It is consistent with how times are displayed elsewhere (e.g. audit logs).

AmeliaIsCoding (changes requested):
Nice, but I think maybe we should stick with displaying 12 hour times. Is there a way to get both the date as DD/MM and the time as 12 hours? Maybe in the `views` prop there is some config we could use that allows this?

maxjohns0n:
> Nice, but I think maybe we should stick with displaying 12 hour times. Is there a way to get both the date as DD/MM and the time as 12 hours? Maybe in the `views` prop there is some config we could use that allows this?

Ok yes I found a way to keep all the times as they were before!

AmeliaIsCoding (approved):
Great! I like that now we only specify eventTimeFormat once :)

minalui (approved):
Looks great :)

Issue #181, opened 2024-03-23
Driver Overview PDF - Driver Name should be Mandatory Field


dan-langridge:
### Reproduction steps

Select Multiple Clients
Download Driver Overview PDF
Leave Driver Name Blank
Submit
Success

### Expected behaviour

Driver Name should be Mandatory Field so we can track who has the parcel

### What browsers are you seeing the problem on?

Safari

### What OS is your computer running on?

macOS

maxjohns0n:
### QA

- Navigate to parcels page
- Select one or more parcels in the table
- From "Actions", choose "Download driver overview"
- [ ] "Download PDF" button is displayed as being disabled
- [ ] Can't click on the button/it doesn't do anything
- Enter some text into the Driver's Name field
- [ ] Button shows as enabled
- [ ] Clicking the button submits the form and generates a correct PDF
- [ ] Deleting all the text from the Driver's Name field disables the button again

PR #314, opened 2024-05-02
#181: Driver name should be mandatory field

maxjohns0n:
## What's changed
Disable submit button when creating driver overview pdf when driver name is blank.
(Followed approach in ShippingLabelModal)

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image image |

## Checklist

no db changes

### DIFF
--- a/src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
+++ b/src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
@@ -32,7 +32,7 @@ const DriverOverviewInput: React.FC<DriverOverviewInputProps> = ({
     return (
         <>
             <Heading>Delivery Information</Heading>
-            <FreeFormTextInput onChange={onDriverNameChange} label="Driver's Name" />
+            <FreeFormTextInput onChange={onDriverNameChange} label="Driver's Name (required)" />
             <DatePicker
                 defaultValue={dayjs()}
                 onChange={onDateChange}
--- a/src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
+++ b/src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
@@ -75,6 +75,8 @@ const DriverOverviewModal: React.FC<ActionModalProps> = (props) => {
 
     const [isDateValid, setIsDateValid] = useState(true);
 
+    const isInputValid = isDateValid && driverName.length > 0;
+
     const onDriverNameChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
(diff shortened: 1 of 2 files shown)


minalui (changes requested):
Looks good - can you add (required) after the drivers name in the placeholder text so that it's more obvious to the user?

maxjohns0n:
> Looks good - can you add (required) after the drivers name in the placeholder text so that it's more obvious to the user?

Done!

minalui (commented):
LGTM 🙂 @AmeliaIsCoding?

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
```diff
@@ -32,7 +32,7 @@ const DriverOverviewInput: React.FC<DriverOverviewInputProps> = ({
     return (
         <>
             <Heading>Delivery Information</Heading>
-            <FreeFormTextInput onChange={onDriverNameChange} label="Driver's Name" />
+            <FreeFormTextInput onChange={onDriverNameChange} label="Driver's Name (required)" />
```
AmeliaIsCoding, line 35: I know Mina asked for the (required) bit, but seeing as shipping labels modal doesn't have this, I'm wondering if we should be consistent so either:

1. no (required) in either modal
2. (required) in both modals

personally I would lean towards 1 because less text and I think it's obvious enough from the disabling/enabling of the download button. But I'm also supportive of 2 if you think it's right @minalui @maxjohns0n  what are your thoughts? i'm not strongly fussed but I think we should be consistent

AmeliaIsCoding (changes requested):
Great, this is exactly the right approach! Just one comment about consistency with the shipping labels modal and once we've decided on an approach and implemented I will approve

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
minalui: Good point - I feel that because there are two inputs in the driver overview modal its better to make it explicit. Happy for the change to be applied to the shipping labels modal.

AmeliaIsCoding: Cool sounds good, @maxjohns0n please also add the (required) text to the shipping labels modal

maxjohns0n: Done! I agree it seems better to be explicit and consistent :)

AmeliaIsCoding (approved):
LGTM

minalui (approved):
LGTM :)

PR #317, opened 2024-05-03
VFB-211: Enforcing that we sort and filter by a DB column and table refactoring

AmeliaIsCoding:
## What's changed

- if we try to sort by a column that's not a column in the DB table, there is an error
- if we try to filter by a column that's not a column in the DB table, there is not an error, but, the code automatically suggests only columns that are in the DB table. I'm genuinely not sure if there is a way to get an error to be thrown - I think this is because it is possible to filter by an arbitary string if we use the ".or" filtering method
- type cleanup: distinguish between server side and client side filters and sorting methods. Table props for filtering and sorting must be consistent with one of these (i.e. you can't have server side filters and client side sorts)
- refactoring of server side paginated tables - parcels, clients, audit logs and users. Split these tables up into files, generally with files for filters, sorts, headers, constants, and the table component itself. General clean up of code in these files.
- handle error instead of throwing error when we fetch the client ID of the selected parcel in the parcel modal

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|imageimage|imageimageimage|

## Checklist

No DB changes.


src/app/parcels/parcelsTable/ParcelsPage.cy.tsx
```diff
@@ -103,11 +98,6 @@ describe("Parcels Page", () => {
             expect(formatDatetimeAsDate(null)).to.eq("-");
         });
 
-        it("datetimeToPackingTimeLabel()", () => {
-            expect(datetimeToPackingTimeLabel("2023-08-04T08:30:00+00:00")).to.eq("AM");
-            expect(datetimeToPackingTimeLabel("2023-08-04T13:30:00+00:00")).to.eq("PM");
-        });
```
AmeliaIsCoding, line 109: This test is no longer needed as the datetimeToPackingTimeLabel function is no longer used

src/components/Tables/Filters.tsx
```diff
@@ -8,25 +8,32 @@ export enum PaginationType {
...
-    | {
-          method: (row: Data, state: State, key: keyof Data) => boolean;
-          paginationType: PaginationType.Client;
-      };
-
-export interface Filter<Data, State> {
+export type ServerSideFilterMethod<DbData extends Record<string, any>, State> = (
+    query: PostgrestFilterBuilder<Database["public"], DbData, any>,
+    state: State
+) => PostgrestFilterBuilder<Database["public"], DbData, any>;
```
AmeliaIsCoding, line 14: the actual change that results in the objective of the ticket being satisfied (in so far as the PR description) is passing in DbData as the second type parameter to the PostgrestFilterBuilder type

src/components/Tables/Table.tsx
```diff
@@ -199,7 +198,11 @@ const defaultColumnStyleOptions = {
     maxWidth: "20rem",
 } as const;
 
-const Table = <Data,>({
+const Table = <
+    Data,
+    PaginationType extends PaginationTypeEnum,
+    DbData extends Record<string, any> = {},
```
AmeliaIsCoding, line 204: we only need DbData if PaginationType is PaginationType.Server, but I'm not sure if its possible to specify that :/ let me know your thoughts

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
```diff
@@ -64,12 +63,36 @@ export const filtersToComponents = <Data,>(
     });
 };
 
-const TableFilterAndExtraColumnsBar = <Data,>(props: Props<Data>): React.ReactElement => {
+const TableFilterAndExtraColumnsBar = <Data, Filter extends CoreFilter>(
+    props: Props<Data, Filter>
+): React.ReactElement => {
+    const handleClear = (): void => {
```
AmeliaIsCoding, line 69: I moved the definition of this function inside this component, instead of passing it as a prop from Table

src/components/Tables/sortMethods.ts
```diff
@@ -0,0 +1,10 @@
+import { Database } from "@/databaseTypesFile";
+import { SortOrder } from "react-data-table-component";
+import { PostgrestFilterBuilder } from "@supabase/postgrest-js";
+
+export type ServerSideSortMethod<DbData extends Record<string, any>> = (
+    query: PostgrestFilterBuilder<Database["public"], DbData, any>,
+    sortDirection: SortOrder
+) => PostgrestFilterBuilder<Database["public"], DbData, any>;
```
AmeliaIsCoding, line 8: again, the second type parameter here is what fulfils the goal of the ticket

src/components/Tables/Filters.tsx
```diff
@@ -44,3 +51,9 @@ export const defaultToString = (value: unknown): string => {
 
     return JSON.stringify(value);
 };
+
+export interface CoreFilter<State = any> {
```
AmeliaIsCoding, line 55: we use this in TableFilterAndExtraColumnsBar.tsx

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
```diff
@@ -47,9 +46,9 @@ const Grow = styled.div`
     flex-grow: 1;
 `;
 
-export const filtersToComponents = <Data,>(
-    filters: Filter<Data, any>[],
-    setFilters: (filters: Filter<Data, any>[]) => void
+export const filtersToComponents = <Filter extends CoreFilter>(
```
AmeliaIsCoding, line 49: we need extends CoreFilter because we need to know that a filter has a filter component

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
```diff
@@ -64,12 +63,36 @@ export const filtersToComponents = <Data,>(
...
+                    ...filter,
+                    state: filter.initialState,
+                }))
+            );
+        props.setAdditionalFilters &&
+            props.additionalFilters &&
+            props.setAdditionalFilters(
+                props.additionalFilters?.map((filter) => ({
+                    ...filter,
+                    state: filter.initialState,
```
AmeliaIsCoding, line 83: we need extends CoreFilter because we need to know that the filter has a state and an initialState

AmeliaIsCoding (commented):
Self-review 1

src/app/parcels/parcelsTable/format.tsx
AmeliaIsCoding, line 1: just moved everything in here and capitalised the components which use useTheme() because they need to be capitalised to use a React hook.

src/databaseUtils.ts
```diff
@@ -31,5 +31,3 @@ export type ViewNames = keyof Views;
 export type ParcelsPlusRow = Views["parcels_plus"]["Row"];
 
 export type ParcelStatus = Schema["status_order"]["event_name"];
-
-export type AuditLogPlusDbRow = ViewSchema["audit_log_plus"];
```
AmeliaIsCoding: I decided to move this out and define the Db Data types in the types file for each server side table. I think this is better because this file will be less cluttered and the types relating to the specific tables will be defined in the types file for that table which I think makes more sense

AmeliaIsCoding (commented):
Self review 2

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -75,22 +76,23 @@ const ParcelsPage: React.FC<{}> = () => {
...
-        getClientIdForSelectedParcel(parcelId)
-            .then((clientId) => setClientIdForSelectedParcel(clientId))
-            .catch((error) => {
-                if (error instanceof Error) {
-                    setErrorMessage(error.message);
-                }
-            });
+        const { clientId, error } = await getClientIdForSelectedParcel(parcelId);
+        if (error) {
+            setModalErrorMessage(getClientIdErrorMessage(error));
```
AmeliaIsCoding, line 92: we need to display this error message in the modal otherwise it's not visible until the modal is closed

src/app/parcels/parcelsTable/fetchParcelTableData.ts
```diff
@@ -263,18 +264,20 @@ export const fetchParcelStatuses = async (): Promise<ParcelStatusesReturnType> =
     return { data: parcelStatusesList, error: null };
 };
 
-export const getClientIdForSelectedParcel = async (parcelId: string): Promise<string> => {
-    const { data, error } = await supabase
-        .from("parcels")
+export const getClientIdForSelectedParcel = async (
```
AmeliaIsCoding, line 267: I moved this function into this file and I handled the error instead of throwing it

AmeliaIsCoding (commented):
Self-review 3

AmeliaIsCoding (commented):
Self-review 4!

src/databaseUtils.ts
rickaw: I actually think `AuditLogPlusDbRow` can stay in this file. I know that parcels_plus stuff happens to be used in just the parcels page, but at the end of the day, `AuditLogPlusDbRow` is a db type, and it doesn't make sense to define this in a file that primarily deals with front-end stuff.

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
```diff
@@ -64,12 +63,36 @@ export const filtersToComponents = <Data,>(
...
+
     const [showMoreFiltersAndHeaders, setShowMoreFiltersAndHeaders] = useState(false);
 
-    const hasPrimaryFilters = props.filters.length !== 0;
+    const hasPrimaryFilters = props.filters && props.filters?.length !== 0 && props.setFilters;
 
-    const hasAdditionalFilters = props.additionalFilters.length !== 0;
+    const hasAdditionalFilters =
+        props.additionalFilters &&
+        props.additionalFilters?.length !== 0 &&
```
rickaw, line 94: can be simplified i think
```suggestion
        props.additionalFilters?.length > 0 &&
```

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
```diff
@@ -87,7 +110,10 @@ const TableFilterAndExtraColumnsBar = <Data,>(props: Props<Data>): React.ReactEl
                 {hasPrimaryFilters && <FilterAltOutlined />}
                 <FilterContainer>
                     <>
-                        {filtersToComponents(props.filters, props.setFilters)}
+                        {props.filters &&
+                            props.filters?.length !== 0 &&
```
rickaw, line 114:
```suggestion
                        { props.filters?.length > 0 &&
```

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
```diff
@@ -117,14 +143,16 @@ const TableFilterAndExtraColumnsBar = <Data,>(props: Props<Data>): React.ReactEl
...
-                            {props.additionalFilters.length > 0 && (
-                                <>
-                                    {filtersToComponents(
-                                        props.additionalFilters,
-                                        props.setAdditionalFilters
-                                    )}
-                                </>
-                            )}
+                            {props.additionalFilters &&
+                                props.additionalFilters?.length !== 0 &&
```
rickaw, line 147:
```suggestion
                            { props.additionalFilters?.length > 0 &&
```

src/components/Tables/Table.tsx
```diff
@@ -386,41 +401,26 @@ const Table = <Data,>({
 
     const rows = dataPortion.map((data, index) => ({ rowId: index, data }));
 
-    const handleClear = (): void => {
```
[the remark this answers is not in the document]
rickaw: nvm i see you've moved it!

src/components/Tables/Table.tsx
AmeliaIsCoding: I like the second solution, have done

src/components/Tables/Filters.tsx
AmeliaIsCoding: CoreFilter is only used in the TableFilterAndExtraColumnsBar to extend the Filter type

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
AmeliaIsCoding: yes, although are these `any`'s okay?

src/components/Tables/Filters.tsx
AmeliaIsCoding: have deleted as per the below suggestion

src/databaseUtils.ts
AmeliaIsCoding: Okay, I've moved all of the DbRow type definitions in here

src/components/Tables/Filters.tsx
```diff
@@ -8,25 +8,32 @@ export enum PaginationType {
...
+
+export interface ClientSideFilter<Data, State> {
     key: keyof Data;
     filterComponent: (state: State, setState: (state: State) => void) => React.ReactElement;
     state: State;
     initialState: State;
-    methodConfig: MethodConfig<Data, State>;
+    method: ClientSideFilterMethod<Data, State>;
     areStatesIdentical: (stateA: State, stateB: State) => boolean;
 }
```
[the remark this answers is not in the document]
AmeliaIsCoding: sure have done

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
AmeliaIsCoding: if I use `!== 0` it works but if I use `> 0` I get an error saying props.additionalFilters.length might be undefined, so have used the former

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
AmeliaIsCoding: as above, I need to use `!==`

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
AmeliaIsCoding: It doesn't recognise that additionalFilters is not undefined if I do this

src/components/Tables/sortMethods.ts
```diff
@@ -0,0 +1,10 @@
+import { Database } from "@/databaseTypesFile";
+import { SortOrder } from "react-data-table-component";
+import { PostgrestFilterBuilder } from "@supabase/postgrest-js";
+
+export type ServerSideSortMethod<DbData extends Record<string, any>> = (
```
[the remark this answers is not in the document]
AmeliaIsCoding: It works instead of `any` in line 5

src/app/admin/usersTable/filters.ts
```diff
@@ -0,0 +1,65 @@
...
+        filterLabel: "User Role",
+        itemLabelsAndKeys: allUserRoles.map((userRole) => [userRole, userRole]),
+        initialCheckedKeys: allUserRoles,
+        method: userRoleSearch,
+    });
+
+    return userRoleFilter;
+};
+
+export const usersFilters: UsersFilter<any>[] = [
```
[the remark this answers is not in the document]
AmeliaIsCoding: the second change would be big. I'm going to go for the first one. will do for the other tables

src/app/clients/clientsTable/types.ts
```diff
@@ -0,0 +1,46 @@
+import { ServerSideFilter, ServerSideFilterMethod } from "@/components/Tables/Filters";
+import { SortState } from "@/components/Tables/Table";
+import { ServerSideSortMethod } from "@/components/Tables/sortMethods";
+import { ViewSchema } from "@/databaseUtils";
+
+export type DbClientRow = ViewSchema["clients_plus"];
+export type ClientsFilterMethod<State = any> = ServerSideFilterMethod<DbClientRow, State>;
+export type ClientsFilter<State = any> = ServerSideFilter<ClientsTableRow, State, DbClientRow>;
```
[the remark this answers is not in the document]
AmeliaIsCoding: Yes, have done. And same for Parcels filters

AmeliaIsCoding (commented):
Self-review post review changes

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
rickaw: ah fair enough - happy to keep it as is!

src/components/Tables/sortMethods.ts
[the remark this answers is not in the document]
rickaw: ideally other `any`s too, but happy to keep them for now for the intereset of time!

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
rickaw: we can't use `unknown` here as Filter extends that type but would be a type with specific generic parameter i.e. that specific filter doesn't conform to `ClientSideFilter<unknown, unknown>` for example. (if i'm making sense)

so yeah happy to use any in this case.

rickaw (approved):

8kristy (approved):

src/components/Tables/sortMethods.ts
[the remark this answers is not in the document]
AmeliaIsCoding: i'm glad we fixed these, i think some of the other `any`s have been kicking around for a while and are a bigger change so maybe something for the future

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
AmeliaIsCoding: very interesting ty

Issue #205, opened 2024-03-26
Parcels Page > Voucher Search Box > Search '?' to find all clients with a missing voucher number


dan-langridge:
### Describe the solution / feature you'd like

We have this functionality on App1 which some users find very helpful.

By entering a question mark '?' into the Voucher Search box, it will show you all clients that have no voucher.

<img width="1478" alt="Screenshot 2024-03-26 at 16 48 14" src="https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/80202270/18bc9fc2-4100-495f-a753-c30bd47b1af3">

rickaw:
## Developer Notes
Should be implemented after #204

KylePUG:
## QA

- Navigate to the "Parcels" page
- Click on "More" under filters
- Click on "Select Columns" and check "Vouchers" to ease visibility
- [ ] Check that there are several parcels that have empty voucher fields and note them down
  - If these don't exist, choose a few parcels and empty the voucher fields by editing the parcel
- [ ] Type "?" into the voucher search field to get the parcels with empty voucher fields

PR #320, opened 2024-05-09
#205: Question mark in voucher search field returns parcels with empty voucher fields

KylePUG:
## What's changed
- Entering a '?' in the search field for "Voucher" now shows all parcels with empty voucher fields

## Screenshots / Videos
Without filtering (for reference):
image

| Before     | After      |
|------------|------------|
|image|image|

## Checklist

If you have made any changes to the database... (no changes)

### DIFF
--- a/src/app/parcels/parcelsTableFilters.ts
+++ b/src/app/parcels/parcelsTableFilters.ts
@@ -77,6 +77,9 @@ export const voucherSearch = (
     query: PostgrestFilterBuilder<Database["public"], any, any>,
     state: string
 ): PostgrestFilterBuilder<Database["public"], any, any> => {
+    if (state === '?') {
+        return query.ilike("voucher_number", ``);
+    }
     return query.ilike("voucher_number", `%${state}%`);
 };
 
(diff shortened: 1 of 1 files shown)


KylePUG:
## Possible pitfalls

This should work for all parcels with **empty** voucher field. This means that if spaces were added accidentally entered into the parcel's information it will not come up when searching "?". I'm not sure if I should cover this case - it seems like padding the beginning and end of a string when inserted into any field to begin with could be something to look at, if it hasn't been done already.

minalui:
Maybe we should consider trimming the voucher number in the parcels form before adding the parcel to the database? Would this be enough to prevent empty voucher numbers made out of spaces? 🤔

KylePUG:
Agreed. I do think that that will be sufficient. Although it also seems like something that all other fields in the parcel form would benefit from. What should be our next steps?

rickaw:
@KylePUG Do you want to make a new ticket on Jira to trim the form inputs, and merge this one in the meantime?

rickaw (approved):

KylePUG:
Done - added to backlog on Jira and merged

Issue #200, opened 2024-03-26
Day Overview PDF - Show Congestion Charge Logo next to applicable Postcodes


dan-langridge:
### Describe the solution / feature you'd like

On the Day Overview PDF - Show the Congestion Charge Logo next to Postcodes that are in the congestion zone

kaydentsoftwire:
QA Steps: 
- navigate to parcels page
- select multiple parcels, including at least: 1 for delivery that is in the congestion charge zone,1 for delivery that is not in the congestion charge zone, 1 for collection that is in the CC zone
- click on 'Actions' and then 'Download Day Overview'
- [ ] Parcel for delivery that is in the CC Zone should have the icon next to the postcode,
- [ ] Parcel for delivery outside of CC Zone should not have the icon
- [ ] Parcel for collection inside of CC Zone should not have the icon
- [ ] Parcel with no postcode recorded
- [ ] When the congestion charge logo is displayed, it displays in-line with the postcode text at approximately the same height

Error message QA - please do this on local dev
- navigate to parcels page
- in congestionCharges.ts, change 'check-congestion-charge' on line 24 to 'failing'
- [ ] Above the table you should see the error message, "Failed to retrieve congestion charge details"+ Log ID
- revert your changes

There is also an error message for the Day Overview PDF, but requires a few more changes to see it..
- duplicate the checkForCongestionCharge function and call is checkForCongestionChargeSecond
- in DayOverviewPdfButton.tsx, change line 84 to call checkForCongestionChargeSecond
- open the Day Overview PDF to check it is still working and showing Congestion Charge
- change checkForCongestionChargeSecond to invoke 'failing'
- Open Day Overview PDF
- [ ] The modal should show ""Failed to retrieve congestion charge data" + Log ID
- revert changes

Alex-Softwire:
QA: Nice work! Seems to work well and well done with the error handling

✅ - Download Day Overview shows congestions charge only when it is supposed to, and is in line with the text.

- Not sure what the line "parcel with no postcode recorded" means in the qa comment?

✅ - Error message appears on the table if congestion charge is not loaded correctly. 
✅ - Error message appears on modal if congestion charge doesn't load (but the initial table call loads). 

[...]

PR #322, opened 2024-05-10
#200: Show Congestion Charge Logo on Day Overview PDF

kaydentsoftwire:
## What's changed
Congestion Charge logo shows on Day Overview PDF next to Postcode if the collection centre is marked as Delivery and the client's home postcode is in the Congestion Zone

If there is an error retrieving congestion charge details from the database, an error is shown above the table and the table data is not shown.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image image |

## Checklist


src/pdf/FontAwesomeIconPdfComponent.tsx
```diff
@@ -4,15 +4,18 @@ import { ReactElement } from "react";
...
 interface FontAwesomeIconProps {
     faIcon: IconDefinition;
+    color?: string;
 }
 
-const FontAwesomeIconPdfComponent = ({ faIcon: { icon } }: FontAwesomeIconProps): ReactElement => {
+const FontAwesomeIconPdfComponent = ({
+    faIcon: { icon },
+    color = "black",
+}: FontAwesomeIconProps): ReactElement => {
```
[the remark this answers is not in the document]
AmeliaIsCoding: Thank you for explaining!

AmeliaIsCoding (changes requested):
Looks good - I like the checkForCongestionCharge function, and the changes in the PDF look good. well done! 2 main things to change, with details in the review comments:

-Return an error from checkForCongestionCharge instead of throwing, and deal with it gracefully
-I think the getCongestionChargeDetailsForParcelsTable and addCongestionChargeDetailsForDayOverview should be moved back to where they were

src/pdf/DayOverview/DayOverviewPdf.tsx
```diff
@@ -22,7 +23,7 @@ interface DayOverviewPdfProps {
 interface CustomSVGProps {
     icon: IconDefinition;
     color: string;
-    fill: boolean;
+    fill?: boolean;
```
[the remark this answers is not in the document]
kaydentsoftwire: Nowhere, whoops! Have changed this back

src/app/congestionCharges.ts
```diff
@@ -0,0 +1,62 @@
...
+    const response = await supabase.functions.invoke("check-congestion-charge", {
+        body: { postcodes: postcodes },
+    });
+
+    if (response.error) {
+        const logId = await logErrorReturnLogId(
+            "Error with congestion charge check",
+            response.error
+        );
+        throw new EdgeFunctionError("congestion charge check", logId);
```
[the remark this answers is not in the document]
kaydentsoftwire: Nice, thank you! Should be done :)

src/app/congestionCharges.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    return parcels;
+};
+
+const checkForCongestionCharge = async (
+    postcodes: (string | null)[]
+): Promise<CongestionChargeDetails[]> => {
+    const response = await supabase.functions.invoke("check-congestion-charge", {
+        body: { postcodes: postcodes },
+    });
```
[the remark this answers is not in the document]
kaydentsoftwire: Good Q! I checked this with null postcodes and supabase returns it as false for congestion charge

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -273,9 +273,16 @@ export const getParcelsByIds = async (
...
+        const logId = await logErrorReturnLogId("Errr");
+        throw new EdgeFunctionError("congestion charge check", logId);
+    }
+
     const { parcelTableRows, error: processParcelDataError } =
-        await processingDataToParcelsTableData(data, congestionCharge);
+        await processingDataToParcelsTableData(data, congestionChargeDetails);
 
     if (processParcelDataError) {
         throw new Error("Failed to process parcels.", { cause: processParcelDataError });
```
kaydentsoftwire, line 288: I did do some reordering of functions in this file which I think is what you can see above - mainly because I got some errors saying things were called before they were defined (not totally sure why?) so I tried to put them in order

kaydentsoftwire: Otherwise this is me changing the data structure that is being passed through to contain { data, error } so the error could be rendered

src/app/congestionCharges.ts
```diff
@@ -0,0 +1,62 @@
...
+    const postcodesWithCongestionChargeDetails = await checkForCongestionCharge(postcodes);
+
+    for (let index = 0; index < parcels.length; index++) {
+        parcels[index].congestionChargeApplies =
+            postcodesWithCongestionChargeDetails[index].congestionCharge;
+    }
+
+    return parcels;
+};
+
```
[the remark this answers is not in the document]
kaydentsoftwire: Cool, done!

src/app/congestionCharges.ts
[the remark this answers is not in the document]
AmeliaIsCoding: Cool, thats good enough!

src/app/parcels/ActionBar/ActionModals/DayOverviewModal.tsx
```diff
@@ -13,6 +13,8 @@ const getPdfErrorMessage = (error: DayOverviewPdfError): string => {
         case "parcelFetchFailed":
             errorMessage = "Failed to fetch parcel data.";
             break;
+        case "failedToRetrieveCongestionChargeDetails":
+            errorMessage = "Failed to retrieve congestion charge data";
```
AmeliaIsCoding, line 17: should there be a break statement after this? (i think it's not strictly needed for the last case, but we should defo be consistent and help out any future devs who might add more cases below this one!)

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -131,10 +107,59 @@ type GetParcelDataAndCountResult =
...
+    }
+
+    if (count === null) {
+        const logId = await logErrorReturnLogId("Error with fetch: Parcels, count is null");
+        throw new DatabaseError("fetch", "parcels", logId);
+    }
+    return count;
+};
+
+const getCongestionChargeDetailsForParcelsTable = async (
```
AmeliaIsCoding, line 145: personal preference so up to you but I think I would put this function above the others because it seems less related to the count and data calls

src/common/congestionCharges.ts
```diff
@@ -0,0 +1,37 @@
...
+    postcodes: (string | null)[]
+): Promise<CongestionChargeReturnType> => {
+    const response = await supabase.functions.invoke("check-congestion-charge", {
+        body: { postcodes: postcodes },
+    });
+
+    if (response.error) {
+        const logId = await logErrorReturnLogId(
+            "Error with congestion charge check",
+            response.error
```
AmeliaIsCoding, line 31: Can you please change this line to `{error: response.error}`, this makes the error object in the log more readable

AmeliaIsCoding (changes requested):
Well done on the error handling! A few small things

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -273,9 +273,16 @@ export const getParcelsByIds = async (
         throw new DatabaseError("fetch", "parcel table", logId);
     }
 
-    const congestionCharge = await getCongestionChargeDetailsForParcels(data, supabase);
+    const { data: congestionChargeDetails, error: congestionChargeError } =
+        await getCongestionChargeDetailsForParcelsTable(data);
+
+    if (congestionChargeError) {
+        const logId = await logErrorReturnLogId("Errr");
+        throw new EdgeFunctionError("congestion charge check", logId);
```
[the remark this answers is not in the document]
AmeliaIsCoding: Although if you're really sick of handling errors, it would be ok for you to make a ticket to cover handling instead of throwing errors for all these functions in `fetchParcelTableData.ts`, as there's quite a lot that needs to be fixed here, and maybe it would be good to do it all at once. Happy for either.

src/app/parcels/fetchParcelTableData.ts
kaydentsoftwire: Sounds good, done

src/app/parcels/fetchParcelTableData.ts
kaydentsoftwire: Seems fine now, I was originally trying to do it so it would still show the data and also the error message if it was just congestion charge missing, so maybe it came up at that stage

src/common/congestionCharges.ts
kaydentsoftwire: Done

src/pdf/DayOverview/DayOverviewPdfButton.tsx
```diff
@@ -66,9 +66,37 @@ export interface DayOverviewPdfData {
...
+    }
+
+    const { data: postcodesWithCongestionChargeDetails, error } =
+        await checkForCongestionCharge(postcodes);
+
+    if (error) {
+        return { data: null, error: error };
+    }
+
+    if (postcodesWithCongestionChargeDetails) {
```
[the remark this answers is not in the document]
kaydentsoftwire: Good point - no, have removed

src/pdf/DayOverview/DayOverviewPdfButton.tsx
```diff
@@ -66,9 +66,37 @@ export interface DayOverviewPdfData {
...
+        for (let index = 0; index < parcels.length; index++) {
+            parcels[index].congestionChargeApplies =
+                postcodesWithCongestionChargeDetails[index].congestionCharge;
+        }
```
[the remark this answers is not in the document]
kaydentsoftwire: Done

src/app/parcels/ActionBar/ActionModals/DayOverviewModal.tsx
kaydentsoftwire: Done

src/app/parcels/fetchParcelTableData.ts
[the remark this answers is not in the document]
kaydentsoftwire: Yeah, to be honest I would rather make a ticket and then perhaps tackle it as a separate task - but handling instead of throwing the errors sounds good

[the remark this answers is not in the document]
kaydentsoftwire: Should the DatabaseErrors also be handled similarly or are they okay as is?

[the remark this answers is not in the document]
kaydentsoftwire: Have made a ticket on Jira

src/pdf/DayOverview/DayOverviewPdfButton.tsx
```diff
@@ -66,9 +66,37 @@ export interface DayOverviewPdfData {
...
 export type DayOverviewPdfError = { type: DayOverviewPdfErrorType; logId: string };
 
+const addCongestionChargeDetailsForDayOverview = async (
+    parcels: ParcelForDayOverview[]
+): Promise<ParcelForDayOverviewResponse> => {
+    const postcodes: (string | null)[] = [];
+
+    for (const parcel of parcels) {
+        if (parcel.client?.address_postcode) {
+            postcodes.push(parcel.client?.address_postcode);
```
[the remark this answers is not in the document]
AmeliaIsCoding: can we add a condition into the if to check if the client is not null? it's not that much better but I think it's a bit more readable to have all the conditional code in an if statement, if there is already an if statement!

src/pdf/DayOverview/DayOverviewPdfButton.tsx
```diff
@@ -75,24 +75,24 @@ const addCongestionChargeDetailsForDayOverview = async (
...
-        }
+        parcel.client?.address_postcode
+            ? postcodes.push(parcel.client?.address_postcode)
+            : postcodes.push(null);
     }
 
     const { data: postcodesWithCongestionChargeDetails, error } =
         await checkForCongestionCharge(postcodes);
 
+    console.log(postcodesWithCongestionChargeDetails);
```
AmeliaIsCoding: remove this

src/pdf/DayOverview/DayOverviewPdfButton.tsx
```diff
@@ -75,24 +75,24 @@ const addCongestionChargeDetailsForDayOverview = async (
     const postcodes: (string | null)[] = [];
 
     for (const parcel of parcels) {
```
AmeliaIsCoding: this for loop can also be a map

src/pdf/DayOverview/DayOverviewPdfButton.tsx
```diff
@@ -75,24 +75,24 @@ const addCongestionChargeDetailsForDayOverview = async (
...
-        if (parcel.client?.address_postcode) {
-            postcodes.push(parcel.client?.address_postcode);
-        }
+        parcel.client?.address_postcode
+            ? postcodes.push(parcel.client?.address_postcode)
+            : postcodes.push(null);
```
AmeliaIsCoding: Do we definitely want to push null postcodes? can you explain why please?

src/pdf/DayOverview/DayOverviewPdfButton.tsx
```diff
@@ -75,24 +75,24 @@ const addCongestionChargeDetailsForDayOverview = async (
     const postcodes: (string | null)[] = [];
 
     for (const parcel of parcels) {
-        if (parcel.client?.address_postcode) {
-            postcodes.push(parcel.client?.address_postcode);
-        }
+        parcel.client?.address_postcode
```
AmeliaIsCoding: Personal preference but I think using if/else is preferred here. Using the ternary operator is preferred when we have an expression

src/app/parcels/fetchParcelTableData.ts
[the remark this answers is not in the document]
AmeliaIsCoding: > Should the DatabaseErrors also be handled similarly or are they okay as is?

they also need to be handled. I don't think we should be throwing errors anywhere (again, recognise you weren't the one to write these! just explaining)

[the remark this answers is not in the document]
AmeliaIsCoding: just unresolving so you see this

AmeliaIsCoding (commented):
Nice! just a few small syntax things in the pdf button file

src/app/parcels/fetchParcelTableData.ts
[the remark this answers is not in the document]
kaydentsoftwire: Ok cool, thanks!

src/pdf/DayOverview/DayOverviewPdfButton.tsx
```diff
@@ -64,9 +66,29 @@ export interface DayOverviewPdfData {
...
 
-type DayOverviewPdfErrorType = ParcelForDayOverviewErrorType;
+type DayOverviewPdfErrorType = ParcelForDayOverviewErrorType | CongestionChargeError;
 export type DayOverviewPdfError = { type: DayOverviewPdfErrorType; logId: string };
 
+const addCongestionChargeDetailsForDayOverview = async (
+    parcels: ParcelForDayOverview[]
+): Promise<ParcelForDayOverviewResponse> => {
+    const postcodes = parcels.map((parcel) => parcel.client?.address_postcode || null);
+
```
kaydentsoftwire, line 93: We need it to return null so that we get a false back for that parcel
Otherwise, for example, we may only get back an object with a length of 3 back from the congestion charge check, even though we are checking 4 parcels, and so the parcels and congestion charge data won't line up

src/pdf/DayOverview/DayOverviewPdfButton.tsx
kaydentsoftwire: reply in my recent self-review

src/pdf/DayOverview/DayOverviewPdfButton.tsx
AmeliaIsCoding: Thats great, thanks

AmeliaIsCoding (approved):
Looks good to me. Can you find out if Ricky/Kristina want to review this before you merge?

src/pdf/DayOverview/DayOverviewPdfButton.tsx
AmeliaIsCoding, line 93: just noticed I think it should be ?? not ||
can you make sure the resulting array is as expected?

AmeliaIsCoding (changes requested):
sorry just noticed 1 thing

src/pdf/DayOverview/DayOverviewPdfButton.tsx
kaydentsoftwire: The array is as expected - in this context I believe the two work in the same way?

kaydentsoftwire: (With a slight difference if the postcode was an empty string)

AmeliaIsCoding: Yeah i agree, happy to leave it however you want, I think I would use ?? but i don't think it matters much
happy for you to ask for reviews from Ricky/Kristina now

kaydentsoftwire: Thanks!

AmeliaIsCoding (approved):

src/app/parcels/ActionBar/ActionModals/DayOverviewModal.tsx
```diff
@@ -13,6 +13,9 @@ const getPdfErrorMessage = (error: DayOverviewPdfError): string => {
...
```
8kristy, line 17: From what I understand if something goes wrong with the charges you don't display any data at all on the pdf, just the error? Don't know if that's too much given that the rest of the parcel data is there - would it still be possible to display the table but add a ? or something next to the deliveries or just have the error at the top of the page in big red letters saying something like "Something went wrong while creating this document, and therefore it is not known if the deliveries are in the congestion zone. Try recreating the document later or reference the postcodes manually"

Unless just showing the error message on any error is the standard, in which case ignore me ;)

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -6,33 +6,22 @@ import { SortState } from "@/components/Tables/Table";
...
 
-    if (response.error) {
-        const logId = await logErrorReturnLogId(
-            "Error with congestion charge check",
-            response.error
-        );
-        throw new EdgeFunctionError("congestion charge check", logId);
-    }
-    return response.data;
+    return congestionChargeDetails;
```
8kristy: Is it possible to just do `return await checkForCongestionCharge(postcodes);`? Or is the variable needed for something?

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -273,9 +273,16 @@ export const getParcelsByIds = async (
         throw new DatabaseError("fetch", "parcel table", logId);
     }
 
-    const congestionCharge = await getCongestionChargeDetailsForParcels(data, supabase);
+    const { data: congestionChargeDetails, error: congestionChargeError } =
+        await getCongestionChargeDetailsForParcelsTable(data);
+
+    if (congestionChargeError) {
+        const logId = await logErrorReturnLogId("Errr");
```
8kristy: Extra r? Looking at the `logErrorReturnLogId` above it you might actually want to make this message descriptive of what the error is - maybe double check how it's used elsewhere but I assume for logs you want a reason for it

src/pdf/DayOverview/DayOverviewPdfButton.tsx
8kristy: Do you have an example for this case? In my head if `client == null` then `parcel.client?.address_postcode == null`, `null || null === null` and same with `address_postcode` being null. By "get a false back for that parcel" do you mean that one of those can return `false` rather than `null`? And would the false not be added to the array in that case? Or is there something that filters falses elsewhere (or even me forgetting how map function works)

src/common/congestionCharges.ts
```diff
@@ -0,0 +1,36 @@
...
+          error: null;
+      }
+    | {
+          data: null;
+          error: { type: CongestionChargeError; logId: string };
+      };
+
+export const checkForCongestionCharge = async (
+    postcodes: (string | null)[]
+): Promise<CongestionChargeReturnType> => {
```
8kristy, line 23: Maybe a line break here for slightly better readability?

8kristy (commented):
Code looking good, a few questions

Also looking at the screenshots - is it easy to change the size of the icon and place it at the end of the postcode, sorta like extra text 
image

Especially worried about what happens to the icon when there's not enough space - does it overflow/shrink if all cells in the row are 1 line long so the postcode cell is very narrow

src/pdf/DayOverview/DayOverviewPdfButton.tsx
kaydentsoftwire: Good point! The comments above were from when I was using a for loop instead of map so probably don't make total sense now. 
"get a false back" was in reference to when it is passed to the congestion charge edge function (where null will return { postcode: null, congestionCharge: false }. 
But it's true that removing `|| null` with map would get basically the same result, with only an added possibility of undefined, done :)

src/app/parcels/ActionBar/ActionModals/DayOverviewModal.tsx
kaydentsoftwire: Yeah I had this chat with Ricky! What you are suggesting was what I was originally aiming to do, but it turned out to be more difficult than expected, so that is one reason it has ended up this way - the other is Ricky said the advantage of showing an error message and no data is that the problem is more likely to be reported and fixed.  But I would be interested in finding out how to make this happen

kaydentsoftwire: as in, I think generally the idea is just showing an error message on any error is the standard, but I also think the reason I have done that here is because I couldn't work out how to do it

src/app/parcels/fetchParcelTableData.ts
kaydentsoftwire: Good point, done

src/app/parcels/fetchParcelTableData.ts
kaydentsoftwire: Done!

src/common/congestionCharges.ts
kaydentsoftwire: If I do this then lint_fix changes it back - I guess it doesn't like the first line to be empty?

src/pdf/FontAwesomeIconPdfComponent.tsx
```diff
@@ -4,17 +4,24 @@ import { ReactElement } from "react";
...
+    marginTop?: string;
 }
 
-const FontAwesomeIconPdfComponent = ({ faIcon: { icon } }: FontAwesomeIconProps): ReactElement => {
+const FontAwesomeIconPdfComponent = ({
+    faIcon: { icon },
+    color = "black",
+    styleWidth = "20px",
+    marginTop = "0px",
+}: FontAwesomeIconProps): ReactElement => {
```
kaydentsoftwire, line 17: Added some more optional parameters with the defaults set to reflect what was already hard-coded (so that the other icons are not affected)
styleWidth to make the congestion charge logo smaller, and marginTop to give a bit of margin above it now that it floats to the top instead of being centred

src/pdf/DayOverview/DayOverviewPdf.tsx
```diff
@@ -55,7 +56,13 @@ const styles = StyleSheet.create({
...
     cellName: { flex: 4 },
-    cellPostcode: { flex: 4 },
+    cellPostcode: {
+        flex: 4,
+        display: "flex",
+        flexDirection: "row",
+        alignItems: "flex-start",
+        paddingTop: "1px",
+    },
     cellTime: { flex: 2 },
```
kaydentsoftwire, line 65: Made the postcode cell into a flex box in order to align both the text and the congestion charge logo at the top next to each other

kaydentsoftwire (commented):
Please see the photos in the description for an updated image of what the logo looks like (smaller and in line)

src/common/congestionCharges.ts
8kristy: I guess the linter doesn't agree with my eyes 😞 yeah just leave it in that case

src/app/parcels/ActionBar/ActionModals/DayOverviewModal.tsx
8kristy: Do you know why is it difficult? Take this from someone who doesn't know much about this codebase lol but I'd probably start in `DayOverviewPdfButton.tsx` where you return `congestionChargeError`, and instead of returning there try to pass that error as a new property onto the PDF (e.g. `listOfNonFatalErrors`). Then in the `DayOverviewPdf.tsx` go over the errors, and if there are any, create messages for each code and add them to the top of the file. Then when you do 
```
 {parcel.collection_centre!.name === "Delivery" &&
                    parcel.congestionChargeApplies && (
                        <FontAwesomeIconPdfComponent faIcon={faCopyright} color="red" />
                    )}
```  
I don't know the exact syntax now and you might need to put in some effort to make it nice (perhaps make a new component?) but aim to say something like 
```
if collection_cnetre.name === delivery
   if congestionChargeError in listOfNonFatalErrors
      use grey ? icon
   else if congestionChargeApplies
      use red C icon
```

Don't spend too much time on it if it doesn't work and leave it as is, but if you'd like to try hopefully some of these ideas can help

Not sure if I agree with "problem will get reported faster and fixed if you return nothing" - it will probably get reported very fast yes, but not necessarily fixed fast 😄 and if it's not something that gets fixed by a refresh and needs actual code changes I don't think people using the system would be happy to not be able to print their list for a while so I think leaning towards a more gentle fail might be better, but I guess it depends who you ask

8kristy (approved):
Good job, love the new smaller icons :)

I left a suggestion on how you might be able to not fail the whole page if only congestion charges break during generation, but I'm approving the PR in case you don't have time for it/it doesn't work (and also a reminder to not get into a rabbit hole if it's too difficult - if it takes more than 30min-1hour to do it's fine to leave it)

If you do decide to change it slack me so I can re-review before merging

PR #324, opened 2024-05-13
VFB-171: Change how age is stored in the database

minalui:
## What's changed
- Store birth_year and birth_month in families table instead of age
- Change seed file to use random birth_years instead of age
- Migration script to change age into birth_years
- Change add/edit client form to add birth year for adults and birth year and month for children
- Client details modal and parcel details modal display age of adults and children in the family
- Change to shopping list pdf to show ages of adults and children
- Put modal action buttons in the footer

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image
image
 | image
image
 |

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/clients/edit/[id]/autofill.ts
```diff
@@ -1,18 +1,11 @@
...
 const isAdult = (member: Schema["families"]): boolean => {
-    return member.age === null || member.age >= 16;
-};
-
-const getNumberAdultsByGender = (
-    family: Schema["families"][],
-    gender: DatabaseEnums["gender"]
-): number => {
-    return family.filter((member) => isAdult(member) && member.gender === gender).length;
+    return member.birth_year <= 2008;
```
Comment (minalui) [inline src/app/clients/edit/[id]/autofill.ts:8]: This should be 2007 to catch any 15 y/o that have not had their birthdays yet

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -147,15 +152,49 @@ export const formatHouseholdFromFamilyDetails = (
...
+                    familyMember.birth_year,
+                    false,
+                    familyMember.birth_month
+                );
+                childDetails.push(`${age} ${familyMember.gender}`);
+            } else {
+                const age = getChildAgeUsingBirthYearAndMonth(familyMember.birth_year, false);
+                childDetails.push(`${age} ${familyMember.gender}`);
+            }
         }
```
minalui, line 198: Essentially:
If child age between 16 and 3 then just display age
Otherwise if they are 2 and below and have a birth month, calculate age using birth month and year.
If no birth month then to use just birth year.

src/common/getCurrentYear.ts
minalui, line 1: Rename file to something more inclusive of the functions

src/common/getCurrentYear.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (totalMonths <= 12) {
+        if (shortHand) {
+            return `${totalMonths}m`;
+        }
+        return `${totalMonths}-months-old`;
+    } else {
+        if (shortHand) {
+            return `${years}`;
+        }
+        return totalMonths < 24 ? `${years}-year-old` : `${years}-years-old`;
```
minalui, line 42: If between 1-2 years then display 1-year-old otherwise display x-years-old

src/components/DataInput/DropdownListInput.tsx
minalui, line 1: Remove unused prop

src/app/clients/form/formSections/NumberAdultsCard.tsx
```diff
@@ -1,59 +1,34 @@
...
-        const allAdultEntriesZero =
-            newAdults.numberFemales === 0 &&
-            newAdults.numberMales === 0 &&
-            newAdults.numberUnknownGender === 0;
-
-        let errorType: Errors = Errors.none;
-        if (invalidAdultEntry) {
-            errorType = Errors.invalid;
-        } else if (allAdultEntriesZero) {
-            errorType = Errors.required;
```
rickaw, line 53: where do we handle these error cases now?

src/app/clients/form/formSections/NumberAdultsCard.tsx
```diff
@@ -1,59 +1,34 @@
...
-        return 0;
-    }
-    if (input.match(numberRegex)) {
-        return parseInt(input);
-    }
-    return -1;
-};
-
-const setNumberAdults = (
+const getAdult = (
```
rickaw, line 18: is getAdult the sensible name for this function?

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -118,13 +123,13 @@ export const formatAddressFromClientDetails = (
...
 export const formatHouseholdFromFamilyDetails = (
-    family: Pick<Schema["families"], "age" | "gender">[]
+    family: Pick<Schema["families"], "birth_year" | "gender">[]
 ): string => {
     let adultCount = 0;
     let childCount = 0;
 
     for (const familyMember of family) {
-        if (familyMember.age === null || familyMember.age >= 16) {
+        if (familyMember.birth_year <= getCurrentYear() - 17) {
```
rickaw, line 132: i think it might be worth pulling this logic of `determining whether a person is an adult` into a function as it's used in a few places with 16 / 17.

src/common/getAgesOfFamily.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export const isChildUsingBirthYear = (birthYear: number): boolean => {
+    return getCurrentYear() - birthYear <= 16;
+};
+
+export const getChildAgeUsingBirthYearAndMonth = (
+    birthYear: number,
+    shortHand: boolean,
+    birthMonth?: number
```
rickaw, line 17: it might make sense to have parameters in this order and make birthMonth nullable? Or at this point, we could just pass a `Person` as a parameter instead?
```suggestion
    birthYear: number,
    birthMonth: number | null
    shortHand: boolean,
```

src/common/getAgesOfFamily.ts
```diff
@@ -0,0 +1,49 @@
...
+            return `${totalMonths}m`;
+        }
+        return `${totalMonths}-month-old`;
+    }
+    if (totalMonths <= 12) {
+        if (shortHand) {
+            return `${totalMonths}m`;
+        }
+        return `${totalMonths}-months-old`;
+    } else {
```
rickaw, line 38: i don't think you need this `else` :)

src/common/formatFamiliesData.ts
```diff
@@ -1,28 +1,84 @@
...
 
 export interface HouseholdSummary {
     householdSize: string;
     genderBreakdown: string;
+    ageAndGenderOfAdults: string;
     numberOfBabies: string;
     ageAndGenderOfChildren: string;
 }
 
 const getChild = (child: Person): string => {
```
rickaw, line 18: is getChild a good name here?

supabase/seed/seed.mts
```diff
@@ -67,7 +67,8 @@ async function generateSeed(): Promise<void> {
     await seed.families(
         (generate) =>
             generate(1000, {
-                age: (ctx) => copycat.int(ctx.seed, { min: 0, max: 100 }),
+                birthYear: (ctx) => copycat.int(ctx.seed, { min: 1901, max: 2024 }),
+                birthMonth: null,
```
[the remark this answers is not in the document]
minalui: I've added babies 2024 Jan - May

src/app/clients/edit/[id]/autofill.ts
```diff
@@ -1,18 +1,11 @@
...
 const isAdult = (member: Schema["families"]): boolean => {
-    return member.age === null || member.age >= 16;
-};
-
-const getNumberAdultsByGender = (
-    family: Schema["families"][],
-    gender: DatabaseEnums["gender"]
-): number => {
-    return family.filter((member) => isAdult(member) && member.gender === gender).length;
+    return member.birth_year <= 2007;
```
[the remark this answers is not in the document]
Comment (minalui) [reply src/app/clients/edit/[id]/autofill.ts]: Yes sorry should have spotted this!

src/app/clients/form/birthYearDropdown.ts
```diff
@@ -0,0 +1,37 @@
+const currentDate = new Date();
+const currentYear = currentDate.getFullYear();
```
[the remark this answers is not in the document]
minalui: Good point, hadn't considered that!

src/app/clients/form/formSections/NumberAdultsCard.tsx
minalui: Overlooked but put back in now

src/app/clients/form/birthYearDropdown.ts
```diff
@@ -1,21 +1,25 @@
-const currentDate = new Date();
-const currentYear = currentDate.getFullYear();
+const getCurrentYear = (): number => {
+    const currentDate = new Date();
+    return currentDate.getFullYear();
+};
```
rickaw: i think this exact logic exists in getAgesOfFamily.ts too, can we commonise this? maybe create a file like `src/common/date.ts`?

src/app/clients/form/birthYearDropdown.ts
```diff
@@ -1,21 +1,25 @@
...
 
-export const adultBirthYears: string[] = Array.from({ length: 120 }, (_, index) => {
-    return (currentYear - index - 16).toString();
-});
-
-export const adultBirthYearList: [string, string][] = adultBirthYears.map((year) => {
-    return [`${year}`, `${year}`];
-});
+export const youngestAdultBirthYear = getAdultBirthYears()[0];
```
rickaw: let's make this a function so that it gets computed every time instead of relying on the value at the build time.
```suggestion
export const getYoungestAdultBirthYear = () => getAdultBirthYears()[0];
```

src/common/formatFamiliesData.ts
```diff
@@ -28,13 +28,13 @@ const getChild = (child: Person): string => {
...
-    if (child.birthMonth) {
-        age = getChildAgeUsingBirthYearAndMonth(child.birthYear, true, child.birthMonth);
-    } else {
-        age = getChildAgeUsingBirthYearAndMonth(child.birthYear, true);
+
+    if (!child.birthMonth) {
+        child.birthMonth = null;
     }
 
+    const age = getChildAgeUsingBirthYearAndMonth(child.birthYear, child.birthMonth, true);
```
rickaw:
```suggestion
    const age = getChildAgeUsingBirthYearAndMonth(child.birthYear, child.birthMonth ?? null, true);
```

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -181,18 +182,26 @@ export const formatBreakdownOfChildrenFromFamilyDetails = (
...
             familyMember.birth_year < getCurrentYear() - 2
         ) {
-            const age = getChildAgeUsingBirthYearAndMonth(familyMember.birth_year, false);
+            const age = getChildAgeUsingBirthYearAndMonth(
+                familyMember.birth_year,
+                familyMember.birth_month,
+                false
+            );
             childDetails.push(`${age} ${familyMember.gender}`);
         } else if (familyMember.birth_year >= getCurrentYear() - 2) {
```
rickaw: should we also have a `isBaby` function?

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -181,18 +182,26 @@ export const formatBreakdownOfChildrenFromFamilyDetails = (
...
                 childDetails.push(`${age} ${familyMember.gender}`);
             } else {
-                const age = getChildAgeUsingBirthYearAndMonth(familyMember.birth_year, false);
+                const age = getChildAgeUsingBirthYearAndMonth(
+                    familyMember.birth_year,
+                    familyMember.birth_month,
+                    false
+                );
                 childDetails.push(`${age} ${familyMember.gender}`);
             }
```
rickaw: i think the if and else do exactly the same thing, so why not just remove the if statement?

src/common/formatFamiliesData.ts
```diff
@@ -1,28 +1,81 @@
...
+            gender = "F";
+            break;
+        case "other":
+            gender = "-";
+            break;
+    }
+
+    if (!child.birthMonth) {
+        child.birthMonth = null;
+    }
```
rickaw: I don't think you need this bit anymore

src/app/clients/form/formSections/NumberAdultsCard.tsx
```diff
@@ -1,59 +1,41 @@
...
-import { ClientCardProps, ClientErrorSetter, ClientSetter } from "../ClientForm";
+import { FormText, StyledCard } from "@/components/Form/formStyling";
+import { ClientCardProps, ClientSetter } from "../ClientForm";
+import DropdownListInput from "@/components/DataInput/DropdownListInput";
+import { getAdultBirthYears } from "@/app/clients/form/birthYearDropdown";
 
-const getNumberAdultsOfGenderDefault = (numberAdultsOfGender: number): string | undefined => {
-    return numberAdultsOfGender === 0 ? undefined : numberAdultsOfGender.toString();
+const numberOfAdultsRange = (value: string): boolean => {
+    return parseInt(value) <= 20 && parseInt(value) > 0;
```
8kristy: Is it possible for `value` to not be a number here?

8kristy: Also should we have some constants at the top of the file for min and max adults?

src/app/clients/form/formSections/NumberChildrenCard.tsx
```diff
@@ -79,32 +90,39 @@ const NumberChildrenCard: React.FC<ClientCardProps> = ({
...
+                                onChange={setChildrenFields(
+                                    fieldSetter,
+                                    fields.children,
+                                    index,
+                                    "birthYear"
+                                )}
+                            />
+                            <DropdownListInput
+                                selectLabelId="children-birth-month-select-label"
+                                labelsAndValues={childBirthMonthList}
```
8kristy: If the current year is selected, should we remove future months? e.g. so you can't enter a baby born in December 2024

src/common/getAgesOfFamily.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (totalMonths <= 12) {
+        if (shortHand) {
+            return `${totalMonths}m`;
+        }
+        return `${totalMonths}-months-old`;
+    }
+    if (shortHand) {
+        return `${years}`;
+    }
+    return totalMonths < 24 ? `${years}-year-old` : `${years}-years-old`;
```
8kristy:
```suggestion
    const years = getCurrentYear() - birthYear;
    const months = birthMonth ? getCurrentMonth() - birthMonth : 0;
    const totalMonths = years * 12 + months;
    
    if (totalMonths <= 1) {
        return shortHand ? `${totalMonths}m` : `${totalMonths}-month-old`;
    }
    if (totalMonths <= 12) {
        return shortHand ? `${totalMonths}m` : return `${totalMonths}-months-old`;
    }
    if (shortHand) {
        return `${years}`;
    }
    return totalMonths < 24 ? `${years}-year-old` : `${years}-years-old`;
```

8kristy: I'd double check that suggestion works as intented but hopefully the shorter version is a bit easier to read

Some other bits
- should it be `totalMonths < 12`? I'd put it as 1-year-old
- should the shorthand for above 2 years be `${years}y` (with a y at the end)?

And generally how much does the age matter for adults? Because without the month the estimates might be inaccurate, e.g. If I put in 2000 as the year but my birthday is in September it will say I'm 24 but that's incorrect. But I don't know how much that matters at all

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -150,15 +157,45 @@ export const formatHouseholdFromFamilyDetails = (
     return `${familyCategory} ${occupantDisplay} (${adultChildBreakdown.join(", ")})`;
 };
 
+export const formatBreakdownOfAdultsFromFamilyDetails = (
+    family: Pick<Schema["families"], "birth_year" | "gender">[]
+): string => {
+    const adultDetails = [];
+
+    for (const familyMember of family) {
+        if (familyMember.birth_year < getCurrentYear() - 16) {
```
8kristy: Would it be better to use `isAdult()` here?

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -150,15 +157,45 @@ export const formatHouseholdFromFamilyDetails = (
...
+                false
+            );
+            childDetails.push(`${age} ${familyMember.gender}`);
+        } else if (isBaby(familyMember.birth_year)) {
+            const age = getChildAgeUsingBirthYearAndMonth(
+                familyMember.birth_year,
+                familyMember.birth_month,
+                false
+            );
+            childDetails.push(`${age} ${familyMember.gender}`);
```
8kristy: The 2 ifs look the same to me? Is it not enough to just have 1 that checks `!isAdult()`? Sorry if I'm missing the difference

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -351,39 +351,37 @@ const ParcelsPage: React.FC<{}> = () => {
...
+                                )}
+                            </Centerer>
+                        }
                     >
                         <OutsideDiv>
                             <ContentDiv>
                                 <Suspense fallback={<ExpandedParcelDetailsFallback />}>
                                     <ExpandedParcelDetails parcelId={selectedParcelId} />
                                 </Suspense>
                             </ContentDiv>
```
8kristy, line 383: Why did this get moved?

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -351,39 +351,37 @@ const ParcelsPage: React.FC<{}> = () => {
...
-                            <ButtonsDiv>
-                                <Centerer>
-                                    <LinkButton link={`/parcels/edit/${selectedParcelId}`}>
-                                        Edit Parcel
-                                    </LinkButton>
-                                    {selectedClientDetails && (
-                                        <>
-                                            <LinkButton
-                                                link={`/clients?clientId=${selectedClientDetails.clientId}`}
-                                                disabled={!selectedClientDetails.isClientActive}
```
8kristy, line 370: I think we should keep these?

src/common/formatFamiliesData.ts
```diff
@@ -1,28 +1,77 @@
...
+    let gender;
+    switch (adult.gender) {
+        case "male":
+            gender = "M";
+            break;
+        case "female":
+            gender = "F";
+            break;
+        case "other":
+            gender = "-";
```
8kristy: I think the dash is a bit confusing, maybe just use "O"? Hopefully that's clear enough with others mentioned above it

src/common/formatFamiliesData.ts
```diff
@@ -1,28 +1,77 @@
...
+            break;
+    }
+    const age = getCurrentYear() - adult.birthYear;
+    return `${age} ${gender}`;
```
8kristy: Is it possible to get the age outside of these functions and then generalise them? I'd probably move the age line to where we call this and pass in the value as a parameter

```
const getPerson = (person: Person, age: string): string => {
    let gender;
    switch (person.gender) {
        case "male":
            gender = "M";
            break;
        case "female":
            gender = "F";
            break;
        case "other":
            gender = "-";
            break;
    }
    return `${age} ${gender}`;

    ...

    ageAndGenderOfAdults: displayList(formattedAdults.map((adult) => getPerson(adult, getCurrentYear() - adult.birthYear))),
    ageAndGenderOfChildren: displayList(formattedChildren.map((child) => getPerson(child, getChildAgeUsingBirthYearAndMonth(child.birthYear, child.birthMonth ?? null, true))))
};
```

src/common/formatFamiliesData.ts
```diff
@@ -1,28 +1,77 @@
...
+            birthYear: child.birth_year,
+        };
+    });
+    const adults = familyData.filter((member) => !isChildUsingBirthYear(member.birth_year));
+    const formattedAdults: Person[] = adults.map((adult) => {
+        return {
+            gender: adult.gender,
+            birthYear: adult.birth_year,
+        };
+    });
```
8kristy: Wonder if you can pull this out into a function and pass in the function as a parameter for the filter (i.e. isChildUsingBirthYear - if passing in !isChildUsingBirthYear doesn't work you could make `isAdultUsingBirthYear(year) => !isChildUsingBirthYear(year)`)

supabase/migrations/20240429145115_add_columns_birth_year_and_month_families_table.sql
```diff
@@ -0,0 +1,9 @@
+alter table "public"."families" add column "birth_month" smallint;
+
+alter table "public"."families" add column "birth_year" smallint;
+
+update "public"."families" set "birth_year" = 2024 - age;
```
8kristy: Is there a way to get current year in sql? I suppose we'll only make this transition this year but just to be safe

src/databaseTypesFile.ts
```diff
@@ -942,7 +945,7 @@ export type Database = {
         }
         Returns: undefined
       }
-      insertClientAndTheirFamily: {
+      insert_client_and_family: {
```
8kristy, line 952: Why do we need to rename this?

supabase/seed/seed.mts
```diff
@@ -72,13 +72,23 @@ async function generateSeed(): Promise<void> {
                 generateFamily(
                     { min: 1, max: 8 },
                     {
-                        age: (ctx) => copycat.int(ctx.seed, { min: 0, max: 100 }),
+                        birthYear: (ctx) => copycat.int(ctx.seed, { min: 1901, max: 2024 }),
```
8kristy: Should we use current year for max and current year - 120 (?)?

supabase/seed/seed.mts
```diff
@@ -72,13 +72,23 @@ async function generateSeed(): Promise<void> {
...
                     }
                 ),
         })
     );
 
+    await seed.families(
+        (generate) =>
+            generate(100, {
+                birthYear: 2024,
+                birthMonth: (ctx) => copycat.int(ctx.seed, { min: 1, max: 5 }),
```
8kristy, line 91: Why only up to 5?

8kristy (commented):
A few comment

src/app/clients/form/formSections/NumberAdultsCard.tsx
minalui: If value is NaN then the function returns false and asks the user to input a valid entry.

src/app/clients/form/formSections/NumberChildrenCard.tsx
minalui: Yes - thanks for spotting this! I've added a check to see if they've selected the current year and created a dynamic list of months depending on the current month

src/app/clients/getExpandedClientDetails.ts
minalui: Yes sorry! I think I handled getting the total months for babies in a separate function before.

src/app/parcels/parcelsTable/ParcelsPage.tsx
minalui: The buttons are currently in the bottom of the page but it should go in the footer so that it is accessible immediately without having to scroll all the way down

src/databaseTypesFile.ts
minalui: Sorry just following supabase convention but I can change it back?

supabase/seed/seed.mts
minalui: Only up to May (current month) otherwise the babies are not going to be born yet.

src/app/parcels/parcelsTable/ParcelsPage.tsx
minalui: Yes sorry I've added them back in the buttons

src/common/formatFamiliesData.ts
minalui: Sure :)

src/common/getAgesOfFamily.ts
minalui: Thanks Kristina - It does look a lot easier to read! 

The only scenarios that may be relevant that I can think of is if they need it for safeguarding children (boundary between 15/16 years old) and or young adults. I don't actually know if this is the case but I do know that agencies that interact with children usually have these. Originally we did think about creating a boundary for children eg. 15-16 years old but it may have looked very crowded on the client details modal if they have a lot of children.

src/common/formatFamiliesData.ts
minalui: Yep sounds good

src/common/formatFamiliesData.ts
minalui: I have created two new functions isChildPerson, isAdultPerson because isChildUsingBirthYear takes in birth_year but we need to take in the person object in the filter. I hope I haven't misunderstood!

supabase/seed/seed.mts
minalui: I've tried to use getCurrentYear() in seed.mts but it has an import error when I try to run generate seed. Can't seem to find a way to do this - any suggestions?

src/app/clients/form/formSections/NumberAdultsCard.tsx
```diff
@@ -1,59 +1,44 @@
...
+import { FormText, StyledCard } from "@/components/Form/formStyling";
+import { ClientCardProps, ClientSetter } from "../ClientForm";
+import DropdownListInput from "@/components/DataInput/DropdownListInput";
+import { getAdultBirthYears } from "@/app/clients/form/birthYearDropdown";
 
-const getNumberAdultsOfGenderDefault = (numberAdultsOfGender: number): string | undefined => {
-    return numberAdultsOfGender === 0 ? undefined : numberAdultsOfGender.toString();
-};
+const minimumNumberOfAdults = 1;
+const maximumNumberOfAdults = 20;
```
8kristy: I think typescript naming convention for constant values is SCREAMING_SNAKE_CASE

src/common/getAgesOfFamily.ts
8kristy: I guess you can't really decide without exact birth date - you don't want to mark a 15 year old as an adult but you don't want to mark an independent 16 year old as a child I think
Does this actually cause problems anywhere? I imagine if there's no coded in part which changes much for this boundary the workers/volunteers can be sensible about it

src/common/formatFamiliesData.ts
8kristy: Oh I meant take the entire thing so something like 
```
const getFormattedPeople= (familyData: Schema["families"], filterFunction: (arg: Array) => Array): Person[] => {
    const people = familyData.filter((member) => filterFunction(member));
    return children.map((child) => {
        return {
            gender: child.gender,
            birthMonth: child.birth_month,
            birthYear: child.birth_year,
        };
    });
};
```
then call it using 
```
const formattedChildren = getFormattedPeople(familyData, isChildPerson);
const formattedAdults =  getFormattedPeople(familyData, isAdultPerson);
```
assuming this works? And then `birthMonth` is null/undefined in either case right? Could add a param to determine that

src/databaseTypesFile.ts
8kristy: Ah nice, fingers crossed it doesn't cause any issues for anyone

supabase/seed/seed.mts
8kristy: Guess you can't import that function but what about just using what's inside of it i.e. just `new Date().getFullYear()`?

8kristy (commented):
Just a few small changes, otherwise looking good

src/common/getAgesOfFamily.ts
minalui: Only in the client and parcel modals and shopping list where it shows how many adults/children there are with their respective ages.

src/app/clients/form/formSections/NumberAdultsCard.tsx
minalui: Never heard of that term! 🤣

supabase/seed/seed.mts
minalui: Done :)

src/common/formatFamiliesData.ts
minalui: I guess birthMonth will depend on if they have a birthMonth stored. If they do, it will be used to calculate the childs age (eg. if they are under 1 years old).

src/app/clients/edit/[id]/autofill.ts
```diff
@@ -25,15 +15,22 @@ const autofill = (
...
+        })
+    );
+
+    const adults = familyData.filter(isAdultPerson).map(
+        (adult): Person => ({
+            gender: adult.gender,
+            birthYear: adult.birth_year,
+            primaryKey: adult.primary_key,
+        })
+    );
```
Comment (8kristy) [inline src/app/clients/edit/[id]/autofill.ts]: Is it possible to use the same `getFormattedPeople` function here, same as in `formatFamiliesData.ts`? Or is there some difference that I can't see on first glance?

8kristy (approved):
One last bit, happy to merge after it's done (or not if it's not possible)

src/app/clients/edit/[id]/autofill.ts
Comment (minalui) [reply src/app/clients/edit/[id]/autofill.ts]: Yes I think the primary key doesn't get used in the form so I can use the same function.

Issue #193, opened 2024-03-26
Modify error message when the email rate limit is exceeded


rickaw:
### Describe the solution / feature you'd like

The error message in https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/issues/177#issuecomment-2018590396 is confusing. If the email rate limit error is encountered, it should display
```
Email rate limit exceeded.
Please contact admin for help.
help@foodbankapp.org
```

maxjohns0n:
Error messages are returned from supabase just as strings, so we chose to add this text when HTTP status 429 is returned (too many requests).

maxjohns0n:
### QA

- From login page, click "forgot your password?" link
- Enter any email address
- Submit the form a few times until the error rate limit exceeded error appears (should be ~3 times).
- [x] Error message should have the text: `Email rate limit exceeded.
Please contact admin for help.
help@foodbankapp.org`
- [x] This should be correctly split across three lines
- Submit the form with the email left blank
- [x] The error message should **not** contain the text `Please contact admin for help.
help@foodbankapp.org`

minalui:
QA passed 🎉

PR #325, opened 2024-05-13
#193: Update error message when email rate limit is exceeded

maxjohns0n:
## What's changed
* In password reset request, when HTTP 429 is received (i.e. email rate limit is exceeded), added help text to prompt the user to contact an admin.
* Preserved newlines in the ErrorMessage component in the AuthPanel.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

No db changes


src/authentication/requestPasswordReset.ts
```diff
@@ -1,10 +1,20 @@
...
 
 interface RequestPasswordResetResponse {
     errorMessage: string | null;
 }
 
+function formatErrorMessage(error: AuthError): string {
+    let errorMessage = error.message;
+    // HTTP 429 is "too many requests". Usually means email rate limit has been exceeded.
+    if (error.status === 429) {
+        errorMessage += ".\nPlease contact admin for help.\nhelp@foodbankapp.org";
```
minalui, line 13: Should we save the email into a variable so that it can be changed easily?

minalui (changes requested):
It looks great - just a small comment

src/authentication/requestPasswordReset.ts
maxjohns0n: Done! Wasn't sure where the best place to put the variable was so I left it in the same function for now :)

src/authentication/requestPasswordReset.ts
```diff
@@ -1,10 +1,21 @@
 import { createClientComponentClient } from "@supabase/auth-helpers-nextjs";
 import { logErrorReturnLogId } from "@/logger/logger";
+import { AuthError } from "@supabase/gotrue-js";
 
 interface RequestPasswordResetResponse {
     errorMessage: string | null;
 }
 
+function formatErrorMessage(error: AuthError): string {
+    const helpEmailAddress = "help@foodbankapp.org";
```
AmeliaIsCoding: let's make a file helpEmailAddress.ts in app/common and move this there

Maybe one day someone will add some more email addresses or other non sensitive useful website data to that file and rename it something more general. But for now it will be okay to just call it that! :)

AmeliaIsCoding (changes requested):
Looks good, one comment

src/authentication/requestPasswordReset.ts
maxjohns0n: Sounds like a good plan! There isn't an app/common at the moment. Did you mean src/common? Or should I create app/common?

AmeliaIsCoding: sorry I meant src/common 
*facepalm*

maxjohns0n: Done!

AmeliaIsCoding (approved):
looks good to me nice job. Check with Ricky/Kristina if they want to review before you merge. :)

src/authentication/requestPasswordReset.ts
```diff
@@ -1,10 +1,21 @@
...
+import { helpEmailAddress } from "@/common/helpEmailAddress";
 
 interface RequestPasswordResetResponse {
     errorMessage: string | null;
 }
 
+function formatErrorMessage(error: AuthError): string {
+    let errorMessage = error.message;
+    // HTTP 429 is "too many requests". Usually means email rate limit has been exceeded.
+    if (error.status === 429) {
```
8kristy: I think it might be nicer to use an enum there instead of a number and a comment (e.g. `if (error.status === HttpStatusCode.TooManyRequests)`). Maybe have a look at a library (looks like axios is in the project and that has status codes https://github.com/axios/axios/blob/v1.x/lib/helpers/HttpStatusCode.js) or if something like this is already in the codebase?

src/authentication/requestPasswordReset.ts
```diff
@@ -1,10 +1,21 @@
...
+        errorMessage += `.\nPlease contact admin for help.\n${helpEmailAddress}`;
```
8kristy: I think instead of having a separate variable it might make sense to just return here - e.g. if you use this method again to write a completely different message that doesn't use the submitted one. So something like
```
if (error.status === HttpStatusCode.TooManyRequests) {
   return error.message + ...
}
return error.message;
```
If there were multiple messages it might also be useful to have it as a switch statement and maybe put this function in a separate file, but I'm assuming a long list of messages here so it's up to you if you want to do these now or leave it for someone else when more cases are added

src/common/helpEmailAddress.ts
```diff
@@ -0,0 +1 @@
+export const helpEmailAddress = "help@foodbankapp.org";
```
8kristy: Is there another place where we have a list of constants like this? If not might be worth generalising this file and calling it something like `contactDetails` or `staticInformation` (I'm sure you can come up with better depending on what we want to use this for)

src/components/AuthPanel.tsx
```diff
@@ -95,6 +95,7 @@ const AuthPanel: React.FC<AuthPanelProps> = ({
         color: ${theme.error};
         font-size: 13px;
         text-align: center;
+        white-space: pre-line;
```
8kristy, line 98: Just curious about general code structure - do you know why this CSS is in a .tsx file rather than being referenced from a stylesheet file?

8kristy (changes requested):
A few minor suggestions :)

src/components/AuthPanel.tsx
maxjohns0n: I think it's because we use the styled-components library. The code here defines a new `ErrorMessage` component which inherits from a `span` but has these extra styles.

src/common/helpEmailAddress.ts
maxjohns0n: I can't find anywhere else with constants like this, so I've renamed it to `contactDetails` for now!

src/authentication/requestPasswordReset.ts
maxjohns0n: I've done the return approach. I agree a switch would be good if we add more cases here but I'm not sure if we'll need to so left it for now.

src/authentication/requestPasswordReset.ts
maxjohns0n: Thanks! The axios enum seems to work well :)

src/components/AuthPanel.tsx
8kristy: Might need to look into this styled-components thing, don't think I've seen it before

8kristy (approved):

minalui (approved):

PR #326, opened 2024-05-13
VFB-212:  Ability to anonymise client data

AmeliaIsCoding:
## What's changed

- Added column `is_active` to `clients` table, true by default, and made personal data fields nullable
- Added delete button to client modal; on clicking, a confirmation dialog is shown
- On confirming delete, `is_active` is set to false, the client's personal information is nullified, and all of their family members are deleted from `families`
- An inactive client is not shown in the clients table
- The client modal for an inactive client shows no personal information, a message saying this client has been deleted, and the client's parcels in a table as usual. The "edit client" , "add parcel" and "delete client" buttons are not displayed
- A parcel for an inactive client is shown in the parcels table, with the client name displayed as {Deleted Client) and the other client information fields displayed as -. It is possible to filter by these values. The icons for flagged for attention and signposting call required are not shown for such parcels.
- The parcel modal for a parcel with an inactive client does not display rows for the client information; there is a message saying that the client has been deleted; and the "see client details" and  "edit client details" buttons are disabled
- Updated RLS policies so that inactive clients cannot be edited; thus the edit client form gives an error message on submitting for an inactive client
- The seed data has been updated so that some seeded clients are inactive, and also, every client has a non-zero family size
- The selected parcel overview, which is used in the action modal, status modal and the audit log modal, displays "Deleted Client @ [Timestamp]", as opposed to "[Postcode] - [Client Name] @ [Timestamp]
- When generating PDFs (except shopping list PDF) for parcels with deleted clients, personal client information is not shown on the PDF
- When generating a map for parcels with deleted clients, these parcels are ignored. If only such parcels are shown, an error is displayed on attempting to generate the map
- When generating the shopping list PDF for parcels with deleted clients, an error is displayed

## Screenshots / Videos
| After      |
|------------|
|image
image
image
image
image
image
image
image
image
image
image

## Checklist

  - Does this require resetting the deployed database? - NO unless we want deactivated clients seeded into the dev site
    - Checkout the dev branch
    - Checkout your branch


src/app/parcels/ParcelsPage.tsx
```diff
@@ -231,7 +238,13 @@ async function getClientIdForSelectedParcel(parcelId: string): Promise<string> {
         throw new Error(message + ` Log ID: ${logId}`);
     }
 
-    return data.client_id;
+    if (data.client == null) {
+        const message = `Failed to find matching client for a parcel with ID ${parcelId}`;
+        const logId = await logErrorReturnLogId(message, { error });
+        throw new Error(message + ` Log ID: ${logId}`);
```
AmeliaIsCoding: I will update this to return an error once VFB-211 has been merged (and before merging this ticket), because that ticket updated this function to return errors instead of throwing

src/app/parcels/parcelsTableFilters.ts
```diff
@@ -61,7 +73,7 @@ export const familySearch = (
         return query.gte("family_count", 10);
     }
     if (stateAsNumber === 1) {
-        return query.lte("family_count", 1);
+        return query.eq("family_count", 1);
```
AmeliaIsCoding: this used to be `lte` because some clients had 0 family members, but this is unrealistic. This conflicted with searching by "-" which should find family size equal to 0,  as this should only be true for deleted clients.
 I updated the seed data so that seeded clients have non-zero family size.

supabase/seed/seed.mts
```diff
@@ -59,19 +60,17 @@ async function generateSeed(): Promise<void> {
             extraInformation: (ctx) => copycat.sentence(ctx.seed, { maxWords: 20 }),
             flaggedForAttention: (ctx) => copycat.bool(ctx.seed),
             signpostingCallRequired: (ctx) => copycat.bool(ctx.seed),
+            isActive: (ctx) => copycat.oneOf(ctx.seed, booleansWeightedToTrue),
+            families: (generate) =>
+                generate(
+                    { min: 1, max: 12 },
+                    { age: (ctx) => copycat.int(ctx.seed, { min: 0, max: 100 }) }
+                ),
```
AmeliaIsCoding: this ensures that each client has a family size between 1 and 12. before, there were plenty of clients with 0 family count. this is unrealistic as the family is supposed to include the client themselves. it is not possible to create a client with 0 family count via the add / edit client form.

src/common/formatClientsData.ts
AmeliaIsCoding, line 1: the `prepareClientSummary` and `prepareRequirementSummary` functions are only used in the shopping list pdf generation, which we prevent occurring for deleted clients. therefore we can use empty strings as a default and not worry about "-" etc.

AmeliaIsCoding (commented):
Self-review

supabase/seed/seed.mts
Alex-Softwire: Am I right in saying the shopping list goes up to a family of size 10. How does this interact with a family of size 11 or 12? Or is 10 actually 10+?

src/app/parcels/parcelsTableFilters.ts
Alex-Softwire: Nice :)

src/app/clients/ClientsPage.tsx
```diff
@@ -171,10 +195,38 @@ const ClientsPage: React.FC<{}> = () => {
...
         return postcodeData ?? nullPostcodeDisplay;
     };
 
+    const onDeleteClient = async (): Promise<void> => {
+        if (clientId) {
+            setDeleteClientErrorMessage(null);
+            const { error: deleteClientError } = await deleteClient(clientId);
+            if (deleteClientError) {
+                setDeleteClientErrorMessage(getDeleteClientErrorMessage(deleteClientError));
+                setIsDeleteClientDialogOpen(false);
```
Alex-Softwire: Would moving setIsDeleteClientDialogOpen to line 220 (and removing from 225) be equivalent (and avoid replication)?

src/app/clients/deleteClient.ts
```diff
@@ -0,0 +1,36 @@
...
+type DeleteClientErrorType = "failedClientDeletion";
+export interface DeleteClientError {
+    type: DeleteClientErrorType;
+    logId: string;
+}
+type DeleteClientResponse = { error: DeleteClientError | null };
+
+const deleteClient = async (clientId: string): Promise<DeleteClientResponse> => {
+    const { error: deleteClientError } = await supabase.rpc("deactivateClient", {
+        clientid: clientId,
```
Alex-Softwire, line 14: Am curious what's caused the difference in case here - does supabase require lowercase and we use camelcase in js? Probably no need to change unless it's an easy fix

src/app/clients/deleteClient.ts
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    const baseAuditLogProps = {
+        action: "delete a client",
+        clientId: clientId,
+        content: { clientId: clientId },
```
Alex-Softwire: If these have the same variable name then we can do { clientId } rather than { clientId: clientId }. Same goes with { logId }. Apologies if I've misremembered and are wrong here

src/app/clients/form/submitFormHelpers.ts
```diff
@@ -153,7 +153,7 @@ export const submitEditClientForm = async (
     if (clientDataAndCount.updatedrows === 0) {
         const logId = await logWarningReturnLogId("Concurrent editing of client");
```
Alex-Softwire: Should this log be updated too to say (or client deleted?). 
Am I right in saying there's no way to tell the difference between the two cases (concurrent editing) or (client deleted)?

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -165,3 +168,36 @@ export const formatBreakdownOfChildrenFromFamilyDetails = (
...
     return childDetails.join(", ");
 };
+
+type IsClientActiveErrorType = "failedClientIsActiveFetch";
+export interface IsClientActiveError {
+    type: IsClientActiveErrorType;
+    logId: string;
+}
+
+type GetClientIsActiveResponse =
```
Alex-Softwire, line 178: I like the more specific type here --> much better than 
{
error: null | IsClientActivatedError,
isActive: null | boolean
}

src/app/parcels/ActionBar/ActionAndStatusBar.cy.tsx
```diff
@@ -155,7 +157,9 @@ describe("Parcels - Action Bar", () => {
                 if (parcel.addressPostcode) {
                     cy.get(".MuiPaper-root").contains(parcel.addressPostcode);
                 }
-                cy.get(".MuiPaper-root").contains(parcel.fullName);
+                if (parcel.fullName) {
```
Alex-Softwire, line 160: Yay for unit tests - do these ones run and pass correctly, or do they still have issue with the next js issue?

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
```diff
@@ -73,6 +73,8 @@ const getPdfErrorMessage = (error: ShoppingListPdfError): string => {
         case "invalidFamilySize":
             errorMessage = "Invalid family size for shopping list PDF.";
             break;
+        case "inactiveClient":
+            errorMessage = "One or more selected parcels belong to inactive clients.";
```
Alex-Softwire, line 77: Should we have break here as we do on the other ones?

src/app/parcels/ExpandedParcelDetails.tsx
```diff
@@ -59,7 +67,13 @@ const ExpandedParcelDetailsView = ({ parcelId }: Props): ReactElement => {
 
             {parcelDetails && (
                 <>
-                    <DataViewer data={parcelDetails.expandedParcelData} />
+                    {!parcelDetails.expandedParcelData.isActive && (
```
Alex-Softwire, line 70: What happens here if isActive is null (due to an error) rather than false. Will it show this parcel belongs to a deleted client where it shouldn't. Or is there protection for this?
Edit: Just seen later where this comes from and it can only be true or false as it's set in getExpandedParcelDetails. However, "clientIsActive" in that file could be null so we might have the same issue? Lmk if you think this is fine.

src/app/parcels/ParcelsPage.cy.tsx
```diff
@@ -37,6 +37,7 @@ const sampleProcessingData: ParcelsPlusRow[] = [
         last_status_timestamp: "2023-08-04T13:30:00+00:00",
         last_status_workflow_order: 1,
         created_at: "2023-12-31T12:00:00+00:00",
+        client_is_active: true,
```
Alex-Softwire: Quick question: when should we use snake case e.g parcel_id?

src/app/parcels/ParcelsPage.tsx
Alex-Softwire: Sounds good

src/app/parcels/ParcelsPage.tsx
```diff
@@ -231,7 +236,13 @@ async function getClientIdForSelectedParcel(parcelId: string): Promise<string> {
         throw new Error(message + ` Log ID: ${logId}`);
     }
 
-    return data.client_id;
+    if (data.client == null) {
```
Alex-Softwire: Can we use === here? == might do some unexpected type conversion that one doesn't want.

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -126,19 +147,29 @@ const getExpandedParcelDetails = async (
...
-    household: string;
-    children: string;
-    packingDate: string;
-    packingSlot: string;
-    method: string;
-    createdAt: string;
-}
+export type ExpandedParcelData =
+    | {
+          isActive: true;
```
Alex-Softwire: Like the union type here :)

src/app/parcels/parcelsTableFilters.ts
```diff
@@ -70,6 +85,12 @@ export const phoneSearch = (
     query: PostgrestFilterBuilder<Database["public"], any, any>,
     state: string
 ): PostgrestFilterBuilder<Database["public"], any, any> => {
+    if (state === "") {
+        return query;
+    }
+    if ("-".toLowerCase().includes(state.toLowerCase())) {
```
Alex-Softwire: What's a lowercase hyphen?

src/common/format.ts
```diff
@@ -68,9 +68,21 @@ export const getReadableWebsiteDataName = (name: string): string =>
...
     addressPostcode: string | null,
-    fullName: string,
-    collectionDatetime: Date | null
-): string =>
-    (addressPostcode ?? nullPostcodeDisplay) +
-    (fullName && ` - ${fullName}`) +
-    (collectionDatetime && ` @ ${dayjs(collectionDatetime).format("DD/MM/YYYY HH:mm")}`);
+    fullName: string | null,
+    collectionDatetime: Date | null,
+    clientIsActive?: boolean | null
```
Alex-Softwire: Why is clientIsActive a question mark, at what points can it be null, and what points can it be undefined?

src/components/Buttons/DeleteButton.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import Button from "@mui/material/Button";
+import { ButtonWrap } from "@/components/Buttons/GeneralButtonParts";
+
+interface Props {
```
Alex-Softwire, line 5: Do you know the difference between interface and type and why we use interface here? (this isn't rhetorical haha, I don't know myself :) )

src/pdf/ShippingLabels/ShippingLabelsPdfButton.tsx
```diff
@@ -123,21 +124,22 @@ const getRequiredData = async (
...
-            full_name: client.full_name,
-            phone_number: client.phone_number,
-            address_1: client.address_1,
-            address_2: client.address_2,
-            address_town: client.address_town,
-            address_county: client.address_county,
-            address_postcode: client.address_postcode,
-            delivery_instructions: client.delivery_instructions,
+            full_name: clientIsActive ? client.full_name ?? "" : displayNameForDeletedClient,
+            phone_number: clientIsActive ? client.phone_number ?? "" : "-",
```
Alex-Softwire, line 137: Is hyphen only used for inactive clients? - when do we use hyphen as supposed to the empty string.

supabase/seed/clientsSeed.mts
```diff
@@ -37,3 +37,5 @@ export const possibleParcelPostCodes = [
     "CB8 9LJ",
     null,
 ];
+
+export const booleansWeightedToTrue = [true, true, true, true, true, true, true, false];
```
Alex-Softwire: was thinking of a neat way of writing this but couldn't think of anything nicer than what you have :)

Alex-Softwire (commented):
Well done! This is a really big change, and looks really good. Just a few questions (and possible suggestions) as much for my understanding of the codebase :)

supabase/seed/seed.mts
AmeliaIsCoding: 10 is actually 10+

src/app/parcels/ActionBar/ActionAndStatusBar.cy.tsx
AmeliaIsCoding: not sure tbh, depends whether or not they use any server components

src/app/parcels/ParcelsPage.cy.tsx
AmeliaIsCoding: the database headers use snake case. we're using it in this component test to replicate getting data from the DB.

supabase/seed/seed.mts
Alex-Softwire: Should it say this in the lists page? - not something to change now if so, but worth noting down somewhere.

src/app/clients/ClientsPage.tsx
AmeliaIsCoding: it's basically equivalent, it means if there's an error the dialog closes before the error message is displayed in the underlying modal, but the lag is basically imperceptible so I'm happy to change

src/app/clients/deleteClient.ts
AmeliaIsCoding: supabase requires lowercase

src/app/clients/form/submitFormHelpers.ts
AmeliaIsCoding: thank you, great catch!
There's no way without making another call to supabase. Ricky and I decided not to do so

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
AmeliaIsCoding: thanks for catching

src/app/parcels/ExpandedParcelDetails.tsx
AmeliaIsCoding: I've fixed this now - the only case where clientIsActive could be undefined was when the parcel had no associated client. I  changed the definition of clientIsActive to use client rather than rawParcelDetails.client, because we check for client being null before the declaration of clientIsActive, and return from the function if it is null. Now typescript knows that clientIsActive is a boolean only.

src/app/parcels/ParcelsPage.tsx
AmeliaIsCoding: thank you! was a typo

src/app/parcels/parcelsTableFilters.ts
AmeliaIsCoding: lmao! have removed

src/common/format.ts
AmeliaIsCoding: I've removed both the ? and the null

src/components/Buttons/DeleteButton.tsx
AmeliaIsCoding: I think we use type when we need to use a union type or similar, and interface when we don't. maybe that's just the convention in our code.

src/pdf/ShippingLabels/ShippingLabelsPdfButton.tsx
AmeliaIsCoding: we only use the hyphen for inactive clients. we use the empty string if the client is active but the field is empty or null.

src/app/clients/ClientsPage.tsx
Alex-Softwire: Thanks, yeah would have thought the change is negligible and nicer to avoid duplication.

src/pdf/ShippingLabels/ShippingLabelsPdfButton.tsx
Alex-Softwire: Thanks, that makes sense!

src/app/parcels/ParcelsPage.cy.tsx
Alex-Softwire: Cool cool.. annoying having to use lots of different cases but that makes sense

src/components/Buttons/DeleteButton.tsx
Alex-Softwire: Fair enough, it looks consistent to the rest of the codebase. Will google an answer at some point.

src/app/clients/form/submitFormHelpers.ts
Alex-Softwire: Happy with the changed error message here.

src/app/clients/deleteClient.ts
Alex-Softwire: Cool cool.. looks a little ugly but not much you can do about it.

src/app/parcels/ExpandedParcelDetails.tsx
Alex-Softwire: Thanks for looking into this! Happy with that solution :)

Alex-Softwire (approved):
Happy with the feedback changes here :).

supabase/seed/seed.mts
AmeliaIsCoding: Yes I think it should, I'll make a ticket

src/app/clients/edit/[id]/autofill.ts
```diff
@@ -35,17 +35,19 @@ const autofill = (
...
-        addressLine1: noPostcode ? "" : clientData.address_1,
-        addressLine2: noPostcode ? "" : clientData.address_2,
-        addressTown: noPostcode ? "" : clientData.address_town,
-        addressCounty: noPostcode ? "" : clientData.address_county,
+        fullName: clientData.full_name ?? "",
+        phoneNumber: clientData.phone_number ?? "",
+        addressLine1: noPostcode ? "" : clientData.address_1 ?? "",
+        addressLine2: noPostcode ? "" : clientData.address_2 ?? "",
+        addressTown: noPostcode ? "" : clientData.address_town ?? "",
+        addressCounty: noPostcode ? "" : clientData.address_county ?? "",
```
Comment (8kristy) [inline src/app/clients/edit/[id]/autofill.ts:50]: From what I understand the postcode also gets deleted - wouldn't this be covered by the `noPostcode ? ""` part already?

src/app/clients/form/ClientForm.tsx
```diff
@@ -151,9 +151,9 @@ const ClientForm: React.FC<Props> = ({ initialFields, initialFormErrors, editCon
                             `Failed to update client and family. Log ID: ${editClientError.logId}`
                         );
                         break;
-                    case "concurrentUpdateConflict":
+                    case "noRowsUpdated":
                         setSubmitErrorMessage(
-                            `Record has been updated recently - please refresh. Log ID: ${editClientError.logId}`
+                            `Update failed. Either the client has been updated recently - please refresh - or the client has been deleted. Log ID: ${editClientError.logId}`
```
8kristy: I find the wording a little confusing. Do you think something like "Update failed, please refresh. If the error persists, the client may have been deleted." (or if there's an error when you refresh the page on the deleted user "The client has been deleted or updated. Please refresh.")

src/app/clients/deleteClient.ts
```diff
@@ -0,0 +1,36 @@
+import { logErrorReturnLogId } from "@/logger/logger";
+import { sendAuditLog } from "@/server/auditLog";
+import supabase from "@/supabaseClient";
+
+type DeleteClientErrorType = "failedClientDeletion";
+export interface DeleteClientError {
```
8kristy, line 6: I kinda feel like the types should maybe sit together and put the interface after them? Assuming that you don't need to define something before you use it
But I also see the benefit of listing them in order, hmm.. Up to you if you want to change it

src/app/parcels/ActionBar/ActionModals/GenerateMapModal.tsx
```diff
@@ -27,7 +27,7 @@ const GenerateMapModal: React.FC<ActionModalProps> = (props) => {
 
     const formattedPostcodes = props.selectedParcels.reduce<string[]>(
         (formattedPostcodes, parcel) => {
-            if (parcel.addressPostcode) {
+            if (parcel.addressPostcode && parcel.addressPostcode !== "-") {
```
8kristy, line 30: Maybe this will get answered as I go through the code but why do we check for "-"? Are the postcodes not nullified after deletion? And then we can just use "-" to display null postcodes
Unless I'm misunderstading and this is not read from the database?

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
```diff
@@ -73,6 +73,9 @@ const getPdfErrorMessage = (error: ShoppingListPdfError): string => {
         case "invalidFamilySize":
             errorMessage = "Invalid family size for shopping list PDF.";
             break;
+        case "inactiveClient":
```
8kristy, line 76: Should this be an error or we just ignore the parcels with inactive clients when making the shopping list and display this as a warning instead?

src/app/parcels/ActionBar/ActionAndStatusBar.cy.tsx
```diff
@@ -36,6 +36,7 @@ describe("Parcels - Action Bar", () => {
             },
             voucherNumber: "123456789",
             createdAt: new Date("2023-12-31T12:00:00+00:00"),
+            clientIsActive: true,
```
8kristy, line 39: Should we have a test with `clientIsActive: false`?

src/components/DataViewer/DataViewer.tsx
```diff
@@ -52,17 +52,22 @@ const formatDisplayValue = (value: valueType): string => {
 
 export interface DataViewerProps {
     data: Data;
+    fieldsToHide?: Partial<[keyof Data]>;
 }
 
-const DataViewer: React.FC<DataViewerProps> = (props) => {
+const DataViewer: React.FC<DataViewerProps> = ({ data, fieldsToHide = [] }) => {
```
8kristy, line 58: Why are we taking the fields out of props? It's a bit more readable now but if you have more fields that signature might get quite long

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -126,19 +147,29 @@ const getExpandedParcelDetails = async (
...
+          voucherNumber: string;
```
8kristy: Is it possible to extend the shorter list to include the extra fields rather than duplicating them? Makes it clearer they're shared and if one updates you don't need to remember to update both

src/common/format.ts
```diff
@@ -68,9 +68,21 @@ export const getReadableWebsiteDataName = (name: string): string =>
...
+            (collectionDatetime && ` @ ${dayjs(collectionDatetime).format("DD/MM/YYYY HH:mm")}`)
+        );
+    }
+    return (
+        displayNameForDeletedClient +
+        (collectionDatetime && ` @ ${dayjs(collectionDatetime).format("DD/MM/YYYY HH:mm")}`)
+    );
+};
+
+export const displayNameForDeletedClient = "Deleted Client";
```
8kristy: Wonder if there's a better location for this, especially since it's imported in a few places

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -155,7 +156,7 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                             <Text>{rowData.address.postcode}</Text>
                         </>
                     ) : (
-                        <Text>{nullPostcodeDisplay}</Text>
+                        <Text>{rowData.clientIsActive ? nullPostcodeDisplay : "-"}</Text>
```
8kristy: Why do we not use "-" for active clients who don't have a postcode?

src/pdf/ShippingLabels/ShippingLabelsPdfButton.tsx
```diff
@@ -131,14 +133,14 @@ const getRequiredData = async (
...
+            address_1: clientIsActive ? client.address_1 ?? "" : "",
+            address_2: clientIsActive ? client.address_2 ?? "" : "",
+            address_town: clientIsActive ? client.address_town ?? "" : "",
```
8kristy, line 140: Is this necessary or would the address be empty for deleted clients and still return ""?

supabase/seed/clientsSeed.mts
8kristy: Maybe like this? `Array(8).fill(true).concat(Array(1).fill(false))` 
Just makes the ratio a bit more obvious imo

supabase/migrations/20240510123247_allow_anonymising_client.sql
```diff
@@ -0,0 +1,119 @@
...
+END;$function$
+;
+
+create policy "Disable updates on inactive clients"
+on "public"."clients"
+as restrictive
+for update
+to authenticated
+using ((is_active = true))
+with check (((is_active = true) OR (is_active = false)));
```
8kristy: These migration files are automatically created right? Just curious why does it say `(is_active = true) OR (is_active = false)` here

src/app/clients/edit/[id]/autofill.ts
Comment (AmeliaIsCoding) [reply src/app/clients/edit/[id]/autofill.ts:50]: I don't think I quite understand what you're asking about. In this PR, the other address fields became nullable. Therefore we must use `clientData.address_1 ?? ""` instead of `clientData.address_1` so that we have a string to display. That's the change I made, but please let me know if I misunderstood and you're asking about the already existing code. :)

src/app/clients/form/ClientForm.tsx
AmeliaIsCoding: Thanks, I prefer that wording too!

src/app/clients/deleteClient.ts
AmeliaIsCoding: I agree it's not perfect, but I think the rest of the codebase mostly has them in this order (or at least the code that I've written does) so I think I'll keep it like this

src/app/parcels/ActionBar/ActionModals/GenerateMapModal.tsx
AmeliaIsCoding: It's a good question - the type of `selectedParcels` here is `ParcelsRow`, which isn't the DB parcel type, but the processed version. during the "processing" the postcode, if null, is set to "-".

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
AmeliaIsCoding: We could do it like this. Ricky suggested the approach I did, but I think it's only strength is its less effort. I'm happy to do it this way if you want, but this will be quite a bit of extra work as we'll have both the case where only parcels with inactive clients are selected, and the case where a mixture of parcels with active and inactive clients are selected. what we display would have to be different. Happy to spend the time on it if you want me to, lmk :)

src/app/parcels/ActionBar/ActionAndStatusBar.cy.tsx
AmeliaIsCoding: These component tests are deprecated and currently failing so I wouldn't be able to test if it worked - see this Slack message https://softwire.slack.com/archives/C05E17YJQN4/p1710868145285059

But I totally agree it would be a good idea to include if we used the component tests

src/components/DataViewer/DataViewer.tsx
AmeliaIsCoding: I'm not quite sure I understand what you mean by the signature? 
I just unpacked the props in order to make it more readable, I thought this was usually preferred

src/app/parcels/getExpandedParcelDetails.ts
AmeliaIsCoding: Good idea, thank you! Have implemented

src/common/format.ts
AmeliaIsCoding: I moved it to the top of this file where we have another string constant. Not sure where else would be better?

src/pdf/DriverOverview/DriverOverviewPdf.tsx
AmeliaIsCoding: Active clients without a postcode can be assumed to be "NFA" (no fixed abode) clients, ie clients without an address. in this case, we want to display the postcode as "NFA". the value of nullPostcode is "NFA". Please see https://github.com/orgs/NorwoodAndBrixtonFoodbank/projects/1?pane=issue&itemId=57550936

src/pdf/ShippingLabels/ShippingLabelsPdfButton.tsx
AmeliaIsCoding: The address fields would indeed be null for a client who has been deleted through the delete button on the front end. However, seeded "deleted" clients have `is_active = false`, but the other fields are not null. (because there's no easy way that we have found in snaplet to seed some fields with values dependant on the value of another field). Therefore we check whether or not to display "" based on whether `clientIsActive` is true. If `clientIsActive` is true, this still doesn't guarantee a non-null value of an address field according to the database schema types, therefore we have to also use `?? ""`

supabase/seed/clientsSeed.mts
AmeliaIsCoding: that's good, thank you!

supabase/migrations/20240510123247_allow_anonymising_client.sql
AmeliaIsCoding: the migration files are typically automatically created, although it is possible to write the SQL manually. in this case it was automatically generated. however, I wrote the with check expression myself. 
The purpose of this RLS policy is to disable editing an inactive client, hence the `using ((is_active = true))`. the using expression applies to the old value of the field, before the update.
The with check expression applies to the new value of the field, after the update. 
If there's no with check expression, the "using" expression is also used for the "with check", but applied to the new value. 
In our case, we want to disable editing an inactive client, but we want to allow the value of `is_active` to be either true or false afterwards. We need to include the with check to avoid "is_active" also being checked to be true after the update, which would prevent deleting clients altogether.
See https://supabase.com/docs/guides/database/postgres/row-level-security for more info

AmeliaIsCoding:
> A few notes and questions, overall good job :)
> 
> One thing - do we want to include deleted clients on day overview or shipping labels? All of the information is empty, so wonder if it would be better to say somewhere "you have selected parcels for clients that no longer exist, their information is not going to be displayed" (or fully error if all selected are deleted users) and exclude those fields
> 
> And as a general tip - I don't know if it was possible to split this up much but it makes it easier for everyone if you keep your PRs shorter - I can go over individual commits but it's more helpful to see what the final submission is, and having a few PRs, e.g. "Add is_active field to db and delete user functionality", "Update UI to deal with deleted users", "Fix document generation issues from deleted users" would make it easier to review since it's more focused. I know it can be annoying being blocked waiting for previous bit/having to rebase stuff on top of your old branches but it is generally better than one large PR

Thank you for reviewing! 

In terms of day overview / shipping labels, I honestly don't have much preference / not entirely sure what the client would prefer. I spoke with Ricky about it and he suggested this method. I can imagine a user wanting either outcome, and I feel like maybe this way might be slightly better because at least this way, the user could cross out or ignore the parcel on the resulting PDF if they did not want it to appear? But the other way round is not possible, ie if we did not display such parcels, the user has no way of getting around that. Lmk your thoughts.

Completely understand! really sorry that it was so much to go through. I think your idea of how to split up this ticket would have worked really well. I feel like I just tend to start coding and it seems too late to split up the ticket once I'm in the middle, but from now on I will be certain to think more before I begin!

src/app/clients/edit/[id]/autofill.ts
Comment (8kristy) [reply src/app/clients/edit/[id]/autofill.ts:50]: Yeah the line that I have in mind is `const noPostcode = clientData.address_postcode === null;` - my assumption is that when you nullify the address, you also nullify the postcode, so it would never reach `clientData.address_xyz` 

Guess then the worry is, are there cases where we have a postcode but no address? Old logic implies no (as you only added the `?? ""` for this case of deleting the user), but it might have been wrong in the first place

Hopefully that makes more sense?

src/app/parcels/ActionBar/ActionModals/GenerateMapModal.tsx
8kristy: Ah yeah thought it might be that way, thanks for clarifying :)

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
8kristy: We had the same debate about the Day Overview pdf on another PR - I agree it's less effort and would get flagged sooner, but it might also inconvenience people for and undefined amount of time. But you're right this PR is already big enough - maybe raise a ticket to change it later?

src/pdf/DriverOverview/DriverOverviewPdf.tsx
8kristy: Ahh I see, I assumed it just means not applicable or something, maybe we should update the naming of that or add a tooltip 😄 unless this is a common abbreviation that only I don't know - can definitely see a volunteer get confused

src/pdf/ShippingLabels/ShippingLabelsPdfButton.tsx
8kristy: Mmmm guess it's not a case we'd encounter in the wild but I see why you need it for test data. Might be worth adding a comment there explaining that so that someone doesn't try fixing it and getting confused with dev data

supabase/migrations/20240510123247_allow_anonymising_client.sql
8kristy: I see, thank you for explaining :) So it runs this policy on the clients table every time there is an update, the request comes from an authenticated entity and we can only change rows where `is_active == true`, and we can change that to either true or false, is that right? Not entirely sure what restrictive does

Also is there a reason for the many brackets? Does it error if you only have `using (is_active = true)` and `with check ((is_active = true) OR (is_active = false))`?

src/components/DataViewer/DataViewer.tsx
8kristy: Method signature :) not sure if you can really call it that here but just this whole thing `const DataViewer: React.FC<DataViewerProps> = ({ data, fieldsToHide = [] })`

Looking at some examples I think it might be the correct way, just in my head more fields == more places to update, and like I mentioned having a massive parameter list. But let's leave it for now

8kristy:
Let's leave the label issue for now, I'll ask what's the preference and if we need to change it we can raise another ticket

> I feel like I just tend to start coding and it seems too late to split up the ticket once I'm in the middle
Honestly relatable, don't worry about it too much, I'm just realising how much pain I inflicted on my reviewers by giving them a 60 file 100 commit PR before 😄

src/app/clients/edit/[id]/autofill.ts
Comment (AmeliaIsCoding) [reply src/app/clients/edit/[id]/autofill.ts:50]: Does this answer your question? :)

|postcode → address_wxyz ↓ | Non-null | Null |
|--------|--------|--------|
| Non-null | Active client with address | Active NFA client|
| Null | Not possible, but DB schema doesn't realise that | Client has been deleted through front end|

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
AmeliaIsCoding: Ticket raised [VFB-231](https://softwiretech.atlassian.net/browse/VFB-231)

src/app/clients/edit/[id]/autofill.ts
Comment (8kristy) [reply src/app/clients/edit/[id]/autofill.ts:50]: Think I looked at this before the other comment about NFA clients, yep makes sense now, thanks :)

src/pdf/DriverOverview/DriverOverviewPdf.tsx
AmeliaIsCoding: Cool, I've renamed `nullPostcodeDisplay` to `displayPostcodeForHomelessClient` as I agree this was unclear!! Not sure if you want me to do anything else? I think the volunteers at the foodbank might be fairly familiar with the abbreviation as I believe they use it already

src/pdf/ShippingLabels/ShippingLabelsPdfButton.tsx
AmeliaIsCoding: Do you mean I should add the comment here, or in seed.mts?

supabase/migrations/20240510123247_allow_anonymising_client.sql
AmeliaIsCoding: Yes that's right! Restrictive means its combined with the other policy on this table (which is only allowing authenticated users to make changes) in such a way that both policies have to be satisfied in order for the change to be allowed. The other option is permissive, in which case only the policy itself has to be satisified in order for the change to be allowed.

supabase added the additional brackets once I added the code into the RLS policy. I just tried to remove them and reset the database, and it seemed to work fine. So I think we should be ok to remove the additional brackets.

src/pdf/DriverOverview/DriverOverviewPdf.tsx
8kristy: If they are then all good, nothing else to do there

src/pdf/ShippingLabels/ShippingLabelsPdfButton.tsx
8kristy: Just here please

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -126,20 +148,30 @@ const getExpandedParcelDetails = async (
...
     household: string;
     children: string;
-    packingDate: string;
-    packingSlot: string;
-    method: string;
-    createdAt: string;
 }
 
+type ExpandedParcelData = ParcelDataForActiveClient | ParcelDataForInactiveClient;
+
```
8kristy, line 174: Very nice

8kristy (commented):
Just add that 1 comment and I think we're done

I asked Dan on the ticket what should we do about the labels, let's wait until after lunch to see if we get a response and if not merge it and we'll show it off tomorrow

8kristy (approved):
Yep good to have it as is, feel free to merge after adding the comment about seeding data :)

src/app/parcels/getExpandedParcelDetails.ts
AmeliaIsCoding: thank you!

AmeliaIsCoding:
> Just add that 1 comment and I think we're done
> 
> I asked Dan on the ticket what should we do about the labels, let's wait until after lunch to see if we get a response and if not merge it and we'll show it off tomorrow

Dan replied saying he's happy with the current implementation.
@8kristy should I keep the ticket I made https://softwiretech.atlassian.net/browse/VFB-231 and make it refer only to shopping list PDFs? or maybe we should get rid of this ticket now

src/pdf/ShippingLabels/ShippingLabelsPdfButton.tsx
AmeliaIsCoding: Have done

8kristy:
I didn't mention the shopping lists but I think if we follow the logic of "this would only happen if someone made a mistake" then I think it's ok to leave things as they are now. I'd leave the ticket but move it to won't do/closed and add an explanation - this might need to be referenced if we decided to hide the parcels like Lincoln mentioned (might still be worth listing the pros and cons)

Issue #182, opened 2024-03-23
Resize Parcels Table Column Width to accommodate text


dan-langridge:
### Describe the solution / feature you'd like

On the parcels table, the column size for 

Family
Postcode
Collection
Packing Date
Time 

are too wide, while 'Status' is too small. (Text is wrapping a lot)
Some tuning required of column width to make all row text readable with not too much blank space.

kaydentsoftwire:
QA steps:
- navigate to parcels page
- [ ] Last Status column should be wider than the others and not overly wrapped 
- [x] When you make the browser shorter some columns should disappear but not overlap
- [x] When you expand the browser to a wider screen they should still remain stretched out across the page, with the contents of Last Status appearing on a single line
- [x] There are no other unusual visual/sizing issues with the column widths

AmeliaIsCoding:
Well done, it looks really good and behaves well on changing the browser size. However, unfortunately there are a couple of long names and statuses that still wrap to the next row on my laptop screen, with the other columns still containing some blank space.
![image](https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/138508884/2e5fe5e9-562e-4b6e-8d4b-02661970a24e)
I would suggest you make the last status and name columns slightly wider still, and the other columns slightly narrower.
@kaydentsoftwire

8kristy:
It's still a bit narrow on smaller screens? 
![image](https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/42096009/5dee0040-016e-4565-bc07-4c88cba69a88)

Do we have separate styling for mobile and desktop? I think I'd aim for something like this 
![image](https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/42096009/ea2f166e-d219-484c-ac2d-9932c887ad32)

(or it doesn't look too bad I think even if we apply the same rules to both)
| Current styling | New styling |
|---|---|
| ![image](https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/42096009/f1644345-f9ff-43da-8c9a-a5ac37e8d1ab)| ![image](https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/42096009/6210fc3f-f631-4447-ae2d-d9bf1470fc8f)|
[...]

PR #333, opened 2024-05-14
#182: Resize Parcels Table Column to accomodate text

kaydentsoftwire:
## What's changed
Adjusted the min and max widths of the table columns so that family, postcode, method, packing date and packing slot are narrower, and last status is wider, accounting for different viewing sizes

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image image | image image |

## Checklist


package-lock.json
rickaw, line 1: hmm what's changed in this file?

src/app/parcels/ParcelsPage.tsx
```diff
@@ -177,26 +177,42 @@ const parcelTableColumnStyleOptions = {
...
+        basis: "5rem",
+        grow: 1,
+        minWidth: "5rem",
     },
     lastStatus: {
-        minWidth: "8rem",
+        basis: "15rem",
+        grow: 5,
     },
 };
```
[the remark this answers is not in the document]
rickaw: Seems like the default column styles are defined in `src/components/Tables/Table.tsx` as `defaultColumnStyleOptions` which is why grow is set to 1 by default. I'm guessing the motivation for that was probably to try and have equal width columns.

If you look closely in `src/components/Tables/Table.tsx`, we define `columns` which we pass to `DataTable` which is a library component. If you look at the type of `columns` it is `CustomColumn` which extends `TableColumn` which extends `TableColumnBase` which tells you the styles you can specify.

Maybe we should explicitly type `parcelTableColumnStyleOptions` i.e.
```ts
const parcelTableColumnStyleOptions: ColumnStyles<ParcelsTableRow> = { ... }
```

This makes me think, do we really need all those `basis`? It doesn't appear in the list of styles that can be specified.

package-lock.json
kaydentsoftwire: Weird haha. This came up as a change after I pulled from dev and ran post checkout

src/app/parcels/ParcelsPage.tsx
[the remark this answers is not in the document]
kaydentsoftwire: We do not! I have removed the unnecessary `basis`

package-lock.json
rickaw: ah seems like it's changed the line endings. If you could go back to this file and change the line ending to LF and commit :) If this doesn't get resolved it's not a big deal, we can still merge it

rickaw (approved):

8kristy (approved):

package-lock.json
kaydentsoftwire: Ahh, done!

Issue #185, opened 2024-03-25
on Driver Overview PDF - add column for number of Parcels per client


dan-langridge:
### Describe the solution / feature you'd like

on the Driver Overview PDF, please add a column to show the number of Parcels for that client.

rickaw:
@dan-langridge The driver overview PDF's table displays one row per parcel that has been selected. e.g. If we have selected three parcels for the same client and created the driver overview PDF, three rows are displayed (see image). Would we still want to add a column to show the number of parcels for that client? Or do we want to show one row per client, and add a column to show the number of parcels for that client? (In this case, should the "Packing Date" column display the packing dates of all the parcels for that client?)


![Image](https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/114006382/cc56fba3-a27a-47ac-a1e1-ae20cd15a570)

dan-langridge:
Sorry some miscommunication comes from using the word Parcels in 2 different places, our fault...
The column i'd like to add is the number of labels printed for that parcel. (So in this scenario, Aaliyah Johnston, we printed 10 Labels, so a column to show 10, so the driver knows this delivery has 10 parcels/boxes/bags associated with it.

rickaw:
@dan-langridge Ah that makes sense, thanks for clarifying!

rickaw:
## Developer Notes
- For each parcel to display on the driver overview, get the latest `Shipping Labels Downloaded` event associated with that parcel and retrieve the number of shipping labels downloaded which is stored in the `event_data` field.
- Add "Number of parcels" column in the Driver Overview PDF which displays the number of shipping labels for that particular parcel.

maxjohns0n:
### QA

- On the parcels page, download some shipping labels for some parcels
- Download the driver overview for those parcels
- [x] PDF has "Number of parcels" column
- [x] Parcels which you downloaded labels for have the correct number in the column
- [x] Parcels which haven't had any labels downloaded show as "No labels downloaded"
- Go back and download a different number of labels for a parcel
- Download a new driver overview
- [x] The new PDF has the updated number of labels (i.e. the most recent value)

kaydentsoftwire:
QA passed

PR #335, opened 2024-05-15
#185: Add column to driver overview for number of labels per parcel

maxjohns0n:
## What's changed
* Driver overview PDF now has a column showing how many labels have been downloaded for each parcel
  * Column is named "Number of Parcels" as per ticket
* Uses the latest "Shipping Labels Downloaded" event for that parcel and gets the number of labels associated with that event

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

### DIFF
--- a/src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
+++ b/src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
@@ -50,7 +50,7 @@ const DriverOverviewInput: React.FC<DriverOverviewInputProps> = ({
 };
 
 const getPdfErrorMessage = (error: DriverOverviewError): string => {
-    let errorMessage: string = "";
+    let errorMessage: string;
     switch (error.type) {
         case "parcelFetchFailed":
             errorMessage = "Failed to fetch parcel data.";
--- a/src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
+++ b/src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
@@ -50,7 +50,7 @@ const DriverOverviewInput: React.FC<DriverOverviewInputProps> = ({
 };
 
 const getPdfErrorMessage = (error: DriverOverviewError): string => {
-    let errorMessage: string;
+    let errorMessage: string = "";
     switch (error.type) {
         case "parcelFetchFailed":
             errorMessage = "Failed to fetch parcel data.";
--- a/src/pdf/DriverOverview/DriverOverviewPdf.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdf.tsx
@@ -114,10 +114,10 @@ const styles = StyleSheet.create({
         width: "20%",
     },
     instructionsColumnWidth: {
-        width: "25%",
+        width: "20%",
     },
     numberOfLabelsColumnWidth: {
-        width: "15%",
+        width: "20%",
     },
 });
 
--- a/src/pdf/DriverOverview/DriverOverviewPdf.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdf.tsx
@@ -16,7 +16,7 @@ export interface DriverOverviewTableData {
     contact?: string;
     packingDate: string | null;
     instructions?: string;
-    numberOfLabels?: number | null;
+    numberOfLabels: number;
 }
 
 export interface DriverOverviewCardDataProps {
--- a/src/pdf/DriverOverview/DriverOverviewPdf.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdf.tsx
@@ -16,6 +16,7 @@ export interface DriverOverviewTableData {
     contact?: string;
     packingDate: string | null;
     instructions?: string;
+    numberOfLabels?: number | null;
 }
 
 export interface DriverOverviewCardDataProps {
@@ -115,6 +116,9 @@ const styles = StyleSheet.create({
     instructionsColumnWidth: {
         width: "25%",
     },
+    numberOfLabelsColumnWidth: {
+        width: "15%",
+    },
 });
 
 const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
@@ -134,6 +138,9 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
             </View>
             <View style={[styles.tableColumn, styles.instructionsColumnWidth]}>
                 <Text>Instructions</Text>
+            </View>
+            <View style={[styles.tableColumn, styles.numberOfLabelsColumnWidth]}>
+                <Text>Number of Parcels</Text>
             </View>
         </View>
     );
@@ -162,10 +169,13 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                     <Text>{rowData.contact}</Text>
                 </View>
                 <View style={[styles.tableColumn, styles.packingDateColumnWidth]}>
-                    <Text>{rowData.packingDate ? rowData.packingDate : "No recorded date"}</Text>
+                    <Text>{rowData.packingDate || "No recorded date"}</Text>
                 </View>
                 <View style={[styles.tableColumn, styles.instructionsColumnWidth]}>
                     <Text>{rowData.instructions}</Text>
+                </View>
+                <View style={[styles.tableColumn, styles.numberOfLabelsColumnWidth]}>
+                    <Text>{rowData.numberOfLabels || "No labels downloaded"}</Text>
                 </View>
             </View>
         );
--- a/src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
@@ -31,7 +31,10 @@ const compareDriverOverviewTableData = (
     return 0;
 };
 
-type ParcelsForDelivery = (Schema["parcels"] & { client: Schema["clients"] })[];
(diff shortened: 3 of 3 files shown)


AmeliaIsCoding:
Can you make the instructions column narrower and the number of parcels column wider please? We want "No labels downloaded" to fit in one row

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -115,6 +116,9 @@ const styles = StyleSheet.create({
     instructionsColumnWidth: {
         width: "25%",
     },
+    numberOfLabelsColumnWidth: {
+        width: "15%",
```
AmeliaIsCoding: i think this being 20% and instructions also being 20% will work well as per my previous comment

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -43,7 +46,32 @@ type ParcelsForDeliveryResponse =
...
+    | {
+          data: Schema["events"][];
+          error: null;
+      }
+    | {
+          data: null;
+          error: { type: ParcelsForDeliveryErrorType; logId: string };
+      };
+
+// Gets all "shipping labels downloaded" events for the given parcels
```
AmeliaIsCoding: can you remove this comment and rename the function `getParcelEvents` to be more descriptive, so that we don't need the comment?

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -43,7 +46,32 @@ type ParcelsForDeliveryResponse =
...
+const getParcelEvents = async (parcelIds: string[]): Promise<ParcelEventResponse> => {
+    const { data, error } = await supabase
+        .from("events")
+        .select("*")
+        .in("parcel_id", parcelIds)
+        .eq("new_parcel_status", "Shipping Labels Downloaded")
+        .order("timestamp", { ascending: false });
+    if (error) {
```
AmeliaIsCoding: can you add an empty line above this if block, and also one below it, for readability

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -67,6 +95,19 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
         dataWithNonNullClients.push({ ...parcel, client: parcel.client });
     }
 
+    const eventResponse = await getParcelEvents(parcelIds);
+    if (eventResponse.error || !eventResponse.data) {
+        return eventResponse;
+    }
+
+    // Find most recent label download event and store the number of labels (the event_data)
```
AmeliaIsCoding: we don't really need this comment, I think the code should aim to be self-explanatory where possible, for readability

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -67,6 +95,19 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
         dataWithNonNullClients.push({ ...parcel, client: parcel.client });
     }
 
+    const eventResponse = await getParcelEvents(parcelIds);
```
AmeliaIsCoding: can you change this to `const {data: eventData, error: eventError} = ....`
(no biggie, just how the rest of the code is written)

maxjohns0n:
> Can you make the instructions column narrower and the number of parcels column wider please? We want "No labels downloaded" to fit in one row

@AmeliaIsCoding Thanks for the suggestion! Done :)

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -43,7 +46,32 @@ type ParcelsForDeliveryResponse =
...
+          data: Schema["events"][];
+          error: null;
+      }
+    | {
+          data: null;
+          error: { type: ParcelsForDeliveryErrorType; logId: string };
+      };
+
+// Gets all "shipping labels downloaded" events for the given parcels
+const getParcelEvents = async (parcelIds: string[]): Promise<ParcelEventResponse> => {
```
AmeliaIsCoding: if this is used for multiple parcels, we should name it getParcelsEvents or something to indicate that (as well as changing it to be more descriptive anyway)

AmeliaIsCoding: Actually I think it makes sense to just put this data fetching into `getParcelsForDelivery`. We can access the events foreign table in the same call. If it's not possible to filter by the `new_parcel_status` and sort by the `timestamp` in this table in the call (it might be), we can do this in the front end.
It's just a bit cleaner

AmeliaIsCoding (changes requested):
Well done - I think it would be best to move the new data fetch call you made into the already existing getParcelsForDelivery. Details in comments, lmk if unclear

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
maxjohns0n: Yeah I couldn't figure out how to do this earlier but I think I've done it now! Lmk if it can be improved further :)

AmeliaIsCoding (approved):
looks great! well done on the tricky supabase call. Happy for you to now ask Ricky/Kristina for a review

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
```diff
@@ -50,7 +50,7 @@ const DriverOverviewInput: React.FC<DriverOverviewInputProps> = ({
 };
 
 const getPdfErrorMessage = (error: DriverOverviewError): string => {
-    let errorMessage: string;
+    let errorMessage: string = "";
```
8kristy: What does this do?

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -64,7 +70,13 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
             );
             return { data: null, error: { type: "noMatchingClient", logId: logId } };
         }
-        dataWithNonNullClients.push({ ...parcel, client: parcel.client });
+
+        let labelCount: number | undefined;
+        if (parcel.events && parcel.events.length > 0 && parcel.events[0].event_data) {
```
8kristy, line 75: Hope js/ts has a standard function for it but you might be able to simplify this to something like `if (!parcel.events.isNullOrEmpty() && parcel.events[0].event_data)`

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -64,7 +70,13 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
...
+            labelCount = Number.parseInt(parcel.events[0].event_data);
+        }
+
+        dataWithNonNullClients.push({ ...parcel, client: parcel.client, labelCount: labelCount });
```
8kristy, line 79: If the `event_data` doesn't exist, does this get a 0, null, undefined or something else? And does that work ok with everything (looks like the type of this is undefined, `labelCount` in `ParcelsForDelivery` is optional, and `numberOfLabels` is null)
Might be worth normalising the types between all of these

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -102,6 +114,7 @@ const getDriverPdfData = async (parcelIds: string[]): Promise<DriverPdfResponse>
             contact: client?.phone_number ?? "",
             packingDate: formatDateToDate(parcel.packing_date) ?? null,
             instructions: client?.delivery_instructions ?? "",
+            numberOfLabels: parcel.labelCount ?? null,
```
8kristy: If you make all the types null as per the other comment you can probably leave this as just `parcel.labelCount`. Although being explicit doesn't hurt - up to you what you want to do with it

8kristy (commented):
A few little bits

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
maxjohns0n: The linter was complaining about it but actually I think my later changes avoided that error anyway. I'll revert this :)

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
maxjohns0n: As far as I can tell there isn't a way to combine a nullish check and an empty array check built in to js/ts, so the first two conditions have to be like this. Lmk if you know of a better way! :)

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
maxjohns0n: It gets an undefined currently. I agree these should be normalised so I've made labelCount non-optional and given it a default value of 0 to simplify things (I think having 0 number of labels makes more sense than having null number of labels). The behaviour should stay the same!

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
8kristy: Hmm yeah doesn't look like it, you could try null coalescence and have something like `parcel.events?.length > 0 && ...` or `parcel.events && parcel.events[0]?.event_data` but I don't think it improves much - if anything it makes it more confusing for the sake of being shorter

8kristy (approved):

PR #336, opened 2024-05-15
VFB-182: Re-navigate to parcel modal on submit and edit of parcel.

Alex-Softwire:
## What's changed
1.  Now navigates to the parcel modal upon parcel form submit
2.  Fixed potential bug when database errors.
3.  Refactored parcel forms code to improve readability. (Have added comments to explain what I've done and why. Please say if I've made a mistake with the refactors / you disagree with any of the changes :) ).

## Screenshots / Videos
Not needed as no ui changes in this pr?

## Checklist

If you have made any changes to the database... ( I have not :) )

### DIFF
--- a/src/app/parcels/add/AddParcelForm.tsx
+++ b/src/app/parcels/add/AddParcelForm.tsx
@@ -16,7 +16,7 @@ import {
 import { ErrorSecondaryText } from "@/app/errorStylingandMessages";
 import supabase from "@/supabaseClient";
 import Title from "@/components/Title/Title";
-import { insertParcel } from "@/app/parcels/form/clientDatabaseFunctions";
+import { insertParcel } from "@/app/parcels/form/submitFormHelpers";
 
 interface AddParcelProps {
     clientId: string;
--- a/src/app/parcels/add/AddParcelForm.tsx
+++ b/src/app/parcels/add/AddParcelForm.tsx
@@ -16,6 +16,7 @@ import {
 import { ErrorSecondaryText } from "@/app/errorStylingandMessages";
 import supabase from "@/supabaseClient";
 import Title from "@/components/Title/Title";
+import { insertParcel } from "@/app/parcels/form/clientDatabaseFunctions";
 
 interface AddParcelProps {
     clientId: string;
@@ -88,7 +89,7 @@ const AddParcels = ({ clientId }: AddParcelProps): React.ReactElement => {
                         initialFields={initialParcelFields}
                         initialFormErrors={initialParcelFormErrors}
                         clientId={clientId}
-                        editMode={false}
+                        writeParcelInfoToDatabase={insertParcel}
                         deliveryPrimaryKey={deliveryPrimaryKey}
                         collectionCentresLabelsAndValues={collectionCentresLabelsAndValues}
                         packingSlotsLabelsAndValues={packingSlotsLabelsAndValues}
--- a/src/app/parcels/edit/EditParcelForm.tsx
+++ b/src/app/parcels/edit/EditParcelForm.tsx
@@ -17,7 +17,7 @@ import supabase from "@/supabaseClient";
 import { Errors } from "@/components/Form/formFunctions";
 import { ErrorSecondaryText } from "@/app/errorStylingandMessages";
 import Title from "@/components/Title/Title";
-import { ParcelDatabaseUpdateRecord, updateParcel } from "@/app/parcels/form/submitFormHelpers";
+import { updateParcel } from "@/app/parcels/form/submitFormHelpers";
 
 interface EditParcelFormProps {
     parcelId: string;
@@ -137,9 +137,7 @@ const EditParcelForm = ({ parcelId }: EditParcelFormProps): React.ReactElement =
                 <ParcelForm
                     initialFields={initialFormFields}
                     initialFormErrors={initialFormErrors}
-                    writeParcelInfoToDatabase={(parcelDatabaseRecord: ParcelDatabaseUpdateRecord) =>
-                        updateParcel(parcelDatabaseRecord, parcelId)
-                    }
+                    writeParcelInfoToDatabase={updateParcel(parcelId)}
                     deliveryPrimaryKey={deliveryKey}
                     collectionCentresLabelsAndValues={collectionCentres}
                     packingSlotsLabelsAndValues={packingSlots}
--- a/src/app/parcels/edit/EditParcelForm.tsx
+++ b/src/app/parcels/edit/EditParcelForm.tsx
@@ -1,22 +1,26 @@
 "use client";
 
 import React, { useEffect, useState } from "react";
-import ParcelForm, { ParcelErrors, ParcelFields, initialParcelFields } from "../form/ParcelForm";
+import ParcelForm, { initialParcelFields, ParcelErrors, ParcelFields } from "../form/ParcelForm";
 import {
     CollectionCentresLabelsAndValues,
-    fetchParcel,
-    getActiveCollectionCentres,
-    PackingSlotsLabelsAndValues,
-    fetchPackingSlotsInfo,
-    ParcelWithCollectionCentreAndPackingSlot,
     FetchCollectionCentresError,
-    PackingSlotsError,
+    fetchPackingSlotsInfo,
+    fetchParcel,
     FetchParcelError,
+    getActiveCollectionCentres,
+    PackingSlotsError,
+    PackingSlotsLabelsAndValues,
+    ParcelWithCollectionCentreAndPackingSlot,
 } from "@/common/fetch";
 import supabase from "@/supabaseClient";
 import { Errors } from "@/components/Form/formFunctions";
 import { ErrorSecondaryText } from "@/app/errorStylingandMessages";
 import Title from "@/components/Title/Title";
+import {
+    ParcelDatabaseUpdateRecord,
+    updateParcel,
+} from "@/app/parcels/form/clientDatabaseFunctions";
 
 interface EditParcelFormProps {
     parcelId: string;
@@ -111,7 +115,6 @@ const EditParcelForm = ({ parcelId }: EditParcelFormProps): React.ReactElement =
             );
             setPackingSlotsIsShown(parcelData.packing_slot?.is_shown);
             setCollectionCentreIsShown(parcelData.collection_centre?.is_shown === true);
-
             setIsLoading(false);
         })();
     }, [parcelId]);
@@ -119,11 +122,11 @@ const EditParcelForm = ({ parcelId }: EditParcelFormProps): React.ReactElement =
     const initialFormErrors: ParcelErrors = {
         voucherNumber: Errors.none,
         packingDate: Errors.none,
-        packingSlot: Errors.none,
+        packingSlot: packingSlotIsShown ? Errors.none : Errors.invalidPackingSlot,
         shippingMethod: Errors.none,
         collectionDate: Errors.none,
         collectionTime: Errors.none,
-        collectionCentre: Errors.none,
+        collectionCentre: collectionCentreIsShown ? Errors.none : Errors.invalidCollectionCentre,
     };
 
     return (
@@ -137,13 +140,12 @@ const EditParcelForm = ({ parcelId }: EditParcelFormProps): React.ReactElement =
                 <ParcelForm
                     initialFields={initialFormFields}
                     initialFormErrors={initialFormErrors}
-                    editMode={true}
-                    parcelId={parcelId}
+                    writeParcelInfoToDatabase={(parcelDatabaseRecord: ParcelDatabaseUpdateRecord) =>
+                        updateParcel(parcelDatabaseRecord, parcelId)
+                    }
                     deliveryPrimaryKey={deliveryKey}
                     collectionCentresLabelsAndValues={collectionCentres}
                     packingSlotsLabelsAndValues={packingSlots}
-                    packingSlotIsShown={packingSlotIsShown}
-                    collectionCentreIsShown={collectionCentreIsShown}
                 />
             )}
         </>
--- a/src/app/parcels/edit/EditParcelForm.tsx
+++ b/src/app/parcels/edit/EditParcelForm.tsx
@@ -17,10 +17,7 @@ import supabase from "@/supabaseClient";
 import { Errors } from "@/components/Form/formFunctions";
 import { ErrorSecondaryText } from "@/app/errorStylingandMessages";
 import Title from "@/components/Title/Title";
-import {
-    ParcelDatabaseUpdateRecord,
-    updateParcel,
-} from "@/app/parcels/form/clientDatabaseFunctions";
+import { ParcelDatabaseUpdateRecord, updateParcel } from "@/app/parcels/form/submitFormHelpers";
 
 interface EditParcelFormProps {
     parcelId: string;
--- a/src/app/parcels/form/ParcelForm.tsx
+++ b/src/app/parcels/form/ParcelForm.tsx
@@ -25,7 +25,7 @@ import CollectionDateCard from "@/app/parcels/form/formSections/CollectionDateCa
 import CollectionTimeCard from "@/app/parcels/form/formSections/CollectionTimeCard";
 import CollectionCentreCard from "@/app/parcels/form/formSections/CollectionCentreCard";
 import {
-    ParcelDatabaseErrors,
+    WriteParcelToDatabaseErrors,
     WriteParcelToDatabaseFunction,
 } from "@/app/parcels/form/submitFormHelpers";
 import { Button, IconButton } from "@mui/material";
@@ -127,7 +127,7 @@ const mergeDateAndTime = (date: string, time: string): Dayjs => {
 const parcelModalRouterPath = (parcelId: string): string => `/parcels?parcelId=${parcelId}`;
 
 const databaseErrorMessageFromErrorType = (
-    errorType: ParcelDatabaseErrors,
+    errorType: WriteParcelToDatabaseErrors,
     logId: string
 ): string => {
     switch (errorType) {
@@ -231,14 +231,10 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
         }
 
         if (error) {
-            /*
-               If the user is trying to edit a parcel that is currently being edited,
-               we want the user to refresh the page to get the current data, so we keep the submit button
-               disabled.
-             */
             if (error.type !== "concurrentUpdateConflict") {
                 setSubmitDisabled(false);
             }
+
             setSubmitErrorMessage(databaseErrorMessageFromErrorType(error.type, error.logId));
         }
     };
--- a/src/app/parcels/form/ParcelForm.tsx
+++ b/src/app/parcels/form/ParcelForm.tsx
@@ -25,13 +25,9 @@ import CollectionDateCard from "@/app/parcels/form/formSections/CollectionDateCa
 import CollectionTimeCard from "@/app/parcels/form/formSections/CollectionTimeCard";
 import CollectionCentreCard from "@/app/parcels/form/formSections/CollectionCentreCard";
 import {
-    InsertParcelErrors,
-    InsertParcelReturnType,
-    ParcelDatabaseInsertRecord,
-    ParcelDatabaseUpdateRecord,
-    UpdateParcelErrors,
-    UpdateParcelReturnType,
-} from "@/app/parcels/form/clientDatabaseFunctions";
+    ParcelDatabaseErrors,
+    WriteParcelToDatabaseFunction,
+} from "@/app/parcels/form/submitFormHelpers";
 import { Button, IconButton } from "@mui/material";
 import { Schema } from "@/databaseUtils";
 import dayjs, { Dayjs } from "dayjs";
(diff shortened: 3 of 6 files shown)


src/app/parcels/edit/EditParcelForm.tsx
```diff
@@ -1,22 +1,26 @@
 "use client";
```
Alex-Softwire, line 1: File renamed to capital letter to match AddParcelForm and other typescript files :)

src/app/parcels/edit/EditParcelForm.tsx
```diff
@@ -111,19 +115,18 @@ const EditParcelForm = ({ parcelId }: EditParcelFormProps): React.ReactElement =
...
-
             setIsLoading(false);
         })();
     }, [parcelId]);
 
     const initialFormErrors: ParcelErrors = {
         voucherNumber: Errors.none,
         packingDate: Errors.none,
-        packingSlot: Errors.none,
+        packingSlot: packingSlotIsShown ? Errors.none : Errors.invalidPackingSlot,
```
Alex-Softwire, line 122: I have moved the logic in showing the initial errors into EditParcelForm from ParcelForm.tsx. This addition should be equivalent to the use states that I've removed, with fewer lines. Lmk if you disagree. The two useStates that I've removed only activate when editMode is true (which is only in editParcelForm), so make sense to just be in EditParcelForm? Also am not sure why they would need to be in a useState? Have tested that when you turn off the packing slot of a parcel in the admin controls, when you click on editParcel, the error message is initially shown.

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -124,20 +130,33 @@ const mergeDateAndTime = (date: string, time: string): Dayjs => {
     return dayjs(date).hour(dayjsTime.hour()).minute(dayjsTime.minute());
 };
 
+const parcelModalRouterPath = (parcelId: string): string => `/parcels?parcelId=${parcelId}`;
+
+const databaseErrorMessageFromErrorType = (
```
Alex-Softwire, line 129: Exported error message calculation to separate function - hopefully this should be neater :)

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -230,36 +230,16 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
-                        errorMessage = `Failed to insert parcel. Log ID: ${insertParcelError.logId}`;
-                        break;
-                }
-                setSubmitErrorMessage(errorMessage);
-                return;
-            }
-        }
-        router.push("/parcels/");
+        const { error, parcelId } = await writeParcelInfoToDatabase(parcelRecord);
         setSubmitDisabled(false);
```
Alex-Softwire: Previously there was a return in line 258, before the setSubmitDisabled. This meant that when there was an error, the submit button gets permanently greyed out, and you can't click it. I think this might be a bug, and should be fixed now, but lmk if this is intended :)

src/app/parcels/add/AddParcelForm.tsx
```diff
@@ -88,7 +89,7 @@ const AddParcels = ({ clientId }: AddParcelProps): React.ReactElement => {
                         initialFields={initialParcelFields}
                         initialFormErrors={initialParcelFormErrors}
                         clientId={clientId}
-                        editMode={false}
+                        writeParcelInfoToDatabase={insertParcel}
```
Alex-Softwire, line 92: We have AddParcelForm, EditParcelForm that builds upon a common ParcelForm. I have refactored this such that all the add/edit specific code are in their respective AddParcelForm/EditParcelForm files, and the ParcelForm contains the common code. This means that we don't need the editMode boolean flag. Lmk if you agree with this idea/change :)

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -219,8 +220,7 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
 
         const isDelivery = fields.shippingMethod === "Delivery";
 
-        const formToAdd = {
-            primary_key: parcelId!,
+        const parcelRecord = {
```
Alex-Softwire, line 217: Renamed so it's the same name as the variable in clientDatabaseFunctions. Primary key removed as it wasn't used I think..

src/app/parcels/add/AddParcelForm.tsx
AmeliaIsCoding: I think this is good

src/app/parcels/edit/EditParcelForm.tsx
AmeliaIsCoding: This is great, nice job!

I think there is the possibility of a user being on the edit parcel page, choosing a valid slot, and another user then turning off that packing slot in the admin panel, then the first user submitting the parcel without error. However, I think this currently possible anyway as I don't think `packingSlotIsShown` is refetched when the database changes (although I haven't checked). So I think this change is okay.

src/app/parcels/form/ParcelForm.tsx
AmeliaIsCoding: Yeah this is good, I don't think the primary key was being used either

src/app/parcels/form/clientDatabaseFunctions.ts
AmeliaIsCoding, line 1: i feel like the name of this file isn't very descriptive / isn't consistent with the client forms, could you change it to something better please? (aware you didn't come up with it)

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -92,13 +99,12 @@ interface ParcelFormProps {
...
-    editMode: boolean;
-    parcelId?: string;
     deliveryPrimaryKey: Schema["collection_centres"]["primary_key"];
     collectionCentresLabelsAndValues: CollectionCentresLabelsAndValues;
     packingSlotsLabelsAndValues: PackingSlotsLabelsAndValues;
-    packingSlotIsShown?: boolean;
-    collectionCentreIsShown?: boolean;
+    writeParcelInfoToDatabase:
+        | ((parcelRecord: ParcelDatabaseUpdateRecord) => Promise<UpdateParcelReturnType>)
+        | ((parcelRecord: ParcelDatabaseInsertRecord) => Promise<InsertParcelReturnType>);
```
AmeliaIsCoding: I think we should define these types in clientDatabaseFunctions.ts, see my comment there

src/app/parcels/form/clientDatabaseFunctions.ts
```diff
@@ -3,11 +3,14 @@ import { InsertSchema, UpdateSchema } from "@/databaseUtils";
 import { logErrorReturnLogId, logWarningReturnLogId } from "@/logger/logger";
 import { AuditLog, sendAuditLog } from "@/server/auditLog";
 
-type ParcelDatabaseInsertRecord = InsertSchema["parcels"];
-type ParcelDatabaseUpdateRecord = UpdateSchema["parcels"];
+export type ParcelDatabaseInsertRecord = InsertSchema["parcels"];
+export type ParcelDatabaseUpdateRecord = UpdateSchema["parcels"];
```
AmeliaIsCoding: Can we actually instead define and export types for the functions that either insert or update a parcel? Then we can use those in the form props, and also in the function definitions below :)
Lmk if you don't know what I mean

src/app/parcels/form/clientDatabaseFunctions.ts
```diff
@@ -3,11 +3,14 @@ import { InsertSchema, UpdateSchema } from "@/databaseUtils";
...
 
-type InsertParcelErrors = "failedToInsertParcel";
-type InsertParcelReturnType = { error: { type: InsertParcelErrors; logId: string } | null };
+export type InsertParcelErrors = "failedToInsertParcel";
+export type InsertParcelReturnType = {
```
AmeliaIsCoding: we would also be able to remove the exports here too

src/app/parcels/form/clientDatabaseFunctions.ts
```diff
@@ -31,15 +34,18 @@ export const insertParcel = async (
...
 
     await sendAuditLog({ ...auditLog, wasSuccess: true, parcelId: data.primary_key });
-    return { error: null };
+    return { error: null, parcelId: data.primary_key };
 };
 
-type UpdateParcelErrors = "failedToUpdateParcel" | "concurrentUpdateConflict";
-type UpdateParcelReturnType = { error: { type: UpdateParcelErrors; logId: string } | null };
+export type UpdateParcelErrors = "failedToUpdateParcel" | "concurrentUpdateConflict";
+export type UpdateParcelReturnType = {
```
AmeliaIsCoding: and we could remove these exports

src/app/parcels/form/clientDatabaseFunctions.ts
```diff
@@ -31,15 +34,18 @@ export const insertParcel = async (
     if (error) {
         const logId = await logErrorReturnLogId("Error with insert: parcel data", error);
         await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
-        return { error: { type: "failedToInsertParcel", logId } };
+        return { error: { type: "failedToInsertParcel", logId }, parcelId: undefined };
```
AmeliaIsCoding: can you use null instead of undefined to be consistent with how we return errors/data elsewhere?
(same for updateParcel)

src/app/parcels/form/clientDatabaseFunctions.ts
```diff
@@ -3,11 +3,14 @@ import { InsertSchema, UpdateSchema } from "@/databaseUtils";
...
-type ParcelDatabaseUpdateRecord = UpdateSchema["parcels"];
+export type ParcelDatabaseInsertRecord = InsertSchema["parcels"];
+export type ParcelDatabaseUpdateRecord = UpdateSchema["parcels"];
 
-type InsertParcelErrors = "failedToInsertParcel";
-type InsertParcelReturnType = { error: { type: InsertParcelErrors; logId: string } | null };
+export type InsertParcelErrors = "failedToInsertParcel";
+export type InsertParcelReturnType = {
+    error: { type: InsertParcelErrors; logId: string } | null;
+    parcelId: string | undefined;
```
AmeliaIsCoding: this is such a nitpick but please change the order of these lines, and in the function, change the return to be `return {parcelId: ..., error: ...}` instead of `return {error: ..., parcelId: ...}`

Similarly for updateParcel.

And also please change the order in line 233 in ParcelForm.tsx

AmeliaIsCoding (changes requested):
I really like the refactoring of the forms! Some comments but they are all small

src/app/parcels/form/clientDatabaseFunctions.ts
Alex-Softwire: Yep, happy do do that, may I ask why? I guess the parcelId is the more important field than the error? Does it match consistency with something?

src/app/parcels/form/clientDatabaseFunctions.ts
Alex-Softwire: We could call it submitFormHelpers to be consistent with the client forms.

Alternatively we could split it into two files called addParcelToDatabase.ts and updateParcelInDatabase.ts or something like that and putting it into the add/edit folders? Though if I was do that, I'd want to do similar to the client forms. Will do the former for now but lmk if you think something else / the latter is better

src/app/parcels/edit/EditParcelForm.tsx
Alex-Softwire: Aah good point, though yeah I don't think it would have updated correctly in its current state. We could always edit this if we do need to change.

src/app/parcels/form/clientDatabaseFunctions.ts
Alex-Softwire: e.g put something like
`export type WriteParcelToDatabaseFunction =
    | ((parcelRecord: ParcelDatabaseUpdateRecord) => Promise<UpdateParcelReturnType>)
    | ((parcelRecord: ParcelDatabaseInsertRecord) => Promise<InsertParcelReturnType>);`
    
 in clientDatabaseFunctions.ts

Alex-Softwire: Am a little confused how to use it in the function definition below (but can save it in the form type)

src/app/parcels/form/clientDatabaseFunctions.ts
Alex-Softwire: Yep happy to do this :) Am used to using undefined for errors but it makes sense for consistency here!

src/app/parcels/form/clientDatabaseFunctions.ts
Alex-Softwire: N.b we still need the ParcelDatabaseUpdateRecord type to be exported as it's needed in Line 140 in EditParcelForm so we can inject the parcelId into the function that we're passing through.

src/app/parcels/form/clientDatabaseFunctions.ts
AmeliaIsCoding: agree completely with your whole comment

src/app/parcels/form/clientDatabaseFunctions.ts
AmeliaIsCoding: cool!

src/app/parcels/form/clientDatabaseFunctions.ts
AmeliaIsCoding: yep, it's just that in the rest of the codebase we return the data and then the error, i suppose the data is more important in some way? but its the consistency that i think we care about

src/app/parcels/form/ParcelForm.tsx
AmeliaIsCoding: I think this was intended, but I don't really have a preference, @rickaw @minalui thoughts?

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -4,10 +4,18 @@ import { logErrorReturnLogId, logWarningReturnLogId } from "@/logger/logger";
...
 type InsertParcelErrors = "failedToInsertParcel";
-type InsertParcelReturnType = { error: { type: InsertParcelErrors; logId: string } | null };
+type InsertParcelReturnType = {
+    error: { type: InsertParcelErrors; logId: string } | null;
+    parcelId: string | null;
+};
+
+export type WriteParcelToDatabaseFunction =
+    | ((parcelRecord: ParcelDatabaseUpdateRecord) => Promise<UpdateParcelReturnType>)
+    | ((parcelRecord: ParcelDatabaseInsertRecord) => Promise<InsertParcelReturnType>);
```
AmeliaIsCoding: I recommend defining these as seperate types like `UpdateParcel` = (parcelRecord: ParcelDatabaseUpdateRecord) => Promise<UpdateParcelReturnType>` and similarly for inserting

src/app/parcels/form/ParcelForm.tsx
Alex-Softwire: It feels a little odd that you can't resubmit in the case of an error (without refreshing the page), especially a currency error, that waiting a bit and resubmitting should fix. That being said, if there's a reason am happy to remove line 228 (which will keep the submit button greyed out in the case of an error).

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -4,10 +4,18 @@ import { logErrorReturnLogId, logWarningReturnLogId } from "@/logger/logger";
...
+export type ParcelDatabaseErrors = InsertParcelErrors | UpdateParcelErrors;
 
 export const insertParcel = async (
```
AmeliaIsCoding: because then we can define this function as `insertParcel: InsertParcel = async (parcelRecord) => {...}`

(and similarly for update)

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -31,15 +39,18 @@ export const insertParcel = async (
     if (error) {
         const logId = await logErrorReturnLogId("Error with insert: parcel data", error);
         await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
-        return { error: { type: "failedToInsertParcel", logId } };
+        return { parcelId: null, error: { type: "failedToInsertParcel", logId } };
     }
 
     await sendAuditLog({ ...auditLog, wasSuccess: true, parcelId: data.primary_key });
-    return { error: null };
+    return { error: null, parcelId: data.primary_key };
```
AmeliaIsCoding: can you change the order of this one too please?

src/app/parcels/form/ParcelForm.tsx
AmeliaIsCoding: I think I agree with you

src/app/parcels/form/clientDatabaseFunctions.ts
AmeliaIsCoding: I've suggested something slightly different in the file

AmeliaIsCoding: I had a look into the line 140 stuff and I can't think of another way to do it so happy to leave it like that for now

AmeliaIsCoding (changes requested):
ok i tried to explain in the comments but i'm not sure how well I did, but basically you can define the types of each of the insert, update functions, and then when you define the functions, you can use those types. and you can export the types of each function, and then in the parcel form props, the writeParcelToDatabase function can be a union of those types. i tried to give an example in the comments. lmk if it doesn't make sense :)

src/app/parcels/form/submitFormHelpers.ts
Alex-Softwire: whoops apologies for missing that one haha

src/app/parcels/form/submitFormHelpers.ts
Alex-Softwire: Nice: that works :), didn't realise you could write function types like that! Having a little bit of trouble with the Update Parcel as one has to inject the parcel id in.

src/app/parcels/form/ParcelForm.tsx
minalui: Thanks for flagging this, originally this was intended so that users don't try to edit the record again until they have up to date data in the form (concurrent editing error). I suppose if we can distinguish between concurrent editing error (we don't allow resubmitting) and other errors (allows resubmitting)?

Alex-Softwire: Yeah I would be happy to implement this - I guess the error message does say please refresh the page so that makes sense

Alex-Softwire: Done :)

src/app/parcels/form/submitFormHelpers.ts
Alex-Softwire: Managed to rewrite the dependency injection in update parcel to hopefully fix this issue!

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -230,36 +224,23 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
+        if (parcelId) {
+            router.push(parcelModalRouterPath(parcelId));
+        }
+
+        if (error) {
+            /*
+               If the user is trying to edit a parcel that is currently being edited,
+               we want the user to refresh the page to get the current data, so we keep the submit button
+               disabled.
+             */
```
AmeliaIsCoding: I think this comment isn't really needed, but up to you

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -3,15 +3,21 @@ import { InsertSchema, UpdateSchema } from "@/databaseUtils";
 import { logErrorReturnLogId, logWarningReturnLogId } from "@/logger/logger";
 import { AuditLog, sendAuditLog } from "@/server/auditLog";
 
+export type WriteParcelToDatabaseFunction = UpdateParcel | InsertParcel;
+export type ParcelDatabaseErrors = InsertParcelErrors | UpdateParcelErrors;
```
AmeliaIsCoding: can we call this `WriteParcelToDatabaseErrors` instead please just to be a bit clearer?

AmeliaIsCoding (approved):
I think I'm happy with this now, barring the 2 small changes I asked, but once you've addressed, happy for Ricky or Kristina to review

src/app/parcels/form/ParcelForm.tsx
Alex-Softwire: Yeah, I added as I thought it might be a little confusing as to why we're doing something different for one specific error but on second thoughts would be happy without as it's a bit long.

src/app/parcels/edit/EditParcelForm.tsx
```diff
@@ -111,19 +112,18 @@ const EditParcelForm = ({ parcelId }: EditParcelFormProps): React.ReactElement =
...
         shippingMethod: Errors.none,
         collectionDate: Errors.none,
         collectionTime: Errors.none,
-        collectionCentre: Errors.none,
+        collectionCentre: collectionCentreIsShown ? Errors.none : Errors.invalidCollectionCentre,
```
8kristy, line 126: Think I'm a bit confused by errors in general, do you mind explaining what is happening here? Why are most of them none, and why would a packing slot or collection centre show an error - is it if that type was deleted and the box in the edit is now blank? Is it affected by whether it's a collection or a delivery (i.e. the extra section is expanded)?

Think an example would be helpful to understand when these would appear

8kristy (approved):
Just a question from me because I don't understand what's going on, otherwise lovely refactoring, feel free to merge when ready

src/app/parcels/edit/EditParcelForm.tsx
Alex-Softwire: Yep: if the parcel has an AM slot for example and an admin deletes (or hides) that slot, when the form is opened the packing slot will be shown as blank and an error message (saying this packing slot no longer exists.. or something like that) will show. This should be equivalent to how it was before, it was originally done via useStates that weren't necessary methinks.

Alex-Softwire: They are mostly none as a parcel that you're editing shouldn't have any errors, as it wouldn't have allowed you to created it in the first place, just those edge cases that could cause it. I don't think collection or delivery should affect things - though the invalidCollectionCentre would only show if the parcel was a collection & the collection centre has subsequently been hidden/removed

PR #341, opened 2024-05-15
VFB-130: Changed shipping label sizes (new)

KylePUG:
## What's changed
 - Changed label size from 200mm to 150mm

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image|image|
|image|image|

## Checklist

If you have made any changes to the database... (no changes)


Alex-Softwire (approved):
Looks good! Rather hard to review a 3 character change haha, but found this comment from dan: 
"Label size would ideally be 150mm x 62mm now, not 200mm x 62mm. It is currently 150mm on dev environment", and it looks like you've changed that so yay!

8kristy (approved):
My one concern is Dan said "label printer which uses a continuous roll of 62mm wide labels", and I wonder if that means it prints sideways and the labels need to be rotated...
I'm sure they can flip it in the print settings, guess just something to keep in mind (or you can double check with someone and update it)

Issue #206, opened 2024-03-26
Default Parcel Page Sorting (Date>Shift>Method>Postcode)


dan-langridge:
### Describe the solution / feature you'd like

By default when you load the Parcels page, the parcels should in the following order:

1. Date
2. Packing Shift (AM/PM/Shift1,2,3,4)
3. Method (Delivery, then Collection)
4. Postcode

So that:
The AM Shift is at the top of the page, with deliveries first (sorted by postcode) and collections second (sorted by postcode)

Then PM list has deliveries first (sorted by postcode) and collections second (sorted by postcode)

e.g.
<img width="626" alt="Screenshot 2024-03-26 at 17 11 24" src="https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/80202270/bd0eae65-bd1f-4b6a-8646-a765c02e21f9">

This can then be overridden by user sorting. But default loading of page should be the above.

AmeliaIsCoding:
Notes for devs:

This should also be the sorting behaviour on sorting by the packing date column.

kaydentsoftwire:
QA steps:
**A note that sorting by postcode uses default sorting, and therefore postcodes beginning E24, for example, will appear before postcodes using E8**
- go to admin / collections centre table, and change 'Clapham Collection Centre' to 'Lapham Collection', changing the first letter of the acronym to 'L', as well. 
- navigate to the parcels page
- [ ] You should see the most recent date, with AM Shift at the top of the page, with deliveries first (sorted by postcode) and collections second (sorted by postcode). If these shifts don't have both delivery/collection or there are not both AM/PM on that date, expand the date range to check this
- [ ] Then PM list has deliveries first (sorted by postcode) and collections second (sorted by postcode), followed by any of Shift 1/2 (in number order (see above for example)
- Select a date range of at least 2 days
- Click to sort by packing date
- [ ] The date should be the earliest date now, but still retain the order - AM (sorted by postcode) then PM (sorted by postcode)
- Reverse the sorting
- [ ] The sorting should return to the same as the default state, with delivery always appearing first, regardless of which way it is ordered

kaydentsoftwire:
Is there a specific way the postcodes should be ordered? Is it useful if it goes, for example,
E1, E12, E2, E23, E3
or should it be in the order
E1, E2, E12, E23, E3
with the same structure replicated in the second half of the postcode?

8kristy:
[...]

PR #343, opened 2024-05-15
#206: Default Parcel Page Sorting

kaydentsoftwire:
## What's changed
The parcels page now has default sorting in this order:

1. Date
2. Packing Shift (AM/PM/Shift1,2,3,4)
3. Method (Delivery, then Collection)
4. Postcode
5. Collection centre

The same sorting applies when sorting by packing date.
Postcode is sorted using default sorting, so E24 1AA will appear before E8 1AA

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

- [x] The migration files are up-to-date with my final set up (npx supabase db diff -f <name_of_migration> should create nothing at this point)
- [x] I have updated the typescript definitions for the database with db:local:generate_types
- [x] I have modified the seed in supabase/seed/seed.mts if appropriate
- [x] If I have modified the seed, I have also generated the seed with npm run db:generate_seed
- [x] With my final set up, I can run npm run dev:reset_supabase without any errors.
- [x] Does this require resetting the deployed database? - NO

    Make sure you have rebased your branch onto dev or merged dev into your branch.
    Checkout the dev branch
    Run npm run post_checkout to reset the database, including the seed data.
    Checkout your branch
    Run npx supabase migrations up --include-all --local to run all outstanding migration files
    Check that the resulting database is what you expect, bar any seed data changes.


kaydentsoftwire (commented):
Just adjusted the sorting using supabase-js's .order in these two places

AmeliaIsCoding:
We need delivery to appear above collection, is this the case?

AmeliaIsCoding (changes requested):
Can I check if the delivery appears above collection, like Dan specified? (I can't really see in the screenshot as it is cut off)

Also, nitpick, but PR titles follow a particular format :)

kaydentsoftwire:
It does, yeah - sorry just realised that today's date doesn't give the most useful example - see here:

image

AmeliaIsCoding (approved):

src/app/parcels/ParcelsPage.tsx
```diff
@@ -129,7 +129,8 @@ const sortableColumns: SortOptions<ParcelsTableRow>[] = [
                 query
                     .order("packing_date", { ascending: sortDirection === "asc" })
                     .order("packing_slot_order")
-                    .order("client_full_name"),
+                    .order("collection_centre_name", { ascending: false })
```
8kristy: From what I understand this sorts by looking at "Delivery" vs "Collection - ..." and when sorted backwards D appears before C. Can we always assume that the centre name will begin with "Collection"? I.e. is it hardcoded somewhere or is it used input? Since users might just input "Lambeth collection centre" and then that will float to the top :p  

If that's the case, do we have a better field by which we can sort, maybe some boolean?

src/app/parcels/ParcelsPage.tsx
```diff
@@ -129,7 +129,8 @@ const sortableColumns: SortOptions<ParcelsTableRow>[] = [
...
+                    .order("client_address_postcode"),
```
8kristy: Do we know what the desired sorting order is with postcodes? Think default JS sorting just looks at the characters so if you gave it a list of these postcodes 
```
["E2 6AA", "E24 6AA", "E8 6AA", "E2 62AA", "E2 6AA", "E2 8AA"]
```
it would sort them as
```
['E2 62AA', 'E2 6AA', 'E2 6AA', 'E2 8AA', 'E24 6AA', 'E8 6AA']
```
whereas I'd expect something more like 
```
['E2 6AA', "E2 6AA", "E2 8AA", "E2 62AA", "E8 6AA", "E24 6AA"]
```
but I don't know how much that matters, especially since postcode are sometimes in very odd places on the map 
image
Might be worth double checking with someone or leaving it for another time, if you do decide to update it it might be worth making a postcode object and shove a comparator function in there?

8kristy (commented):
Just some unclear bits, probably just me not knowing how things work

src/app/parcels/ParcelsPage.tsx
kaydentsoftwire: Hmm good points, I suspect that the important / useful thing is to have the postcodes grouped together, e.g all E1 postcodes appear next to each other in the list. But I have left a comment on the ticket to see if there is a specific order that they should be in

8kristy: Seems like the preference is to have them ordered like I suggested, but Dan said it's ok if we keep them alphabetical. If you can come up with a very easy solution to sort them correctly it would be nice, but feel free to leave it as is :)

supabase/migrations/20240517141758_add_is_delivery_bool_to_collection_centre_table.sql
```diff
@@ -0,0 +1,33 @@
...
+    last_status.event_data AS last_status_event_data,
+    last_status."timestamp" AS last_status_timestamp,
+    last_status.workflow_order AS last_status_workflow_order
+   FROM (((((parcels
+     LEFT JOIN collection_centres ON ((parcels.collection_centre = collection_centres.primary_key)))
+     LEFT JOIN clients ON ((parcels.client_id = clients.primary_key)))
+     LEFT JOIN packing_slots ON ((parcels.packing_slot = packing_slots.primary_key)))
+     LEFT JOIN family_count ON ((family_count.family_id = clients.family_id)))
+     LEFT JOIN last_status ON ((last_status.parcel_id = parcels.primary_key)))
+  ORDER BY parcels.packing_date DESC;
```
kaydentsoftwire, line 37: I created a migration that adds a is_delivery boolean to the collections centre table and then recreates the parcel_plus table with this boolean pulled into it
The files above are updating types and the test to allow for this new column

supabase/seed/collectionCentresSeed.mts
```diff
@@ -10,5 +10,6 @@ export const collectionCentres = [
     {
         name: "Delivery",
         acronym: "DLVR",
+        isDelivery: true,
     },
 ];
```
kaydentsoftwire, line 17: the seed is updated to incorporate the new column. The column has a default value of false, as the presumption is that any new collection centres will not be delivery (only Delivery will be delivery!) but the Delivery column is seeded as true.

src/app/parcels/parcelsTable/sortableColumns.ts
```diff
@@ -38,7 +38,8 @@ const parcelsSortableColumns: SortOptions<ParcelsTableRow, ParcelsSortMethod>[]
             query
                 .order("packing_date", { ascending: sortDirection === "asc" })
                 .order("packing_slot_order")
-                .order("client_full_name"),
+                .order("is_delivery", { ascending: false })
+                .order("client_address_postcode"),
     },
     {
         key: "packingSlot",
```
kaydentsoftwire, line 46: updates sorting by packing date to reflect order 
    Date
    Packing Shift (AM/PM/Shift1,2,3,4)
    Method (Delivery, then Collection)
    Postcode

supabase/migrations/20240517141758_add_is_delivery_bool_to_collection_centre_table.sql
```diff
@@ -0,0 +1,38 @@
+alter table "public"."collection_centres" add column "is_delivery" boolean not null default false;
+
+update "public"."collection_centres"
+       set is_delivery=true
+       where name='Delivery';
+
```
kaydentsoftwire, line 6: migration creates the 'is_delivery' column and updates the row where the name is 'Delivery' to be true; everything else defaults to false

src/app/parcels/parcelsTable/fetchParcelTableData.ts
```diff
@@ -46,7 +46,8 @@ const getParcelsQuery = (
         query = query
             .order("packing_date", { ascending: false })
             .order("packing_slot_order")
-            .order("client_full_name");
+            .order("is_delivery", { ascending: false })
```
8kristy, line 49: I think after is_delivery you also want to sort alphabetically by collection centre name?

8kristy: Also would it be difficult and would it make sense to fix this here? https://softwiretech.atlassian.net/browse/VFB-150
Ideally this should also apply to the default sorting if it doesn't already
Think it would make the most sense to sort
Default: date -> slot -> is_delivery -> collection centre name for is_delivery == false, has congestion charge if is_delivery == true
By Method: is_delivery -> collection centre name for is_delivery == false, has congestion charge if is_delivery == true and then reverse of that if it's sorted the other way

src/app/parcels/parcelsTable/sortableColumns.ts
8kristy: Update to sort by collection centre name/congestion charge here too

supabase/seed/collectionCentresSeed.mts
8kristy: Might be worth explicitly stating that collection centres are false as otherwise someone might assume it's null

8kristy (commented):
Nice work on the db changes :) Just missing the collection centre name ordering, and then congestion charges but let me know if it's too difficult and we can do that as the current ticket separately at some point

8kristy (approved):

PR #345, opened 2024-05-16
VFB-224: Update local storage

rickaw:


supabase/tools/createStorage.ts
```diff
@@ -23,21 +23,24 @@ async function createPublicBucket(supabase: SupabaseClient, name: string): Promi
...
-async function uploadCongestionZoneFile(supabase: SupabaseClient): Promise<void> {
+async function uploadCongestionZoneFile(
+    supabase: SupabaseClient,
+    bucketName: string
+): Promise<void> {
     const file = fs.readFileSync("supabase/tools/congestion-zone-postcodes.txt");
 
     const { error } = await supabase.storage
-        .from("public")
+        .from(bucketName)
```
8kristy, line 37: Is it different for creating buckets and the "from" part needs to be the actual bucket name?

supabase/tools/createStorage.ts
```diff
@@ -23,21 +23,24 @@ async function createPublicBucket(supabase: SupabaseClient, name: string): Promi
     });
 
     if (error) {
-        throw new Error(`Failed to create bucket called ${name}, error: ${error}`);
+        throw new Error(`Failed to create bucket called ${name}, error: ${error.message}`);
```
8kristy, line 26: I assume this was just wrong before?

supabase/tools/createStorage.ts
rickaw: this is inconsistent with how we access the bucket in the edge function but this is in fact the syntax that works when i tested locally

rickaw: which is a big shame

supabase/tools/createStorage.ts
rickaw: yup yup the error message came out as `[Object object]` so showing the message instead now

8kristy (approved):

Issue #159, opened 2024-03-18
Attempting to print duplicate 'Shipping Labels' or 'Shopping List' needs a warning message


dan-langridge:
### Describe the solution / feature you'd like

If printing Labels or Shopping list for a second time, a warning message needs to appear before the action is run: 
	“The following postcodes have already been printed: E3 2AB. Are you sure you want to print again?”

AmeliaIsCoding:
Notes for devs:
I think this best solution will be to check the last status of the selected parcel, and if it is Shipping Labels Downloaded / Shopping List Downloaded, display the warning message.

maxjohns0n:
> Notes for devs: I think this best solution will be to check the last status of the selected parcel, and if it is Shipping Labels Downloaded / Shopping List Downloaded, display the warning message.

We agreed to instead check the `events` table to see if the parcel has ever had any Shipping Labels Downloaded / Shopping List Downloaded events. This is in case there has been one but it isn't the most recent event for that parcel.

maxjohns0n:
### QA

- Download some shipping labels for a couple of parcels with different postcodes
- [x] When trying to download again, a warning should appear in the modal
- [x] The postcodes of the parcels you already downloaded should be listed
- [x] Any parcels selected which don't already have any labels downloaded should not be listed
- [x] When no parcels with labels downloaded are selected, the warning doesn't appear
- [x] Postcodes should not be duplicated in the warning message, even if there are multiple parcels selected with that postcode
- [x] Repeat for shopping list downloads

KylePUG:
QA passed!

I do wonder though if more specificity is needed, e.g. when a lot of parcels with the same postcode is downloaded but only one of them has been downloaded before - it will be quite hard to figure out which it is if it isn't displayed in "Last Status". Maybe an optional 'more info' dropdown in the warning message that lists the parcels already downloaded or the number of parcels that are already downloaded for each postcode? Possibly something for another ticket though.

Anyway, this seems to do what's required so well done! :)

maxjohns0n:
> I do wonder though if more specificity is needed, e.g. when a lot of parcels with the same postcode is downloaded but only one of them has been downloaded before - it will be quite hard to figure out which it is if it isn't displayed in "Last Status". Maybe an optional 'more info' dropdown in the warning message that lists the parcels already downloaded or the number of parcels that are already downloaded for each postcode? Possibly something for another ticket though.


[...]

PR #346, opened 2024-05-16
#159: Add warning when downloading duplicate shipping labels or shopping lists

maxjohns0n:
## What's changed
* When trying to download a shipping label or shopping list for a parcel which already has one, display a warning.
* Implemented in a `DuplicateDownloadWarning` component which will look for a given event (e.g. "Shipping Labels Downloaded") and find all the parcels in a list which have at least one of these events.
* This replaces the previous similar feature for shopping lists.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

No db changes


src/components/Modal/Modal.tsx
```diff
@@ -7,7 +7,7 @@ import { faClose } from "@fortawesome/free-solid-svg-icons";
 import Icon from "@/components/Icons/Icon";
 
 export const StyledDialog = styled(Dialog)`
-    & .MuiPaper-root {
+    & .MuiPaper-root:first-of-type {
```
maxjohns0n, line 10: This seemed to be intended to modify the style of the main modal box, but actually selected all `MuiPaper-root` classes inside the modal. This was messing with the styling of my `Alert` so I changed this to just select the first `MuiPaper-root` class, which should just affect the modal itself now.

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
```diff
@@ -1,49 +1,13 @@
...
 import ShoppingListPdfButton from "@/pdf/ShoppingList/ShoppingListPdfButton";
 import { ShoppingListPdfError } from "@/pdf/ShoppingList/getShoppingListData";
 import { sendAuditLog } from "@/server/auditLog";
-import { nullPostcodeDisplay } from "@/common/format";
-
-interface ShoppingListsConfirmationProps {
-    selectedParcels: ParcelsTableRow[];
-}
-
-const ShoppingListsConfirmation: React.FC<ShoppingListsConfirmationProps> = ({
```
maxjohns0n, line 22: Replacing this with the new implementation so it is consistent

src/app/parcels/fetchParcelTableData.ts
```diff
@@ -322,3 +323,21 @@ export const fetchParcelStatuses = async (): Promise<ParcelStatusesReturnType> =
...
     return { data: parcelStatusesList, error: null };
 };
+
+export const getParcelsWithEvent = async (
+    targetEventName: StatusType,
+    parcelIds: string[]
+): Promise<GetDbParcelDataResult> => {
+    const { data, error } = await supabase
+        .from("parcels_plus")
+        .select("*, events!inner(new_parcel_status)")
```
rickaw: Would this work if a parcel has no events associated with it?

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const DuplicateDownloadWarning: React.FC<DuplicateDownloadWarningProps> = async (props) => {
+    const parcelIds = props.parcels.map((parcel) => parcel.parcelId);
+    const { parcels, error } = await getParcelsWithEvent(props.targetEventName, parcelIds);
+    if (error) {
+        return;
+    }
+
+    const postcodes = parcels.map((parcel) => parcel.client_address_postcode);
+    const uniquePostcodes = Array.from(new Set(postcodes));
```
rickaw: I feel like this task of finding whether there'd be duplicate events should be delegated to the parent element. It feels a little confusing to see `<DuplicateDownloadWarning ...>` in the parent when the warning doesn't always exist. What do you think?

src/app/parcels/fetchParcelTableData.ts
maxjohns0n: I think so yeah. The inner join means only parcels with events are returned, which is all we want here

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
maxjohns0n: I agree it could be confusing. My thinking was to have all the functionality in one component which can then be shared by both the shipping label and shopping list modals so that we could avoid repeating code/logic between the two. So I guess that's the tradeoff - do you have a preference? Or is there another way to do it?

src/app/parcels/parcelsTable/fetchParcelTableData.ts
```diff
@@ -280,3 +281,21 @@ export const getClientIdForParcel = async (parcelId: string): Promise<FetchClien
 
     return { clientId: clientIdData.client_id, error: null };
 };
+
+export const getParcelsWithEvent = async (
```
AmeliaIsCoding: I'd maybe change the name to getParcelsByEvent, to indicate we're filtering by the event instead of returning the parcel plus the event, however if you think this is best I'm happy with this too!

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
8kristy: I'm thinking that the postcode retrieval should be a helper function (probably near `getParcelsWithEvent`?) so then you can reuse it for both, and then this component would only contain bits for the styled alert i.e. 
```
const StyledAlert = styled(Alert)`
    border-radius: 4px;
`;

return (
            <StyledAlert severity="warning">
                The following postcodes have already been downloaded: {uniquePostcodes.join(", ")}.
                Are you sure you want to print again?
            </StyledAlert>
        )
    );
```
and then add the component with `duplicateDownloadedPostcodes.lenght > 0 && </DuplicateDownloadWarning>` so it's clear it only appears when we have duplicate downloads. What do you think?

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+    return (
+        uniquePostcodes.length > 0 && (
+            <StyledAlert severity="warning">
+                The following postcodes have already been downloaded: {uniquePostcodes.join(", ")}.
+                Are you sure you want to print again?
+            </StyledAlert>
+        )
+    );
```
8kristy: Wording like that to me implies a confirmation window
Does the warning appear after you press print or just when you open the modal? I think it's better if it appears when you press the button, and then the button text turns into "Confirm Download PDF"
If it's the other way this might need some rewording but I think I prefer the confirmation way (I'm actually not sure - is it meant to show a warning like that or have another window pop up? I guess both patterns get used in the wild)

Let me know your thoughts maybe I'm misinterpreting something

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
```diff
@@ -1,49 +1,13 @@
...
 import SelectedParcelsOverview from "../SelectedParcelsOverview";
-import { ParcelsTableRow } from "../../parcelsTable/types";
-import { StatusType, getStatusErrorMessageWithLogId } from "../Statuses";
+import { getStatusErrorMessageWithLogId } from "../Statuses";
 import ShoppingListPdfButton from "@/pdf/ShoppingList/ShoppingListPdfButton";
 import { ShoppingListPdfError } from "@/pdf/ShoppingList/getShoppingListData";
 import { sendAuditLog } from "@/server/auditLog";
-import { nullPostcodeDisplay } from "@/common/format";
-
-interface ShoppingListsConfirmationProps {
```
8kristy, line 18: Ticket asked to add warning to shipping labels and shopping lists - is it just the case that shopping lists had a warning and we only needed it on shipping labels? Does this have the same behaviour as the new method?

src/app/parcels/parcelsTable/fetchParcelTableData.ts
```diff
@@ -280,3 +281,21 @@ export const getClientIdForParcel = async (parcelId: string): Promise<FetchClien
...
+    targetEventName: StatusType,
+    parcelIds: string[]
+): Promise<GetDbParcelDataResult> => {
+    const { data, error } = await supabase
+        .from("parcels_plus")
+        .select("*, events!inner(new_parcel_status)")
```
8kristy: I'm not super familiar with this syntax but do we need to select * or can we just select parcel id, postcode and event status? (assuming those are not the only fields returned by the query)

8kristy (commented):
The new approach is neat but might need a bit of restructuring, I wonder if we need a bit more inspiration from the old approach

Also in the screenshots the download button is greyed out? Is it meant to be like that, can you click it?

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
maxjohns0n: Currently it just appears when you open the modal. Adjusting the wording would be much simpler than changing the button's behaviour but I'm happy to do whichever you think is better. I guess you could also argue that this modal is a kind of confirmation anyway but I would tend to agree with you that it implies a separate confirmation step.

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
maxjohns0n: Yes it turned out the shopping list already had one, but it had slightly different behaviour to what we decided on for this ticket so I shared the new component across both modals.

src/app/parcels/parcelsTable/fetchParcelTableData.ts
maxjohns0n: I'm also not too familiar, but when I tried to just select some of the fields, there was a type error because it no longer matched the full schema of the `parcels_plus` view. I think other queries generally just used a `select *` so I copied that. There may be a better workaround I'm not aware of though!

maxjohns0n:
> Also in the screenshots the download button is greyed out? Is it meant to be like that, can you click it?

The button is greyed out simply because the Quantity field hasn't been filled in yet in the screenshots.

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
maxjohns0n: Ok I've tried this out, lmk what you think!

src/app/parcels/parcelsTable/fetchParcelTableData.ts
AmeliaIsCoding: I would suggest making your own type using the `parcels_plus` schema and `Pick` to avoid having to fetch so much data. Then we can use that type in the return type of the function

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
AmeliaIsCoding: Personally I feel showing the text before clicking the button is what Dan wanted, as this is the behaviour that exists already in another action

AmeliaIsCoding (commented):
suggestions

src/app/parcels/parcelsTable/fetchParcelTableData.ts
maxjohns0n: Thanks! I realised the only field we actually needed was the postcode, so I've changed it to just fetch that now, and moved the uniqueness stuff into the same function now that it's just getting postcodes.

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
```diff
@@ -110,11 +80,25 @@ const ShoppingListModal: React.FC<ActionModalProps> = (props) => {
...
+                "Shopping List Downloaded",
+                parcelIds
+            );
+            if (error) {
+                return;
+            }
+            setDuplicateDownloadedPostcodes(postcodes);
+        };
+        fetchData();
+    }, [parcelIds]);
```
8kristy: Can you move this into a separate helper file (not sure if an appropriate one already exists)? So this will become
```
   const getDuplicateDownloadedPostcodes = async (parcelIds: string[], eventName: string): Promise<void> => {
            const { postcodes, error } = await getParcelPostcodesByEvent(
                "Shopping List Downloaded",
                parcelIds
            );         
            return error ? null : postcodes;
        };
```
somewhere, and then in this file and shipping labels you can do
```
const parcelIds = props.selectedParcels.map((parcel) => parcel.parcelId);
const duplicateDownloadedPostcodes = getDuplicateDownloadedPostcodes(parcelIds, "Shopping List Downloaded")

```

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+const StyledAlert = styled(Alert)`
+    border-radius: 4px;
+`;
+
+const DuplicateDownloadWarning: React.FC<DuplicateDownloadWarningProps> = (props) => {
+    return (
+        <StyledAlert severity="warning">
+            The following postcodes have already been downloaded:{" "}
```
8kristy: What is the `{" "}` for?

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
```diff
@@ -0,0 +1,28 @@
...
+            {props.postcodes
+                .map((postcode) => postcode ?? displayPostcodeForHomelessClient)
```
8kristy: Should we just ignore the ones that don't have a postcode?

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
```diff
@@ -0,0 +1,28 @@
...
+                .join(", ")}
+            you sure you want to print again?
```
8kristy:
```suggestion
                .join(", ")}.
            Are you sure you want to print again?
```

8kristy (commented):
Couple minor things

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
8kristy: Left a comment on review :)

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
8kristy: Ok, let's keep it as is :)

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
maxjohns0n: The issue I had was that you can't `await` anything in a client component, which I would need to do for this helper function to return the `string[]` I want instead of a Promise. So I tried to get around this by using an effect and state. Do you know of a better workaround?

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
maxjohns0n: I was in a fight with the linter and lost :( it added this so that the whitespace didn't get ignored because it was at the end of the line. I've reorganised that code now to make it clearer.

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
maxjohns0n: I wasn't sure about this. Originally I did ignore it, but I merged this change from #326 and then thought that the postcodes shown here really just represent parcels which already have labels/lists printed. So if a parcel is going to have a duplicate label we should still flag that, even if it doesn't have a postcode. What do you think?

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
maxjohns0n: Done!

src/app/parcels/ActionBar/DuplicateDownloadWarning.tsx
8kristy: So if the client is deleted and has a `-` as postcode I don't think we need to warn against that - the labels that come out are blank and shouldn't cause any issues (and hopefully it shouldn't really happen anyways). From what I imagine happens now is it shows NFA for both and we'd want to have something like 
```
if isActive then  
   if postcode then 
      postcode 
   else 
      displayPostcodeForHomelessClient 
else 
    don't add 
```
but that might be complicated

Not too sure about NFA - maybe it is good to point it out even if it is a bit confusing

So maybe leaving it as is is the best way to go about it

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
8kristy: Ah, I see. You can still factor out the "guts" of the function to a helper, and then wrap the call in a function:
```
    const fetchData = async (): Promise<void> => {
        setDuplicateDownloadedPostcodes(
            await getDuplicateDownloadedPostcodes(parcelIds, "Shipping Labels Downloaded")
        );
    };

    fetchData();
```
(no need for the useEffect, i think)

It might feel a little silly but it gives better transparency where `getDuplicateDownloadedPostcodes` is used if someone changes it

8kristy: Or even 
```

    let duplicateDownloadedPostcodes = null;

    const fetchData = async (): Promise<void> => {
        duplicateDownloadedPostcodes = await getDuplicateDownloadedPostcodes(
            parcelIds,
            "Shipping Labels Downloaded"
        );
    };

    fetchData();
```
and remove the state

8kristy (commented):
One last suggestion about pulling out a function and then we're good to merge I think :)

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
maxjohns0n: Ok I think I've managed to do it! It turned out the effect (& state) were required so that it updates when you change the selected parcels, otherwise it just fetches once when the page loads (so doesn't work!). But I've made it as simple as possible.

8kristy: Makes sense, looks great now :)

8kristy (approved):

Issue #152, opened 2024-03-14
Add dividing lines between packing slots or dates when ordering by those columns


dan-langridge:
### Describe the solution / feature you'd like

In the warehouse we have 2 packing shifts when parcels are packed, either in the morning (AM) or afternoon (PM). 

However at Christmas, we change from 2 Shifts (AM/PM) to 4 Shift (Shift 1, Shift 2, Shift 3, Shift 4). 

Admin needs to be able to edit these to reflect what season we are in. 

We also have a Shift called Supplies (people who don't get food, but get items like Plates/Cutlery/Bedding/Clothes, and a Shift called Spanish. We use packing shifts as a natural divide on the parcels page to help the manager see the different amount of parcels in each shift, and to say "This person is special and in a category" or, for the callers team, to say, hey, don't ring this client unless you speak spanish...or to our spanish caller, you can easily find all the clients you need to ring in the Spanish section on the page.

All Shifts should have a dividing line between them (like current app) to make it really obvious where the divide is between different shifts

<img width="1411" alt="Screenshot 2024-03-14 at 22 42 13" src="https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/80202270/929d81e9-5d09-462f-84c8-6ed043f85967">

rickaw:
Hey @dan-langridge, thanks for raising this. 

- The ability to add packing slots, and to swtich on and off particular slots is currently being worked on currently with [VFB-37](https://softwiretech.atlassian.net/browse/VFB-37).
- The greeen dividing line between shifts is something new. We'll get working on that soon. Just to clarify, do we want to show the divider only when sorting the parcels by the packing slots ("Time" column in your image)? i.e. No need to show the line when sorting by something else?

dan-langridge:
Thanks Ricky.

I agree with your point, the dividing line would only be required when sorting by:
- Packing Slot (AM/PM)
- Date (if we are viewing parcels across multiple days for example, helpful to have a divider between days)

Thanks
Dan

KylePUG:
## QA 

- Use light mode for this first section
- Navigate to parcels page and sort by packing slots
- [x] Green dividing lines should appear between rows of distinct packing slots
- Increase date range across multiple days and sort by packing dates
- [x] Green dividing lines should appear between rows of distinct days 
- [x] Thinner light green dividing lines should also appear between rows of distinct packing slots
[...]

PR #349, opened 2024-05-17
#152: Add packing slot/date dividing line

KylePUG:
## What's changed
- Added dividing line between distinct packing slots/dates when sorting by packing slots/dates

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image|imageSorting by Packing Date|
| |imageSorting by Packing Slot|

## Checklist

If you have made any changes to the database... (no changes


AmeliaIsCoding:
The screenshots look really good, I'm really impressed you managed to do this! however, I think there's been a slight misinterpretation of the ticket. Since sorting by the date means sorting by date, then packing slot, then other fields, I think when we sort by either date or packing slot, in both cases, we should have a dividing line between each packing slot, and between dates. (maybe the line between dates could be thicker or a darker colour than the line between packing slots).

However, your interpretation might be right instead? @rickaw what do you think Dan meant?

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -117,6 +120,41 @@ const ParcelsPage: React.FC<{}> = () => {
         })();
     }, []);
 
+    function getValueBoundaries<DataType>(values: (DataType | null)[]): number[] {
```
AmeliaIsCoding: lets use an arrow function as this is what we tend to use in the code

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -117,6 +120,41 @@ const ParcelsPage: React.FC<{}> = () => {
...
+        return getValueBoundaries<string>(packingSlotValues);
+    };
+
+    const searchForPackingDateBreakPoints = (parcelsTableRows: ParcelsTableRow[]): number[] => {
+        const packingSlotValues = parcelsTableRows.map((row) => {
+            return row.packingDate?.getDate() ?? null;
+        });
+
+        return getValueBoundaries<number>(packingSlotValues);
+    };
```
AmeliaIsCoding: can you put all of the code form lines 123-157 into a different file in the same parent folder and export it, if possible? you could call it `conditionalStyling.ts` perhaps?

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -147,12 +185,19 @@ const ParcelsPage: React.FC<{}> = () => {
             } else {
                 setParcelsDataPortion(data.parcelTableRows);
                 setFilteredParcelCount(data.count);
+                if (sortedColumn === "packingSlot") {
+                    setParcelRowBreakPoints(searchForPackingSlotBreakPoints(data.parcelTableRows));
+                } else if (sortedColumn === "packingDate") {
+                    setParcelRowBreakPoints(searchForPackingDateBreakPoints(data.parcelTableRows));
+                } else {
```
AmeliaIsCoding: instead of having and if, else if, else structure here, maybe we could define some switch case statement elsewhere (maybe in `conditionalStyling.ts` that takes in the sorted column and returns the function that searches for the break points for that column?

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -49,6 +49,9 @@ const ParcelsPage: React.FC<{}> = () => {
     const [filteredParcelCount, setFilteredParcelCount] = useState<number>(0);
     const [selectedParcelId, setSelectedParcelId] = useState<string | null>(null);
     const [clientIdForSelectedParcel, setClientIdForSelectedParcel] = useState<string | null>(null);
+    const [parcelRowBreakPoints, setParcelRowBreakPoints] = useState<number[]>([]);
+
+    const [sortedColumn, setSortedColumn] = useState<string>("");
```
AmeliaIsCoding: I think this state should be of type `keyof ParcelsTableRow`, which restricts the string to be one of the column keys

EDIT: actually I don't think this state is needed, as this information is already contained in the `sortState` state, correct? we can just use that state, to avoid us having two sources of truth.

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -117,6 +120,41 @@ const ParcelsPage: React.FC<{}> = () => {
         })();
     }, []);
 
+    function getValueBoundaries<DataType>(values: (DataType | null)[]): number[] {
+        const valueSet = new Set(values);
```
AmeliaIsCoding: what is valueSet? can we be a bit more descriptive please?

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -117,6 +120,41 @@ const ParcelsPage: React.FC<{}> = () => {
...
+
+        const boundaries: number[] = [];
+        valueSet.forEach((value) => {
+            if (value === values[0]) {
+                return;
+            }
+
+            const indexOfFirstOccurrence = values.indexOf(value);
+            if (indexOfFirstOccurrence === -1) {
+                //
```
AmeliaIsCoding: what is the purpose of this comment? can we restructure this logic so that we only have an if, and not an if/else with an empty if?

src/app/parcels/parcelsTable/ParcelsPage.tsx
AmeliaIsCoding: instead of using a generic, we could just use a union type in the props, such as `values: number[] | string[]` (include nulls if you need)
We don't refer to `DataType` in the function body, so we don't need a generic

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -117,6 +120,41 @@ const ParcelsPage: React.FC<{}> = () => {
...
+            } else {
+                boundaries.push(indexOfFirstOccurrence);
+            }
+        });
+        return boundaries;
+    }
+
+    const searchForPackingSlotBreakPoints = (parcelsTableRows: ParcelsTableRow[]): number[] => {
+        const packingSlotValues = parcelsTableRows.map((row) => {
+            return row.packingSlot;
```
AmeliaIsCoding: you don't need the `{return ...}` here. you can just do `.map((row) => row.packingSlot)` and similarly elsewhere.

src/components/Tables/Table.tsx
```diff
@@ -149,13 +149,15 @@ interface Props<Data, DbData extends Record<string, any>, PaginationType> {
...
     >;
+    parcelRowBreakPoints?: number[]; //This is a list of indices where the line will be drawn when sorting by packing date/slot
     defaultShownHeaders?: readonly (keyof Data)[];
     toggleableHeaders?: readonly (keyof Data)[];
     columnDisplayFunctions?: ColumnDisplayFunctions<Data>;
     columnStyleOptions?: ColumnStyles<Data>;
     editableConfig: EditableConfig<Data>;
     onRowClick?: OnRowClickFunction<Data>;
     pointerOnHover?: boolean;
+    setSortedColumn?: (sortedColumn: string) => void;
```
AmeliaIsCoding: we should be able to remove this additional prop if we use the sortState to extract the sorted column. If for whatever reason we do need it, please move this prop into the sortConfig.

src/components/Tables/Table.tsx
```diff
@@ -149,13 +149,15 @@ interface Props<Data, DbData extends Record<string, any>, PaginationType> {
             ? ClientSideFilter<Data, any>
             : ServerSideFilter<Data, any, DbData>
     >;
+    parcelRowBreakPoints?: number[]; //This is a list of indices where the line will be drawn when sorting by packing date/slot
```
AmeliaIsCoding: 1. this is the Table component, so we shouldn't refer to parcels here, as the Table component is used for other tables, not just the parcels table.
2. We don't need this comment, the code should aim to be self-explanatory, and I think `rowBreakPoints` is a self-explanatory variable name

AmeliaIsCoding (changes requested):
Well done! some comments, mostly 
1. move the code that gets the boundaries outside of the tables file
2. remove the extra state for the column name, as this is already contained within the sortState state
3. remove the extra prop for setting the column name in the table component

I'll have another look at the getValueBoundaries code once it's in another file outside of ParcelsPage.tsx

Also, tiny thing, but we've decided to use the PR title naming convention of "#XXX: Description", ie with a colon :)

src/app/parcels/parcelsTable/ParcelsPage.tsx
KylePUG: was going to include a `throw` if the value cannot be found but that basically cannot happen so I removed it

KylePUG:
Made the changes above! Ready to be reviewed again :)

AmeliaIsCoding (commented):
this all looks really good! thank you for making the changes that I requested, i think you understand what I asked really well! 

Did you see the comment I made before my first review, about Dan's intentions with the packing slot/date dividing lines?

AmeliaIsCoding:
> Made the changes above! Ready to be reviewed again :)

Also btw it's possible to rerequest review using the button on the PR at the top right hand side! :)

KylePUG:
> Did you see the comment I made before my first review, about Dan's intentions with the packing slot/date dividing lines?

Yes. I was waiting for Ricky's response before, but I can just get on this - should not take very long

KylePUG:
This is how it looks now. I have also changed the getRowBoundaries algorithm as the previous one doesn't work for this additional case :)

image

src/app/parcels/parcelsTable/ParcelsPage.tsx
```diff
@@ -42,13 +42,15 @@ import {
...
 import { DbParcelRow } from "@/databaseUtils";
+import { searchForBreakPoints } from "./conditionalStyling";
 
 const ParcelsPage: React.FC<{}> = () => {
     const [isLoading, setIsLoading] = useState(true);
     const [parcelsDataPortion, setParcelsDataPortion] = useState<ParcelsTableRow[]>([]);
     const [filteredParcelCount, setFilteredParcelCount] = useState<number>(0);
     const [selectedParcelId, setSelectedParcelId] = useState<string | null>(null);
     const [clientIdForSelectedParcel, setClientIdForSelectedParcel] = useState<string | null>(null);
+    const [parcelRowBreakPoints, setParcelRowBreakPoints] = useState<number[][]>([[], []]);
```
AmeliaIsCoding: I think the [][] type might not be the clearest here. Eg., by looking at this file, I don't really know whether the outer array corresponds to the packing slots, or the inner array. Also, this isn't very sustainable if we wanted to add more dividing lines. Can you think of a better data structure to use here? we want it to 1. be clear which array of numbers corresponds to which column that we're using to find the breakpoints and 2. be sustainable if we wanted to add more dividing lines and 3. cope with the case where there are dividing lines between the same 2 rows according to more than one measure (ie between 2 dates the packing slot also changes, we need to cope with that)

src/components/Tables/Table.tsx
```diff
@@ -618,6 +620,28 @@ const TableStyling = styled.div`
...
+                border-top: solid #77bda0;
+            }`
+            )
+            .join()}
+
+    ${(props) =>
+        props.rowBreakPoints
+            ?.slice()[0]
+            .map(
+                (darkGreenBreakPoint) =>
```
AmeliaIsCoding: We also want this code to be able to deal with an arbitrary number of columns that we're using to find the breakpoints by, and we don't want the component to be the one to deal with the colours/significance of each such column, but the ParcelsTable file, so that we can use the breakpoints feature in other tables as flexibly as possible in the future if we want to. So maybe the data structure that I was talking about should also store the colour of the dividing line.

src/components/Tables/Table.tsx
```diff
@@ -618,6 +620,28 @@ const TableStyling = styled.div`
...
         background-color: ${(props) => props.theme.main.background[0]};
     }
+
+    ${(props) =>
+        props.rowBreakPoints
+            ?.slice()[1]
+            .map(
+                (lightGreenBreakPoint) =>
+                    `& .rdt_TableRow:nth-child(${lightGreenBreakPoint + 1}) {
+                border-top: solid #77bda0;
```
AmeliaIsCoding: Can we try and use a colour thats stored within `props.theme`?

AmeliaIsCoding (changes requested):
Looks good, but requested a different data structure for the breakpoints in ParcelsPage - details in comments

src/app/parcels/parcelsTable/conditionalStyling.ts
```diff
@@ -24,19 +25,44 @@ const searchForPackingDateBreakPoints = (parcelsTableRows: ParcelsTableRow[]): n
...
+): BreakPointConfig[] => {
+    switch (headerKey) {
+        case "packingSlot": {
+            const packingSlotBreakPointConfig = {
+                name: "packingSlot",
+                breakPoints: searchForPackingSlotBreakPoints(parcelsTableRows),
+                colour: theme.primary.background[3],
+                thickness: "5pt",
+            };
+            return [packingSlotBreakPointConfig];
```
AmeliaIsCoding: why not just return [{....}] instead of assigning the constant? i think it's slightly confusing because you have a constant of the same name below but slightly different properties.

src/app/parcels/parcelsTable/conditionalStyling.ts
```diff
@@ -24,19 +25,44 @@ const searchForPackingDateBreakPoints = (parcelsTableRows: ParcelsTableRow[]): n
...
-                searchForPackingSlotBreakPoints(parcelsTableRows),
-            ];
+export type BreakPointConfig = {
+    name: string;
+    breakPoints: number[];
+    colour: string;
+    thickness: string;
+};
+
+export const searchForBreakPoints = <TableRow extends ParcelsTableRow>(
```
AmeliaIsCoding: can you explain the use of this type please? why TableRow extends ParcelsTableRow, why not just ParcelsTableRow?

AmeliaIsCoding (changes requested):
Looks really good well done! just a couple of small things

src/app/parcels/parcelsTable/conditionalStyling.ts
KylePUG: could just leave it as ParcelsTableRow tbf. i was thinking that if `searchForBreakPoints` will be used in the future for other tables it might be more easily generalisable? let me know if you want me to just leave it as `ParcelsTableRow` :)

src/app/parcels/parcelsTable/conditionalStyling.ts
```diff
@@ -0,0 +1,68 @@
...
+    const packingSlotValues = parcelsTableRows.map((row) => row.packingSlot);
+    return getRowBoundaries(packingSlotValues);
+};
+
+const searchForPackingDateBreakPoints = (parcelsTableRows: ParcelsTableRow[]): number[] => {
+    const packingSlotValues = parcelsTableRows.map((row) => row.packingDate?.getDate() ?? null);
+    return getRowBoundaries(packingSlotValues);
+};
+
+export type BreakPointConfig = {
```
AmeliaIsCoding: this type is relevant to the table component and would work for any table, so let's put this in the table component rather than the parcelsTable folder

src/app/parcels/parcelsTable/conditionalStyling.ts
AmeliaIsCoding: what this generic type currently does is it means that when we call this function, the type of the table row has to extend the type of ParcelsTableRow, this won't help us be more generalisable

also, searchForPackingDateBreakPoints takes in a type "ParcelsTableRow", and we refer to column headers from that type in the switch case.
so your function is only going to work for data of that type

I think there are 2 options: make your function work with ParcelsTableRow only, or make your function work with a generic TableRow type. your function currently only works with ParcelsTableRow, and i don't think it's worth the effort to get it to work with a generic type for now

I would just remove the generic and use ParcelsTableRow

src/app/parcels/parcelsTable/conditionalStyling.ts
AmeliaIsCoding: you resolved this but I can't see the change, maybe you forgot to push?

src/app/parcels/parcelsTable/conditionalStyling.ts
```diff
@@ -0,0 +1,68 @@
+import { DefaultTheme } from "styled-components/dist/types";
+import { ParcelsTableRow } from "./types";
+
+const getRowBoundaries = (columnValues: (string | number | null)[]): number[] => {
+    let temp = columnValues[0];
```
AmeliaIsCoding: what is `temp`? can we use a more descriptive name?

src/app/parcels/parcelsTable/conditionalStyling.ts
```diff
@@ -0,0 +1,68 @@
...
+    const boundaries: number[] = [];
+
+    columnValues.forEach((columnValue, index) => {
+        if (columnValue !== temp) {
```
AmeliaIsCoding: do we even need temp? can we use 

```typescript
columnValues.forEach((columnValue, index) => {
    const previousIndex = index - 1 >= 0  ? index -1 : 0;
    const previousValue = columnValues[previousIndex];
    if (columnValue !== previousValue) {
        ...
    }
etc
``` 

This avoids the use of `let` which I think is discouraged to use where possible

AmeliaIsCoding (changes requested):

KylePUG:
Solved the desync bug! Ready for review :)

src/app/parcels/parcelsTable/conditionalStyling.ts
```diff
@@ -0,0 +1,58 @@
+import { ParcelsTableRow } from "./types";
+import { BreakPointConfig } from "@/components/Tables/Table";
+
+const getRowBoundaries = (columnValues: (string | number | null)[]): number[] => {
+    const boundaries: number[] = [];
+
+    columnValues.forEach((columnValue, index) => {
+        const previousIndex = index - 1 >= 0 ? index - 1 : 0;
+        const previousValue = columnValues[previousIndex];
+        if (columnValue !== previousValue) {
```
8kristy:
```suggestion
        if (columnValue !== columnValues[previousIndex]) {
```

src/components/Tables/Table.tsx
```diff
@@ -401,6 +421,23 @@ const Table = <
...
+        },
+        dateAndSlotSecondary: {
+            colour: theme.primary.background[2],
+            thickness: "2.5pt",
+        },
+        slotPrimary: {
+            colour: theme.primary.background[3],
+            thickness: "5pt",
+        },
+    };
```
8kristy: Should we move this to `conditionalStyling.ts` and use the objects directly rather than the key?

src/components/Tables/Table.tsx
```diff
@@ -132,6 +133,22 @@ export type EditableConfig<Data> =
...
+export type DividingLineStyleOptions = {
+    dateAndSlotPrimary: DividingLineStyle;
+    dateAndSlotSecondary: DividingLineStyle;
+    slotPrimary: DividingLineStyle;
+};
+
+type DividingLineStyle = {
+    colour: string;
+    thickness: string;
+};
```
8kristy: Not sure if this should also go to `conditionalStyling.ts` - see what works and looks better

src/components/Tables/Table.tsx
```diff
@@ -618,6 +661,19 @@ const TableStyling = styled.div`
...
+            .map((breakPointConfig) => {
+                return breakPointConfig.breakPoints
+                    .map(
+                        (breakPoint) => `& .rdt_TableRow:nth-child(${breakPoint + 1}) {
+                            border-top: ${props.dividingLineStyleOptions[breakPointConfig.dividingLineStyle].thickness} solid ${props.dividingLineStyleOptions[breakPointConfig.dividingLineStyle].colour};
+                        }`
+                    )
+                    .join();
+            })
+            .join()}
```
8kristy, line 654: This is quite a lengthy bit, might be worth adding a comment about what this does and how

8kristy (commented):
Couple minor bits

src/app/parcels/parcelsTable/conditionalStyling.ts
KylePUG: Done

src/components/Tables/Table.tsx
KylePUG: I agree that this bit belongs in `conditionalStyling.ts`. I can't actually include this within the `searchForBreakPoints` function (in `conditionalStyling.ts`) as this will lead to the same problem with the table and dividing lines desyncing. I have now kept most of the code similar but moved this section dealing with styling into `conditionalStyling.ts` as a function that can be called. Let me know if this is okay :)

src/components/Tables/Table.tsx
KylePUG: done

src/components/Tables/Table.tsx
KylePUG: done

src/components/Tables/Table.tsx
8kristy: Ah that's annoying that it gets desynced but yeah this is ok

8kristy (approved):

AmeliaIsCoding (approved):

PR #352, opened 2024-05-17
VFB-225: Remove fallback routing to improve login refresh issue.

Alex-Softwire:
## What's changed
- Unnecessary fallback rerouting has been removed. This should improve the login refresh error.
- If post_checkout is run with the user still signed on, instead of a refresh loop, now the login button initially doesn't work (though works when you refresh the page)
- Although not a full fix, this is an improvement so is worth adding. 

## Screenshots / Videos
no ui changes

## Checklist

If you have made any changes to the database... (none made :) )

### DIFF
--- a/docs/common-problems.md
+++ b/docs/common-problems.md
@@ -10,6 +10,9 @@
 ### Fails to create users when resetting database
 - Try again, sometimes it can take a few times to work
 
+### Login button does nothing after running post_checkout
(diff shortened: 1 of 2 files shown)


AmeliaIsCoding (approved):
This looks good to me, although I don't know much about the routing stuff. Well done on this!

8kristy (approved):
Does look like something that needs a different solution, but hopefully this is something that will only happen locally and refreshing the page sounds less annoying than it restarting every 3s

PR #354, opened 2024-05-17
VFB-229: Fix bug to allow adding parcels for collection

AmeliaIsCoding:
## What's changed
Fixed a bug that preventing adding parcels for collection.

## Screenshots / Videos
| Before     |
|------------|
|image|

## Checklist

### DIFF
--- a/src/app/parcels/form/ParcelForm.tsx
+++ b/src/app/parcels/form/ParcelForm.tsx
@@ -187,7 +187,15 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
         setSubmitDisabled(true);
         let inputError;
         if (fields.shippingMethod === "Collection") {
-            inputError = checkErrorOnSubmit(formErrors, setFormErrors);
+            inputError = checkErrorOnSubmit(formErrors, setFormErrors, [
+                "voucherNumber",
+                "packingDate",
+                "packingSlot",
+                "shippingMethod",
+                "collectionDate",
+                "collectionTime",
+                "collectionCentre",
+            ]);
         } else {
             inputError = checkErrorOnSubmit(formErrors, setFormErrors, [
                 "voucherNumber",
--- a/src/app/parcels/form/ParcelForm.tsx
+++ b/src/app/parcels/form/ParcelForm.tsx
@@ -187,15 +187,7 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
         setSubmitDisabled(true);
         let inputError;
         if (fields.shippingMethod === "Collection") {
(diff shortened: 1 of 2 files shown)


src/app/parcels/form/ParcelForm.tsx
```diff
@@ -187,7 +187,15 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
         setSubmitDisabled(true);
         let inputError;
         if (fields.shippingMethod === "Collection") {
-            inputError = checkErrorOnSubmit(formErrors, setFormErrors);
+            inputError = checkErrorOnSubmit(formErrors, setFormErrors, [
```
Alex-Softwire: Am a little confused as to why this will make a difference? 
Lines 215 and 216 of formFunctions.ts
` for (const [errorKey, error] of Object.entries(errorType)) {
        if (!keysToCheck || keysToCheck.includes(errorKey)) {`
if keysToCheck is empty, then all the keys (in errorType) will be checked. Doesn't errorType contain the same set of 7 keys that are now in the added array. So in both cases all 7 keys will be checked?

Alex-Softwire (commented):
Nice quick work into solving this issue, and raising a pr so fast! Well done :)
Am just a tad confused as to what caused the underlying problem given it was sometimes working?

src/app/parcels/form/ParcelForm.tsx
AmeliaIsCoding: yeah good point. the error was occurring because another key was accidentally added to errorType. I've fixed this, and removed the above change.

src/components/Form/formFunctions.ts
```diff
@@ -209,6 +209,7 @@ export const checkErrorOnSubmit = <
     errorSetter: (errors: SpecificErrors) => void,
     keysToCheck?: string[]
 ): boolean => {
+    console.log(errorType);
```
Alex-Softwire: Can this be removed please :)

Alex-Softwire (approved):
This makes so much sense!!!

So.. if you entered an incorrect date at any point, or typed the date rather than clicked on the calendar menu, it created a new error called "key". As the collection mode checks all errors rather than an array (subset) of errors, it checks this new "key" error. This new "key" error cannot be override if a correct date is entered.

This also explains why it was inconsistent (different people entering dates differently), and that your original fix would have worked (but didn't find the underlying cause).

src/components/Form/formFunctions.ts
AmeliaIsCoding: ty for catching!

8kristy (approved):

PR #355, opened 2024-05-20
#236: Display action list including delete parcel if user is admin

minalui:
## What's changed
- Check whether user role and only display delete parcels if user is admin, manager or staff

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

No changes to database

### DIFF
--- a/src/app/parcels/ActionBar/Actions.tsx
+++ b/src/app/parcels/ActionBar/Actions.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import React, { useContext, useState } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import Menu from "@mui/material/Menu/Menu";
 import MenuList from "@mui/material/MenuList/MenuList";
 import MenuItem from "@mui/material/MenuItem/MenuItem";
@@ -30,50 +30,36 @@ export type ActionName =
     | "Download Driver Overview"
     | "Delete Parcel";
 
-type AvailableActionsType = {
-    [actionKey in ActionName]: {
-        newStatus: StatusType;
-    };
+type ActionTypes = {
+    actionName: ActionName;
+    newStatus: StatusType;
+    availableToVolunteers: boolean;
 };
 
-export const allAvailableActions: AvailableActionsType = {
-    "Download Day Overview": {
+const availableActions: ActionTypes[] = [
+    {
+        actionName: "Download Day Overview",
         newStatus: "Day Overview Downloaded",
+        availableToVolunteers: true,
     },
-    "Download Shopping Lists": {
+    {
+        actionName: "Download Shopping Lists",
         newStatus: "Shopping List Downloaded",
+        availableToVolunteers: true,
     },
-    "Download Shipping Labels": {
+    {
+        actionName: "Download Shipping Labels",
         newStatus: "Shipping Labels Downloaded",
+        availableToVolunteers: true,
     },
-    "Generate Map": {
-        newStatus: "Map Generated",
-    },
-    "Download Driver Overview": {
+    { actionName: "Generate Map", newStatus: "Map Generated", availableToVolunteers: true },
+    {
+        actionName: "Download Driver Overview",
         newStatus: "Driver Overview Downloaded",
+        availableToVolunteers: true,
     },
-    "Delete Parcel": {
-        newStatus: "Parcel Deleted",
-    },
-};
-
-export const volunteerAvailableActions: Omit<AvailableActionsType, "Delete Parcel"> = {
-    "Download Day Overview": {
-        newStatus: "Day Overview Downloaded",
-    },
-    "Download Shopping Lists": {
-        newStatus: "Shopping List Downloaded",
-    },
-    "Download Shipping Labels": {
-        newStatus: "Shipping Labels Downloaded",
-    },
-    "Generate Map": {
-        newStatus: "Map Generated",
-    },
-    "Download Driver Overview": {
-        newStatus: "Driver Overview Downloaded",
-    },
-};
+    { actionName: "Delete Parcel", newStatus: "Parcel Deleted", availableToVolunteers: false },
+];
 
 interface Props {
     fetchSelectedParcels: () => Promise<ParcelsTableRow[]>;
@@ -112,7 +98,19 @@ const Actions: React.FC<Props> = ({
 }) => {
     const [selectedParcels, setSelectedParcels] = useState<ParcelsTableRow[]>([]);
     const [modalToDisplay, setModalToDisplay] = useState<ActionName | null>(null);
+    const [availableActionsForUserRole, setAvailableActionsForUserRole] = useState<ActionTypes[]>(
+        []
+    );
     const { role } = useContext(RoleUpdateContext);
+
+    useEffect(() => {
+        if (role === "volunteer") {
+            const actionList = availableActions.filter((action) => action.availableToVolunteers);
+            setAvailableActionsForUserRole(actionList);
+        } else {
+            setAvailableActionsForUserRole(availableActions);
+        }
+    }, [role]);
 
     const onModalClose = (): void => {
         setModalToDisplay(null);
@@ -146,12 +144,10 @@ const Actions: React.FC<Props> = ({
 
     return (
         <>
-            {Object.entries(
-                role !== "volunteer" ? allAvailableActions : volunteerAvailableActions
-            ).map(([key, value]) => {
+            {availableActionsForUserRole.map(({ actionName, newStatus }) => {
                 return (
-                    modalToDisplay === key &&
-                    getActionModal(key, {
+                    modalToDisplay === actionName &&
+                    getActionModal(actionName, {
                         isOpen: true,
                         onClose: onModalClose,
                         selectedParcels: selectedParcels,
@@ -159,7 +155,7 @@ const Actions: React.FC<Props> = ({
                         headerId: "action-modal-header",
                         actionName: modalToDisplay,
                         updateParcelStatuses: updateParcelStatuses,
-                        newStatus: value.newStatus,
+                        newStatus: newStatus,
                     })
                 );
             })}
@@ -170,19 +166,17 @@ const Actions: React.FC<Props> = ({
                     anchorEl={actionAnchorElement}
                 >
                     <MenuList id="action-menu">
-                        {Object.entries(
-                            role !== "volunteer" ? allAvailableActions : volunteerAvailableActions
-                        ).map(([key]) => {
+                        {availableActionsForUserRole.map(({ actionName }) => {
                             return (
                                 <MenuItem
-                                    key={key}
+                                    key={actionName}
                                     onClick={onMenuItemClick(
-                                        key as ActionName,
+                                        actionName as ActionName,
                                         isNotAtLeastOne,
                                         errorMessage
                                     )}
                                 >
-                                    {key}
+                                    {actionName}
                                 </MenuItem>
                             );
                         })}
--- a/src/app/parcels/ActionBar/Actions.tsx
+++ b/src/app/parcels/ActionBar/Actions.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import Menu from "@mui/material/Menu/Menu";
 import MenuList from "@mui/material/MenuList/MenuList";
 import MenuItem from "@mui/material/MenuItem/MenuItem";
@@ -14,6 +14,7 @@ import GenerateMapModal from "./ActionModals/GenerateMapModal";
 import ShippingLabelModal from "./ActionModals/ShippingLabelModal";
 import ShoppingListModal from "./ActionModals/ShoppingListModal";
 import { UpdateParcelStatuses } from "./ActionAndStatusBar";
+import { checkUserIsAdmin } from "@/app/parcels/ActionBar/ActionModals/adminActions";
 
 const isNotAtLeastOne = (value: number): boolean => {
     return value < 1;
@@ -35,7 +36,7 @@ type AvailableActionsType = {
     };
 };
 
-export const availableActions: AvailableActionsType = {
+export const adminAvailableActions: AvailableActionsType = {
     "Download Day Overview": {
         newStatus: "Day Overview Downloaded",
     },
@@ -53,6 +54,24 @@ export const availableActions: AvailableActionsType = {
     },
     "Delete Parcel": {
         newStatus: "Parcel Deleted",
+    },
+};
+
+export const nonAdminAvailableActions: Omit<AvailableActionsType, "Delete Parcel"> = {
+    "Download Day Overview": {
+        newStatus: "Day Overview Downloaded",
+    },
+    "Download Shopping Lists": {
+        newStatus: "Shopping List Downloaded",
+    },
+    "Download Shipping Labels": {
+        newStatus: "Shipping Labels Downloaded",
+    },
+    "Generate Map": {
+        newStatus: "Map Generated",
+    },
+    "Download Driver Overview": {
+        newStatus: "Driver Overview Downloaded",
     },
 };
 
@@ -93,6 +112,20 @@ const Actions: React.FC<Props> = ({
 }) => {
     const [selectedParcels, setSelectedParcels] = useState<ParcelsTableRow[]>([]);
     const [modalToDisplay, setModalToDisplay] = useState<ActionName | null>(null);
+    const [userIsAdmin, setUserIsAdmin] = useState<boolean>(false);
+
+    useEffect(() => {
+        const isAdmin = async (): Promise<void> => {
+            const { isSuccess, failureReason } = await checkUserIsAdmin();
+            if (failureReason) {
+                setModalError(`Error authenticating as admin: ${failureReason}`);
+            } else {
+                setUserIsAdmin(isSuccess);
+            }
+        };
+
+        void isAdmin();
+    }, []);
 
     const onModalClose = (): void => {
         setModalToDisplay(null);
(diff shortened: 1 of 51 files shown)


src/server/authenticateAdminUser.ts
```diff
@@ -1,7 +1,7 @@
 import { getSupabaseServerComponentClient } from "@/supabaseServer";
 import { fetchUserRole } from "@/common/fetchUserRole";
 
-type Authenticated =
+export type Authenticated =
```
Alex-Softwire: Why have these types been exported? I can't see their usage in the above file?

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -56,6 +57,24 @@ export const availableActions: AvailableActionsType = {
     },
 };
 
+export const volunteerAvailableActions: Omit<AvailableActionsType, "Delete Parcel"> = {
```
Alex-Softwire: Ooh didn't know the omit type was a thing in typescript -- nice :)

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -56,6 +57,24 @@ export const availableActions: AvailableActionsType = {
...
+    "Download Day Overview": {
+        newStatus: "Day Overview Downloaded",
```
Alex-Softwire: Is there a way we can avoid repetition here -> am just a little worried that if new actions are added, and new roles have access to different actions, then we could end having issues. e.g copy paste errors meaning that newStatus is not synced between different roles. 
Lmk know if you disagree with this / couldn't see a way to do this.

src/app/parcels/ActionBar/Actions.tsx
Alex-Softwire: Why is this exported? Where else is it used in?

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -35,7 +36,7 @@ type AvailableActionsType = {
     };
 };
 
-export const availableActions: AvailableActionsType = {
+export const allAvailableActions: AvailableActionsType = {
```
Alex-Softwire: I know you didn't set this up :). Why were the actions set to be a key-value pair, in their only use Lines 149 and 173 we do Object.Entries which turns the actions from a key-value pair to an array. Would it make more sense to be an array in the first place, so we don't need the transition? 
e.g. [{actionName: "Download Day Overview", newStatus: "Day Overview Downloaded"}, ...] 
Feel free to ignore this if you don't think it's neater - or there is a good reason for it to be done like this :)

Alex-Softwire (commented):
Nice work with this :), looks like this will stop volunteers deleting parcels. Just a couple suggestions to avoid repetition, especially if we add more actions and more roles that limit these actions. Let me know if you disagree with any of the suggestions :)

src/server/authenticateAdminUser.ts
minalui: Sorry! I used the authenticate as admin function before I used the role context and forgot to change them back!

Alex-Softwire: No worries haha, happy to spot something :)

src/app/parcels/ActionBar/Actions.tsx
minalui: Removed :)

src/app/parcels/ActionBar/Actions.tsx
minalui: I've added a new "availableToVolunteers" to the list of available actions and the list gets filtered in an useEffect. Please let me know if there is a better approach 👍

src/app/parcels/ActionBar/Actions.tsx
minalui: Good point 👍 I've changed it into an array

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -29,32 +30,36 @@ export type ActionName =
...
-    };
+type ActionTypes = {
+    actionName: ActionName;
+    newStatus: StatusType;
+    availableToVolunteers: boolean;
 };
 
-export const availableActions: AvailableActionsType = {
-    "Download Day Overview": {
+const availableActions: ActionTypes[] = [
```
Alex-Softwire, line 40: Nice well done, I like this a lot! This (the refactor) makes it much easier to add a new action, see which actions are restricted to volunteers, and we're not duplicating newStatus. Also we don't need to do Object.entries, and [key] unnecessarily which makes it more readable :)

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -29,32 +30,36 @@ export type ActionName =
...
+    {
+        actionName: "Download Shipping Labels",
         newStatus: "Shipping Labels Downloaded",
+        availableToVolunteers: true,
     },
-    "Generate Map": {
-        newStatus: "Map Generated",
-    },
-    "Download Driver Overview": {
+    { actionName: "Generate Map", newStatus: "Map Generated", availableToVolunteers: true },
```
Alex-Softwire: Is Eslint forcing it on one line? Would the linter complain if you put that object on three lines like the other ones just so it looks nicer and consistent. Annoying if it does.

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -148,17 +166,17 @@ const Actions: React.FC<Props> = ({
...
                     <MenuList id="action-menu">
-                        {Object.entries(availableActions).map(([key]) => {
+                        {availableActionsForUserRole.map(({ actionName }) => {
                             return (
                                 <MenuItem
-                                    key={key}
+                                    key={actionName}
                                     onClick={onMenuItemClick(
-                                        key as ActionName,
+                                        actionName as ActionName,
```
Alex-Softwire: Why do we need the type assertion "as ActionName", shouldn't typescript know it is an ActionName?

Edit - was it a relic of the old key value map -- another reason for this refactor :)

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -126,18 +144,18 @@ const Actions: React.FC<Props> = ({
 
     return (
         <>
-            {Object.entries(availableActions).map(([key, value]) => {
+            {availableActionsForUserRole.map(({ actionName, newStatus }) => {
```
Alex-Softwire, line 155: I like that you split this object up here, makes it look much neater :)

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -93,6 +98,19 @@ const Actions: React.FC<Props> = ({
 }) => {
     const [selectedParcels, setSelectedParcels] = useState<ParcelsTableRow[]>([]);
     const [modalToDisplay, setModalToDisplay] = useState<ActionName | null>(null);
+    const [availableActionsForUserRole, setAvailableActionsForUserRole] = useState<ActionTypes[]>(
```
Alex-Softwire: Do we need this to be a useState / useEffect? Is the role likely to change while the user has the page still open - if the role changes it should cause a re-render anyway.

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -93,6 +98,19 @@ const Actions: React.FC<Props> = ({
...
+        []
+    );
+    const { role } = useContext(RoleUpdateContext);
+
+    useEffect(() => {
+        if (role === "volunteer") {
```
Alex-Softwire: Out of interest, why are we singling out volunteers. The ticket name is "Display action list including delete parcel if user is admin" which would imply that only admins can delete parcels. What about staff / managers or the other roles - what should they have access to?

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -29,32 +30,36 @@ export type ActionName =
...
 
-type AvailableActionsType = {
-    [actionKey in ActionName]: {
-        newStatus: StatusType;
-    };
+type ActionTypes = {
+    actionName: ActionName;
+    newStatus: StatusType;
+    availableToVolunteers: boolean;
 };
```
AmeliaIsCoding, line 38: if we wanted to have the availability of actions dependant on other roles, we'd have to create more properties in this object for each role. this isn't ideal. maybe we could have a switch case statement elsewhere (perhaps in roles?) that takes in a role and returns which actions are available to that role?

src/app/parcels/ActionBar/Actions.tsx
Alex-Softwire: To make this clearer, we could change availableToVolunteers to rolesAvailableTo = ["admin","staff"...]. I would have thought that would simplify the code too? Lmk if that doesn't work.

Alex-Softwire (commented):
Very happy with this refactor :) Think it's a definite improvement to the codebase - just a couple more comments/suggestions which might simplify things.

AmeliaIsCoding (changes requested):
Nice, I like the change away from the key value pair list that we used to have. I suggested a different way to deal with finding the available actions for a role that I think might be a bit more sustainable. :)

src/app/parcels/ActionBar/Actions.tsx
Alex-Softwire: Would have thought one could do it without - just defining the actions in terms of the role.

src/app/parcels/ActionBar/Actions.tsx
Alex-Softwire: what about just having rolesAvailableTo = [...] as a property on this? Though I guess both our suggestions would work haha

AmeliaIsCoding: yeah I think that is also a better solution than currently, but I think it might get a bit messy if we add lots more roles, I could see the possibility of someone forgetting to copy and paste a new role into another action if ygm. If we were to use a switch case we could also have a nice default that allows all roles, which might be quite nice and neat. However, I don't think there's that much difference, and this also has the advantage of not having to make a separate function. So I don't mind that much.

src/app/parcels/ActionBar/Actions.tsx
minalui: Working now 👍

src/app/parcels/ActionBar/Actions.tsx
minalui: Yes that's what I initially thought too but Dan commented after I've written the QA that manager, staff and admin should be able to delete parcels too. https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/issues/236#issuecomment-2119949498

I think that is a good suggestion - let me try implementing

src/app/parcels/ActionBar/Actions.tsx
Alex-Softwire: Yeah I defo see what you mean and that's definitely an advantage of doing it that way. Though you still have the issue with forgetting to add the role onto the switch statement & it may be harder to couple the two together (filter by the results of the switch statement). That being said, there's pros and cons with both & you've convinced me that I'd be happy to see either (or a 3rd suggestion)!

src/app/parcels/ActionBar/Actions.tsx
Alex-Softwire: Ah that makes sense, apologies for not seeing this! Could you update the what's changed in the pr to match this :)

src/app/parcels/ActionBar/Actions.tsx
minalui: Yes makes sense thanks!

src/app/parcels/ActionBar/Actions.tsx
minalui: Thanks both - great points. I've kept it as a list of roles that have access to the action for now. Happy to make further changes as necessary.

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -29,32 +31,44 @@ export type ActionName =
...
+    availableToRole: UserRole[];
 };
 
-export const availableActions: AvailableActionsType = {
-    "Download Day Overview": {
+const availableActions: ActionTypes[] = [
+    {
+        actionName: "Download Day Overview",
         newStatus: "Day Overview Downloaded",
+        availableToRole: ["volunteer", "staff", "manager", "admin"],
```
Alex-Softwire: Is there an exported array of all roles anywhere? If so, could we use that for the ones that are available to all roles -just to make it easier if a new role is added? Happy to leave if there isn't.

Alex-Softwire (approved):
Nice work :) Thanks for these changes, think I'm pretty happy with this! Just the one question about using a predefined array of roles for the all role actions (just to avoid issues if a new role is added). Feel free to pass on once done though.

src/app/parcels/ActionBar/Actions.tsx
Alex-Softwire: or just add 
`const allRoles = ["volunteer", "staff"...]`, though if we have a more common list somewhere else that would be better.

8kristy: Not sure if we're talking about the same thing here but I agree that we should have some role groups, e.g.
```
everyone = ["volunteer", "staff", "management", "admin"]
org = ["staff", "management", "admin"]
adminOnly = ["admin"]
``` 
and put those in a file somewhere in case we suddenly have many roles and many groups

Wonder if we should make it configurable in the db somewhere (e.g. a new `roles` table) rather than hardcoding 🤔 both the lists and which list is assigned to each action, although that sounds like extra UI work so might be better to leave it and add if requested, only laying the foundations here

I think having groups like that would fix the problem of "we made a new role and forgot to add it to an action" - you'd have to forget the entire group

8kristy (commented):
Just a suggestion on role lists

Also looks like your end to end tests are failing?

Poke me when you want me to have another look

src/app/parcels/ActionBar/Actions.tsx
Alex-Softwire: Yep! That was what I was talking about, happy with adding a list to a separate file like the above. There is another point in the codebase where we go allUserRoles = [..] so commonising that would be nice. The db option would probably be better eventually but probably out of scope for this ticket?

minalui: Thanks both - I've put the lists in roles.tsx and replaced allUserRoles with the new list.

src/app/parcels/ActionBar/Actions.tsx
```diff
@@ -98,19 +107,16 @@ const Actions: React.FC<Props> = ({
...
-            setAvailableActionsForUserRole(availableActions);
+    const getAvailableActionsForUserRole = (): ActionTypes[] => {
+        if (role === null) {
+            return [];
         }
-    }, [role]);
+        return availableActions.filter((action) => {
+            return action.availableToRole.includes(role);
+        });
+    };
```
AmeliaIsCoding, line 119: i feel like this function doesn't need to be a function but a constant - then we won't have to call it twice?

src/app/roles.tsx
AmeliaIsCoding, line 1: Would it be a good idea to use these in the auth logic as well (what users can access what pages)?

AmeliaIsCoding (changes requested):
Looks nice, couple small things

8kristy (approved):
Looks good :) Can do Amelia's suggestion then merge, for the pages auth maybe raise a new ticket?

src/app/parcels/ActionBar/Actions.tsx
minalui: Good point - I've created a constant using this function because I need to exclude role = null because action.availableToRole does not include null (typing issue).

src/app/roles.tsx
minalui: Ticket created https://softwiretech.atlassian.net/browse/VFB-230

AmeliaIsCoding: personally I feel like this won't take long at all and might be a bit confusing to someone who picks up this ticket - so maybe worth doing in this ticket - the relevant code is in app/roles.tsx lines 25-46

AmeliaIsCoding (commented):
LGTM but i feel like maybe it's worth using the role groups in the pages auth already instead of putting it in another ticket, as it will be a small change. Not that fussed though, happy for you to not!

src/app/roles.tsx
minalui: Sorry didn't look into it as Kristina said to raise a ticket - I'll do this now

minalui: I've updated using the lists - not sure it's more readable but it's good to condense the staff and manager roles together. Let me know if there are other ways of improving.

src/app/roles.tsx
```diff
@@ -27,22 +27,20 @@ const pathsOnlyShownToStaffAndAbove = ["/lists"] as const;
...
-        case "staff":
-            return [...pathsShownToAllAuthenticatedUsers, ...pathsOnlyShownToStaffAndAbove];
-        case "volunteer":
-            return pathsShownToAllAuthenticatedUsers;
-        case null:
-            return pathsNotRequiringLogin;
+    if (adminRoles.includes(role as AdminRolesType)) {
+        return [
+            ...pathsShownToAllAuthenticatedUsers,
+            ...pathsOnlyShownToStaffAndAbove,
```
AmeliaIsCoding, line 33: I feel like we should rename this constants to `pathsOnlyShownToOrganisationAndAbove` so it's consistent with the new role group naming system

src/app/roles.tsx
```diff
@@ -27,22 +27,20 @@ const pathsOnlyShownToStaffAndAbove = ["/lists"] as const;
...
-            ];
-        case "manager":
-            return [...pathsShownToAllAuthenticatedUsers, ...pathsOnlyShownToStaffAndAbove];
-        case "staff":
-            return [...pathsShownToAllAuthenticatedUsers, ...pathsOnlyShownToStaffAndAbove];
-        case "volunteer":
-            return pathsShownToAllAuthenticatedUsers;
-        case null:
-            return pathsNotRequiringLogin;
+    if (adminRoles.includes(role as AdminRolesType)) {
```
AmeliaIsCoding, line 30: is there a way to get around this typecasting?

AmeliaIsCoding (changes requested):
looks good, i like the new `getShownPagesByRole`, just a couple of things

src/app/roles.tsx
minalui: It's not happy without - let me know if there are suggestions to get around it

AmeliaIsCoding: i'm not sure, if you can't thinking of anything maybe its ok to merge?

8kristy (approved):

AmeliaIsCoding (approved):

PR #357, opened 2024-05-24
#182: Resize parcels table column width

kaydentsoftwire:
## What's changed
Parcel table column widths are adjusted - full name and status columns proportionally larger, accounting for a smaller screen, and ensuring that the icon column shrinks on a smaller screen

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image image | image image |

## Checklist


src/app/parcels/parcelsTable/styles.ts
```diff
@@ -11,38 +11,42 @@ export const PreTableControls = styled.div`
 
 export const parcelTableColumnStyleOptions = {
     iconsColumn: {
-        width: "4rem",
+        maxWidth: "4rem",
+        minWidth: "0.7rem",
+        grow: 0.5,
     },
```
kaydentsoftwire, line 17: Allows icons column to shrink when viewed on a smaller screen

src/app/parcels/parcelsTable/styles.ts
```diff
@@ -11,38 +11,42 @@ export const PreTableControls = styled.div`
...
     fullName: {
-        grow: 1.3,
-        minWidth: "8rem",
+        grow: 1.2,
+        minWidth: "6.5rem",
     },
```
kaydentsoftwire, line 21: The other grow values for middle columns have shrunk so full name is proportionally larger in relation to the middle columns, but smaller in relation to the last status column to account for a smaller screen

src/app/parcels/parcelsTable/styles.ts
```diff
@@ -11,38 +11,42 @@ export const PreTableControls = styled.div`
...
-        minWidth: "5rem",
+        grow: 0.5,
+        minWidth: "4rem",
     },
     lastStatus: {
         grow: 3,
         maxWidth: "30rem",
+        minWidth: "10rem",
     },
 };
```
kaydentsoftwire, line 51: gave last status a min width to make it larger on a smaller screen, and adjusted the sizing of the other columns to be proportionally smaller

src/app/parcels/parcelsTable/styles.ts
```diff
@@ -11,38 +11,42 @@ export const PreTableControls = styled.div`
...
+        grow: 0.6,
         minWidth: "6rem",
     },
     packingDate: {
         hide: 800,
-        grow: 0.8,
-        minWidth: "6rem",
+        grow: 0.7,
+        minWidth: "7rem",
     },
```
kaydentsoftwire, line 40: min width here as increased as the last digit of the year in the date was wrapping

src/app/parcels/parcelsTable/styles.ts
```diff
@@ -11,38 +11,42 @@ export const PreTableControls = styled.div`
...
         minWidth: "6rem",
     },
     addressPostcode: {
         hide: 800,
-        grow: 0.8,
-        minWidth: "6rem",
+        grow: 0.6,
+        minWidth: "5rem",
+        maxWidth: "6rem",
     },
```
kaydentsoftwire, line 31: this is to keep the postcode size smaller to fit with the character length but also allow a little bit of stretch for wider screens

src/app/parcels/parcelsTable/styles.ts
8kristy: Is that just space around icons or icons themselves?

8kristy:
It is quite stretched out on the big screen but I don't think we should shrink the smaller columns even more - think it would look a bit silly if we squeezed everything on the left and then had a massive status column. One thing that bothers me is how Postcode and Method are very close together - is it possible to separate them, or somehow distribute the columns evenly?
image

Maybe also give postcode a little bit more room - the longer ones start wrapping around 1220px width
image

Otherwise looks really good 😊

src/app/parcels/parcelsTable/styles.ts
kaydentsoftwire: Just the space around

kaydentsoftwire:
Made the adjustments to the postcode column, so it now looks like this on a larger screen.
image

And this on a smaller screen (no wrapping on postcode):

image

8kristy (approved):

PR #358, opened 2024-05-28
VFB-212-FIX: Ability to anonymise client data

AmeliaIsCoding:
## What's changed
- Fix duplicate audit log when deleting client
- Fix parcels table filters for seeded deleted clients: when we search for a parcel with a deleted client, instead of looking for null values in the column being searched in the database, we look for parcels with a client with `is_active = false`
- Fix parcels table sorts for seeded deleted clients: when we sort by a client info field, parcels with deleted clients are put to the bottom on first click of that column header, and they're put to the top on second click of that column header. This is done based on the value of `is_active` rather than the value of the field
- Fix default and packing date sorts so that the sort order is by:
        Date
        Packing Shift (AM/PM/Shift1,2,3,4)
        Method (Delivery, then Collection)
        Client inactivity (active clients, then inactive clients)
        Postcode
        Collection centre

## Checklist


src/app/clients/clientsTable/ClientsPage.tsx
```diff
@@ -156,6 +156,7 @@ const ClientsPage: React.FC<{}> = () => {
             }
             router.push("/clients");
         }
+        return;
```
8kristy: What does this fix?

src/app/clients/clientsTable/filters.ts
```diff
@@ -4,11 +4,20 @@ import { ClientsFilter, ClientsFilterMethod } from "./types";
 import { fullNameSearch, phoneSearch, postcodeSearch } from "@/common/databaseFilters";
 import { DbClientRow } from "@/databaseUtils";
 
-const clientsFullNameSearch: ClientsFilterMethod = fullNameSearch<DbClientRow>("full_name");
+const clientsFullNameSearch: ClientsFilterMethod = fullNameSearch<DbClientRow>(
+    "full_name",
+    "is_active"
```
8kristy, line 9: I can see why this is needed, but just out of curiousity what happens if you don't include this in the filter?

src/app/clients/deleteClient.ts
```diff
@@ -24,12 +24,11 @@ const deleteClient = async (clientId: string): Promise<DeleteClientResponse> =>
         const logId = await logErrorReturnLogId(`error with delete: client ${clientId}`, {
             error: deleteClientError,
         });
-        void sendAuditLog({ ...baseAuditLogProps, wasSuccess: false, logId });
```
8kristy, line 27: Don't think I've seen `void` being used like this before - is this the thing that caused multiple logs? If so do you know why?

src/app/parcels/parcelsTable/filters.ts
```diff
@@ -21,18 +21,22 @@ import { parcelTableHeaderKeysAndLabels } from "./headers";
 import { DbParcelRow } from "@/databaseUtils";
 import { fullNameSearch, phoneSearch, postcodeSearch } from "@/common/databaseFilters";
 
-const parcelsFullNameSearch: ParcelsFilterMethod<string> =
-    fullNameSearch<DbParcelRow>("client_full_name");
+const parcelsFullNameSearch: ParcelsFilterMethod<string> = fullNameSearch<DbParcelRow>(
+    "client_full_name",
+    "client_is_active"
```
8kristy, line 26: Does this remove deleted parcels from the search? I think if the deleted parcel filter is on we might still want to show them?

src/common/databaseFilters.ts
```diff
@@ -4,46 +4,55 @@ import { DbClientRow, DbParcelRow } from "@/databaseUtils";
...
             return query;
         }
         if (parcelsPageDeletedClientDisplayName.toLowerCase().includes(state.toLowerCase())) {
-            return query.or(`${columnLabel}.ilike.%${state}%, ${columnLabel}.is.null`);
+            return query.or(
+                `${fullNameColumnLabel}.ilike.%${state}%, ${clientIsActiveColumnLabel}.eq.false`
+            );
         }
-        return query.ilike(`${columnLabel}`, `%${state}%`);
+        return query.ilike(`${fullNameColumnLabel}`, `%${state}%`);
```
8kristy, line 19: What does this do? Does the last bit need to do anything with `clientIsActiveColumnLabel`?

src/components/Modal/ConfirmDialog.tsx
```diff
@@ -31,7 +31,7 @@ const ConfirmDialog: React.FC<Props> = ({ isOpen, message, onCancel, onConfirm }
             <ModalInner>
                 <Text>{message}</Text>
                 <Button onClick={onConfirm} variant="contained" color="primary">
-                    <p>Confirm</p>
+                    Confirm
```
8kristy, line 34: Why do we need to remove this?

8kristy (commented):
Just a few questions 

Could you give me a summary of the filter changes? Think I'm a bit lost as to what we want to achieve

src/app/clients/clientsTable/ClientsPage.tsx
AmeliaIsCoding: nothing but I thought it would be better practice to include, can remove if you want

src/app/clients/clientsTable/filters.ts
AmeliaIsCoding: we need the is_active column name in order to apply filters to it

src/app/clients/deleteClient.ts
AmeliaIsCoding: Yes this is what caused the problem, but I'm not exactly sure why/how. I spent ages trying to figure it out :(

src/app/parcels/parcelsTable/filters.ts
AmeliaIsCoding: we do remove parcels with deleted clients from the search in the very specific case of when we search for "NFA" parcels, because in this case we want to show parcels with a null postcode and client_is_active set to true. (to avoid showing parcels with a null postcode and client_is_active set to false, as these parcels belong to deleted clients and are not NFA).

src/components/Modal/ConfirmDialog.tsx
AmeliaIsCoding: It's just more consistent with the other components, I can add it back in if you want, it makes no change

src/app/parcels/parcelsTable/filters.ts
AmeliaIsCoding: Also please note that a deleted parcel (aka a parcel with last status = "Parcel Deleted") is not the same as a parcel with a deleted client :)

AmeliaIsCoding:
Ok filter changes amount basically to: when we search for a parcel with a deleted client, either by
- searching for name "(Client Deleted)"
- searching for postcode "-"
- searching for family count "-"
- searching for phone number "-"
Instead of looking for null values in these columns in the database, we look for parcels with a client with `is_active = false`
Therefore seeded deleted clients, who have `is_active = false` but non-null values in these columns, are correctly displayed according to the filters

src/common/databaseFilters.ts
AmeliaIsCoding: I'm not sure which line you're referring to but line 19 is just a variable being renamed. lines 15-17 means when we search for a parcel with name ("Deleted Client") we look for parcels in the DB with clients with `is_active = false` rather than parcels with a null client name

src/app/clients/clientsTable/ClientsPage.tsx
8kristy: I personally find it a bit confusing at the end of the function - usually I'd put a return in an if statement if we don't want to execute the rest of the function so it being here makes me feel suspicious, so maybe remove it

src/app/parcels/parcelsTable/filters.ts
8kristy: I think it isn't now but we might change that as part of https://softwiretech.atlassian.net/browse/VFB-233

I think we haven't fully decided yet what we're doing with deleted clients' parcels so happy to exclude them from the search either way

src/app/parcels/parcelsTable/sortableColumns.ts
```diff
@@ -36,6 +44,7 @@ const parcelsSortableColumns: SortOptions<ParcelsTableRow, ParcelsSortMethod>[]
         key: "packingDate",
         sortMethod: (query, sortDirection) =>
             query
+                .order("client_is_active")
```
8kristy: Does this need to be `ascending: sortDirection !== "asc"`? From what I understand all these just push inactive clients to the end?

8kristy (commented):
Filtering makes sense now, thanks :) Just remove the return and then the bit about ascending

src/app/clients/clientsTable/ClientsPage.tsx
AmeliaIsCoding: Sure, will do!

src/app/parcels/parcelsTable/filters.ts
AmeliaIsCoding: okay cool :)

src/app/parcels/parcelsTable/sortableColumns.ts
AmeliaIsCoding: Thanks for catching this - I've fixed this so that, when sorting by this field or the default sorting, the sorting is now

Date (ascending or descending depending on how many times the header has been clicked)
Packing Shift (AM/PM/Shift1,2,3,4)
Method (Delivery, then Collection)
Client inactivity (active clients, then inactive clients)
Postcode
Collection centre

I think this is the expected / desired behaviour

src/app/parcels/parcelsTable/sortableColumns.ts
```diff
@@ -1,6 +1,15 @@
...
 import { ParcelsTableRow, ParcelsSortMethod } from "./types";
 
+export const defaultParcelsSort: ParcelsSortMethod = (query, sortDirection) =>
+    query
+        .order("packing_date", { ascending: sortDirection === "asc" })
+        .order("packing_slot_order")
+        .order("is_delivery", { ascending: false })
+        .order("client_is_active", { ascending: false })
+        .order("client_address_postcode")
+        .order("collection_centre_name");
```
AmeliaIsCoding, line 11: This slightly changes the default sort, but this is a correction - the default sort should always have been exactly the same as the packing date sort

AmeliaIsCoding (commented):
self-review

src/app/parcels/parcelsTable/sortableColumns.ts
8kristy: I think that's correct - and it covers accidentally revealing deleted postcodes, right? Re #206

8kristy (approved):
Maybe double check the default sorting conversation but I think it's all correct

src/app/parcels/parcelsTable/sortableColumns.ts
AmeliaIsCoding: yes, because we sort by the client inactivity first, then the sensitive info field (postcode), the postcode is never revealed

Issue #196, opened 2024-03-26
On Driver Overview PDF - Indicate which parcels are 'Collection' and which parcels are 'Deliveries'


dan-langridge:
### Describe the solution / feature you'd like

If one of our drivers is taking all the Collections parcels AND doing some deliveries, the Driver Overview PDF that is created for them does not show which parcels are collections and which are deliveries, so might home deliver a parcel that was supposed to be dropped off at a Collection Centre.

On Driver Overview PDF, can we Indicate which parcels are 'Collections' and which parcels are 'Deliveries'.

Group Collections and mark them with the name of the collection centre and the Feet Icon. 
Group Deliveries and mark them the Van Icon.

kaydentsoftwire:
QA steps:
Note that the postcode sorting should match javascript default sorting (with E8 coming before E24, for example), which reflects how the sorting works elsewhere
- go to parcels page and select some parcels for collection and some for delivery
- under actions, click 'Download Driver Overview'
- [x] a pdf with two tables should appear
- [x] Deliveries should have a van icon next to it and collections should have a footprints icon next to it
- [x] The collections table should have a column 'Collection Centre'
- [x] The deliveries table should have an 'Address' column
- [x] The deliveries should match what was labelled as delivery, and the collections should match what was labelled as collections
- [x] Deliveries should be ordered by postcode
- [x] Collections should be ordered by collection centre
- go to parcels page and select just deliveries then download driver overview
- [x] only the deliveries table should appear, with deliveries ordered by postcode 
- go to parcels page and select just collections then download driver overview
- [x] only the collections table should appear, with collections grouped/ordered by collection centre

Error testing - to be done on local.
- go to parcels table in the database
- look at the client in your top row and search for them in the parcel table
- select their parcel and click 'Download driver overview'
- when the modal is up, delete the collection centre for the parcel in the database (just the id in the parcel's row, not the whole collection centre)
- click Download PDF on the modal
- [x] you should see the error, "Failed to find a collection centre for one or more of the parcels. LogId:" + log ID

minalui:
QA passed! 🎉

Stuart-Knowles-SW:
Has been in UAT for > 3 months, presumed done

PR #359, opened 2024-05-28
#196: Show collections or deliveries on driver overview

kaydentsoftwire:
## What's changed
The driver overview PDF groups parcels by collection and delivery
It shows two separate tables with the name 'Collections' or 'Deliveries' with the icon
Collections show the collection centre name and Deliveries show the address
On the PDF, collections are ordered / grouped by collection centre and deliveries are ordered by postcode

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| Collections and deliveries together ordered by name image | Deliveries ordered by postcode image Collections ordered by collection centre image When collection or delivery only, the other table is not shown image |

## Checklist

### DIFF
--- a/src/pdf/DriverOverview/DriverOverviewPdf.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdf.tsx
@@ -131,7 +131,7 @@ const styles = StyleSheet.create({
         marginRight: "5px",
         marginBottom: "5px",
         alignSelf: "flex-start",
-    }
+    },
 });
 
 const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
@@ -179,9 +179,9 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                 <Text>Number of Parcels</Text>
             </View>
         </View>
-    )
+    );
 
-    const createRowData = (rowData: DriverOverviewTableData, index: number) => {
+    const createRowData = (rowData: DriverOverviewTableData, index: number): React.JSX.Element => {
         return (
             // eslint-disable-next-line react/no-array-index-key
             <View key={index} style={[styles.tableRow, styles.flexRow]} wrap={false}>
@@ -189,22 +189,25 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                     <Text>{rowData.name}</Text>
                 </View>
                 <View style={[styles.tableColumn, styles.addressColumnWidth]}>
-                    { rowData.collection_centre.isDelivery ? rowData.address.postcode || rowData.clientIsActive ? (
-                        <>
-                            <Text>{rowData.address.line1}</Text>
-                            <Text>{rowData.address.line2}</Text>
-                            <Text>{rowData.address.town}</Text>
-                            <Text>{rowData.address.county}</Text>
-                            <Text>{rowData.address.postcode}</Text>
-                        </>
+                    {rowData.collection_centre.isDelivery ? (
+                        rowData.address.postcode || rowData.clientIsActive ? (
+                            <>
+                                <Text>{rowData.address.line1}</Text>
+                                <Text>{rowData.address.line2}</Text>
+                                <Text>{rowData.address.town}</Text>
+                                <Text>{rowData.address.county}</Text>
+                                <Text>{rowData.address.postcode}</Text>
+                            </>
+                        ) : (
+                            <Text>
+                                {rowData.clientIsActive ? displayPostcodeForHomelessClient : "-"}
+                            </Text>
+                        )
                     ) : (
-                        <Text>
-                            {rowData.clientIsActive ? displayPostcodeForHomelessClient : "-"}
-                        </Text>
-                    ):  <>
+                        <>
                             <Text>{rowData.collection_centre.name}</Text>
                         </>
-                    }
+                    )}
                 </View>
                 <View style={[styles.tableColumn, styles.contactColumnWidth]}>
                     <Text>{rowData.contact}</Text>
@@ -220,10 +223,10 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                 </View>
             </View>
         );
-    }
+    };
 
-    const collections = data.tableData[0].map(createRowData)
-    const deliveries = data.tableData[1].map(createRowData)
+    const collections = data.tableData[0].map(createRowData);
+    const deliveries = data.tableData[1].map(createRowData);
 
     return (
         <Document>
@@ -247,12 +250,14 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                 <View style={[styles.h2text, styles.collectionOrDeliveryHeader, styles.flexRow]}>
                     <Text style={styles.collectionOrDeliveryHeader}>Deliveries</Text>
                     <FontAwesomeIconPdfComponent faIcon={faTruck}></FontAwesomeIconPdfComponent>
-                    </View>
+                </View>
                 <View style={[styles.flexColumn, { width: "100%" }]}>{deliveryHeader}</View>
                 <View style={[styles.tableSection, styles.flexColumn]}>{deliveries}</View>
                 <View style={[styles.h2text, styles.collectionOrDeliveryHeader, styles.flexRow]}>
                     <Text style={[styles.collectionOrDeliveryHeader]}>Collections</Text>
-                    <FontAwesomeIconPdfComponent faIcon={faShoePrints}></FontAwesomeIconPdfComponent>
+                    <FontAwesomeIconPdfComponent
+                        faIcon={faShoePrints}
+                    ></FontAwesomeIconPdfComponent>
                 </View>
                 <View style={[styles.flexColumn, { width: "100%" }]}>{collectionHeader}</View>
                 <View style={[styles.tableSection, styles.flexColumn]}>{collections}</View>
--- a/src/pdf/DriverOverview/DriverOverviewPdf.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdf.tsx
@@ -146,14 +146,14 @@ const styles = StyleSheet.create({
 });
 
 const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
-    const createHeader = (category: string): React.JSX.Element => {
+    const createHeader = (category: Method): React.JSX.Element => {
         return (
             <View style={[styles.tableHeader, styles.flexRow]}>
                 <View style={[styles.tableColumn, styles.nameColumnWidth]}>
                     <Text>Name</Text>
                 </View>
                 <View style={[styles.tableColumn, styles.addressColumnWidth]}>
-                    <Text>{category === "Collection" ? "Collection Centre" : "Address"}</Text>
+                    <Text>{category === Method.Collection ? "Collection Centre" : "Address"}</Text>
                 </View>
                 <View style={[styles.tableColumn, styles.contactColumnWidth]}>
                     <Text>Contact</Text>
--- a/src/pdf/DriverOverview/DriverOverviewPdf.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdf.tsx
@@ -40,7 +40,7 @@ type CreateHeaderProps = "Delivery" | "Collection";
 export type DriverOverviewTablesData = {
     collections: DriverOverviewRowData[];
     deliveries: DriverOverviewRowData[];
-}
+};
 
 const styles = StyleSheet.create({
     container: {
@@ -236,7 +236,7 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                         COMPLETION OF DELIVERIES
                     </Text>
                 </View>
-                {deliveries.length ?
+                {deliveries.length ? (
                     <View style={styles.tableContainer}>
                         <View
                             style={[
@@ -255,10 +255,8 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                         </View>
                         <View style={[styles.tableSection, styles.flexColumn]}>{deliveries}</View>
                     </View>
-                    :
-                    null
-                }
-                {collections.length ?
+                ) : null}
+                {collections.length ? (
                     <View style={styles.tableContainer}>
                         <View
                             style={[
@@ -277,9 +275,7 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                         </View>
                         <View style={[styles.tableSection, styles.flexColumn]}>{collections}</View>
                     </View>
-                    :
-                    null
-                } 
+                ) : null}
             </Page>
         </Document>
     );
--- a/src/pdf/DriverOverview/DriverOverviewPdf.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdf.tsx
@@ -18,7 +18,10 @@ export interface DriverOverviewTableData {
     instructions?: string;
     clientIsActive: boolean;
     numberOfLabels: number;
-    isDelivery: boolean;
+    collection_centre: {
+        name: string;
+        isDelivery: boolean;
+    };
 }
 
 export interface DriverOverviewCardDataProps {
@@ -125,7 +128,7 @@ const styles = StyleSheet.create({
 });
 
 const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
-    const header = (
+    const deliveryHeader = (
         <View style={[styles.tableHeader, styles.flexRow]}>
             <View style={[styles.tableColumn, styles.nameColumnWidth]}>
                 <Text>Name</Text>
@@ -148,6 +151,29 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
         </View>
     );
 
+    const collectionHeader = (
+        <View style={[styles.tableHeader, styles.flexRow]}>
+            <View style={[styles.tableColumn, styles.nameColumnWidth]}>
+                <Text>Name</Text>
+            </View>
+            <View style={[styles.tableColumn, styles.addressColumnWidth]}>
+                <Text>Collection Centre</Text>
+            </View>
+            <View style={[styles.tableColumn, styles.contactColumnWidth]}>
+                <Text>Contact</Text>
+            </View>
+            <View style={[styles.tableColumn, styles.packingDateColumnWidth]}>
+                <Text>Packing Date</Text>
+            </View>
+            <View style={[styles.tableColumn, styles.instructionsColumnWidth]}>
+                <Text>Instructions</Text>
+            </View>
+            <View style={[styles.tableColumn, styles.numberOfLabelsColumnWidth]}>
+                <Text>Number of Parcels</Text>
+            </View>
+        </View>
+    )
+
     const createRowData = (rowData: DriverOverviewTableData, index: number) => {
         return (
             // eslint-disable-next-line react/no-array-index-key
@@ -156,20 +182,22 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                     <Text>{rowData.name}</Text>
                 </View>
                 <View style={[styles.tableColumn, styles.addressColumnWidth]}>
-                    {rowData.address.postcode || rowData.clientIsActive ? (
+                    { rowData.collection_centre.isDelivery ? rowData.address.postcode || rowData.clientIsActive ? (
                         <>
                             <Text>{rowData.address.line1}</Text>
                             <Text>{rowData.address.line2}</Text>
                             <Text>{rowData.address.town}</Text>
                             <Text>{rowData.address.county}</Text>
                             <Text>{rowData.address.postcode}</Text>
-                            <Text>{rowData.isDelivery ? "Delivery" : "Collection"}</Text>
                         </>
                     ) : (
                         <Text>
                             {rowData.clientIsActive ? displayPostcodeForHomelessClient : "-"}
                         </Text>
-                    )}
+                    ):  <>
+                            <Text>{rowData.collection_centre.name}</Text>
+                        </>
+                    }
                 </View>
                 <View style={[styles.tableColumn, styles.contactColumnWidth]}>
                     <Text>{rowData.contact}</Text>
@@ -187,8 +215,8 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
         );
     }
 
-    const deliveries = data.tableData[0].map(createRowData)
-    const collections = data.tableData[1].map(createRowData)
+    const collections = data.tableData[0].map(createRowData)
+    const deliveries = data.tableData[1].map(createRowData)
 
     return (
         <Document>
@@ -209,9 +237,9 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                         COMPLETION OF DELIVERIES
                     </Text>
                 </View>
-                <View style={[styles.flexColumn, { width: "100%" }]}>{header}</View>
+                <View style={[styles.flexColumn, { width: "100%" }]}>{deliveryHeader}</View>
                 <View style={[styles.tableSection, styles.flexColumn]}>{deliveries}</View>
-                <View style={[styles.flexColumn, { width: "100%" }]}>{header}</View>
+                <View style={[styles.flexColumn, { width: "100%" }]}>{collectionHeader}</View>
                 <View style={[styles.tableSection, styles.flexColumn]}>{collections}</View>
             </Page>
         </Document>
--- a/src/pdf/DriverOverview/DriverOverviewPdf.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdf.tsx
@@ -3,6 +3,8 @@
 import React from "react";
 import { Text, Document, Page, View, StyleSheet, Image } from "@react-pdf/renderer";
 import { displayPostcodeForHomelessClient } from "@/common/format";
+import { faTruck, faShoePrints } from "@fortawesome/free-solid-svg-icons";
+import FontAwesomeIconPdfComponent from "../FontAwesomeIconPdfComponent";
 
 export interface DriverOverviewTableData {
     name: string;
@@ -125,6 +127,11 @@ const styles = StyleSheet.create({
     numberOfLabelsColumnWidth: {
         width: "20%",
     },
+    collectionOrDeliveryHeader: {
+        marginRight: "5px",
+        marginBottom: "5px",
+        alignSelf: "flex-start",
+    }
 });
 
 const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
@@ -237,8 +244,16 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                         COMPLETION OF DELIVERIES
                     </Text>
                 </View>
+                <View style={[styles.h2text, styles.collectionOrDeliveryHeader, styles.flexRow]}>
+                    <Text style={styles.collectionOrDeliveryHeader}>Deliveries</Text>
+                    <FontAwesomeIconPdfComponent faIcon={faTruck}></FontAwesomeIconPdfComponent>
+                    </View>
                 <View style={[styles.flexColumn, { width: "100%" }]}>{deliveryHeader}</View>
                 <View style={[styles.tableSection, styles.flexColumn]}>{deliveries}</View>
+                <View style={[styles.h2text, styles.collectionOrDeliveryHeader, styles.flexRow]}>
+                    <Text style={[styles.collectionOrDeliveryHeader]}>Collections</Text>
+                    <FontAwesomeIconPdfComponent faIcon={faShoePrints}></FontAwesomeIconPdfComponent>
+                </View>
                 <View style={[styles.flexColumn, { width: "100%" }]}>{collectionHeader}</View>
                 <View style={[styles.tableSection, styles.flexColumn]}>{collections}</View>
             </Page>
--- a/src/pdf/DriverOverview/DriverOverviewPdf.tsx
+++ b/src/pdf/DriverOverview/DriverOverviewPdf.tsx
@@ -37,6 +37,8 @@ interface DriverOverviewCardProps {
     data: DriverOverviewCardDataProps;
 }
 
+type CreateHeaderProps = "Delivery" | "Collection";
+
 const styles = StyleSheet.create({
     container: {
         padding: 25,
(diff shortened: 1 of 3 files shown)


src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -3,6 +3,8 @@
 import React from "react";
 import { Text, Document, Page, View, StyleSheet, Image } from "@react-pdf/renderer";
 import { displayPostcodeForHomelessClient } from "@/common/format";
+import { faTruck, faShoePrints } from "@fortawesome/free-solid-svg-icons";
+import FontAwesomeIconPdfComponent from "../FontAwesomeIconPdfComponent";
```
kaydentsoftwire, line 8: importing these to show the delivery and collection icons

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -18,12 +20,16 @@ export interface DriverOverviewTableData {
     instructions?: string;
     clientIsActive: boolean;
     numberOfLabels: number;
+    collection_centre: {
+        name: string;
+        isDelivery: boolean;
+    };
 }
```
kaydentsoftwire, line 26: updating the interface as collection centre name (and whether it is a delivery) is passed through to separate the deliveries and collections and also display the collection centre name if it is a collection

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -18,12 +20,16 @@ export interface DriverOverviewTableData {
...
 
 export interface DriverOverviewCardDataProps {
     driverName: string;
     date: Date;
-    tableData: DriverOverviewTableData[];
+    tableData: DriverOverviewTableData[][];
     message: string;
```
kaydentsoftwire, line 31: This is updated as we pass an array of collections followed by an array of deliveries

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -76,6 +82,7 @@ const styles = StyleSheet.create({
         width: "100%",
         borderTop: "none",
         borderBottom: "1px solid black",
+        marginBottom: "15px",
     },
```
kaydentsoftwire, line 97: This creates space between the two tables

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -120,10 +127,15 @@ const styles = StyleSheet.create({
     numberOfLabelsColumnWidth: {
         width: "20%",
     },
+    collectionOrDeliveryHeader: {
+        marginRight: "5px",
+        marginBottom: "5px",
+        alignSelf: "flex-start",
+    },
 });
```
kaydentsoftwire, line 146: Formatting the 'Deliveries' and 'Collections' headings

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -120,52 +129,65 @@ const styles = StyleSheet.create({
...
+                <View style={[styles.tableColumn, styles.packingDateColumnWidth]}>
+                    <Text>Packing Date</Text>
+                </View>
+                <View style={[styles.tableColumn, styles.instructionsColumnWidth]}>
+                    <Text>Instructions</Text>
+                </View>
+                <View style={[styles.tableColumn, styles.numberOfLabelsColumnWidth]}>
+                    <Text>Number of Parcels</Text>
+                </View>
             </View>
```
kaydentsoftwire, line 170: This function takes in a string - either "Delivery" or "Collection" (I decided against a boolean to make it more readable). If Delivery it creates the header Address, and if it is "Collection" it creates the header collection centre

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -120,52 +129,65 @@ const styles = StyleSheet.create({
...
+                                <Text>{rowData.address.county}</Text>
+                                <Text>{rowData.address.postcode}</Text>
+                            </>
+                        ) : (
+                            <Text>
+                                {rowData.clientIsActive ? displayPostcodeForHomelessClient : "-"}
+                            </Text>
+                        )
+                    ) : (
                         <>
```
kaydentsoftwire, line 197: I added in a ternary operator to check if it is Delivery - in which case do the logic it was already doing - and if not then display the collection centre name

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -182,7 +204,12 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                 </View>
             </View>
         );
-    });
```
kaydentsoftwire, line 185: changed table into a function createRow which can be called twice to create two separate tables of deliveries and collections

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -182,7 +204,12 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
...
+    };
+
+    const deliveriesHeader = createHeader("Delivery");
+    const collectionsHeader = createHeader("Collection");
+    const collections = data.tableData[0].map(createRow);
+    const deliveries = data.tableData[1].map(createRow);
```
kaydentsoftwire, line 248: calling the methods to create the headers and tables

kaydentsoftwire: tableData[0] contains the collections and tableData[1] contains the deliveries

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -14,7 +14,7 @@ import { PdfDataFetchResponse } from "../common";
 interface DriverOverviewData {
     driverName: string;
     date: Date;
-    tableData: DriverOverviewTableData[];
+    tableData: DriverOverviewTableData[][];
     message: string;
```
kaydentsoftwire, line 21: The interface here changes because the tableData now contains two arrays - one of collections and one of deliveries

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -94,32 +108,48 @@ type DriverPdfResponse =
...
+        instructions: clientIsActive ? client?.delivery_instructions ?? "" : "-",
+        clientIsActive: clientIsActive,
+        numberOfLabels: parcel.labelCount,
+        collection_centre: {
+            name: parcel.collection_centre?.name,
+            isDelivery: parcel.collection_centre?.is_delivery,
+        },
+    });
+};
+
```
kaydentsoftwire, line 138: this is called twice, once to populate the collections array and once to populate the deliveries array

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -51,9 +52,10 @@ type ParcelsForDeliveryErrorType = "parcelFetchFailed" | "noMatchingClient";
 const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDeliveryResponse> => {
     const { data, error } = await supabase
         .from("parcels")
-        .select("*, client:clients(*), events(event_data)")
+        .select("*, client:clients(*), events(event_data), collection_centre:collection_centres(*)")
         .in("primary_key", parcelIds)
         .limit(1, { foreignTable: "clients" })
+        .limit(1, { foreignTable: "collection_centres" })
         .eq("events.new_parcel_status", "Shipping Labels Downloaded")
         .order("timestamp", { foreignTable: "events", ascending: false });
```
kaydentsoftwire, line 54: Also selecting the collection centre data from the database

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -31,14 +31,15 @@ const compareDriverOverviewTableData = (
     return 0;
 };
 
-type ParcelsForDelivery = (Schema["parcels"] & {
+type ParcelForDelivery = Schema["parcels"] & {
     client: Schema["clients"];
+    collection_centre: Schema["collection_centres"];
     labelCount: number;
-})[];
+};
```
kaydentsoftwire, line 28: changed the type to not be an array so that it can be reused more flexibly, sometimes used below for the individual parcel and sometimes as an array

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -31,14 +31,15 @@ const compareDriverOverviewTableData = (
...
 
 type ParcelsForDeliveryResponse =
     | {
-          data: ParcelsForDelivery;
+          data: ParcelForDelivery[];
           error: null;
```
kaydentsoftwire, line 33: updated as the type is now singular

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -203,8 +230,20 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
...
+                <View style={[styles.tableSection, styles.flexColumn]}>{deliveries}</View>
+                <View style={[styles.h2text, styles.collectionOrDeliveryHeader, styles.flexRow]}>
+                    <Text style={[styles.collectionOrDeliveryHeader]}>Collections</Text>
+                    <FontAwesomeIconPdfComponent
+                        faIcon={faShoePrints}
+                    ></FontAwesomeIconPdfComponent>
+                </View>
+                <View style={[styles.flexColumn, { width: "100%" }]}>{collectionsHeader}</View>
+                <View style={[styles.tableSection, styles.flexColumn]}>{collections}</View>
             </Page>
```
kaydentsoftwire, line 272: Now shows two headers and two tables + the titles above

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -71,20 +73,32 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
             return { data: null, error: { type: "noMatchingClient", logId: logId } };
         }
 
+        if (parcel.collection_centre === null) {
+            const logId = await logErrorReturnLogId(
+                "Error with fetch: Parcels. No collection centre found"
+            );
+            return { data: null, error: { type: "noCollectionCentre", logId: logId } };
+        }
+
```
kaydentsoftwire, line 75: I added in this error for if no collection centre was found (which includes delivery), but I don't know how to test / QA this as there isn't a way of deleting a collection centre?

kaydentsoftwire (commented):
Worth noting that as it is, if only collections or deliveries are selected, the header to the other one appears but the table is empty. If preferred, I could replace this with a message that says 'No collections/deliveries selected'

src/pdf/DriverOverview/DriverOverviewPdf.tsx
8kristy: I'm a little confused by this structure. Do you think it would be better to make address optional, and then just have 2 new fields: `collectionCentreName?: string` and `isDelivery: boolean`?

src/pdf/DriverOverview/DriverOverviewPdf.tsx
8kristy: Not a fan of the double array - it's not immediately obvious what's in each and it might be difficult to remember which index contains what. Maybe create another type which has 2 `DriverOverviewTableData[]`, e.g.

```
type DriverOverviewParcelData {
    collections: DriverOverviewTableData[];
    deliveries: DriverOverviewTableData[];
}
```

src/pdf/DriverOverview/DriverOverviewPdf.tsx
8kristy: I think `isDelivery` boolean is still readable 

But if you don't like that idea I'd change `CreateHeaderProps` to be an enum - lower chance of typos

src/pdf/DriverOverview/DriverOverviewPdf.tsx
8kristy: Do the collection centres have addresses? From the ticket "Group Collections and mark them with the name of the collection centre and the Feet Icon." I interpreted that as this structure

Deliveries van_icon
table

Collection - centre 1 feet_icon
table

Collection - centre 2 feet_icon
table
...

in which case it might be sensible to keep the address and put in the one of the collection centre itself

Although I do prefer your implementation to that and if the collection centres don't have an address it's probably best to keep as is

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -203,8 +230,20 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
...
+                    <FontAwesomeIconPdfComponent faIcon={faTruck}></FontAwesomeIconPdfComponent>
+                </View>
+                <View style={[styles.flexColumn, { width: "100%" }]}>{deliveriesHeader}</View>
+                <View style={[styles.tableSection, styles.flexColumn]}>{deliveries}</View>
+                <View style={[styles.h2text, styles.collectionOrDeliveryHeader, styles.flexRow]}>
+                    <Text style={[styles.collectionOrDeliveryHeader]}>Collections</Text>
+                    <FontAwesomeIconPdfComponent
+                        faIcon={faShoePrints}
+                    ></FontAwesomeIconPdfComponent>
+                </View>
```
8kristy: Think I like the idea of saying "No deliveries/collections" instead of an empty table if the array is empty (or maybe not even showing the section at all)

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
8kristy: How does this behave with deliveries? If those don't have a collection centre I think we shouldn't throw an error (but we should if it's isDelivery==false and collectionCentre==null)

In terms of testing - QA may be tricky but I think you can manually delete the value of the field in the database through supabase UI (i.e. not the actual collection centre but just the value for some specific record just make it null)

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
8kristy: Do you think we can call it once then split it into two arrays based on whether it's a delivery or not?

8kristy: Do like that big thing as a separate function though - but might need a better name, maybe something like `transformParcelToDriverOverviewTableData`?

8kristy (commented):
A few comments

src/pdf/DriverOverview/DriverOverviewPdf.tsx
kaydentsoftwire: I haven't made the address optional because the way the data is put into this structure, I don't think there would be a case where there could actually be no address (though it would maybe have some fields that were null or an empty string). I didn't think it was worth changing how that works in order to make it optional, but let me know what you think?
Similarly collection_centre_name for a delivery is actually "Delivery" so I have not made that optional

src/pdf/DriverOverview/DriverOverviewPdf.tsx
kaydentsoftwire: I renamed DriverOverviewTableData to DriverOverviewRowData because we create an array of rows rather than an array of tables, which made more sense to me. And then created a version of this type called DriverOverviewTablesData (because it contains the tables)

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -18,19 +20,31 @@ export interface DriverOverviewTableData {
...
 
 interface DriverOverviewCardProps {
     data: DriverOverviewCardDataProps;
 }
 
+enum Method {
+    Delivery = "Delivery",
+    Collection = "Collection",
+}
+
```
kaydentsoftwire, line 42: used the enum as recommended!

src/pdf/DriverOverview/DriverOverviewPdf.tsx
kaydentsoftwire: I decided on the enum - partly because I've not used them before(!), partly because the bit I didn't really like/think was readable was calling the method as createHeader(true) or createHeader(false).

src/pdf/DriverOverview/DriverOverviewPdf.tsx
kaydentsoftwire: Collection Centres do not have addresses so I have kept it as this for now

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
kaydentsoftwire: Deliveries have a collection centre ("Delivery") so doesn't throw an error

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
kaydentsoftwire: I have renamed this and also separated out a new function below that is called once and contains the logic that assembles it into client information
also I was wrong above - it is actually called many times as the parcels are mapped over and every parcel calls this!

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -14,55 +17,46 @@ import { PdfDataFetchResponse } from "../common";
...
-    if (first.name < second.name) {
-        return -1;
-    }
-    if (first.name > second.name) {
-        return 1;
-    }
-    return 0;
-};
-
-type ParcelsForDelivery = (Schema["parcels"] & {
```
kaydentsoftwire, line 34: This was enforcing a sort by client name which I removed

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -14,55 +17,46 @@ import { PdfDataFetchResponse } from "../common";
...
         .from("parcels")
-        .select("*, client:clients(*), events(event_data)")
+        .select("*, client:clients(*), events(event_data), collection_centre:collection_centres(*)")
         .in("primary_key", parcelIds)
         .limit(1, { foreignTable: "clients" })
+        .limit(1, { foreignTable: "collection_centres" })
         .eq("events.new_parcel_status", "Shipping Labels Downloaded")
+        .order("name", { foreignTable: "collection_centres"})
+        .order("address_postcode", { foreignTable: "clients" })
         .order("timestamp", { foreignTable: "events", ascending: false });
```
kaydentsoftwire: Updated to sort by collection centre name and then by postcode (so that collection centres will appear grouped by postcode on the pdf, and deliveries will appear sorted by postcode in the javascript default sorting way)

kaydentsoftwire (commented):
I updated the sorting in this iteration so that the PDF is sorted by collection centre and postcode, not by name - image above is updated to show the changes as well as not showing a table with no content.
Also many of these comments were meant to be responses to your comments, Kristina, but now your comments aren't showing above them - so I hope it's still clear!

src/pdf/DriverOverview/DriverOverviewPdf.tsx
kaydentsoftwire: I went for not showing the sections that don't have any content

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -14,55 +17,69 @@ import { PdfDataFetchResponse } from "../common";
...
 const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDeliveryResponse> => {
     const { data, error } = await supabase
         .from("parcels")
-        .select("*, client:clients(*), events(event_data)")
+        .select("*, client:clients(*), events(event_data), collection_centre:collection_centres(*)")
         .in("primary_key", parcelIds)
         .limit(1, { foreignTable: "clients" })
+        .limit(1, { foreignTable: "collection_centres" })
         .eq("events.new_parcel_status", "Shipping Labels Downloaded")
-        .order("timestamp", { foreignTable: "events", ascending: false });
```
kaydentsoftwire, line 54: Do not believe we need to order by timestamp ie. most recent event as we are ordering by postcode and collection centre. I tried implementing the sort here through .order but it was very inconsistent in what it gave me so I moved to manually sorting

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -14,55 +17,69 @@ import { PdfDataFetchResponse } from "../common";
...
+    }
+    if (first.address.postcode) {
+        return 1;
+    }
+    if (second.address.postcode) {
+        return -1;
+    }
+    return 0;
+};
+
```
kaydentsoftwire: essentially the idea here is implementing javascripts default sorting for the postcode - I thought this makes sense as it is what we are doing elsewhere

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
kaydentsoftwire: And then changed to sort by collection centre name instead!

src/pdf/DriverOverview/DriverOverviewPdf.tsx
8kristy: That's fair, I still think it might be clearer to take out the fields out of the `collection_centre` structure - `isDelivery` doesn't make much sense to me under `collection_centre`
Unless there's a reason why they have to be that way?

8kristy: Oh wait you've done that haven't you, silly github not showing me changes on the comment >:(

src/pdf/DriverOverview/DriverOverviewPdf.tsx
8kristy: Yep that's perfect, thank you!

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -120,52 +138,65 @@ const styles = StyleSheet.create({
...
-            <View style={[styles.tableColumn, styles.numberOfLabelsColumnWidth]}>
-                <Text>Number of Parcels</Text>
+    const createHeader = (category: string): React.JSX.Element => {
+        return (
+            <View style={[styles.tableHeader, styles.flexRow]}>
+                <View style={[styles.tableColumn, styles.nameColumnWidth]}>
+                    <Text>Name</Text>
+                </View>
+                <View style={[styles.tableColumn, styles.addressColumnWidth]}>
+                    <Text>{category === "Collection" ? "Collection Centre" : "Address"}</Text>
```
8kristy:
```suggestion
    const createHeader = (category: Method): React.JSX.Element => {
        return (
            <View style={[styles.tableHeader, styles.flexRow]}>
                <View style={[styles.tableColumn, styles.nameColumnWidth]}>
                    <Text>Name</Text>
                </View>
                <View style={[styles.tableColumn, styles.addressColumnWidth]}>
                    <Text>{category === Method.Collection ? "Collection Centre" : "Address"}</Text>
```

8kristy: Since you mentioned you haven't used enums before - it's best practice to never look at the actual values of them, i.e. all the comparison etc is done through the enum itself. This is good because the compiler can catch any problems like you using a value that shouldn't be available or having a typo and then spending ages trying to find the issue. It also means you can update the enum value in one place and it will change everywhere rather than having to update everything manually. Very powerful tool!

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
8kristy: Yep think that's a good idea - I think you can simplify this function to 
```
const comparePostcodes = (first: DriverOverviewRowData, second: DriverOverviewRowData): number => {
    if (first.address.postcode && second.address.postcode) {
        return first.address.postcode.localeCompare(second.address.postcode);
    }
    if (!first.address.postcode && !second.address.postcode) {
        return 0;
    }
    return first.address.postcode ? 1 : -1;
};
```

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
8kristy: Just curious how was it inconsistent?

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -94,32 +123,56 @@ type DriverPdfResponse =
...
+            deliveries.push(transformRowToDriverOverviewTableData(parcel));
+        } else {
+            collections.push(transformRowToDriverOverviewTableData(parcel));
+        }
+    }
+
+    return {
+        collections: collections.sort(compareCollectionCentres),
+        deliveries: deliveries.sort(comparePostcodes),
+    };
```
8kristy: You could maybe do
```
const transformedParcels = parcels.map(p => transformRowToDriverOverviewTableData(p));

return {
    collections = transformedParcels.filter(p => !p.isDelivery).sort(compareCollectionCentres),
    deliveries = transformedParcels.filter(p => p.isDelivery).sort(comparePostcodes),
}
```
but it's up to you - I'm just a fan of array functions

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -203,8 +239,46 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
...
+                            ]}
+                        >
+                            <Text style={styles.collectionOrDeliveryHeader}>Deliveries</Text>
+                            <FontAwesomeIconPdfComponent
+                                faIcon={faTruck}
+                            ></FontAwesomeIconPdfComponent>
+                        </View>
+                        <View style={[styles.flexColumn, { width: "100%" }]}>
+                            {deliveriesHeader}
+                        </View>
```
8kristy: You can pull this out as a separate component with headerName, icon, header and itemList as params (maybe more those are the differences that I see) and then reuse it

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -94,32 +123,56 @@ type DriverPdfResponse =
...
-            },
-            contact: clientIsActive ? client?.phone_number ?? "" : "-",
-            packingDate: formatDateToDate(parcel.packing_date) ?? null,
-            instructions: clientIsActive ? client?.delivery_instructions ?? "" : "-",
-            clientIsActive: clientIsActive,
-            numberOfLabels: parcel.labelCount,
-        });
-    }
-    clientInformation.sort(compareDriverOverviewTableData);
+    const clientInformation = transformParcelDataToClientInformation(parcels);
```
8kristy: Are clientInformation and transformParcelDataToClientInformation accurate names here? They looks like the transform parcel data to me

8kristy (commented):
Few potential code improvements

I do want to look into sorting from db a bit more since that would be nicer/easier than custom methods but I understand if it doesn't work

Otherwise looks very good!

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
kaydentsoftwire: It basically didn't seem to be working (I wasn't able to pin down why). It seemed to be working at first but then when I selected a larger amount of data the collection centres were all in different orders as were the postcodes. My nearest guess is that something was happening later in the code to rearrange the order. If that is the case, maybe with the changes you have suggested it will work :)

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -14,55 +17,46 @@ import { PdfDataFetchResponse } from "../common";
...
-        .select("*, client:clients(*), events(event_data)")
+        .select("*, client:clients(*), events(event_data), collection_centre:collection_centres(*), collection_centres(name), clients(address_postcode)")
         .in("primary_key", parcelIds)
         .limit(1, { foreignTable: "clients" })
+        .limit(1, { foreignTable: "collection_centres" })
         .eq("events.new_parcel_status", "Shipping Labels Downloaded")
-        .order("timestamp", { foreignTable: "events", ascending: false });
+        .order("collection_centres(name)")
+        .order("clients(address_postcode)");
+
```
kaydentsoftwire, line 54: If you're interested, found the solution for this here: https://github.com/supabase/supabase-js/issues/97

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
kaydentsoftwire: Done!

kaydentsoftwire (commented):
Improved the code as previous review and changed manual sorting to database sorting

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -94,33 +101,48 @@ type DriverPdfResponse =
 
 type DriverPdfErrorType = ParcelsForDeliveryErrorType;
 
+const transformRowToDriverOverviewTableData = (
+    parcel: ParcelForDelivery
```
kaydentsoftwire, line 105: renamed this method

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
```diff
@@ -94,33 +101,48 @@ type DriverPdfResponse =
...
-            instructions: clientIsActive ? client?.delivery_instructions ?? "" : "-",
-            clientIsActive: clientIsActive,
-            numberOfLabels: parcel.labelCount,
-        });
-    }
-    clientInformation.sort(compareDriverOverviewTableData);
-    return { data: clientInformation, error: null };
+    const tableData = transformParcelDataToTableData(parcels);
+    return { data: tableData, error: null };
 };
```
kaydentsoftwire, line 147: renamed clientInformation to tableData

src/pdf/DriverOverview/DriverOverviewPdf.tsx
kaydentsoftwire: Done, thank you

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
kaydentsoftwire: Renamed in latest :)

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -205,8 +267,8 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                         COMPLETION OF DELIVERIES
                     </Text>
                 </View>
-                <View style={[styles.flexColumn, { width: "100%" }]}>{header}</View>
-                <View style={[styles.tableSection, styles.flexColumn]}>{table}</View>
+                {deliveries.length ? deliveriesTable : null}
+                {collections.length ? collectionsTable : null}
```
8kristy: Think this is the standard way of  doing optional components in react?
```suggestion
                {deliveries.length ?? deliveriesTable}
                {collections.length ?? collectionsTable}
```

8kristy (approved):
Tiny suggestion feel free to merge after fixing it

well done!!

PR #360, opened 2024-05-28
VFB-201: Remove new packing slot from table if error on creation

maxjohns0n:
## What's changed
If there is an error when creating a new packing slot, the new row now disappears.

## Screenshots / Videos
When adding "test" slot. Now is not shown once the error has occurred.
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

### DIFF
--- a/src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
+++ b/src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
@@ -165,6 +165,7 @@ const PackingSlotsTable: React.FC = () => {
                 setErrorMessage(
                     `Failed to add the packing slot. Log ID: ${insertPackingSlotError.logId}`
                 );
+                setRows((rows) => rows.slice(0, -1));
                 void sendAuditLog({
                     ...baseAuditLog,
                     wasSuccess: false,


src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
```diff
@@ -165,6 +165,7 @@ const PackingSlotsTable: React.FC = () => {
                 setErrorMessage(
                     `Failed to add the packing slot. Log ID: ${insertPackingSlotError.logId}`
                 );
+                setRows((rows) => rows.slice(0, -1));
```
maxjohns0n, line 168: Am I right in assuming the new slot will always be at the end of the array? (The code which adds it does always add it at the end, the user can't reorder them until the creation is done, and I don't think there is any array sorting going on.)

8kristy: I think that's a fair assumption, but is it possible to prevent the front end from adding this to the list at all unless we know that the DB call succeeded?

maxjohns0n: I don't think so as far as I understand. When you click the button to add a new packing slot, it adds a new editable row to the table, which the user then fills in. And I think you have to do that by adding it to this array. So if we didn't add to the array then I'd imagine we couldn't fill out its details and save it.

8kristy: Ahh I see, that makes sense

8kristy (approved):

Issue #180, opened 2024-03-23
Driver Overview PDF - Not Creating a Status


dan-langridge:
### Reproduction steps

When I:

- Select Multiple Clients
- Download Driver Overview PDF
- Submit
- Status of Clients does not update

### Expected behaviour

Status of Clients should update to:
Out for Delivery with [Name of Driver] [Date]

### What browsers are you seeing the problem on?

Safari

### What OS is your computer running on?

macOS

rickaw:
## Developer Notes
- [VFB-175](https://softwiretech.atlassian.net/browse/VFB-175) covered the creation of a new status
- The outstanding task is to change the new status name from `Driver Overview Downloaded` to `Out for Delivery with [Name of Driver] [Date]`

AmeliaIsCoding:
Notes: 
-Delete driver overview downloaded status as this is no longer needed
-When driver overview downloaded, new status should be "Out for Delivery" with the event data containing the driver name and date

AmeliaIsCoding:
### QA

- Select a parcel with its last status in the past
- Download the driver overview for the parcel, inputting a driver name and a date of your choice
- [x]  The parcel's status should update to Out for Delivery (with [Name of Driver]) @ [DD/MM/YYYY] where the name and date are correct to what you input
- [x] In the parcel modal, in the events table, the most recent event should display be Out for Delivery (with [Name of Driver]) @ [DD/MM/YYYY]
- [x] Repeat but enter the driver name "     John     " --> the driver name in the both the status and the PDF should be "John", ie with whitespace trimmed
- [x] "Driver Overview Downloaded" is not visible as a status in the front end
- [x] No events in the `events` in supabase have event name "Driver Overview Downloaded"
- [x] In the `status_order` table, there is no row with status "Driver Overview Downloaded"
- [x] In local, some seeded parcels are shown in the front end with last status "Out for Delivery (with [Name of Driver]) @ [DD/MM/YYYY]

rickaw:
@dan-langridge In the parcels table's "last status" column, we currently show the name of the last status along with the date the status was set (see image). We also allow the user to specify the delivery date when they download the driver overview, which goes on the driver overview PDF, but this date is not stored for any other purposes yet.

[...]

PR #361, opened 2024-05-28
#180: Driver overview status change

AmeliaIsCoding:
## What's changed

- Change "Driver Overview Downloaded" status to "Out for Delivery"
- Events with this status also have "with [Driver Name]" in the event data
- When generating a driver overview PDF, the date of the event is set to the delivery date, not necessarily today's date, and this is shown in the front end parcels table

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image|image|

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
```diff
@@ -87,12 +91,17 @@ const DriverOverviewModal: React.FC<ActionModalProps> = (props) => {
...
+        setDriverName(null);
         setErrorMessage(null);
     };
 
     const onPdfCreationCompleted = async (): Promise<void> => {
-        const { error } = await props.updateParcelStatuses(props.selectedParcels, props.newStatus);
+        const { error } = await props.updateParcelStatuses(
+            props.selectedParcels,
+            props.newStatus,
+            `with ${driverName ?? "Unknown Driver"}`,
```
8kristy: Isn't the driver name required? If we set it to null maybe we can instead cover cases where driver name is just whitespace

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
```diff
@@ -70,15 +70,19 @@ const DriverOverviewModal: React.FC<ActionModalProps> = (props) => {
...
     const [date, setDate] = useState(dayjs());
 
     const [isDateValid, setIsDateValid] = useState(true);
 
-    const isInputValid = isDateValid && driverName.length > 0;
+    const isInputValid = isDateValid && driverName;
 
     const onDriverNameChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
-        setDriverName(event.target.value);
+        if (event.target.value === "") {
```
8kristy: Referencing the other comment it may make sense to change this to e.g. `event.target.value.trim().length === 0`

Although in an ideal world the name field would probably have some regex to validate that the name has letters in it since now you can still set it as `.` or something - maybe raise a ticket?

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
```diff
@@ -70,15 +70,19 @@ const DriverOverviewModal: React.FC<ActionModalProps> = (props) => {
...
+            setDriverName(null);
+        } else {
+            setDriverName(event.target.value);
```
8kristy: Maybe trim whitespace here too?

src/app/parcels/format.ts
```diff
@@ -0,0 +1,4 @@
+import { StatusType } from "./ActionBar/Statuses";
+
+export const formatEventName = (newStatus: StatusType, eventData: string | null): string =>
+    newStatus + (eventData ? ` (${eventData})` : "");
```
8kristy:
```suggestion
    `${newStatus} ${eventData ? `(${eventData})` : ""}`
```

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -190,7 +190,7 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
                 <View style={[styles.infoAndLogoContainer, styles.flexRow]}>
                     <View style={styles.informationContainer}>
                         <Text style={styles.h1text}>Driver Overview</Text>
-                        <Text style={styles.h2text}>Driver Name: {data.driverName}</Text>
+                        <Text style={styles.h2text}>Driver Name: {data.driverName ?? ""}</Text>
```
8kristy: Do we want this to say "Unknown driver" here like in the other place? May be worth pulling out into a constant somewhere

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
```diff
@@ -0,0 +1,11 @@
+update public.events set new_parcel_status = 'Out for Delivery', event_data = 'with John' where new_parcel_status = 'Driver Overview Downloaded';
+
+delete from public.status_order where event_name = 'Driver Overview Downloaded';
+
+update public.status_order set workflow_order = 12 where event_name = 'Map Generated';
+update public.status_order set workflow_order = 13 where event_name = 'Out for Delivery';
```
8kristy, line 6: Looks like `Driver Overview Downloaded` is currently 12 - should we keep it as that and avoid changing other values?

8kristy (commented):
Nice work, just a few bits about null/required driver names

8kristy:
Also very random tip but I noticed you have some repeated stuff in your git history (even though it just looks like you did 2 things and named them the same) - you don't need to do anything now since everything gets squashed on merge anyways, but if you ever need/want to keep your git history a bit cleaner, you can use `git rebase -i <target_branch>`. It's a bit tricky but you can "rewrite history" by squashing some of your commits together, reordering and renaming them (it's very easy to mess up with all the conflicts that pop up so just remember that you can always revert with `git reset --hard origin/<your_branch_name>` - as long as you haven't pushed the changes to remote!!)

Not sure how much you know about rebase but it's an alternative to merging - just puts your commits on top of the branch on which you're rebasing (so you can just do `git rebase <target_branch>` - the `-i` just puts you in interactive mode which allows you to mess with the commits)

Just thought it might be something interesting :)

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
AmeliaIsCoding: The driver name is not required - ie, its possible to open the driver overview modal, choose a date and download the PDF without typing anything in the driver name box. In this case `driverName` is null. But I agree that it would be good to also use "Unknown Driver" if `driverName` is just whitespace - I think this will be the case if I use `trim` as in your other comment

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
AmeliaIsCoding: i'll raise a ticket. it also might be nice to display some sort of user feedback if the driver name is invalid.
ticket: https://softwiretech.atlassian.net/browse/VFB-240

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
AmeliaIsCoding: Sorry, completely ignore me, the driver name is required

AmeliaIsCoding: Do you want me to make any changes here? since the download button is only enabled when the input is valid (ie the driver name is non null) it should not be possible for this function to be called when the driverName is null, however TS doesn't realise this, it thinks driverName could still be null

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
AmeliaIsCoding: Have done, good idea

src/pdf/DriverOverview/DriverOverviewPdf.tsx
AmeliaIsCoding: Sure

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
AmeliaIsCoding: I think it would be best to delete the Driver Overview Downloaded status as it now should not be used for anything and has become meaningless

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
8kristy: Does it allow you to delete all of the nulls from `string | null`? We can probably just return everything to the state it was in before and then all the trimming etc issues are going to be covered by the validation ticket you raised

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
8kristy: Sorry I worded that very badly 😄 I meant set `Out for Delivery` to 12 to replace `Driver Overview Downloaded` and then avoid updating `Map Generated` and all the statuses afterwards so they keep their current order number

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
AmeliaIsCoding: You mean change the type of the driverName state back to `string` only and default it to `""`?

src/app/parcels/ActionBar/Statuses.tsx
```diff
@@ -16,9 +16,8 @@ import { fetchParcelStatuses } from "@/app/parcels/parcelsTable/fetchParcelTable
 export type StatusType = ParcelStatus[][number];
 
 const nonMenuStatuses: StatusType[] = [
-    "Driver Overview Downloaded", //Generated when driver overview pdf downloaded
     "Map Generated", //Generated when maps generated
-    "Out for Delivery",
```
8kristy, line 21: What was this used for in the past? If we update the database should we reorder this to match?

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
AmeliaIsCoding: okay, we could - just so you know we already have a status called "Out for Delivery" with status_order 14, so the result of this is that we'd have a gap in the statuses, ie a status with order 13, a status with order 15, and statuses either side, but none with order 14 - i thought this might be abit confusing to future devs, although it won't affect anything

image

src/app/parcels/ActionBar/Statuses.tsx
AmeliaIsCoding: the purpose of this array is just to store the statuses which should not be visible in the status dropdown (the dropdown which allows us to directly change the statuses of parcels)
"Out for Delivery" was already in this array, all I did was remove "Driver Overview Downloaded" as this is no longer a status, and move "Out for Delivery" from the bottom of the array to where the "Driver Overview Downloaded" entry was. the order of this array doesn't actually matter.

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
8kristy: Yeah just noticed - did we update the status to old "Out for Delivery"? I agree with you on the gap but I'm confused by 2 out for deliveries even more

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
8kristy: Yep although I'm happy with the current solution too

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
8kristy: Is this just something that exists in seeded data but we can't manually update to it and there's no event that triggered it (until now where downloading the driver overview changes it)?

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
AmeliaIsCoding: Ok, I see how the old solution was simpler, but I like the use of null because it represents a driver name that hasn't been entered yet or is invalid and I think that makes more sense to represent as null rather than an empty string

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
AmeliaIsCoding: No there is only 1 "Out for Delivery" - this is what the `status_order` table changes to
image

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
AmeliaIsCoding: (i'll keep it as current solution)

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
8kristy: Ok so to summarize
- We had Driver Overview Downloaded and Out for Delivery
- Status updated to Driver Overview Downloaded when you download the pdf
- Out for Delivery doesn't seem to be used anywhere and just appears in the seeded data?
- You used the unused Out for Delivery to update to when we download the pdf and deleted Driver Overview Downloaded
- You shifted the numbers so there's no gap where Driver Overview Downloaded was
Is this correct?
If yes then happy with stuff, guess just worried about Out for Delivery being used for something else

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
```diff
@@ -70,15 +71,19 @@ const DriverOverviewModal: React.FC<ActionModalProps> = (props) => {
...
-    const isInputValid = isDateValid && driverName.length > 0;
+    const isInputValid = isDateValid && driverName;
 
     const onDriverNameChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
-        setDriverName(event.target.value);
+        if (event.target.value.trim().length === 0) {
+            setDriverName(null);
+        } else {
+            setDriverName(event.target.value.trim());
+        }
```
8kristy:
```suggestion
        let trimmedDriverName = event.target.value.trim();
        setDriverName(trimmedDriverName.length !== 0 ? trimmedDriverName ? null);
```

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
```diff
@@ -0,0 +1,11 @@
+update public.events set new_parcel_status = 'Out for Delivery', event_data = 'with John' where new_parcel_status = 'Driver Overview Downloaded';
```
8kristy:
```suggestion
update public.events set new_parcel_status = 'Out for Delivery', event_data = 'with John' where new_parcel_status = 'Driver Overview Downloaded' or new_parcel_status = 'Out for Delivery';
```

8kristy: Just for those seeded bits - hopefully it doesn't get stuck in a loop?

8kristy (approved):
Little suggestion on readability (if you think it's worse feel free to leave it - it looked better in my head) and db status updates (assuming it doesn't break anything and my assumptions on the status are correct)

If I understood the old use of Out for Delivery correctly then I'm happy to merge this, if not let me know and we can spend a little more time figuring it out

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
AmeliaIsCoding: Yes that's correct - Out for Delivery was not used elsewhere

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
AmeliaIsCoding: looks good

supabase/migrations/20240529104554_remove_driver_overview_downloaded_status.sql
AmeliaIsCoding: good catch, this will make sure existing "Out for Delivery" events include the event data

Issue #302, opened 2024-04-26
Bulk Edit: Change Date of Multiple Parcels at once


dan-langridge:
### Describe the solution / feature you'd like

There are some occasions where we need to move multiples parcels onto a different packing day.
The ability to select multiple parcels and change the packing date for all of them at once would be handy.
- Select all parcels
- Action
- Change Date
- Change Date popup appears with calendar picker 
- Submit

dan-langridge:
A separate action to change the packing slot is a brilliant idea, thanks!Kind RegardsDan Langridge(+44) 07540 696683On 29 Apr 2024, at 09:26, Ricky Kawagishi ***@***.***> wrote:﻿
@dan-langridge Thanks for this. Would you also like to be able to pick the packing slot whilst changing the packing date? Or perhaps a separate action to set the packing slot for selected parcels?

rickaw:
## Developer Notes
- Add a parcel action to set packing date of selected parcels
- Add a parcel adtion to set packing slot of selected parcels

KylePUG:
QA:

- [ ] Check that packing slot/date is changed:
  - [x] Select a few parcels
  - [x] Under 'Actions' dropdown, click 'Change Packing Slot/Date'
  - [x] Before selecting a packing slot/date (i.e. leave the dropdown/date field blank), click change - a warning message should show - "Please choose a valid packing slot/date."
  - [x] Choose a packing slot/date to switch to and click 'Change'
  - [x] A success message (in the modal) should be shown - "Packing Slot/Date Changed to <packing_slot/packing_date>"
  - [x] Verify that the packing slot/date has been changed 
  - [x] A new status should also be seen after the change (can be seen on last status or in the modal on clicking the parcel) "Packing Slot/Date Changed (new packing slot/date: <packing_slot/packing_date> @ <date_of_change>"
  - [x] Repeat the steps for packing date/slot (note that a wider date range filter will be required to verify changes for packing date)
  - NOTE: Bug when changing packing slots (stay on success screen for a few seconds and refetch error messages come in) raised on JIRA https://softwiretech.atlassian.net/browse/VFB-246?atlOrigin=eyJpIjoiYWJiNGE2MzM0NDIyNDkxMzgxZjI4YmMxNzhlMWUzNDYiLCJwIjoiaiJ9

- [x] Check audit logs:
  - [x] After the changes made in the previous section, audit logs should be created. These are found under Admin > View Audit Logs
[...]

PR #366, opened 2024-05-29
#302: Bulk edit of parcel packing date and slot

KylePUG:
## What's changed

- Added the ability to bulk edit parcel packing slots and dates under Actions
  - Added under actions, 'Change Packing Date' and 'Change Packing Slot'
  - These will open modals that prompt user to choose a new packing date/slot
  - on change these will modify all selected parcels to the desired packing date/slot
- These both send audit logs that are viewable under the audit log tables
- These will also update "Last Status"

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/parcels/ActionBar/ActionModals/CommonDateAndSlot.tsx
```diff
@@ -0,0 +1,102 @@
...
+            action = "change packing date";
+            break;
+        case "packingSlot":
+            updateResponse = await supabase
+                .from("parcels")
+                .update({ packing_slot: packingDateOrSlotData }, { count: "exact" })
+                .eq("primary_key", parcel.parcelId);
+            action = "change packing slot";
+            break;
+    }
```
8kristy: Could we generalize this bit?

```
const updateParcelPackingDateOrSlot = (fieldToUpdate: {field: packingDateOrSlotData}, actionMessage: string : PostgrestSingleResponse<null> => {
    return await supabase
                .from("parcels")
                .update(fieldToUpdate, { count: "exact" })
                .eq("primary_key", parcel.parcelId);
            action = actionMessage;
}

...

switch (updateField) {
        case "packingDate":
            updateResponse = updateParcelPackingDateOrSlot({ packing_date: packingDateOrSlotData }, "change packing date");
            break;
        case "packingSlot":
            updateResponse = updateParcelPackingDateOrSlot({ packing_slot: packingDateOrSlotData }, "change packing slot");
            break;
    }
```

src/app/parcels/ActionBar/ActionModals/CommonDateAndSlot.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { logErrorReturnLogId, logWarningReturnLogId } from "@/logger/logger";
+import supabase from "@/supabaseClient";
+import { PostgrestSingleResponse } from "@supabase/supabase-js";
+
+export const getUpdateErrorMessage = (error: FetchParcelError | UpdateParcelError): string => {
+    let errorMessage: string;
+    switch (error.type) {
+        case "noMatchingParcels":
+            errorMessage = "No parcel in the database matches the selected parcel.";
+            break;
```
8kristy, line 20: Do we need this one?

src/app/parcels/ActionBar/ActionModals/CommonDateAndSlot.tsx
```diff
@@ -0,0 +1,102 @@
...
+        case "failedToFetchParcel":
+            errorMessage = "Failed to fetch parcel data.";
+            break;
```
8kristy, line 23: Is this one missing to log after fetching parcel?

src/app/parcels/ActionBar/ActionModals/DateChangeModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const onDateChange = async (): Promise<void> => {
+        if (!date) {
+            setWarningMessage("Please choose a valid packing date.");
+            return;
+        }
+        const newPackingDate = getDbDate(dayjs(date));
+        let packingDateUpdateError: FetchParcelError | UpdateParcelError | null = null;
+        for (const parcel of props.selectedParcels) {
+            (async () => {
+                const { error } = await packingDateOrSlotUpdate(
```
8kristy: Does this not also return parcelId?

src/app/parcels/ActionBar/ActionModals/DateChangeModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+                if (error) {
+                    packingDateUpdateError = error;
+                    return;
+                }
+            })();
+        }
+
+        const { error: statusUpdateError } = await props.updateParcelStatuses(
+            props.selectedParcels,
+            props.newStatus,
```
8kristy, line 49: Where does it get the new status from?

src/app/parcels/ActionBar/ActionModals/DateChangeModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+            `new packing date: ${newPackingDate}`,
+            "change packing date"
+        );
+        if (packingDateUpdateError) {
+            setErrorMessage(getUpdateErrorMessage(packingDateUpdateError));
+        } else if (statusUpdateError) {
+            setErrorMessage(getStatusErrorMessageWithLogId(statusUpdateError));
```
8kristy: Before I ran `post_checkout` I'd get an error but the parcels got updated - I don't think we should roll back the ones that succeeded, but is there a way of maybe saying "x updates failed"?

src/app/parcels/ActionBar/ActionModals/SlotChangeModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+interface SlotInputProps {
+    packingSlotsLabelsAndValues: PackingSlotsLabelsAndValues;
+    setSlot: (slot: string) => void;
+}
+
+const SlotChangeInput: React.FC<SlotInputProps> = ({ packingSlotsLabelsAndValues, setSlot }) => {
+    return (
+        <>
+            <Heading>What date would you like to change to?</Heading>
```
8kristy:
```suggestion
            <Heading>What slot would you like to change to?</Heading>
```

src/app/parcels/ActionBar/ActionModals/SlotChangeModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+            );
+            setDisplayModal(false);
+            return;
+        }
+        setPackingSlots(packingSlotsData);
+    })();
+
+    const onSlotChange = async (): Promise<void> => {
+        if (slot === "") {
+            setWarningMessage("Please choose a valid packing slot.");
```
8kristy: Is this on slot change or on submit?

src/app/parcels/ActionBar/ActionModals/SlotChangeModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+                }
+            })();
+        }
+
+        let newPackingSlotText: string = "";
+        packingSlots.forEach((packingSlot) => {
+            if (packingSlot[1] === slot) {
+                newPackingSlotText = packingSlot[0];
+            }
+        }); //since 'slot' is the key for the packing slot, this algorithm is used to just recover the value, i.e. 'AM', 'PM, etc.
```
8kristy: Not sure what this does? If it's just the name of the slot, since we update them all to the same value, can't we just use packingSlots[0][0]? Can probably just put it straight in the string

src/app/parcels/ActionBar/ActionModals/SlotChangeModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+            props.selectedParcels,
+            props.newStatus,
+            `new packing slot: ${newPackingSlotText}`,
+            "change packing slot"
+        );
+        if (packingSlotUpdateError) {
+            setErrorMessage(getUpdateErrorMessage(packingSlotUpdateError));
+        } else if (statusUpdateError) {
+            setErrorMessage(getStatusErrorMessageWithLogId(statusUpdateError));
+        } else {
```
8kristy: Same thing with error appearing but updating here

src/app/parcels/ActionBar/ActionAndStatusBar.tsx
```diff
@@ -19,6 +19,7 @@ export type UpdateParcelStatuses = (
     parcels: ParcelsTableRow[],
     newStatus: StatusType,
```
8kristy, line 20: Was date not used anywhere before?

src/app/parcels/ActionBar/Statuses.tsx
```diff
@@ -36,24 +38,28 @@ export const saveParcelStatus = async (
...
 ): Promise<SaveParcelStatusResult> => {
     const timestamp = (date ?? dayjs()).toISOString();
     const eventsToInsert = parcelIds
-        .map((parcelId: string) => {
+        .map((parcelId: string, index) => {
             return {
                 new_parcel_status: statusName,
                 parcel_id: parcelId,
                 event_data: statusEventData,
+                client_id: clientIds?.slice()[index],
```
8kristy: What does the slice do here?

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
```diff
@@ -98,6 +98,7 @@ const DriverOverviewModal: React.FC<ActionModalProps> = (props) => {
             props.selectedParcels,
             props.newStatus,
             `with ${driverName ?? displayNameForNullDriverName}`,
+            undefined,
```
8kristy, line 104: May be useful to fetch clientIds here too and pass them in if possible?

8kristy (commented):
A few small bits but well done, very well structured

Could you please carefully review the error issues think they may cause some odd behaviour

src/app/parcels/ActionBar/ActionModals/CommonDateAndSlot.tsx
KylePUG: done

src/app/parcels/ActionBar/ActionModals/CommonDateAndSlot.tsx
KylePUG: `fetchParcel` can output two errors, one of them is this. I believe that it is necessary here to cover all bases?

src/app/parcels/ActionBar/ActionModals/CommonDateAndSlot.tsx
KylePUG: I thought it might've been unnecessary, but on second though any action should be logged whether successful or not so adding this.

src/app/parcels/ActionBar/ActionModals/DateChangeModal.tsx
KylePUG: these can be found in parcels > ActionBar > ActionModals > Actions.tsx 

The list of available actions are in `availableActions` which are used to give newStatus when a specific MenuItem is clicked in the Actions dropdown

src/app/parcels/ActionBar/ActionModals/SlotChangeModal.tsx
KylePUG: done

src/app/parcels/ActionBar/ActionModals/SlotChangeModal.tsx
KylePUG: Agree that this can be confusing - the button says "Change" at the moment hence the name choice. I will change this to onSubmit for clarity :)

src/app/parcels/ActionBar/ActionModals/SlotChangeModal.tsx
KylePUG: packingSlots is a list of all available slots with the structure `{{"AM", <AM key>}, {"PM", <PM key>}, ...}`. These change based on the allowed packing slots as set by the admin. Strangely enough the only information that is retained when retrieving the user input from the dropdown is the key, which is stored in `slot`. This algorithm is just a quick and dirty way to get the corresponding string value back. 

The code where the information on the string (i.e. "AM") is lost is in components > DataInput > DropdownListInput.tsx where in the MenuItem component `label` is not stored. It might be possible to pass both `label` and `value` into the component (I will have to check) - this would require a change on everything that uses dropdown lists. Everything else seems to be compatible which is why I just used the above algorithm, but let me know if you want me to find a solution that is more robust :)

src/app/parcels/ActionBar/ActionAndStatusBar.tsx
KylePUG: I'm not too sure regarding this - this came from a merge conflict - from what I can see work from #180  ?

src/app/parcels/ActionBar/Statuses.tsx
KylePUG: Really I just want `clientIds[index]`. But since `clientIds` is `string[] | undefined`, I needed the `?`. I don't think that typescript likes the `[index]` right after the `?` so `.slice()` helps with typescript understanding what I want. Is there a better way to do this?

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
KylePUG: This resulted from merging with #180 as well I think, but yes sure I can get on this!

src/app/parcels/ActionBar/ActionModals/CommonDateAndSlot.tsx
8kristy: Ah I see it's an existing thing used elsewhere, thanks!

src/app/parcels/ActionBar/ActionModals/DateChangeModal.tsx
KylePUG: It does, I'll try to implement it with the error message abt "x updates failed"

src/app/parcels/ActionBar/ActionModals/DateChangeModal.tsx
KylePUG: Will get on this

src/app/parcels/ActionBar/ActionModals/SlotChangeModal.tsx
8kristy: Guess if it requires changing dropdowns overall we can leave it 😄 A nicer way might be
```
let newPackingSlotText: string = packingSlots.find(packingSlot => packingSlot [1] === slot)[0];
```

src/app/parcels/ActionBar/ActionAndStatusBar.tsx
8kristy: Yeah think it's meant to stay there

src/app/parcels/ActionBar/Statuses.tsx
8kristy: You can use `clientIds?.at(index)` I think

8kristy (approved):

src/app/parcels/ActionBar/ActionModals/DriverOverviewModal.tsx
KylePUG: on second look this field is for `action` - this is not needed for driver overview :)

Issue #233, opened 2024-04-04
Parcels Collection Time Slot - Change from Free Text to Admin chosen time slots for each centre


dan-langridge:
### Describe the solution / feature you'd like

Currently, when adding a Parcel for a collection, the Collection Time Slot is Free Text 

Change this to:
- On the admin page > collection centres > for each centre, allow the admin to decide some time slots when the centre is open:
- e.g. VXL = 10:00, 10:15, 10:30, 10:45, 11:00.....13:45
-Then, when a Volunteer Adds new parcel for collection, when they choose a collection time, show a drop down list with the pre-chosen time slots

minalui:
Dev notes:
- Added time_slot column to collection centres that is of type collection_timeslot_type[]
- Collection_timeslot_type : {time: time without time zone; is_active: boolean}
- Added a time slots column to collection centres table in the admin panel with a button that opens the collection slots modal
- Modal opens to show a list of checkboxes with corresponding times
- Checkboxes can be checked or unchecked to demonstrate opening hours (checked = open, unchecked = closed)
- Save button to save changes to db
- Selected collection datetime displayed in parcels modal
- Available collection slot displayed when collection centre is selected

minalui:
QA

As an admin user
- [x] I should see a collection centres table in the admin panel with a new column collection slots.
- [x] When I click on edit collection slots, it opens a modal to allow me to select and deselect when the collection centre is open
- [x] When I click save, my changes are saved into collection centre table in the database
Ticket raised for generating two audit logs [https://softwiretech.atlassian.net/browse/VFB-254](https://softwiretech.atlassian.net/browse/VFB-254)
- [x] If I create a new collection centre, the edit collection centre button is disabled until I click save

As a regular user
- [x] I can see "COLLECTION DATE TIME" in the parcels modal
- [x] When I edit/add a parcel, I can see the collection slots available for each collection centre (please edit some time slots to check)
- [x] If a collection centre does not have any collection slots set then an empty dropdown appears. Ticket raised for error handling [https://softwiretech.atlassian.net/jira/software/c/projects/VFB/boards/862/backlog?selectedIssue=VFB-247](https://softwiretech.atlassian.net/jira/software/c/projects/VFB/boards/862/backlog?selectedIssue=VFB-247)
- [x] If I save a parcel, the changes to collection slots are reflected in the parcels modal

KylePUG:
QA check:

[...]

PR #367, opened 2024-05-30
#233 parcels collection time slot

minalui:
## What's changed
- Added time_slot column to collection centres that is of type collection_timeslot_type[] 
- Collection_timeslot_type : {time: time without time zone; is_active: boolean}
- Added a time slots column to collection centres table in the admin panel with a button that opens the collection slots modal
- Modal opens to show a list of checkboxes with corresponding times
- Checkboxes can be checked or unchecked to demonstrate opening hours (checked = open, unchecked = closed)
- Save button to save changes to db
- Selected collection datetime displayed in parcels modal
- Available collection slot displayed when collection centre is selected

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image
image
image
image
 | 
image
image
image
image
image
image
 |

## Checklist

  - N/A I have modified the seed in `supabase/seed/seed.mts` if appropriate
  - N/A If I have modified the seed, I have also generated the seed with `npm run db:generate_seed` 
    - Checkout the dev branch
    - Checkout your branch


src/app/admin/collectionCentresTable/CollectionCentreActions.ts
```diff
@@ -17,7 +23,27 @@ type FetchCollectionCentresResult =
           error: { type: "failedToFetchCollectionCentres"; logId: string };
       };
 
-export const fetchCollectionCentres = async (): Promise<FetchCollectionCentresResult> => {
+export const defaultCollectionTimeSlots: Schema["collection_centres"]["time_slots"] = [
```
minalui: This is not needed anymore as default is set in the db

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -58,16 +78,50 @@ function getBaseAuditLogForCollectionCentreAction(
...
 }
 
+function getBaseAuditLogForCollectionCentreTimeSlots(
+    action: string,
+    timeSlotsWithPrimaryKey: FormattedTimeSlotsWithPrimaryKey
+): Pick<AuditLog, "action" | "content" | "collectionCentreId"> {
+    return {
+        action,
+        content: {
+            timeSlots: "Hello",
```
minalui: This needs to be json of the timeslots

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -58,16 +78,50 @@ function getBaseAuditLogForCollectionCentreAction(
...
+    };
+}
+
+const formatCollectionCentreTimeSlotDbData = (
+    row: CollectionCentresTableRow
+): FormattedTimeSlotsWithPrimaryKey => {
+    const formattedTimeSlots = row.timeSlots.map((timeSlot) => {
+        return {
+            time: formatTimeStringToHoursAndMinutes(timeSlot.time !== null ? timeSlot.time : ""),
+            isActive: timeSlot.is_active !== null ? timeSlot.is_active : true,
```
minalui: Not sure if this is the right approach - technically they shouldn't be null

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -80,6 +134,10 @@ const CollectionCentresTable: React.FC = () => {
         void getCollectionCentresForTable();
     }, [getCollectionCentresForTable]);
 
+    useEffect(() => {
+        console.log(timeSlotModalData);
```
minalui: Don't need this console.log anymore

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -99,6 +157,63 @@ const CollectionCentresTable: React.FC = () => {
...
+            void sendAuditLog({
+                ...baseAuditLog,
+                wasSuccess: false,
+                logId: updateTimeSlotError.logId,
+            });
+        }
+
+        void sendAuditLog({ ...baseAuditLog, wasSuccess: true });
+        setTimeSlotModalIsOpen(false);
+        getCollectionCentresForTable();
```
minalui: Is this needed?

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -174,7 +289,10 @@ const CollectionCentresTable: React.FC = () => {
 
         if (newRow.isNew) {
             const { data: newCollectionCentreData, error: newCollectionCentreError } =
-                await addNewCollectionCentre(newRow);
+                await addNewCollectionCentre({
+                    ...newRow,
+                    timeSlots: defaultCollectionTimeSlots,
```
minalui: Not needed anymore as default set in db

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -324,6 +467,49 @@ const CollectionCentresTable: React.FC = () => {
...
+                        </>
+                    }
+                    isOpen={timeSlotModalIsOpen}
+                    onClose={() => {
+                        setTimeSlotModalIsOpen(false);
+                    }}
+                    headerId="expandedCollectionCentreTimeSlotsModal"
+                    footer={
+                        <Centerer>
+                            <Button onClick={handleModalSaveClick}>Save</Button>
```
minalui: Make this button green

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -120,7 +127,8 @@ const noCollectionFormSections = [
 
 const mergeDateAndTime = (date: string, time: string): Dayjs => {
     // dayjs objects are immutable so the setter methods return a new object
-    const dayjsTime = dayjs(time);
+    dayjs.extend(customParseFormat);
+    const dayjsTime = dayjs(time, "HH:mm");
```
minalui: This customParseFormat allows time in HH:mm to be parsed

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -267,6 +297,7 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
                             fields={fields}
                             collectionCentresLabelsAndValues={collectionCentresLabelsAndValues}
                             packingSlotsLabelsAndValues={packingSlotsLabelsAndValues}
+                            collectionTimeSlotsLabelsAndValues={collectionSlotsLabelsAndValues}
```
minalui, line 306: Add a drop down value of "Please select collection centre to see available times" etc. if collectionSlotLabelsAndValues are not available

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -126,6 +126,7 @@ const getExpandedParcelDetails = async (
                     method: rawParcelDetails.collection_centre?.is_shown
                         ? rawParcelDetails.collection_centre?.name
                         : `${rawParcelDetails.collection_centre?.name} (inactive)`,
+                    collectionDateTime: formatDateTime(rawParcelDetails.collection_datetime),
```
minalui, line 130: What if rawParcelDetails.collection_datetime is null for example in delivery?

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -150,6 +151,7 @@ const getExpandedParcelDetails = async (
                 method: rawParcelDetails.collection_centre?.is_shown
                     ? rawParcelDetails.collection_centre?.name
                     : `${rawParcelDetails.collection_centre?.name} (inactive)`,
+                collectionDateTime: rawParcelDetails.collection_datetime,
```
minalui, line 155: What if rawParcelDetails.collection_datetime was null?

src/common/fetch.ts
```diff
@@ -134,6 +135,54 @@ export const getActiveCollectionCentres = async (
...
+        .select("time_slots")
+        .eq("primary_key", collectionCentrePrimaryKey)
+        .single();
+
+    if (error || !data) {
+        const logId = await logErrorReturnLogId(
+            "Error with fetch: Collection time slots data",
+            error
+        );
+        return { data: null, error: { type: "collectionTimeSlotsFetchFailed", logId: logId } };
```
minalui, line 177: Maybe no data should have a different error type

src/common/format.ts
```diff
@@ -86,3 +86,7 @@ export const getParcelOverviewString = (
         (collectionDatetime && ` @ ${dayjs(collectionDatetime).format("DD/MM/YYYY HH:mm")}`)
     );
 };
+
+export const formatTimeStringToHoursAndMinutes = (timeString: string): string => {
```
minalui, line 101: This is to change 10:00:00 into 10:00

supabase/seed/collectionCentresSeed.mts
```diff
@@ -15,3 +15,4 @@ export const collectionCentres = [
         isDelivery: true,
     },
 ];
+
```
minalui: Line added after lint_fix

minalui (commented):
Self-review

src/app/parcels/getExpandedParcelDetails.ts
minalui: It will display as COLLECTION DATE TIME -

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
8kristy: Guess it's nice to have safety, although I'd set it to false if it did happen to be null

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -99,6 +155,62 @@ const CollectionCentresTable: React.FC = () => {
...
+                `Failed to update the collection centre time slots. Log ID: ${updateTimeSlotError.logId}`
+            );
+            void sendAuditLog({
+                ...baseAuditLog,
+                wasSuccess: false,
+                logId: updateTimeSlotError.logId,
+            });
+        }
+
+        void sendAuditLog({ ...baseAuditLog, wasSuccess: true });
```
8kristy: Does this generate additional audit logs? I think this is the problem Amelia was having - she changed `void` to `await` and it fixed it. May be worth updating these as well?

src/common/format.ts
8kristy: Feels a bit dodgy to do it like that, especially if someone changes the format. Maybe convert to an actual date and then format the time? 
```
const time = new Date(timeString);
return `${time.getHours()}:${time.getMinutes()}`
```

src/app/parcels/form/formSections/CollectionTimeCard.tsx
```diff
@@ -1,35 +1,40 @@
...
+    collectionTimeSlotsLabelsAndValues,
 }) => {
     return (
         <GenericFormCard
-            title="Collection Time"
+            title="Collection Slots"
             required={true}
-            text="What time is the client collecting their parcel?"
+            text="What slot is the client collecting their parcel?"
         >
```
8kristy: I think it's ok to leave it as time since that's what users see it as, just in a dropdown instead of free form

supabase/migrations/20240530143601_add_column_time_slot_collection_centre_table.sql
```diff
@@ -0,0 +1,23 @@
...
+     ROW('11:45:00', TRUE)::collection_timeslot_type,
+     ROW('12:00:00', TRUE)::collection_timeslot_type,
+     ROW('12:15:00', TRUE)::collection_timeslot_type,
+     ROW('12:30:00', TRUE)::collection_timeslot_type,
+     ROW('12:45:00', TRUE)::collection_timeslot_type,
+     ROW('13:00:00', TRUE)::collection_timeslot_type,
+     ROW('13:15:00', TRUE)::collection_timeslot_type,
+     ROW('13:30:00', TRUE)::collection_timeslot_type,
+     ROW('13:45:00', TRUE)::collection_timeslot_type,
+     ROW('14:00:00', TRUE)::collection_timeslot_type
```
8kristy: I think the default should just be empty and the admin can add new fields from the interface

I don't see functionality to add new timeslots, only hide/show them? Could you please add a "add timeslot" button to the edit timeslots modal and change its title

Would it be difficult to seed a few timeslots for dev environments?

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -159,6 +167,13 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
     const [submitDisabled, setSubmitDisabled] = useState(false);
     const [modalIsOpen, setModalIsOpen] = useState<boolean>(false);
     const [clientDetails, setClientDetails] = useState<ExpandedClientData | null>(null);
+    const [collectionSlotsLabelsAndValues, setCollectionSlotsLabelsAndValues] =
+        useState<CollectionTimeSlotsLabelsAndValues>([
+            [
+                "Please select a collection centre for available collection slots",
+                "Collection centre not selected",
```
8kristy: Is this meant to be displayed if you open collection slots before selecting a centre? It just shows me a blank row (which I think I prefer - wouldn't want users selecting an error message. Could we put this as a validation error maybe?)

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -99,6 +155,62 @@ const CollectionCentresTable: React.FC = () => {
...
+                    time: timeSlot.time,
+                    isActive: (timeSlot.isActive = !timeSlot.isActive),
+                };
+            } else {
+                return {
+                    time: timeSlot.time,
+                    isActive: timeSlot.isActive,
+                };
+            }
+        });
```
8kristy: I think it might be more efficient to find the correct entry and update the value of that in place rather than updating the whole list?
```
const updatedTime = event.currentTarget.parentElement?.parentElement?.innerText;
let timeslot = timeSlotModalData.timeSlots.find(t => t.time === updatedTime);
timeslot.isActive = !timeslot.isActive;
```

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -250,6 +364,31 @@ const CollectionCentresTable: React.FC = () => {
             editable: true,
             renderHeader: (params) => <Header {...params} />,
         },
+        {
+            field: "collectionSlot",
+            type: "actions",
+            headerName: "Collection Slots",
+            flex: 1,
+            renderHeader: (params) => <Header {...params} />,
+            renderCell: (params) => {
```
8kristy, line 437: Just trying to understand the code - do you know where it gets these params from and how does it know the row primary key?

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -324,6 +463,51 @@ const CollectionCentresTable: React.FC = () => {
                     hideFooter
                 />
             )}
+            {timeSlotModalIsOpen && (
+                <Modal
+                    header={
+                        <>
+                            <Icon icon={faShoePrints} color={theme.primary.largeForeground[2]} />{" "}
+                            Select Collection Centre Time Slots
```
8kristy: Update header to "Edit Collection Centre Time Slots"

src/app/parcels/form/formSections/CollectionTimeCard.tsx
```diff
@@ -1,35 +1,40 @@
...
-                    onChange={(value) => {
-                        onChangeDateOrTime(fieldSetter, errorSetter, "collectionTime", value);
-                    }}
-                    label="Time"
-                    value={fields.collectionTime ? dayjs(fields.collectionTime) : null}
+                <DropdownListInput
+                    selectLabelId="collection-slot-select-label"
+                    labelsAndValues={collectionTimeSlotsLabelsAndValues}
+                    listTitle="Collection Slot"
+                    defaultValue={fields.collectionSlot ? fields.collectionSlot : ""}
```
8kristy: Is this the bit that sets the slot value when you edit the parcel? Right now if I edit a parcel to select a slot, save, edit again and don't do anything the collection slot is blank. If I press save it changes it to 20:04:50 or 20:04:20 on different parcels - don't really know where it's getting that time from? Maybe the old value from the time picker? Could you fix it to retrieve the correct value please

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -174,6 +189,47 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
+                if (data.length === 0) {
+                    setFormErrors({
+                        ...initialFormErrors,
+                        collectionSlot: Errors.noCollectionSlotsSet,
+                    });
+                } else {
+                    setFormErrors({
+                        ...initialFormErrors,
+                        collectionSlot: Errors.initial,
+                    });
```
8kristy: Does this one show up if you press submit without selecting a timeslot?

8kristy (commented):
I think we need functionality to add slots from the admin UI, without touching the database so please add a button for that

Not sure if the default slot picker behaviour is correct?

And very small thing but noticed there's a double scroll bar on smaller screens on the modal 
image

Otherwise looks nice - I like the edit time slots button (I imagined a list of them just right there which would be very clunky) and besides some bugs the actual time slot dropdown is good

src/common/format.ts
minalui: I think because timeString is HH:mm:ss it becomes invalid date when you put it in new Date(timeString).
I have now used dayjs instead but I need to pass in a 0 otherwise it becomes a single digit.

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
minalui: It's not happened here but I will change them to awaits

supabase/migrations/20240530143601_add_column_time_slot_collection_centre_table.sql
minalui: I can set default to empty but won't be able to seed them because there's a known bug with snaplet with arrays. Will add the buttons now.

src/app/parcels/form/ParcelForm.tsx
minalui: Yes - just thought I'd cover all bases 🥲 The form currently does check that it has a value but only if it's previously null eg. if they already had a time for collection but not matching the slots it would still save. Should I edit this to remove the previous collection time so they have to select a new time?

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
minalui: Not 100% but my understanding is that the rows get passed into DataGrid and when an action button is clicked on that row it passed which row on to the params. Row is defined as type CollectionCentresTableRow and the primary key is params.row.id.

src/app/parcels/form/ParcelForm.tsx
minalui: Yes, if there are no collection slots for the particular collection centre then an error displays: "There are no collection slots set for this collection centre, please select a different collection centre or contact admin." on the card but once a collection centre with time slot data is selected then it resets the error to "required".

src/app/parcels/form/formSections/CollectionTimeCard.tsx
minalui: I figured it was because when the parcel info gets passed down the time is not formatted correctly, but sorted now :)

src/common/format.ts
8kristy: Should we pad the hour as well?

```suggestion
    const hours = String(dayjsTime.hour()).padStart(2, "0");
    const minutes = String(dayjsTime.minute()).padStart(2, "0");
    return `${hours}:${minutes}`;
```

or

```
const padNumberWithLeadingZeros = (number: number, places: number) : String => {
    return String(number).padStart(places, "0");
}

export const formatTimeStringToHoursAndMinutes = (timeString: string): string => {
    dayjs.extend(customParseFormat);
    const dayjsTime = dayjs(timeString, "HH:mm:ss");
    return `${padNumberWithLeadingZeros(dayjsTime.hour())}:${padNumberWithLeadingZeros(dayjsTime.minute())}`;
};
```
mix and match with what you think is the nicest outcome

src/app/parcels/form/ParcelForm.tsx
8kristy: I guess this is something that will only happen in our dev environments right? Cause the live env is going to only have slots
I don't think we should remove the previous time in case they don't want to edit it, but it would be nice to see it somewhere explicitly - either put into the dropdown as a value for that specific parcel or just some text saying "previously selected time hh:mm" above the dropdown
And then if we picked an actual slot it should show that slot preselected

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -99,6 +173,111 @@ const CollectionCentresTable: React.FC = () => {
...
+    const handleTimeSlotCheckBoxChange = (event: React.SyntheticEvent<Element, Event>): void => {
+        if (!timeSlotModalData) {
+            return;
+        }
+
+        const updatedTime = event.currentTarget.parentElement?.parentElement?.innerText;
+        const timeSlotIndex = timeSlotModalData.timeSlots.findIndex(
+            (slot) => slot.time === updatedTime
+        );
+        const timeSlot = timeSlotModalData.timeSlots.find((slot) => slot.time === updatedTime);
```
8kristy: I think if we have index we can just use that instead of searching again
```suggestion
        const timeSlot = timeSlotModalData.timeSlots[timeSlotIndex]
```

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -99,6 +173,111 @@ const CollectionCentresTable: React.FC = () => {
...
+        }
+
+        timeSlot.isActive = !timeSlot.isActive;
+
+        addNewTimeSlotToTimeSlotModalData(timeSlotModalData, timeSlot);
+        const newTimeSlotsArray = [
+            ...timeSlotModalData.timeSlots.slice(0, timeSlotIndex),
+            timeSlot,
+            ...timeSlotModalData.timeSlots.slice(timeSlotIndex + 1),
+        ];
```
8kristy: Is this to insert it in the right place? Would it be nicer to just append it and then sort? (assuming default time sorting is sensible)

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -99,6 +173,111 @@ const CollectionCentresTable: React.FC = () => {
...
+            (slot) => slot.time === updatedTime
+        );
+        const timeSlot = timeSlotModalData.timeSlots.find((slot) => slot.time === updatedTime);
+        if (!timeSlot) {
+            return;
+        }
+
+        timeSlot.isActive = !timeSlot.isActive;
+
+        addNewTimeSlotToTimeSlotModalData(timeSlotModalData, timeSlot);
```
8kristy: Why do we have add new slot logic in here? Is there something not immediately obvious that prevents you from just using `setTimeSlotModalData(timeSlotModalData.timeSlots);` after toggling the timeslot?

src/app/parcels/form/formSections/CollectionSlotCard.tsx
```diff
@@ -0,0 +1,40 @@
...
+    errorSetter,
+    fieldSetter,
+    formErrors,
+    fields,
+    collectionTimeSlotsLabelsAndValues,
+}) => {
+    return (
+        <GenericFormCard
+            title="Collection Slots"
+            required={true}
```
8kristy, line 23: Seems like when you edit a parcel, if you don't update the value it's filled in but won't allow you to submit because this field is required? Works fine if I change it to another time though

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
8kristy: Just going over the commits, sounds like it made duplicate rows? Do you know why it does that?

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -99,6 +173,111 @@ const CollectionCentresTable: React.FC = () => {
...
+            ...existingTimeSlotData,
+            timeSlots: newTimeSlotArray,
+        };
+
+        setTimeSlotModalData(updatedTimeSlotModalData);
+    };
+
+    const handleSaveSlotClick = async (): Promise<void> => {
+        setTimeSlotModalErrorMessage(null);
+        if (timeSlotModalData === null || collectionTimeSlotValue === undefined) {
```
8kristy, line 234: I think it may be initially not setting the value? i.e. if I press add, it by default fills it up with the current time, however nothing happens if I press save. If I change the value myself to the time (i.e. interact with the field and update values) then it works

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -99,6 +173,111 @@ const CollectionCentresTable: React.FC = () => {
...
+            return;
+        }
+        const newTimeSlot: FormattedTimeSlot = {
+            time: formatDayjsToHoursAndMinutes(collectionTimeSlotValue),
+            isActive: false,
```
8kristy: I think they should be active by default when you add them

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -99,6 +173,111 @@ const CollectionCentresTable: React.FC = () => {
...
+        };
+
+        if (checkIfSlotExists(timeSlotModalData, newTimeSlot)) {
+            setTimeSlotModalErrorMessage(
+                "This time slot already exists. Please select a different time."
+            );
```
8kristy: Can we make a separate error for this and put it under the edit slot instead of the main div?

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -174,6 +184,44 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
+                if (error) {
+                    let errorMessage;
+                    switch (error.type) {
+                        case "collectionTimeSlotsFetchFailed":
+                            errorMessage = "Failed to fetch collection time slots";
+                            break;
+                    }
+                    setSubmitErrorMessage(`${errorMessage}. Log ID: ${error.logId}`);
+                    return;
+                }
```
8kristy, line 204: Don't know if you can see it but github seems to be complaining: `Type '"noCollectionTimeSlotsData"' is not comparable to type '"collectionTimeSlotsFetchFailed"'.`
Not sure what's wrong with it but thought I'll bring attention to it

src/common/format.ts
```diff
@@ -39,6 +40,18 @@ export const formatDatetimeAsDate = (datetime: Date | string | null): string =>
...
+    if (datetime instanceof Date) {
+        return datetime.toLocaleDateString("en-GB");
+    }
+
+    if (datetime === null || isNaN(Date.parse(datetime))) {
+        return "-";
+    }
+
+    return new Date(datetime).toLocaleTimeString("en-GB");
+};
```
8kristy: If it happens to be Date, does it break if you try to convert? i.e. just have 
```
if (datetime === null || isNaN(Date.parse(datetime))) {
    return "-";
}

return new Date(datetime).toLocaleTimeString("en-GB");
```

8kristy (commented):
A few little bugs and suggestions; main question is about toggling the state I think

Code looks good!

src/app/parcels/form/ParcelForm.tsx
minalui: I've removed "noCollectionTimeSlotsData" from the return type getActiveTimeSlotsForCollectionCentre but forgot to remove it from the switch - it's passing now :)

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
minalui: I've had a go but I think we still need to slice out the original element in the array before appending otherwise it would be duplicated.

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
minalui: Sorry that is my mistake. It doesn't need that line. Originally I used the same function but I realised if I used the same function it added the new slot onto the array and caused a duplicate

src/app/parcels/form/ParcelForm.tsx
minalui: I'm not sure there is a good way to do this and make collection slots required.

When we added the packing slots - if the original packing slot was not available - there would be an error saying the original packing slot is no longer available, please select a different slot. If they didn't want to change it then they could leave the form.

I feel that either:
We keep it similar to packing slots to show an error message and tell them to select an available slot
or
Make collection slots not required?

src/app/parcels/form/formSections/CollectionSlotCard.tsx
minalui: Please see comment above re showing previous selected time

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
minalui: I've setCollectionTimeSlotValue(dayjs(collectionTimeSlotValue)) when the add button is clicked so that collectionTimeSlotValue state has a value.

src/common/format.ts
minalui: I think Date.parse doesn't like it but I've removed Date as a parameter as its only used in the format parcel details bit

src/app/parcels/form/ParcelForm.tsx
minalui: Currently the dropdown is set when a collection centre is selected.
If a collection centre is selected, I don't think its a good idea to put it in the drop down (feels a bit buggy to me)
We can put it as text in the card but they won't be able to save it because the dropdown is expecting a value unless we make it not required.

8kristy: I think matching what packing slot does is the best idea for consistency

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
8kristy: I gave it a try and this seems to work without creating duplicates
```
 timeSlot.isActive = !timeSlot.isActive;

        const updatedTimeSlotData: FormattedTimeSlotsWithPrimaryKey = {
            ...timeSlotModalData,
            timeSlots: timeSlotModalData.timeSlots,
        };

        setTimeSlotModalData(updatedTimeSlotData);
```
anything that looks like it could break?
I don't really know why we need `updatedTimeSlotData` given that we're overriding the timeSlots data in there with the same data that's currently in `timeSlotModalData` since we updated in place - that way if you click checkbox nothing happens. Probably some react state magic?

minalui: Thanks I got really confused too 🙈

8kristy (approved):

Issue #356, opened 2024-05-22
Trend/statistics reporting for clients and parcels


dan-langridge:
### Describe the solution / feature you'd like

Every week, I create a report for the FoodBank managers on the number of parcels we had the following week, the family sizes and also the number of pets. See spreadsheet attached.

Note that these parcels are only ones that have successfully been packed and delivered. So parcels that have the status Called and No response, Parcel Denied etc are omitted from these stats.

If the app could have a new page called Reports where this table exists that would be great. Only viewable by Admins/Managers/Staff. Or a way to create the last weeks stats which I copy and paste into the google spreadsheet. Or worse case (which is what I currently do), a way to download the raw data from the database and I extract the data myself)

### Additional context

<img width="1208" alt="Screenshot 2024-05-22 at 16 48 08" src="https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/80202270/00517e1b-7723-495f-a984-dcfad2017a69">

Alex-Softwire:
Changelog: 

- Have added a reports table available to both admin & staff. 
- It currently counts parcels that have at some point been assigned the status "Parcel Completed", "Fulfilled With Trussel Trust" and "Delivered" - it shouldn't double count parcels, if the status' have been assigned twice. 
- It will group completed parcels by week, using the timestamp of the time that they were completed. 
- It should show everything from the first completed parcel to now.
- Deleted (but still delivered) parcels will be included
- Deleted clients are currently included in the total but not any of the family size or pet totals (as that information is no longer known) - By saving report data, this may be fixed in a subsequent ticket. 


Although a download to csv file button may be added later - for now one can do the following on the supabase console to download the raw data:
![image](https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/assets/107864173/0fb61241-26c0-492c-b67c-eec49ff57c2e)

Alex-Softwire:
QA: 

- Check everything in the change-log is valid 
This could include checking:
- Updating status' of parcels to see if they appear in the table. See whether the pet categories are calculated correctly.
- Testing whether the table is accessible to volunteers (it should not be).
- Testing the table layout on small browser sizes (though a future styling ticket may come later to fix this).
- Adding a completed parcel in the past (e.g 2022) to see if it will include data from it.

kaydentsoftwire:
The table looks good and is not viewable as a volunteer 👍  but I think there may be a bug in the code somewhere here
[...]

PR #368, opened 2024-05-30
#356: Add reports page and table

Alex-Softwire:
## What's changed
- Added reports view to the backend
- Added reports table to display that view

## Screenshots / Videos
image
image

## Checklist

    - Checkout the dev branch
    - Checkout your branch


supabase/migrations/20240528091033_add_completed_parcels_view.sql
```diff
@@ -0,0 +1,30 @@
+CREATE OR REPLACE VIEW completed_parcels AS
```
Alex-Softwire: This sql is a little bit gnarly and might need some explanation - any suggestions as to what to do here, or is it fine?

supabase/migrations/20240528091033_add_completed_parcels_view.sql
```diff
@@ -0,0 +1,30 @@
...
+WITH completed_events AS (
+    SELECT
+        events.parcel_id,
```
Alex-Softwire: The following parcels are counted ->
Parcels that have (at some point) had the event "Delivered", "Parcel Collected", "Fulfilled with Trussell Trust" AND the most recent event is NOT "Parcel Deleted" - they are dated in the reports table by the timestamp of the most recent time that they have been set to "Delivered", "Parcel Collected" or "Fulfilled with Trussell Trust".

Alex-Softwire: Dating by the time that they are delivered/collected makes the most sense to me as we only care about delivered/collected parcels, but I might be wrong on that assumption.

supabase/migrations/20240529091033_add_reports_view.sql
```diff
@@ -0,0 +1,38 @@
+CREATE OR REPLACE VIEW reports AS
+WITH first_completed_parcel AS (
```
Alex-Softwire: We include every week from the first completed parcel to now.

src/app/reports/reportsTable/ReportsPage.tsx
```diff
@@ -0,0 +1,130 @@
+"use client";
```
Alex-Softwire, line 1: Most of this code is based off the way ClientsPage was implemented. Same with the getReportsData.

src/app/reports/reportsTable/getReportsData.ts
```diff
@@ -0,0 +1,103 @@
...
+    let query = supabase.from("reports").select("*");
+
+    if (sortState.sortEnabled && sortState.column.sortMethod) {
+        query = sortState.column.sortMethod(query, sortState.sortDirection);
+    } else {
+        query.order("week_commencing", { ascending: false });
+    }
+
+    query = query.range(startIndex, endIndex);
+    query = query.abortSignal(abortSignal);
```
8kristy: Why do we immediately replace it? Is the first line redundant or am I just missing some inner workings?

src/app/reports/reportsTable/getReportsData.ts
```diff
@@ -0,0 +1,103 @@
...
+        const logId = await logErrorReturnLogId("Error with fetch: reports table", {
+            error: reportsError,
+        });
+        return { error: { type: "failedToFetchReportsTable", logId }, data: null };
+    }
+
+    const reportsData: ReportsTableRow[] = reports.map((report) => {
+        return {
+            weekCommencing: report.week_commencing ?? "",
+            total: report.total_parcels ?? 0,
```
8kristy: May be worth moving the totals to the end to make it consistent

src/app/reports/reportsTable/getReportsData.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const getReportsCount = async (
+    supabase: Supabase,
+    abortSignal: AbortSignal
+): Promise<GetReportsCountReturnType> => {
+    let query = supabase.from("reports").select("*", { count: "exact", head: true });
+
+    query = query.abortSignal(abortSignal);
+
+    const { count, error: reportsError } = await query;
```
8kristy, line 84: Should count have type/field name?

src/app/reports/reportsTable/types.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export type GetReportsCountReturnType =
+    | {
+          error: { type: GetReportsDataAndCountErrorType; logId: string };
+          data: null;
+      }
+    | {
+          error: null;
+          data: number;
+      };
```
8kristy: Could we order the data in these the same, e.g.
```
{
    data: <data>;
    error: null;
}
| {
    data: null;
    error: <error>
}
```
maybe check what's done elsewhere and if it's not consistent decide on the best one here just so that they're the same

src/app/reports/reportsTable/types.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export interface ReportsTableRow {
+    weekCommencing: string;
+    total: number;
```
8kristy: Move totals to the right place for consistency

src/app/roles.tsx
```diff
@@ -22,7 +22,7 @@ const pathsShownToAllAuthenticatedUsers = [
     "/set-password",
 ] as const;
```
8kristy, line 24: I think volunteers are not meant to see it - only admin, managers and staff - so think it needs to be removed from this list

supabase/migrations/20240530291033_add_completed_parcels_view.sql
```diff
@@ -0,0 +1,30 @@
+CREATE OR REPLACE VIEW completed_parcels AS
+WITH completed_events AS (
+    SELECT
+        events.parcel_id,
+        MAX(events.timestamp) AS completed_timestamp
```
8kristy: Why is it `MAX(events.timestamp)`?

supabase/migrations/20240530291033_add_completed_parcels_view.sql
```diff
@@ -0,0 +1,30 @@
...
+        events
+    WHERE
+        new_parcel_status IN (
+            SELECT
+                event_name
+            FROM
+                status_order
+            WHERE
+                is_successfully_completed_event
+        )
```
8kristy: Think you can remove the new `is_successfully_completed_event` column and just use 
```
WHERE new_parcel_status LIKE 'Parcel Collected' or new_parcel_status LIKE 'Fulfilled with Trussell Trust' or new_parcel_status like 'Delivered'
```

supabase/migrations/20240530291033_add_completed_parcels_view.sql
```diff
@@ -0,0 +1,30 @@
...
+    completed_events.parcel_id,
+    completed_timestamp,
+    family_count,
+    pet_food
+FROM
+    completed_events
+        INNER JOIN
+    parcels_plus ON parcels_plus.parcel_id = completed_events.parcel_id AND NOT parcels_plus.last_status_event_name = 'Parcel Deleted'
+        LEFT JOIN
+    clients ON parcels_plus.client_id = clients.primary_key
```
8kristy: Simplest SQL I can come up with for this is 
```
SELECT events.parcel_id, events.timestamp AS completed_timestamp, family_count, pet_food
FROM events 
INNER JOIN parcels_plus ON events.parcel_id = parcels_plus.parcel_id
LEFT JOIN clients ON parcels_plus.client_id = clients.primary_key
WHERE new_parcel_status LIKE 'Parcel Collected' or new_parcel_status LIKE 'Fulfilled with Trussell Trust' or new_parcel_status like 'Delivered'
```

but I don't know if it works in views and also don't know if it returns useful data at all

supabase/migrations/20240530391033_add_reports_view.sql
```diff
@@ -0,0 +1,38 @@
...
+),
+list_of_weeks AS (
+ SELECT generate_series(
+        0,
+        52 * (EXTRACT(YEAR FROM CURRENT_DATE) - EXTRACT(YEAR FROM start_date)) + EXTRACT(WEEK FROM CURRENT_DATE) - EXTRACT(WEEK FROM start_date))
+     AS number_of_weeks_ago
+ FROM first_completed_parcel
+)
+SELECT
+    TO_CHAR(date_trunc('week', CURRENT_DATE) - (number_of_weeks_ago * INTERVAL '1 week'), 'YYYY-MM-DD') as week_commencing,
```
8kristy: May be worth adding a comment on what this is doing

8kristy (commented):
Nice work, I'll look into simplifying and verifying the SQL a bit more but well done on the page and getting the data overall!

src/app/roles.tsx
Alex-Softwire: Hmm it has been  -> though I did that in one of my later commits whoops so might have been post review? Thanks for spotting though :)

src/app/reports/reportsTable/types.ts
Alex-Softwire: Yep, done! Will also fix this in clients equivalent as that is what this is based off :)

Alex-Softwire: Will have a check to see if it's in even more places.

src/app/reports/reportsTable/types.ts
Alex-Softwire: Whoops, relic of the past, thx for spotting :)

supabase/migrations/20240530291033_add_completed_parcels_view.sql
Alex-Softwire: Yep I definitely can. I had it like that before, though spoke with @AmeliaIsCoding about this and she made the point that having it as a separate column means that we can more easily edit what status' count as completed. E.g if they start working with a new trust, it's less likely to be missed off the list and unnoticed. Am happy to switch it back though if you disagree with this :)

If so - would `WHERE new_parcel_status IN ('Parcel Collected', 'Fulfilled with Trussel Trust', 'Delivered')` be a good way of writing that? Thanks!

src/app/reports/reportsTable/getReportsData.ts
Alex-Softwire: Will do!! Thanks of spotting

supabase/migrations/20240530291033_add_completed_parcels_view.sql
Alex-Softwire: Will have a test of this :)

src/app/reports/reportsTable/getReportsData.ts
Alex-Softwire: Hmm... am not sure what's going on here, it's like that on the other files. It's definitely not fully replacing it as it's doing query = query.method(), not sure if that's any different to doing query.method(). Will investigate.

src/app/reports/reportsTable/getReportsData.ts
Alex-Softwire: IntelliJ/Typescript seems to know it's number | null, presumably as it's from a count query. I could give it an alias reportsCount or something like that but not sure if that adds anything? Thanks :)

supabase/migrations/20240530291033_add_completed_parcels_view.sql
Alex-Softwire: Have just tested and it does work! Though there's a couple of edge cases that it doesn't cover, but am not sure how much that they matter:
1. Parcels will be double counted if they get set to multiple successful states, although this probably shouldn't happen, there's nothing to stop you in the website from setting delivered twice, delivered then collected. That's why I group_by parcel_id to ensure uniqueness, and set to MAX(timestamp) so the completion time is set to the last time the parcel is set to completed. 
2. Parcels will not be removed if they are deleted (why I added AND NOT parcels_plus.last_status_event_name = 'Parcel Deleted'.

Lmk if any of these edge cases actually matter cause agree that the above is a nice simplification in that case. If we keep the above query - should probably add a comment to explain why I've done it. Thanks!

supabase/migrations/20240530291033_add_completed_parcels_view.sql
Alex-Softwire: See below comment :) -> So we don't double count parcels if they are set to multiple "successful" states, there's nothing in the frontend stopping this happening and I would have though it may happen by accident a few times. MAX(timestamp) is just an aggregate function - if there are two successful states, we pick the latest.     Not sure how important this is though, happy to remove if the simplicity is more important!

src/app/reports/reportsTable/getReportsData.ts
Alex-Softwire: Had an investigation - doesn't look to be a difference between the two query = query.order() and doing query.order(). I think the latter is nicer as it's less confusing - will switch this and other cases to that if that makes sense.

supabase/migrations/20240530391033_add_reports_view.sql
Alex-Softwire: Done - added a few comments to explain the slightly gnarly sql here.

supabase/migrations/20240530291033_add_completed_parcels_view.sql
Alex-Softwire: Just noticed another edge case - Deleted Clients. Currently they count in the total but not any of the family sized (as that info has been removed), or any of the pets data (as that info has been removed). Is this how it should work?

src/app/reports/reportsTable/getReportsData.ts
8kristy: Think I like how it was done above, just a bit more consistent
```
const { data: reports, error: reportsError } = await query;
```
just change reports to count

src/app/reports/reportsTable/types.ts
```diff
@@ -0,0 +1,54 @@
...
+      };
+
+export type GetReportsCountReturnType =
+    | {
+          data: null;
+          error: { type: GetReportsDataAndCountErrorType; logId: string };
+      }
+    | {
+          data: number;
+          error: null;
```
8kristy:
```suggestion
        data: number;
        error: null;
      }
    | {
        data: null;
        error: { type: GetReportsDataAndCountErrorType; logId: string };
```

supabase/migrations/20240530291033_add_completed_parcels_view.sql
8kristy: Nope that makes sense let's leave, but if possible maybe add a comment explaining it

supabase/migrations/20240530291033_add_completed_parcels_view.sql
8kristy: Ah that's a pretty good point, let's leave it

src/app/reports/reportsTable/getReportsData.ts
Alex-Softwire: I'm a little confused -> do you mean to replace it with this:

`const { data: count, error: reportsError } = await query;`

I think this isn't equivalent to what I have as that's setting the data variable to count, rather than using count variable.

The query is `const query = supabase.from("reports").select("*", { count: "exact", head: true });` which returns data, count and error. Data would be all elements of the table, while count is the number of them - so they wouldnt' do the same thing. Lmk if I'm misunderstanding :)

supabase/migrations/20240530291033_add_completed_parcels_view.sql
8kristy: The double counting does matter so I'd put the MAX(timestamp back), and then if we leave the extra column then the condition would be just `is_successfully_completed_event`

The deleted parcels and clients are a bit more interesting. I think that we should keep the deleted parcels in the data simply because in my head we only care about the stats of how many parcels were deleted - if we delete some old parcels from 5 years ago but suddenly want to find that report it will all show zeros. So I think we ignore deleted parcels and if it was completed successfully we add it to the data

The same logic applies to the deleted clients, however we don't have all the available data there. This is probably where the "let's have a table and update it with new data each week" approach would've been more useful. I'm almost tempted to add an "unknown" column but I think for now we just include them in the total and make a note somewhere that this may happen, and then raise a ticket to handle it in the future

supabase/migrations/20240530291033_add_completed_parcels_view.sql
```diff
@@ -0,0 +1,32 @@
...
+            SELECT
+                event_name
+            FROM
+                status_order
+            WHERE
+                is_successfully_completed_event
+        )
+    -- Ensures each parcel is only counted once.
+    GROUP BY
+        events.parcel_id
```
8kristy: Oh yeah would this not solve the double counting issue with multiple statuses?

src/app/reports/reportsTable/types.ts
Alex-Softwire: Am a little confused as to what the change is here - what is the space for?

supabase/migrations/20240530391033_add_reports_view.sql
```diff
@@ -0,0 +1,50 @@
+CREATE OR REPLACE VIEW reports AS
+
+-- Gets the date of the first completed parcel
```
8kristy: Think this one is fairly self explanatory - can delete the comment
```suggestion
```

src/app/reports/reportsTable/ReportsPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <>
+            {isLoadingForFirstTime ? (
+                <Centerer>
+                    <CircularProgress aria-label="table-initial-progress-bar" />
+                </Centerer>
+            ) : (
+                <>
+                    {errorMessage && <ErrorSecondaryText>{errorMessage}</ErrorSecondaryText>}
+                    <TableSurface>
+                        <ServerPaginatedTable<ReportsTableRow, DbReportRow>
```
8kristy, line 101: Where can you style this? I think the week commencing column was a bit too narrow on smaller screens and overlapping the family header

8kristy (commented):
Just need to think about deleted clients and fix the styling a bit otherwise I think mostly good to go

src/app/reports/reportsTable/ReportsPage.tsx
Alex-Softwire: Can have a look into this. Though were we planning on saving styling (such as the line to demarcate the pets section) to a future ticket? Though I guess this is more important?

supabase/migrations/20240530291033_add_completed_parcels_view.sql
Alex-Softwire: Yeah it would, that and the MAX line are required together to work. If you group by parcel_id, the things you select now have to be aggregate functions that take in multiple rows and output a single value e.g MAX(timestamp). Just writing timestamp would cause a syntax error. Is that what you mean here?

src/app/reports/reportsTable/types.ts
8kristy: Oh space is an accident, the change is to swap the order so that the null data is at the bottom

supabase/migrations/20240530291033_add_completed_parcels_view.sql
8kristy: Oh yeah that is how SQL works isn't it 😅 Thanks for explaining, I'm a bit rusty

src/app/reports/reportsTable/ReportsPage.tsx
8kristy: I think the main point is that it should be simpler - have a look at #182 
If it's not simple let me know and we can leave it for now

Alex-Softwire: Had a look, looks doable for that one column, might require another ticket to fully polish the table?

src/app/reports/reportsTable/getReportsData.ts
8kristy: Ah sorry I missed the extra value, let's leave it

supabase/migrations/20240530291033_add_completed_parcels_view.sql
Alex-Softwire: Cool cool - happy to remove the deleted parcels clause, that makes sense.  Though the flip side of this is that there is now no way of removing parcels from being counted by the reports. Will do that change though.

Re deleted clients, we saying leave it as it is for now? Yeah I see that disadvantage of using a view like I have. One could have mix of live data from the last month and stored data before then, or just a stored table (with a similar sql definition) and a cron job, but I think that sounds like a future ticket? Happy to make those extra tickets to tidy this up once this is merged.

src/app/reports/reportsTable/types.ts
Alex-Softwire: Hmm I thought the pattern in the codebase was to put data first (independently of nullness) - I think I was corrected on this in my first ticket. That being said, with a ctrl + f, I think we've done a bit of both in the codebase with maybe slightly more like what I've suggested. Though happy to switch the order if you think so though :)

src/app/reports/reportsTable/ReportsPage.tsx
Alex-Softwire: Done! Thanks for the suggestion, was easier than I thought!

Alex-Softwire: Have added another screen shot in this case.

supabase/migrations/20240530291033_add_completed_parcels_view.sql
Alex-Softwire: No worries, had no idea about any of those types of sql statements until this ticket, so it's been good to learn

src/app/reports/reportsTable/types.ts
8kristy: Sorry I worded that badly - data stays at the top of each object but the object which has null data should go in the bottom I think, so 
```
{
data: not null
error: null
}
| {
 data: null
 error: not null
}
```

Alex-Softwire: Aah I see what you mean now -- will go and change it

supabase/migrations/20240530291033_add_completed_parcels_view.sql
8kristy: Yeah I think we leave it as is (doubt they'll be deleting clients anytime soon) and then create a ticket to aggregate a "permanent" report somewhere

src/app/reports/reportsTable/types.ts
Alex-Softwire: Done!

8kristy (approved):
Just update migration file timestamps to more recent so they get applied in the right order and we're good to go, well done :)

Issue #303, opened 2024-04-26
Client modal - Free text box for storing notes about the client


dan-langridge:
### Describe the solution / feature you'd like

Sometimes it would be helpful to store notes against a client, for example that they received a special item (not on the shopping list) in their last parcel like a Microwave, or any other notes that would be helpful to store.
- Open a client modal
- Scroll down to new NOTES text box
- Edit text
- Submit

rickaw:
## Developer Notes
- Add a `notes` text field for clients
- Display the notes in client detail modal
- Add the notes field on the add / edit client form

AmeliaIsCoding:
## Also included in the ticket
- Free text box for editing client notes directly from client modal

AmeliaIsCoding:
## QA

- [x] In dev database `clients` table, there is a column `notes` of type string
- [x] In local database `clients` table, there is a column `notes` which is filled by a random string
- [x] In dev site, in Clients page, clicking on a client row to open its modal shows a field for Notes
- [x] There is no field for "Is Active"
- [x] The notes field is an editable text box with buttons for cancel and save
- [x] On typing something in the box and saving, closing and reopening the modal, the changes persist
- [x] On typing something in the box and cancelling, the text in the box reverts to the original text without needing to close and reopen
- [x] In the front end audit logs table in Admin page, an audit log row has been added with action "edit client notes", a link to the client whose notes you edited, and content populated with `clientId` and the notes that you added
- In local site, change the file src/app/c;lients/updateClientNotes.ts to force an error on submission - eg, change the table that is updated to a different string like "error expected". Try and edit the notes of a client as before. An error should be displayed in the modal that includes the log ID
- [x] The error disappears on each of the following: typing something else in the notes box, clicking "cancel", closing and reopening the modal
- [x] In the front end audit logs table in Admin page, an audit log row has been added with action "edit client notes", a link to the client whose notes you edited, and content populated with `clientId` and the notes that you added, and was success false, and log ID populated
- [x] Back in dev site, open a client modal, click "edit details", and scroll to the bottom. The notes should be visible as a text field, and editing the content saves correctly on submit
- [x] Create a new client; scroll to the bottom; the notes should be visible as a text field, with the content saving correctly when the client is made

8kristy:
Moving back to to-do, implementation here https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/pull/370 but needs review/tweaking

[...]

PR #370, opened 2024-05-31
#303: Add client notes

AmeliaIsCoding:
## What's changed

- added client notes field as an optional string (@minalui)
- client notes are editable in both edit/add client form (@minalui)
- and also directly in the client modal as a free text box (@AmeliaIsCoding)
- seeded clients include notes (@AmeliaIsCoding)
- removed 'is active' field from client modal, as this is implicit in whether or not the client info is displayed (@AmeliaIsCoding)

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|imageimage|imageimage|

## Checklist

    - Checkout the dev branch
    - Checkout your branch


AmeliaIsCoding (commented):
Self-review

src/app/clients/ExpandedClientDetails.tsx
```diff
@@ -17,20 +30,91 @@ const DeletedText = styled.div`
...
-        </>
+    const [notes, setNotes] = useState<string | null>(originalNotes);
+    const [errorMessage, setErrorMessage] = useState<string | null>(null);
+    const [isLoading, setIsLoading] = useState<boolean>(true);
+
+    const loadData = useCallback(() => {
+        (async () => {
+            setIsLoading(true);
+            setClientDetails(await getExpandedClientDetails(clientId));
+            setClientParcelsDetails(await getClientParcelsDetails(clientId));
```
stonelink: If `displayClientsParcels` is false, should this be called?

src/databaseTypesFile.ts
```diff
@@ -956,6 +959,13 @@ export type Database = {
         }
         Returns: string
       }
+      insertClientAndTheirFamily: {
```
stonelink: Surprising to see a new Function`insertClientAndTheirFamily` here. 

It looks like we should just be sticking with `insert_client_and_family`, which is the procedure that is still called in `submitFormHelpers` - see comment in the migration file below.

supabase/migrations/20240531142938_add_client_notes.sql
```diff
@@ -0,0 +1,104 @@
+alter table "public"."clients" add column "notes" text;
+
+set check_function_bodies = off;
+
+CREATE OR REPLACE FUNCTION public."insertClientAndTheirFamily"(clientrecord jsonb, familymembers jsonb)
```
stonelink: Hi @AmeliaIsCoding 
I know this was a while ago, but...

I'm unsure why the db Function in this migration is called `insertClientAndTheirFamily` - the rest of the PR looks like we should be changing the existing procedure `insert_client_and_family` (which is called from submitFormHelpers.ts.

stonelink: I think what's happened is [this commit](https://github.com/NorwoodAndBrixtonFoodbank/nbf-website/commit/2f5c9f774b35ed756358e30b69e9665acb1762a1) was merged by @minalui after this PR was raised, which changed the name of the Function.

Could I please check my thoughts with you both?

I _think_ I should:
- merge `dev` into this branch (so that `insert_client_and_family` has replaced `insertClientAndTheirFamily`)
- change the Function name in this this sql file
- apply the outstanding migrations to my local DB
  - I think I should be able to use `npx supbase db push` because my local DB matches `dev`
  - have you been rebuilding your local DB from scratch each time?
- regenerate the types file using `npm run db:local:generate_types` (because in this PR it has the camelCase function name)

Thanks

src/app/clients/ExpandedClientDetails.tsx
stonelink: This is false in ParcelForm.tsx. 

I think in that case `clientParcelsDetails` should be left as null.

supabase/migrations/20240531142938_add_client_notes.sql
minalui: Hi Lincoln,

Sorry if I'm a little rusty - it has been a little while. @AmeliaIsCoding please feel free to correct if I've made a mistake.
I think I renamed it to insert_client_and_family to match the supabase syntax when I was refactoring the function to add birth year instead of age. It was one of my last tickets so my thought is it maybe the change wasn't merged yet?

I think your method should work - depending on the date of the migration file, you may need to change it to a later date. 

What I did previously to make changes to db:
1. Switch to `dev` branch
2. `npm run dev:reset_supabase`
3. Switch to current branch
4. Make changes needed in supabase (I think Amelia has added the client notes into the function)
5. `npx supabase db diff -f <name_of_migration>` to migrate the changes
6. `npm run db:local:generate_types` to generate the types
7. Then check if the migration can happen succesfully

stonelink: Thanks Mina :)

stonelink (approved):

PR #372, opened 2024-06-04
VFB-258: Fix status modal calendar bug

Alex-Softwire:
## What's changed
When you add a status to a parcel, the calendar to chose the date had an issue. It was picking a date about + or - 2 weeks away from the one you chose. This is because it was setting the day of the week not the day of the month.

## Checklist


src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -39,7 +39,7 @@ const StatusesBarModal: React.FC<StatusesBarModalProps> = (props) => {
             date
                 .set("year", newDate?.year() ?? date.year())
                 .set("month", newDate?.month() ?? date.month())
-                .set("day", newDate?.day() ?? date.day())
```
Alex-Softwire, line 42: .day sets the day of the week not the day of the month. So when you picked a date on the calendar, it would pick the wrong date in the same month & same day of the week.

supabase/migrations/20240710232517_remove_clientid_from_events.sql
stonelink, line 1: Having client Id in the events table wasn't necessary (because an event is linked to a parcel, which has a client) and broke #159

stonelink (approved):

PR #409, opened 2024-07-11
VFB-283 implement full typescript linting

Stuart-Knowles-SW:
The `typescript-eslint` recommended rules were never added to eslint config, which lead to some fairly shaky types. This adds that rule and fixes all of the problems that don't require major rewrites.

## Benefits
- Increased type safety in the current code
  - Specifically, there were a lot of complicated nested interfaces that made it look like we had type safety but at their base they were just an any type so nothing was really being checked
- Will maintain higher standards of type safety going forward
- Fewer bugs in the future

## Risks
- Bugs sneaking from many many changes

## Mitigations
- Typechecking and linting pass, and are stricter than before
- I've UAT'd as much as I can, focusing on components such as filters which had the most changes
- Almost all of the changes are purely to types, so should have no runtime impact
- Tests still pass

## Problems I didn't fix
- The implementation of the Filter Types are based around interfaces which led to some type conflicts that were going to be very difficult to resolve, but logically can be ignored. For now, I ignored them with comments. In the long term we should think about rewriting them with classes
- Not all of the problems were solved in the most beautiful way possible. There is a marginal increase in complexity here, however in return we get much stronger type safety.

## Info
The main culprits for bad practice were:
- non-null assertions:
    ```
    const val: client | null
    console.log(client!.name)
    ```
- `any` types
- redundant type definitions, e.g, `const foo: string = "I'm definitely a string"`
- using `{}` in situations you shouldn't

Of those, the first two are the ones that really needed fixing because they could very realistically lead to runtime bugs

### DIFF
--- a/src/app/roles.tsx
+++ b/src/app/roles.tsx
@@ -55,7 +55,9 @@ interface RoleUpdateContextType {
 
 export const RoleUpdateContext = createContext<RoleUpdateContextType>({
     role: null,
-    setRole: (_role) => {},
+    setRole: (_role) => {
+        throw new Error("Context implementation not provided");
+    },
 });
 
 export const RoleManager: React.FC<Props> = ({ children }) => {
--- a/src/components/Tables/Table.tsx
+++ b/src/components/Tables/Table.tsx
@@ -2,9 +2,9 @@
 
 import Icon from "@/components/Icons/Icon";
 import {
-    ClientSideFilter,
+    DistributeClientFilter,
+    DistributeServerFilter,
     PaginationType as PaginationTypeEnum,
-    ServerSideFilter,
 } from "@/components/Tables/Filters";
 import TableFilterAndExtraColumnsBar from "@/components/Tables/TableFilterAndExtraColumnsBar";
 import {
@@ -25,6 +25,8 @@ import {
     DividingLineStyleOptions,
     getDividingLineStyleOptions,
 } from "@/app/parcels/parcelsTable/conditionalStyling";
+import { PostgrestFilterBuilder } from "@supabase/postgrest-js";
+import { Database } from "@/databaseTypesFile";
 
 export type TableHeaders<Data> = readonly (readonly [keyof Data, string])[];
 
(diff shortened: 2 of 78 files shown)


src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
stonelink, line 1: On line 56 of this file (I can't attach a comment to the line), there's a call to `logErrorReturnLogId()` with the same structure as the calls in `getCurrentUser.ts` and elsewhere in this PR.

That is, it looks like it should be calling the 3-parameter version:
```
logErrorReturnLogId("Error with fetch: Parcels", {}, error)
```

Have I misunderstood? I think I saw a handful like this.

src/components/Tables/sortMethods.ts
```diff
@@ -3,8 +3,8 @@ import { SortOrder } from "react-data-table-component";
 import { PostgrestFilterBuilder } from "@supabase/postgrest-js";
 
 export type ServerSideSortMethod<DbData extends Record<string, unknown>> = (
-    query: PostgrestFilterBuilder<Database["public"], DbData, any>,
-    sortDirection: SortOrder
-) => PostgrestFilterBuilder<Database["public"], DbData, any>;
+    sortDirection: SortOrder,
```
stonelink, line 6: Why have these parameters swapped?

src/components/Tables/Table.tsx
```diff
@@ -482,12 +491,12 @@ const Table = <
                         onChangePage={
                             paginationConfig.enablePagination
                                 ? paginationConfig.onPageChange
-                                : () => {}
+                                : () => undefined
```
stonelink, line 494: Interesting!
(My typescript knowledge is rather incomplete)

src/app/roles.tsx
```diff
@@ -55,7 +55,9 @@ interface RoleUpdateContextType {
 
 export const RoleUpdateContext = createContext<RoleUpdateContextType>({
     role: null,
-    setRole: (_role) => {},
+    setRole: (_role) => {
+        throw new Error("Context implementation not provided");
```
stonelink, line 59: Thanks for this

stonelink (approved):
Impressive, and an insightful journey through the codebase!

src/pdf/DriverOverview/DriverOverviewPdfButton.tsx
Stuart-Knowles-SW: Our logging and especially error logging is something I'd like to improve generally, and it's one of the less elegant parts of this PR. The short answer here is that supabases `PostgrestError` class apparently isn't a subclass of `Error`, which makes it a bit weird.

I've updated this to put the error object in `{}`, so it will be rendered the same in the logs, good catch!

src/components/Tables/sortMethods.ts
Stuart-Knowles-SW: The client sort method has `sortDirection` first (it doesn't have a query), and it's very useful for type purposes for the two methods to be compatible with the type of the first argument matching

PR #410, opened 2024-07-12
VFB-227 Resolve Supabase security advisory errors

### LINKED ISSUE #272: See date/time when a User last logged into their account
### Is your feature request related to a problem? If so, please describe.

_No response_

### Describe the solution / feature you'd like

On the > Admin Page > Users Table
Add a column which shows a date and time stamp for when each user last logged in.

This would be helpful for identifying in-active accounts.

### Additional context

_No response_

stonelink:
There were two types of security advisory errors:
- views had the the default SECURITY_DEFINER state, meaning that the view had Postgres-level permissions even if the invoking user doesn't have them
- contents of the auth table were being exposed in the `users_plus` view

Because the fix for the `users_plus` issue was to add more info to the `profiles` table, I also added each user's Last Sign In Date, which resolves #272 

## Checklist
- [X] I have self-reviewed this PR
- [X] Make sure you've tested via `npm run test:e2e`

  - [X] I have updated the typescript definitions for the database with `npm run db:local:generate_types`
  - [-] I have modified the seed in `supabase/seed/seed.mts` if appropriate
  - [X] With my final set up, I can run `npm run dev:reset_supabase` without any errors.
  - [-] If I have added constraints e.g. `set not null`, I have taken reasonable measures to ensure constraint violations do not happen when the migration occurs. 
    - Checkout the dev branch
    - Checkout your branch

### DIFF
--- a/supabase/tools/createUserForEachRole.ts
+++ b/supabase/tools/createUserForEachRole.ts
@@ -34,7 +34,7 @@ async function createUserForEachRole(): Promise<void> {
         role: "volunteer",
         firstName: "volunteer",
         lastName: "person",
-        telephoneNumber: "078888888888",
+        telephoneNumber: "020866666666",
     });
     await createUser(supabase, {
         email: "manager@example.com",
@@ -42,7 +42,7 @@ async function createUserForEachRole(): Promise<void> {
         role: "manager",
         firstName: "manager",
         lastName: "person",
-        telephoneNumber: "078888888888",
+        telephoneNumber: "075555555555",
     });
     await createUser(supabase, {
         email: "staff@example.com",
@@ -50,7 +50,7 @@ async function createUserForEachRole(): Promise<void> {
         role: "staff",
         firstName: "staff",
         lastName: "person",
-        telephoneNumber: "078888888888",
+        telephoneNumber: "074444444444",
     });
 }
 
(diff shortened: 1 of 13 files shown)


supabase/tools/createUserForEachRole.ts
```diff
@@ -69,12 +69,15 @@ async function createUser(supabase: SupabaseClient, userProfile: userProfile): P
     }
 
     if (data) {
+        const nowTimestamp = new Date().toISOString();
```
Stuart-Knowles-SW: Could we use a constant time for this? I doubt it'll matter much in practice, but I have a preference for seed data (which this sort of is) to be static. Not too fussed though

Stuart-Knowles-SW (approved):
One tiny comment but nothing too important. Overall looks great!

supabase/tools/createUserForEachRole.ts
stonelink: Good point, fixed.

stonelink:
## QA steps
Login as different roles (admin, volunteer) in turn and check that all fields in in tables and modals that are populated from the following are working as expected:
[ ] parcels - including events/status
[ ] clients - including family details
[ ] reports

Admins only:
[ ] users - including checking 'last sign in' time
[ ] audit log

PR #414, opened 2024-07-15
VFB-245 fix parcel page client names with closing brackets

LeaHe188:
## What's changed
In parcel page, when we search names for the closing bracket, the expected result is to show deleted users and the users whose names contain a ")".

After looking into the documentation of supabase, it seems supabase can return a promise that satisfies the query. But given the seemingly unmatched datatype (i.e. Promise and DBQuery), and that the ")" seems more like a programmer slang, we might just want to ban the use of ")", and any other reserved characters stated here in the PostgREST documentation: https://postgrest.org/en/stable/references/api/url_grammar.html#reserved-characters.

On further investigation, only alphanumeric words are allowed in the input. If any non-alphanumeric words appear, they will be deleted.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| 
before
 | 
after
 |

## Checklist

    - Checkout the dev branch
    - Checkout your branch

---


src/common/databaseFilters.ts
```diff
@@ -8,15 +8,16 @@ export const fullNameSearch = <DbData extends DbClientRow | DbParcelRow>(
     clientIsActiveColumnLabel: Extract<keyof DbData, "is_active" | "client_is_active">
 ): ServerSideFilterMethod<DbData, string> => {
     return (query, state) => {
-        if (state === "") {
+        const newState = state.replace(/[,.():/]/g, "");
```
Stuart-Knowles-SW, line 11: Can we call this something like `cleanState` to better reflect its purpose?

src/common/databaseFilters.ts
```diff
@@ -8,15 +8,16 @@ export const fullNameSearch = <DbData extends DbClientRow | DbParcelRow>(
...
         }
-        if (parcelsPageDeletedClientDisplayName.toLowerCase().includes(state.toLowerCase())) {
+        if (parcelsPageDeletedClientDisplayName.toLowerCase().includes(newState.toLowerCase())) {
             return query.or(
-                `${fullNameColumnLabel}.ilike.%${state}%, ${clientIsActiveColumnLabel}.eq.false`
+                `${fullNameColumnLabel}.ilike.%${newState}%, ${clientIsActiveColumnLabel}.eq.false`
             );
         }
-        return query.ilike(`${fullNameColumnLabel}`, `%${state}%`);
+        return query.ilike(`${fullNameColumnLabel}`, `%${newState}%`);
```
Stuart-Knowles-SW, line 20: I know this was here from before, but can you remove the string interpolation on `fullNameColumnLabel` and just pass it as a variable?

Stuart-Knowles-SW (commented):
Looks good, just a couple of little comments. Before I approve, can you check whether there are other filters or search components in the codebase where similar query formatting happens which we might want to fix?

src/common/databaseFilters.ts
Stuart-Knowles-SW, line 11: Are these the only special characters we want to remove? Or would we be better off specifying a list of allowed characters?

LeaHe188: Yea sure, and filters also appear in the client page so I'll look into it

LeaHe188: Yea the documentation said these are the only ones that may disrupt the url, but since the filter is for names we might want to only prompt for alphanumeric input / alphabet?

Stuart-Knowles-SW: Can you link me that documentation?

Stuart-Knowles-SW:
Also, just some PR hygiene:

- The PR title should also include the ticket number
- It would be helpful if the "What's changed" description also included a little bit of the purpose behind the ticket and links to relevant documentation if it's non-standard
- When you're doing the checklists, `[x]` is rendered as an actual checkbox which is probably a nicer way to go than the ticks

src/common/databaseFilters.ts
LeaHe188: Not being the supabase documentation but the PostgREST one:
https://postgrest.org/en/stable/references/api/url_grammar.html#reserved-characters

found in post:
https://stackoverflow.com/questions/76065534/how-to-escape-string-characters-in-supabase-js-or-query

src/common/databaseFilters.ts
LeaHe188: this is done:)

src/common/databaseFilters.ts
LeaHe188: changed into alphanumeric input only and the filter I made changes to were the base filters, so the client page applied as well

PR #420, opened 2024-07-16
VFB-262 render single item from wiki table

rjesson-softwire:
## What's changed
create info page with template accordion
display top row of wiki table with custom hyperlink format
add markdown style hyperlink embedding
fix excessive async fetching
hid accordion if no items in wiki table
add 2 component tests

hyperlink formatting:
standard markdown link formatting, e.g
<google.com> or [google](google.com)

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| *no before* | image |

## Checklist
- [x ] I have documented the testing steps for QA
- [x ] Make sure you've verified it works via `npm run build` and `npm run start`
- [ x] Make sure you've fixed all linting problems with `npm run lint_fix`
- [ x] Make sure you've tested via `npm run test:e2e`

---

* **Introduction of InfoPage and Ancillary Additions**
A new file titled `InfoPage.tsx` was added, which contains several key features. These include a function known as `formatContent` that refines data, a `getTopWikiRow` function that retrieves high-priority information, and an `InfoPage` feature which displays the collective information. Adding to this, a `WikiAccordian` component was introduced along with `WikiAccordian.cy.tsx`, a dedicated file for testing this component.
* **Incorporation of WikiAccordian Component**
A new file by the name `WikiAccordian.tsx` was created which includes a feature known as the `WikiAccordian` component. 
* **Addition of Info Component in New File `page.tsx`**
A new `page.tsx` file was introduced which comprises an `Info` component that will handle and display information.
* **Access Allowance Update**
A route modification was done in the `roles.tsx` file where "/info" was appended to the `pathsShownToAllAuthenticatedUsers` list - This essentially means that all verified users will now have access to the "/info" page.
* **Type Addition in Utilities**
The `DbWikiRow` type was introduced to the `databaseUtils.ts` file to streamline data management within the database.


src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+                        {plainPart}
+                    </a>
+                );
+            }
+        } else {
+            return part;
+        }
+    });
+    return formattedContent;
+};
```
Stuart-Knowles-SW: Can we move this into the `WikiAccordion` file? The logic feels like it should be internal to that component

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+) => {
+    const topRowContentParts = topRowContent.split(/(\[.*\]\(.*\)|<.*>)/);
+    let counter = 0;
+    const formattedContent = topRowContentParts.map((part, index) => {
+        if (index % 2 === 1) {
+            const plainPart = part.slice(1, -1);
+            if (plainPart.includes("](")) {
+                const items = plainPart.split(/\]\(/);
+                return (
+                    <a href={items[1]} key={counter++}>
```
Stuart-Knowles-SW: This isn't a safe way to key the map because it's not an intrinsic property of the item. A proper way to do this would be to pre-process the raw content into an array of objects that look something like
```
{ content: string;
href?: string;
key: string;}
```

where the key is something like a random uuid -- and cache the result of that the result of that to only be recalculated when the data is refetched.

Then, you map those objects in WikiAccordion, rendering them as an anchor if `href` is defined or plain text otherwise, with the key as the key attribute.

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+        error: Error | null;
+    };
+
+    if (query.error) {
+        throw query.error;
+    }
+    return query.data;
+};
+
+export const formatContent: (topRowContent: string) => (string | React.JSX.Element)[] = (
```
Stuart-Knowles-SW: This is a fairly nasty return type, I'd prefer if we wrapped the plain text in Fragments so it's all JSX if this is the route we're going (though, as I've said elsewhere, I think there's cleaner options)

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,69 @@
+"use client";
```
Stuart-Knowles-SW: This page would be a prime candidate for server side rendering, how much work would that be to implement?

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const getTopWikiRow = async (supabase: Supabase): Promise<DbWikiRow> => {
+    const query = (await supabase.from("wiki").select("*").limit(1).single()) as {
+        data: DbWikiRow;
+        error: Error | null;
+    };
+
+    if (query.error) {
+        throw query.error;
+    }
```
Stuart-Knowles-SW: Ideally we don't want uncaught errors, this should just be propagated through, and an error message rendered (and the error logged) if the error is not null

src/app/info/WikiAccordian.tsx
```diff
@@ -0,0 +1,31 @@
...
+import AccordionDetails from "@mui/material/AccordionDetails";
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
+import { DbWikiRow } from "@/databaseUtils";
+
+interface AccordianProps {
+    topRow: DbWikiRow;
+    formattedContent: (string | React.JSX.Element)[];
+}
+
+const WikiAccordian: React.FC<AccordianProps> = (props) => {
```
Stuart-Knowles-SW: Can we have a further think about naming here? To me, it's unclear what the scope of `WikiAccordion` is. It could be something for the whole wiki page, or just a part of a single item. I'd prefer `WikiItem` or `WikiEntry` here. The general rule is "name for usage, not implementation"

src/app/info/WikiAccordian.tsx
```diff
@@ -0,0 +1,31 @@
...
+    return (
+        <>
+            {props.topRow.order !== -1 && (
```
Stuart-Knowles-SW: If the top level of a component is a conditional, that conditional should sit where the component is used, not here

src/app/info/WikiAccordian.tsx
Stuart-Knowles-SW: In general I prefer the expanded syntax, i.e.:
`...React.FC<AccordianProps> = ({ topRow, formattedContent }) => {`

I know that both are used in the codebase, however, so I'm not going to insist on it.

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,69 @@
+"use client";
+
+import React, { useState, useEffect } from "react";
+import WikiAccordian from "./WikiAccordian";
```
Stuart-Knowles-SW: I think we prefer absolute imports, with `@` in this codebase

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+    });
+    return formattedContent;
+};
+
+const InfoPage: React.FC = () => {
+    const [topRow, setTopRow]: [DbWikiRow, (a: DbWikiRow) => void] = useState({
+        title: "",
+        content: "",
+        order: -1,
+    });
```
Stuart-Knowles-SW: Can we initialise this to `null`? It forces better type safety and makes comparisons more clear. Also, the correct way to type state is:
`const [state, setState] = useState<Type>(initial_value)`

src/app/info/WikiAccordian.tsx
```diff
@@ -0,0 +1,31 @@
+import Accordion from "@mui/material/Accordion";
+import AccordionSummary from "@mui/material/AccordionSummary";
+import AccordionDetails from "@mui/material/AccordionDetails";
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
+import { DbWikiRow } from "@/databaseUtils";
+
+interface AccordianProps {
+    topRow: DbWikiRow;
```
Stuart-Knowles-SW: What's the reasoning behind calling this `topRow`?

Stuart-Knowles-SW (changes requested):
Good first go! Quite a few things to improve, feel free to come talk to me if any of them don't make sense :)

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import { Supabase } from "@/supabaseUtils";
+import { DbWikiRow } from "@/databaseUtils";
+import supabase from "@/supabaseClient";
+import Error from "next/error";
+
+interface QueryType {
+  data: DbWikiRow | null;
+  error: Error | null;
+}
```
Stuart-Knowles-SW: strictly, these are likely to be undefined not null, i.e.:
```
interface QueryType {
  data?: DbWikiRow;
  error?: Error;
}
```

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,47 @@
...
+        fetchData();
+      }, []);
+    } catch (e: any) {
+      return (
+      <>
+        <h3>Error Occured</h3>
+        <p>{e}</p>
+      </>
+      );
+    }
```
Stuart-Knowles-SW: Although this is valid code, wrapping hooks is in general a bit of an anti-pattern.
The pattern elsewhere in the codebase is to have two state variables, one for the error and one for the data, set them both in useEffect, then render conditionally based on them. `src/app/clients/clientsTable/ClientsPage.tsx` has some good examples for comparison

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    return (topRow && topRow.order !== -1) ? <WikiItems row={topRow}/> : <></>;
```
Stuart-Knowles-SW: order should never be -1 now I think, so this can be more concisely written
`return topRow && <wikiItems row={topRow}/>`

src/app/info/WikiItems.tsx
```diff
@@ -0,0 +1,56 @@
...
+            }
+        } else {
+            return {content: part, key: crypto.randomUUID()}
+        }
+    });
+    return contentParts.map((part: ContentPart)=>{return part.href ? <a href={part.href} key={part.key}>{part.content}</a> : <span key={part.key}>{part.content}</span>})
+};
+
+const WikiItems: React.FC<AccordianProps> = (props) => {
+    const formattedContent = formatContent(props.row.content);
```
Stuart-Knowles-SW: We should wrap this in a `useMemo` call to stop it being recalculated every page refresh

src/app/info/WikiItems.tsx
```diff
@@ -0,0 +1,56 @@
...
+                return {content: plainPart, href: plainPart, key: crypto.randomUUID()}
+            }
+        } else {
+            return {content: part, key: crypto.randomUUID()}
+        }
+    });
+    return contentParts.map((part: ContentPart)=>{return part.href ? <a href={part.href} key={part.key}>{part.content}</a> : <span key={part.key}>{part.content}</span>})
+};
+
+const WikiItems: React.FC<AccordianProps> = (props) => {
```
Stuart-Knowles-SW: Sorry to be such a pain about the naming. Can this be `WikiItem` singular?

src/app/info/WikiItems.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const formattedContent = formatContent(props.row.content);
+    return (
+        <>
```
Stuart-Knowles-SW: These fragment tags are redundant now :)

Stuart-Knowles-SW (commented):
Getting a lot closer now, still a few minor changes though :)

src/app/info/InfoPage.tsx
rjesson-softwire: switched to SSR so no hooks anymore

src/app/info/WikiItems.tsx
rjesson-softwire: have wrapped and website functions, unsure if having intended effect though

src/app/info/InfoPage.tsx
rjesson-softwire: removed the topRow condition as error is handled beforehand implying that data not be not null. However -1 is our default value for order, functions like an 'invalid' flag for a given wiki so that it wont be displayed. In the next ticket we filter out any -1 orders but since only one item is displayed in this ticket this is sufficient.

src/app/info/InfoPage.tsx
```diff
@@ -1,47 +1,41 @@
...
-        <p>{e}</p>
-      </>
-      );
-    }
-
-    return (topRow && topRow.order !== -1) ? <WikiItems row={topRow}/> : <></>;
+export async function getServerSideProps() {
+  const query: QueryType = await getTopWikiRow(supabase);
+  return query
+}
```
Stuart-Knowles-SW: It looks like you're trying to use [server side props like this](https://nextjs.org/docs/pages/building-your-application/data-fetching/get-server-side-props), but I think there's a few implementation details we need to iron out. Lets have a chat about it :)

src/app/info/WikiItem.tsx
```diff
@@ -35,22 +38,24 @@ export const formatContent: (topRowContent: string) => (React.JSX.Element)[] = (
...
-                >
-                    {props.row.title}
-                </AccordionSummary>
-                <AccordionDetails>{formattedContent}</AccordionDetails>
-            </Accordion>
-        </>
-    );
+const WikiItem: React.FC<AccordianProps> = (props) => {
+    const visibleWikiItem = useMemo(() => {
+        const formattedContent = formatContent(props.row.content);
```
Stuart-Knowles-SW: I've realised I was wrong here, we don't want use memo because this can be server side rendered. However, lets have a chat about use memo and its usage because this isn't quite right here.

src/app/info/InfoPage.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const InfoPage: React.FC<QueryType> = ({data, error}) => {
+  if (error) {
+    console.error(error)
+    return (<h3>Error occured when fetching data.</h3>)
+  } else {
+    const topRow = data;
+    return (topRow.order !== -1) ? <WikiItem row={topRow}/> : <></>;
```
Stuart-Knowles-SW: I still don't quite understand in what situation `topRow.order` can equal -1?

Stuart-Knowles-SW (approved):
LGTM! Thanks for all the hard work on this, I know it took longer than expected to get over the line

PR #426, opened 2024-07-18
VFB-277 Add drop down on lists page to toggle between list types

ebinezer04:
## What's changed
Added 'Hotel' and 'Regular' buttons which filter the lists page by selected list type. Adding a new item adds it to the currently selected list.

## Screenshots / Videos
 Before     
 image 
After
image

## Checklist

---

Delete the command below if you don't want this to be generated.


src/app/lists/EditModal.tsx
```diff
@@ -134,12 +135,15 @@ const EditModal: React.FC<Props> = ({ data, onClose }) => {
...
-    const onSubmit = async (): Promise<void> => {
+    const onSubmit = async (currentList: ListName): Promise<void> => {
         if (toSubmit === null) {
             return;
         }
 
         if (data === null) {
+            currentList.toLowerCase() === "regular"
+                ? (toSubmit.list_type = "regular")
+                : (toSubmit.list_type = "hotel");
```
Stuart-Knowles-SW: Could this be written as `toSubmit.list_type = currentList.toLowerCase()`?

src/app/lists/EditModal.tsx
```diff
@@ -164,8 +168,8 @@ const EditModal: React.FC<Props> = ({ data, onClose }) => {
 
     const Footer = (
         <>
-            <Button variant="contained" color="primary" onClick={onSubmit}>
-                Submit
+            <Button variant="contained" color="primary" onClick={() => onSubmit(currentList)}>
+                Submit {currentList}
```
Stuart-Knowles-SW: Is there a better way we can signpost this to the user? I imagine without context it isn't immediately clear what "submit regular" means.

I reckon something closer to the header, maybe just below it, that tells the user what list they're on (`` <p> `Current List: ${currentList}` </p>`` or something)

src/app/lists/ListDataview.tsx
```diff
@@ -135,13 +137,16 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
...
+    currentList,
+    setCurrentList,
 }) => {
     const [modal, setModal] = useState<EditModalState>();
     const [toDelete, setToDelete] = useState<number | null>(null);
     // need another setState otherwise the modal content changes before the close animation finishes
     const [toDeleteModalOpen, setToDeleteModalOpen] = useState<boolean>(false);
     const [listData, setListData] = useState<ListRow[]>(listOfIngredients);
     const [primaryFilters, setPrimaryFilters] = useState<ListFilter[]>(filters);
+    const [listStateAnchorElement, setListStateAnchorElement] = useState<HTMLElement | null>(null);
```
Stuart-Knowles-SW: Just a heads up, it's not good practice to store React elements in state, because it prevents React from properly tracking them for updates and refreshes. It often works, but should be avoided where possible because it's a easy way for bugs to slip in in the future.

In this case, it's required by MaterialUI, so you've done the right thing here, but just to let you know this is an anti-pattern in general.

However, I think we could use more clear names, because it's not obvious at a first glance what purpose this serves

src/app/lists/ListStates.tsx
```diff
@@ -0,0 +1,49 @@
...
+                                setCurrentList(swapListNameValue(currentList));
+                                setListStateAnchorElement(null);
+                            }}
+                        >
+                            {swapListNameValue(currentList)}
+                        </MenuItem>
+                    </MenuList>
+                </Menu>
+            )}
+        </>
```
Stuart-Knowles-SW: If you have a conditional rendering at the top level based on an optional argument, it should live outside the component. That means that `ListStates` should have `listStateAnchorElement` as a mandatory property, there should be no conditional here, then it should be called as
```
{listStateAnchorElement &&
    <ListStates {...props} />
}
```

src/app/lists/ListStates.tsx
```diff
@@ -0,0 +1,49 @@
...
+    setListStateAnchorElement: React.Dispatch<React.SetStateAction<HTMLElement | null>>;
+    currentList: ListName;
+    setCurrentList: React.Dispatch<React.SetStateAction<ListName>>;
+}
+
+const swapListNameValue = (currentListName: ListName): ListName => {
+    return currentListName === "Regular" ? "Hotel" : "Regular";
+};
+
+const ListStates: React.FC<Props> = ({
```
Stuart-Knowles-SW: It's not immediately obvious to me what this component does from its name, could we have something a bit more descriptive (of purpose, not implementation)

src/components/Tables/Table.tsx
```diff
@@ -122,12 +123,16 @@ export type FilterConfig<Filter> =
     | {
           primaryFiltersShown: false;
           additionalFiltersShown: false;
+          listChoiceButton: false;
```
Stuart-Knowles-SW: This is a generic component, can we avoid adding specific properties to it?

The fundamental thing we're adding to the table is a special filter which _must_ have a value. If that can be fit into `primary_filters`, great! If not, then can we add a generic (and optional) field for mandatory filters.

Stuart-Knowles-SW (commented):
Good start! A few little things about convention and naming, etc., but the biggest thing to look at is keeping the generic components generic.

src/app/lists/ListsPage.tsx
```diff
@@ -75,33 +82,60 @@ const formatListData = (listsData: Schema["lists"][]): ListRow[] => {
...
+        label: "Item",
+        headers: listsHeaderKeysAndLabels,
+        method: filterRowByText,
+    }),
+    buttonGroupFilter({
+        key: "listType",
+        filterLabel: "",
+        filterOptions: ["regular", "hotel"],
+        initialActiveFilter: "regular",
+        //method: (row: ListRow, state: string, key: keyof ListRow) => true
```
rjesson-softwire: remove comment

src/app/lists/ListDataview.tsx
```diff
@@ -323,6 +316,12 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
                         additionalFiltersShown: false,
                     }}
                 />
+                <EditModal
+                    onClose={() => setModal(undefined)}
+                    data={modal}
+                    key={modal?.primary_key}
+                    currentList={primaryFilters[1].state === "hotel" ? "hotel" : "regular"}
```
Stuart-Knowles-SW: This looks like a hold-over from capitalisation problems, can this just be `primaryFilters[1].state`?

src/components/Tables/ButtonFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+interface ButtonProps {
+    activeFilter: string;
+    buttonLabel: string;
+    setState: (state: string) => void;
+}
+
+const FilterButton: React.FC<ButtonProps> = (buttonProps) => {
+    const active = buttonProps.activeFilter === buttonProps.buttonLabel;
```
Stuart-Knowles-SW: very minor point, but I'm a fan of boolean naming conventions where boolean names start with either `is` or `should` (or similar). So here I'd prefer `isActive` over `active`

src/components/Tables/ButtonFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+    );
+};
+
+export const buttonGroupFilter = <Data,>({
+    key,
+    filterLabel,
+    filterOptions,
+    initialActiveFilter,
+    method,
+    persistOnClear = false,
```
Stuart-Knowles-SW: similarly here, can we go with `shouldPersistOnClear`

src/components/Tables/Table.tsx
```diff
@@ -127,15 +127,15 @@ export type FilterConfig<Filter> =
...
+          setAdditionalFilters: (additionalFilters: Filter[]) => void;
       }
     | {
           primaryFiltersShown: true;
           primaryFilters: Filter[];
           setPrimaryFilters: (primaryFilters: Filter[]) => void;
-          additionalFiltersShown: true;
-          additionalFilters: Filter[];
-          setAdditionalFilters: (additionalFilters: Filter[]) => void;
+          additionalFiltersShown: false;
```
Stuart-Knowles-SW: It looks like you've just swapped these around, can we revert this to avoid spurious changes

src/components/Tables/TableFilterAndExtraColumnsBar.tsx
```diff
@@ -73,18 +73,30 @@ function TableFilterAndExtraColumnsBar<Data, Filter extends FilterBase<Data, Sta
...
+                props.filters?.map((filter) =>
+                    filter.persistOnClear
+                        ? {
+                              ...filter,
+                          }
+                        : {
+                              ...filter,
+                              state: filter.initialState,
+                          }
+                )
```
Stuart-Knowles-SW: The syntax `{ ...filter }` is (usually) equivalent to `filter`, so we can simplify this by writing:
```
                props.filters?.map((filter) =>
                    filter.persistOnClear
                        ? filter
                        : {
                              ...filter,
                              state: filter.initialState,
                          }
                )
```

And the same below :)

Stuart-Knowles-SW (commented):
A few little style things but this is looking much better now :)

Stuart-Knowles-SW (approved):
Thanks for all the hard work on this, looks great now!

PR #429, opened 2024-07-18
VFB-291 Install Jest and write a unit test

LeaHe188:
## What's changed
Tested getClientParcelsData.ts using Jest.

## Checklist

    - Checkout the dev branch
    - Checkout your branch


package.json
```diff
@@ -89,12 +99,24 @@
     "start-server-and-test": "^2.0.0",
     "styled-components": "^6.0.4",
     "supabase": "^1.142.2",
-    "typescript": "^5.2.2",
```
Stuart-Knowles-SW, line 92: Is there a reason we've moved this out of dependencies into devDependencies?

src/app/clients/getClientParcelsData.test.ts
```diff
@@ -0,0 +1,37 @@
...
+jest.mock("@/logger/logger", () => ({
+    logErrorReturnLogId: jest.fn(() => Promise.resolve(logID)),
+}));
+
+jest.mock("@/supabaseClient", () => {
+    return { default: jest.fn() };
+});
+
+describe("test for getClientparcelsData", () => {
+    test("Should format the data correctly", () => {
```
Stuart-Knowles-SW: Can we use the `it(...` rather than `test`, it's a minor readability thing, but tends to be preferred

Stuart-Knowles-SW (commented):
Looks good, nice work! A couple of things to fix then I can approve it

package.json
LeaHe188: there were conflicts in package.json when I rebase, and I accepted both comments, shall I add this so that the dependencies are not disrupted?

src/app/clients/getClientParcelsData.test.ts
LeaHe188: this is done

package.json
LeaHe188: I've moved it back to dependencies, no errors occurred

Stuart-Knowles-SW (approved):
LGTM :D

PR #433, opened 2024-07-19
VFB-302 fix top row item name to show when deleting from list

rjesson-softwire:
## What's changed
changed logic so when deleting the top item from the lists table, the message appears correctly.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

### DIFF
--- a/src/app/lists/ListDataview.tsx
+++ b/src/app/lists/ListDataview.tsx
@@ -279,7 +279,7 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
         <>
             <ConfirmDialog
                 message={`Are you sure you want to delete ${
-                    toDelete ? listOfIngredients[toDelete].itemName : ""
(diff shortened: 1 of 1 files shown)


src/app/lists/ListDataview.tsx
```diff
@@ -279,7 +279,7 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
         <>
             <ConfirmDialog
                 message={`Are you sure you want to delete ${
-                    toDelete ? listOfIngredients[toDelete].itemName : ""
+                    toDelete != null ? listOfIngredients[toDelete].itemName : ""
```
Stuart-Knowles-SW: You've dropped the second equals 🙈

Stuart-Knowles-SW (approved):
Looks like you dropped the second equals 🙈

PR #434, opened 2024-07-19
VFB-266 Add wiki update button view conditional on user role

SoftJoelWire:
## What's changed
Added update buttons to the wiki page which are conditional on the role of the user (appear only for admins and managers) and give an alert when clicked

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| 
image
| 
Screenshot 2024-07-19 155657
|

## Checklist

### DIFF
--- a/src/app/info/WikiItems.tsx
+++ b/src/app/info/WikiItems.tsx
@@ -4,16 +4,15 @@ import AccordionDetails from "@mui/material/AccordionDetails";
 import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
 import { DbWikiRow } from "@/databaseUtils";
 import UpdateIcon from "@mui/icons-material/Update";
-import {
-    AccordionWrapper,
-    WikiUpdateButton,
-    buttonAlert,
-    AdminManagerDependent,
-    AccordionSurface,
-} from "@/app/info/lib/ClientSideHelpers";
+import { buttonAlert, AdminManagerDependent, AccordionSurface } from "@/app/info/ClientSideHelpers";
+import { AccordionWrapper, WikiUpdateButton } from "@/app/info/StyleComponents";
 
-interface AccordianProps {
+interface AccordionProps {
     rows: DbWikiRow[];
+}
+
+interface AccordionContentProps {
+    row: DbWikiRow;
 }
 
 interface ContentPart {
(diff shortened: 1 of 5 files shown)


src/app/info/lib/ClientSideHelpers.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+interface RoleProps {
+    children?: React.ReactNode;
+}
+
+export const AdminManagerDependent: React.FC<RoleProps> = (props) => {
+    const { role } = useContext(RoleUpdateContext);
+
+    return <>{(role === "admin" || role === "manager") && props.children}</>;
+};
```
Stuart-Knowles-SW: My understanding is that this is the only part of this file (also the alert I guess but that's temporary) which needs to be client side. Can we put the  styled components in their own server side fille and keep these here?

src/components/Tables/TableSurface.tsx
```diff
@@ -10,7 +10,7 @@ const StyledPaper = styled(Paper)`
     border-radius: 1rem;
 `;
 
-interface Props {
+export interface Props {
```
Stuart-Knowles-SW, line 13: For something this generically named and small, I'd rather you just redefine it where you need it. Good job reusing code, but this is a case where we should keep it simple

src/app/info/WikiItems.tsx
```diff
@@ -49,20 +56,27 @@ const WikiItems: React.FC<AccordianProps> = (props) => {
...
+                                    id="panel1-header"
+                                >
+                                    <h2>{row.title}</h2>
+                                </AccordionSummary>
+                                <AccordionDetails>
+                                    {convertContentToElements(row.content)}
+                                </AccordionDetails>
+                            </Accordion>
+                        </AccordionSurface>
+                    </AccordionWrapper>
```
Stuart-Knowles-SW: Can we separate this out into its own component? It can stay in the same file, but as it grows I think we can benefit from keeping it separate

Stuart-Knowles-SW (commented):
Good work, a few minor changes but mainly just moving stuff around

Stuart-Knowles-SW (commented):
As discussed in person, I've requested a few changes to naming to better fit the "Name for usage, not implementation" principle, then this should be good to go :)

Stuart-Knowles-SW (approved):
Looks great now, thanks for doing those refactorings and renamings!

PR #437, opened 2024-07-23
VFB 264 updating wiki item title and content

SoftJoelWire:
## What's changed
Can now edit, update, delete and cancel edits on the info page for wiki items.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| 
image
|image|

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/info/ClientSideHelpers.tsx
```diff
@@ -1,18 +1,33 @@
...
-export const buttonAlert = (): void => {
-    alert("Wiki item has been updated!");
+export const enterEditMode = (setEditMode: (editMode: boolean) => void): void => {
+    setEditMode(true);
 };
 
-interface RoleProps {
-    children?: React.ReactNode;
+interface EditProps {
+    editMode: boolean;
```
rjesson-softwire: since this is a boolean, could change name to something along the lines of isInEditMode

src/app/info/DefaultWikiItemView.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+interface DefaultViewProps {
+    rowRef?: DbWikiRow;
+    setEditModeRef: (editMode: boolean) => void;
+}
+
+const DefaultWikiItemView: React.FC<DefaultViewProps> = ({ rowRef, setEditModeRef }) => {
+    return (
+        <>
+            {rowRef && (
```
rjesson-softwire: would we be able to move the rowRef condition to where this component is called, rather than in the top line of the component?

src/app/info/ClientSideHelpers.tsx
```diff
@@ -1,18 +1,33 @@
...
+    row?: DbWikiRow;
 }
 
-export const AdminManagerDependent: React.FC<RoleProps> = (props) => {
-    const { role } = useContext(RoleUpdateContext);
-
-    return <>{(role === "admin" || role === "manager") && props.children}</>;
+export const EditModeDependentItem: React.FC<EditProps> = ({ editMode, row }) => {
+    const [rowRef, setRowRef] = useState<DbWikiRow | undefined>(row);
+    const [editModeRef, setEditModeRef] = useState(editMode);
```
rjesson-softwire: as above - could switch to boolean naming convention to make it clearer this is a boolean?

src/app/info/EditWikiItemView.tsx
```diff
@@ -0,0 +1,130 @@
...
+    setEditModeRef: (editMode: boolean) => void;
+}
+
+const deleteWikiItem = async (
+    setEditModeRef: (editMode: boolean) => void,
+    setRowRef: (row?: DbWikiRow) => void,
+    rowRef?: DbWikiRow
+): Promise<void> => {
+    if (rowRef) {
+        const result = confirm("Confirm deletion of this item?");
```
rjesson-softwire: minor point: could change naming of result to something like confirmation

src/app/info/EditWikiItemView.tsx
```diff
@@ -0,0 +1,130 @@
...
+        if (result) {
+            const response = await supabase
+                .from("wiki")
+                .delete()
+                .match({ wiki_key: rowRef.wiki_key });
+            if (response.error) {
+                logErrorReturnLogId("error fetching wiki data", response.error);
```
rjesson-softwire: could you change the audit log message here to be about deleting an item from wiki rather than fetching data?

src/app/info/EditWikiItemView.tsx
```diff
@@ -0,0 +1,130 @@
...
+    setRowRef: (row: DbWikiRow) => void
+): Promise<void> => {
+    const result = confirm("Confirm update of this item?");
+    if (result) {
+        const response = await supabase
+            .from("wiki")
+            .update({ title: newTitle, content: newContent })
+            .match({ wiki_key: rowRef.wiki_key });
+        if (response.error) {
+            logErrorReturnLogId("error fetching wiki data", response.error);
```
rjesson-softwire: as above

src/app/info/EditWikiItemView.tsx
```diff
@@ -0,0 +1,130 @@
...
+            row_order: rowRef.row_order,
+            wiki_key: rowRef.wiki_key,
+        });
+    }
+};
+
+const EditWikiItemView: React.FC<EditViewProps> = ({ rowRef, setRowRef, setEditModeRef }) => {
+    return (
+        <>
+            {rowRef && (
```
rjesson-softwire: as above

rjesson-softwire (changes requested):

src/app/info/ClientSideHelpers.tsx
```diff
@@ -5,29 +5,30 @@ import { useState } from "react";
 import DefaultWikiItemView from "@/app/info/DefaultWikiItemView";
 import EditWikiItemView from "@/app/info/EditWikiItemView";
 
-export const enterEditMode = (setEditMode: (editMode: boolean) => void): void => {
-    setEditMode(true);
+export const enterEditMode = (setIsInEditMode: (editMode: boolean) => void): void => {
```
rjesson-softwire: for consistency there is one editMode left here to be changed

src/app/info/EditWikiItemView.tsx
```diff
@@ -10,54 +10,54 @@ import supabase from "@/supabaseClient";
 import { logErrorReturnLogId } from "@/logger/logger";
 
 interface EditViewProps {
-    rowRef?: DbWikiRow;
+    rowRef: DbWikiRow;
     setRowRef: (row?: DbWikiRow) => void;
-    setEditModeRef: (editMode: boolean) => void;
+    setIsInEditModeRef: (editMode: boolean) => void;
```
rjesson-softwire: as above with editMode

src/app/info/EditWikiItemView.tsx
```diff
@@ -10,54 +10,54 @@ import supabase from "@/supabaseClient";
...
 }
 
 const deleteWikiItem = async (
-    setEditModeRef: (editMode: boolean) => void,
+    setIsInEditModeRef: (editMode: boolean) => void,
```
rjesson-softwire: and here

src/app/info/ClientSideHelpers.tsx
```diff
@@ -5,29 +5,30 @@ import { useState } from "react";
...
-            {!editModeRef ? (
-                <DefaultWikiItemView rowRef={rowRef} setEditModeRef={setEditModeRef} />
-            ) : (
-                <EditWikiItemView
-                    rowRef={rowRef}
-                    setRowRef={setRowRef}
-                    setEditModeRef={setEditModeRef}
-                />
-            )}
+            {rowRef &&
```
rjesson-softwire: would we be able to move the rowRef condition even further out? i don't think EditModeDependentItem needs to be called if row is undefined

src/app/info/EditWikiItemView.tsx
```diff
@@ -10,54 +10,54 @@ import supabase from "@/supabaseClient";
...
-        if (result) {
+        const confirmation = confirm("Confirm deletion of this item?");
+        if (confirmation) {
             const response = await supabase
                 .from("wiki")
                 .delete()
                 .match({ wiki_key: rowRef.wiki_key });
             if (response.error) {
-                logErrorReturnLogId("error fetching wiki data", response.error);
+                logErrorReturnLogId("error deleting item", response.error);
```
rjesson-softwire: sorry to ask for an edit on this again - could we still specify that its a wiki row we're deleting rather than a possibly too generic 'item'

src/app/info/EditWikiItemView.tsx
```diff
@@ -10,54 +10,54 @@ import supabase from "@/supabaseClient";
...
-    if (result) {
+    const confirmation = confirm("Confirm update of this item?");
+    if (confirmation) {
         const response = await supabase
             .from("wiki")
             .update({ title: newTitle, content: newContent })
             .match({ wiki_key: rowRef.wiki_key });
         if (response.error) {
-            logErrorReturnLogId("error fetching wiki data", response.error);
+            logErrorReturnLogId("error updating item", response.error);
```
rjesson-softwire: as above

rjesson-softwire (changes requested):

src/app/info/ClientSideHelpers.tsx
SoftJoelWire: rowRef can only become undefined once deleted in EditWikiItemView. This is the highest level client side file in the info route and we need to store rowRef as a ref so that it can be dynamically updated (deleted). We cannot store this ref any higher as that would require storing it on a SSR page which does not support the useState hook.

rjesson-softwire (commented):
LGTM! 👍

src/app/info/EditWikiItemView.tsx
```diff
@@ -0,0 +1,126 @@
...
+                        defaultValue={rowRef.content}
+                        multiline
+                        rows={4}
+                        margin="normal"
+                        fullWidth={true}
+                    />
+                </div>
+                <WikiEditModeButton onClick={() => cancelWikiItemEdit(setIsInEditModeRef)}>
+                    {" "}
+                    <CancelIcon />{" "}
```
Stuart-Knowles-SW: I think this might have been the linter misinterpreting whitespace?

src/app/info/EditWikiItemView.tsx
```diff
@@ -0,0 +1,126 @@
...
+                            rowRef,
+                            title_input.value,
+                            content_input.value,
+                            setIsInEditModeRef,
+                            setRowRef
+                        );
+                    }}
+                >
+                    {" "}
+                    <SaveAltIcon />{" "}
```
Stuart-Knowles-SW: And here

src/app/info/ClientSideHelpers.tsx
```diff
@@ -1,18 +1,34 @@
...
+export const EditModeDependentItem: React.FC<EditProps> = ({ isInEditMode: isInEditMode, row }) => {
+    const [rowRef, setRowRef] = useState<DbWikiRow | undefined>(row);
+    const [isInEditModeRef, setIsInEditModeRef] = useState(isInEditMode);
+    return (
+        <>
+            {rowRef &&
+                (!isInEditModeRef ? (
+                    <DefaultWikiItemView rowRef={rowRef} setIsInEditModeRef={setIsInEditModeRef} />
+                ) : (
+                    <EditWikiItemView
```
Stuart-Knowles-SW: "Default" isn't very clear by itself what the intention is. I think names like `WikiItemDisplay` and `WikiItemEdit` could be more clear

src/app/info/ClientSideHelpers.tsx
```diff
@@ -1,18 +1,34 @@
...
-    const { role } = useContext(RoleUpdateContext);
-
-    return <>{(role === "admin" || role === "manager") && props.children}</>;
+export const EditModeDependentItem: React.FC<EditProps> = ({ isInEditMode: isInEditMode, row }) => {
+    const [rowRef, setRowRef] = useState<DbWikiRow | undefined>(row);
+    const [isInEditModeRef, setIsInEditModeRef] = useState(isInEditMode);
+    return (
+        <>
+            {rowRef &&
+                (!isInEditModeRef ? (
```
Stuart-Knowles-SW: just for simplicity, can we remove the `!` and reverse the order of the options?

src/app/info/ClientSideHelpers.tsx
```diff
@@ -1,18 +1,34 @@
...
+    row?: DbWikiRow;
 }
 
-export const AdminManagerDependent: React.FC<RoleProps> = (props) => {
-    const { role } = useContext(RoleUpdateContext);
-
-    return <>{(role === "admin" || role === "manager") && props.children}</>;
+export const EditModeDependentItem: React.FC<EditProps> = ({ isInEditMode: isInEditMode, row }) => {
+    const [rowRef, setRowRef] = useState<DbWikiRow | undefined>(row);
+    const [isInEditModeRef, setIsInEditModeRef] = useState(isInEditMode);
```
Stuart-Knowles-SW: Refs are something very specific in react, can we avoid using it in names of things that aren't refs? I think
```
    const [rowData, setRowData] = useState<DbWikiRow | undefined>(row);
    const [isInEditMode, setIsInEditMode] = useState(isInEditMode);
```
Should be fine

src/app/info/DefaultWikiItemView.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { WikiItemAccordionSurface, WikiUpdateDataButton } from "@/app/info/StyleComponents";
+import UpdateIcon from "@mui/icons-material/Update";
+import { Accordion, AccordionDetails, AccordionSummary } from "@mui/material";
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
+import { convertContentToElements } from "@/app/info/WikiItems";
+import AdminManagerDependent from "@/app/info/AdminManagerDependent";
+
+interface DefaultViewProps {
+    rowRef: DbWikiRow;
+    setIsInEditModeRef: (isInEditMode: boolean) => void;
```
Stuart-Knowles-SW: Given that we only ever call this with `true` in this file, can we change it to an argument-less function then pass it through in the calling component?
So here it's
```
interface DefaultViewProps {
    rowRef: DbWikiRow;
    openEditMode: () => void;
...
                <WikiUpdateDataButton
                    onClick={openEditMode}
                >
```
and where this is called:
`<DefaultWikiItemView rowData={row} openEditMode={() => setIsinEditMode(true)} />`

src/app/info/EditWikiItemView.tsx
```diff
@@ -0,0 +1,126 @@
...
+    setRowRef: (row?: DbWikiRow) => void;
+    setIsInEditModeRef: (isInEditMode: boolean) => void;
+}
+
+const deleteWikiItem = async (
+    setIsInEditModeRef: (isInEditMode: boolean) => void,
+    setRowRef: (row?: DbWikiRow) => void,
+    rowRef?: DbWikiRow
+): Promise<void> => {
+    if (rowRef) {
```
Stuart-Knowles-SW: rowRef seems like it's always defined when this is called?

src/app/info/EditWikiItemView.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+interface EditViewProps {
+    rowRef: DbWikiRow;
+    setRowRef: (row?: DbWikiRow) => void;
+    setIsInEditModeRef: (isInEditMode: boolean) => void;
+}
+
+const deleteWikiItem = async (
+    setIsInEditModeRef: (isInEditMode: boolean) => void,
+    setRowRef: (row?: DbWikiRow) => void,
```
Stuart-Knowles-SW: Similarly here with argument-less functions

src/app/info/ClientSideHelpers.tsx
```diff
@@ -1,18 +1,34 @@
...
+        <>
+            {rowRef &&
+                (!isInEditModeRef ? (
+                    <DefaultWikiItemView rowRef={rowRef} setIsInEditModeRef={setIsInEditModeRef} />
+                ) : (
+                    <EditWikiItemView
+                        rowRef={rowRef}
+                        setRowRef={setRowRef}
+                        setIsInEditModeRef={setIsInEditModeRef}
+                    />
```
Stuart-Knowles-SW: Can we also wrap this in an admin/manager requirement? It can still be accessed by staff currently with a little bit of "behind the scenes work", even if the database calls get blocked

src/app/info/EditWikiItemView.tsx
Stuart-Knowles-SW, line 1: This is mostly just preference on my part, but can we change some icons? Can we switch the position of the cancel and delete buttons (so cancel appears in the same place that edit does), and switch the edit icon to a pencil icon?

Stuart-Knowles-SW (commented):
Few little bits to look at, but nothing huge about the structure

src/app/info/EditWikiItemView.tsx
SoftJoelWire: line 34 of WikiItemEdit sets the rowData to be undefined on deletion

src/app/info/EditWikiItemView.tsx
SoftJoelWire: Similar reason as above

src/app/info/AddWikiItemButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+interface WikiRowQuerySuccessType {
+    data: DbWikiRow;
+    error: null;
+}
+
+interface WikiRowQueryFailureType {
+    data: null;
+    error: PostgrestError;
+}
+export type WikiRowQueryType = WikiRowQuerySuccessType | WikiRowQueryFailureType;
```
rjesson-softwire, line 19: is this the best place for these types to be defined? would it make sense to move them to say the fetch file

src/app/info/EditModeDependentItem.tsx
```diff
@@ -0,0 +1,41 @@
...
+    row?: DbWikiRow;
+    sortedRows: DbWikiRow[];
+}
+
+const EditModeDependentItem: React.FC<EditProps> = ({ row, sortedRows }) => {
+    const [rowData, setrowData] = useState<DbWikiRow | undefined>(row);
+    const [isInEditMode, setIsInEditMode] = useState<boolean>(false);
+    return (
+        <>
+            {rowData !== undefined &&
```
rjesson-softwire: the first line of this component is a conditional, can we move the conditional out to the place where the component is called

src/app/info/WikiItemEdit.tsx
```diff
@@ -0,0 +1,140 @@
...
+    sortedRows: DbWikiRow[];
+}
+
+const WikiItemEdit: React.FC<WikiItemEditProps> = ({
+    rowData,
+    setRowData,
+    setIsInEditMode,
+    sortedRows,
+}) => {
+    const deleteWikiItemWithoutConfirmation = async (): Promise<void> => {
```
rjesson-softwire: don't love the naming of this function - i get it's to make it clear that this function doesn't include the confirmation, but it feels a little like it's implying there is no confirmation at all, rather than it being elsewhere.

rjesson-softwire (dismissed):

src/app/info/WikiItems.tsx
```diff
@@ -52,40 +48,40 @@ export const convertContentToElements = (rowContent: string): React.JSX.Element[
...
+    );
+
+    const wikiItemsEndRef = useRef<HTMLDivElement>(null);
+    useEffect(() => {
+        if (wikiItemsEndRef.current) {
+            wikiItemsEndRef.current.scrollIntoView({
+                behavior: "smooth",
+            });
+        }
+    }, [sortedRows]);
```
Stuart-Knowles-SW: Nice!

src/app/info/WikiItemEdit.tsx
```diff
@@ -0,0 +1,138 @@
...
+            .from("wiki")
+            .delete()
+            .match({ wiki_key: rowData.wiki_key })) as WikiRowQueryType;
+        if (deleteResponse.error) {
+            logErrorReturnLogId("error deleting wiki row item", deleteResponse.error);
+        }
+        const indexToRemove: number = sortedRows.findIndex(
+            (row) => row.wiki_key === rowData.wiki_key
+        );
+        sortedRows.splice(indexToRemove, 1);
```
Stuart-Knowles-SW: I think this should be handled by a state setter, not by mutating the state

src/app/info/WikiItemEdit.tsx
```diff
@@ -0,0 +1,138 @@
...
+import supabase from "@/supabaseClient";
+import { logErrorReturnLogId } from "@/logger/logger";
+import { WikiRowQueryType } from "@/app/info/AddWikiItemButton";
+
+interface WikiItemEditProps {
+    rowData: DbWikiRow;
+    setRowData: (row?: DbWikiRow) => void;
+    setIsInEditMode: (isInEditMode: boolean) => void;
+    sortedRows: DbWikiRow[];
+}
```
Stuart-Knowles-SW: Two questions here: Firstly, does the edit component really need the entire array of `sortedRows` as a property? And secondly, can `setIsInEditMode` be replaced by `closeEditMode` since we only ever call it with false here

src/app/info/AddWikiItemButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+    };
+
+    return (
+        <ButtonMargin>
+            <Button variant="contained" onClick={addWikiItem}>
+                + Add
+            </Button>
+        </ButtonMargin>
+    );
+};
```
Stuart-Knowles-SW: There's a bug here, something to do with `setSortedRows` I think. If you edit an item and save it, then click "Add item", it duplicates the item you just edited. This only happens in the front end, nothing changes in the database.

src/app/info/AddWikiItemButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+    setSortedRows: (rows: DbWikiRow[]) => void;
+}
+
+const AddWikiItemButton: React.FC<AddWikiItemButtonProps> = ({ sortedRows, setSortedRows }) => {
+    const addWikiItem = async (): Promise<void> => {
+        if (
+            sortedRows.filter((row) => {
+                return !row.title && !row.content;
+            }).length === 0
+        ) {
```
Stuart-Knowles-SW: If this is the only thing we need sorted rows for, can we calculate that in the parent rather than passing the whole array through?

Stuart-Knowles-SW (commented):
Really nice, you've clearly thought a lot about functionality and edge cases. Going forward, I'd like to try and adhere to the principle that each component should only be passed the properties it _needs_. The `AddWikiItemButton` is a good example: it doesn't need the whole array of sorted rows, justwhether any of them are without content, so we should calculate that in the parent.

Also there's one weird bug with adding items, I think it's from you mutating state directly rather than using a setter but I haven't confirmed that

package-lock.json
```diff
@@ -17585,4 +17585,4 @@
       }
     }
   }
-}
+}
```
Stuart-Knowles-SW: This looks like a spurious change, can we remove it?

src/app/info/AddWikiItemButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+        if (!doesEmptyRowExist) {
+            const { data, error } = (await supabase
+                .from("wiki")
+                .insert({})
+                .select()
+                .single()) as WikiRowQueryType;
+            error
+                ? logErrorReturnLogId("error inserting and fetching new data", error)
+                : appendNewRow(data, -1);
+        }
```
Stuart-Knowles-SW: This is a very nit-picky, stylistic comment, but can we use the early-return pattern here? The idea is to do validation checks at the start of a function and return if they aren't met. This allows you to nest as little code inside if statements as possible, and to keep the main purpose of the function at the root, handling the edge cases in `if`s. In practice, that'll look like this:
```
    const addWikiItem = async (): Promise<void> => {
        if (doesEmptyRowExist) {
            return;
        }
        const { data, error } = (await supabase
            .from("wiki")
            .insert({})
            .select()
            .single()) as WikiRowQueryType;
        error
            ? logErrorReturnLogId("error inserting and fetching new data", error)
            : appendNewRow(data, -1);
    };
```

src/app/info/EditModeDependentItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+    appendNewRow: (newRow: DbWikiRow, index: number) => void;
+    removeRow: (row: DbWikiRow) => number;
+}
+
+const EditModeDependentItem: React.FC<EditProps> = ({ row, appendNewRow, removeRow }) => {
+    const [rowData, setrowData] = useState<DbWikiRow | undefined>(row);
+    const [isInEditMode, setIsInEditMode] = useState<boolean>(false);
+    return (
+        <>
+            {rowData !== undefined &&
```
Stuart-Knowles-SW: Since `undefined` evaluates to false I think we can simplify this to `{rowData && ...`

src/app/info/WikiItems.tsx
```diff
@@ -52,40 +49,83 @@ export const convertContentToElements = (rowContent: string): React.JSX.Element[
...
+
+    const [doesEmptyRowExist, setDoesEmptyRowExist] = useState<boolean>(false);
+
+    useMemo(() => {
+        setDoesEmptyRowExist(
+            displayRows.filter((row) => {
+                return !row.title && !row.content;
+            }).length !== 0
+        );
+    }, [displayRows]);
```
Stuart-Knowles-SW: `useMemo` is for caching values, so we can completely avoid state here (I've also rewritten the function for a bit more conciseness):
```
    const doesEmptyRowExist = useMemo(
        () => displayRows.some((row) => !row.title && !row.content),
        [displayRows]
    );
```

Stuart-Knowles-SW (commented):
A few more, but we're nearly over the line

Stuart-Knowles-SW (approved):
LGTM!

PR #440, opened 2024-07-24
VFB-280 update client creation to specify default list type

rjesson-softwire:
## What's changed
create new mandatory dropdown field in client creation / edit forms to specify which default list type the client belongs to. This information is stored in the clients table in the database on submit.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

    - Checkout the dev branch
    - Checkout your branch

---

* **Introduction of List Type Selection Feature during Form Submission**
The addition of the `chooseListType` function now enables users to select the type of list as "Regular" while filling out a form, enhancing the user's experience and providing a more personalized use of the service.

* **Inclusion of List Type Option in Various Client Forms**
The introduction of the `listType` field in the form pages of 'add new client' and 'edit client,' provides the functionality for users to select the type of list while adding or editing a Client in the application. An option for list type input is now present in the client form released with the `listTypeCard.tsx` component.

* **Ensuring Default List Type in Client and Family Operations**
The creation of new migration files support the inclusion of default list types in the insert and update SQL functions. This change facilitates maintaining uniformity and standardization throughout the client and family functions by always including a default list type.

### DIFF
--- a/src/app/clients/form/ClientForm.tsx
+++ b/src/app/clients/form/ClientForm.tsx
@@ -62,7 +62,7 @@ export interface ClientFields extends Fields {
     numberOfAdults: number;
     children: Person[];
     numberOfChildren: number;
-    listType: ListName;
+    listType: ListName | "";
     dietaryRequirements: BooleanGroup;
     feminineProducts: BooleanGroup;
     babyProducts: boolean | null;
--- a/src/app/clients/form/ClientForm.tsx
+++ b/src/app/clients/form/ClientForm.tsx
@@ -39,6 +39,7 @@ import Title from "@/components/Title/Title";
 import ClientNotesCard from "@/app/clients/form/formSections/ClientNotes";
 import { getCurrentYear } from "@/common/date";
 import { youngestAdultBirthYear } from "@/app/clients/form/birthYearDropdown";
+import ListTypeCard from "./formSections/ListTypeCard";
 
 interface Props {
     initialFields: ClientFields;
@@ -60,6 +61,7 @@ export interface ClientFields extends Fields {
     numberOfAdults: number;
     children: Person[];
     numberOfChildren: number;
+    listType: string;
     dietaryRequirements: BooleanGroup;
     feminineProducts: BooleanGroup;
     babyProducts: boolean | null;
@@ -81,6 +83,7 @@ export interface ClientErrors extends FormErrors<ClientFields> {
     addressPostcode: Errors;
     numberOfAdults: Errors;
     numberOfChildren: Errors;
+    listType: Errors;
     nappySize: Errors;
 }
 
@@ -94,6 +97,7 @@ const formSections = [
     AddressCard,
     NumberAdultsCard,
     NumberChildrenCard,
+    ListTypeCard,
     DietaryRequirementCard,
     FeminineProductCard,
     BabyProductCard,
@@ -152,6 +156,9 @@ const ClientForm: React.FC<Props> = ({ initialFields, initialFormErrors, editCon
     const errorSetter = createSetter(setFormErrors, formErrors);
 
     const submitForm = async (): Promise<void> => {
+
+        console.log("Submitting form:", fields.listType);
+
         setSubmitDisabled(true);
 
         const inputError = checkErrorOnSubmit(formErrors, setFormErrors);
--- a/src/app/clients/form/ClientForm.tsx
+++ b/src/app/clients/form/ClientForm.tsx
@@ -40,6 +40,7 @@ import ClientNotesCard from "@/app/clients/form/formSections/ClientNotes";
 import { getCurrentYear } from "@/common/date";
 import { youngestAdultBirthYear } from "@/app/clients/form/birthYearDropdown";
 import ListTypeCard from "./formSections/ListTypeCard";
+import { ListName } from "@/app/lists/ListDataview";
 
 interface Props {
     initialFields: ClientFields;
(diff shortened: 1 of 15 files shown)


src/app/clients/form/formSections/ListTypeCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const listTypeLabelsAndValues: ListTypeLabelsAndValues = [
+        ["Regular", "regular"],
+        ["Hotel", "hotel"],
+    ];
+
+    return (
+        <GenericFormCard
+            title="List Type"
+            required={true}
+            text="From which list should parcels for this client be packed using?"
```
ebinezer04: Happy with the actual code, solution looks good. 

Would there be a better way to word this list type question?

ebinezer04 (commented):
Code looks good :)

ebinezer04 (approved):

src/common/fetch.ts
```diff
@@ -195,6 +196,10 @@ export const getActiveTimeSlotsForCollectionCentre = async (
     return { data: activeTimeSlots, error: null };
 };
 
+export type ListName = Database["public"]["Enums"]["list_type"];
```
Stuart-Knowles-SW: If you import the `Enums` object from the Database types file, you can write this as `export type ListName = Enums<"list_type">;`, which I think is slightly more concise and understandable, but I'm not hugely fussed by it if you prefer this way

src/common/fetch.ts
```diff
@@ -195,6 +196,10 @@ export const getActiveTimeSlotsForCollectionCentre = async (
...
+
+export const listTypes = ["regular", "hotel"] as ListName[];
```
Stuart-Knowles-SW: Can we keep this simple and just call them `ListType` and `LIST_TYPES_ARRAY`? In hindsight, choosing a name that ends in `type` may not have been a brilliant idea, but we want to make it clear the two are linked. The all caps syntax is fairly standard for constant values.

We can also improve the type safety of the array in two ways: firstly, by using the variable declaration `const listTypes: ListName[]` rather than `as ListName[]`: the former means that if the expression we write on the right hand side isn't the right type the compiler will complain, whereas the latter overrides any typechecking (try it, add anything into the array as it is and it won't complain). Secondly, we can use an `as const` assertion to make sure the compiler doesn't let us mutate the array, this is slightly different than `as Type`, in that it tells Typescript to use the most narrow type available for the variable.

Putting all of this together, the whole two lines would look like:
```
export type ListType = Enums<"list_type">;

export const LIST_TYPES_ARRAY: ListType[] = ["regular", "hotel"] as const;
```

Stuart-Knowles-SW: That's a bit of a dense comment, lmk if you want to chat through it 😅

src/app/clients/form/formSections/ListTypeCard.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import { errorText, valueOnChangeDropdownList } from "@/components/Form/formFunctions";
+import GenericFormCard from "@/components/Form/GenericFormCard";
+import { ErrorText } from "@/components/Form/formStyling";
+import DropdownListInput from "@/components/DataInput/DropdownListInput";
+import { ClientCardProps } from "../ClientForm";
+import { listTypes } from "@/common/fetch";
+import { capitaliseWords } from "@/common/format";
+
+type ListTypeLabelsAndValues = [string, string][];
```
Stuart-Knowles-SW, line 10: I really like this. It's not a brilliant type where it's defined in `DropdownListInput.tsx` – we tend to prefer objects to arrays when the length is not variable – but creating an explicit type here makes the usage in this file really clear, which is very helpful :D

src/app/clients/form/ClientForm.tsx
```diff
@@ -60,6 +62,7 @@ export interface ClientFields extends Fields {
     numberOfAdults: number;
     children: Person[];
     numberOfChildren: number;
+    listType: ListName | "";
```
Stuart-Knowles-SW: How difficult would it be to have this as `ListName | null`? In general, null is the closest thing we have to the absence of a value, and here it has the benefit that we can't accidentally do string operations on an empty string which would then lose type information. At display time we'd then write something like `value ?? ""` to convert it to an empty string before display

Stuart-Knowles-SW (commented):
Looks good, just a few little changes :)

Stuart-Knowles-SW:
Also, we do an interactive rebase to remove the commits from VFB-277 before we merge?

src/app/clients/form/formSections/ListTypeCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <GenericFormCard
+            title="List Type"
+            required={true}
+            text="Which list should be used by default to pack parcels for this client?"
+        >
+            <DropdownListInput
+                selectLabelId="list-type-select-label"
+                labelsAndValues={listTypeLabelsAndValues}
+                listTitle="List Type"
+                defaultValue={fields.listType ? fields.listType : ""}
```
Stuart-Knowles-SW: We can use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) here (the other nice thing about using `Type | null`): 
`    defaultValue={fields.listType ?? ""}`

Stuart-Knowles-SW (approved):
One tiny change, but not one I need to see again before merging!

PR #443, opened 2024-07-29
VFB-279 create correct shopping list

rjesson-softwire:
## What's changed
The 'download shopping list' action on the parcels page now creates a pdf that only shows items which are in the list type of the parcel.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image|

## Checklist

---

* **A New Way to Identify Lists Implemented**
  - The code in the files `fetch.ts`, `getParcelsData.ts` and `shoppingListPdfDataProps.ts` was updated to include a new reference for `ListName` from `"@/app/lists/ListDataview"`. This means the program will now be using a specific reference to identify various lists.
  
* **Changes to Shopping List PDF Representation**
  - The `ShoppingListPdf.tsx` file has undergone several changes, aimed at using a new type of item representation, identified as `ItemForPDF`, instead of the previously used `Item`. This is an enhancement to provide more detailed or specific information in the shopping list PDF document.
  
* **Parcel Information Enhancement**
  - The `getParcelsData.ts` was updated to include a new `listType` property in the `ParcelInfo` interface and the function `prepareParcelInfo` now sets this new property. This augments the parcel information with the type of list it associated with.
  
* **Filtering of Shopping List Data**
  - The code to generate shopping list data (`getShoppingListData.ts`) has been revised. Now, it includes a new filtering function that extracts items based on the type of list they belong to.
  
* **Improved Item Interface and Shopping List PDF Data**
  - The `Item` interface has been renamed and modified to `ItemWithListType` in `shoppingListPdfDataProps.ts`. Additionally, different types and functions were updated to use `ItemForPDF` or `ItemWithListType` instead of `Item`. This shift allows for a more precise representation of items in the shopping list PDF.
  
* **New Item Filtering Function**
  - A newly added function `filterItemsbyListType` in `shoppingListPdfDataProps.ts` allows filtering an array of items based on a given list type, enabling more specific data handling based on list categories.
  
These adjustments overall improve how lists and items are represented and how shopping list data is generated and manipulated, making it more efficient and aligned with specific list types.

### DIFF
--- a/src/pdf/ShoppingList/shoppingListPdfDataProps.ts
+++ b/src/pdf/ShoppingList/shoppingListPdfDataProps.ts
@@ -7,15 +7,14 @@ import { Schema } from "@/databaseUtils";
 import { logErrorReturnLogId } from "@/logger/logger";
 import { ListName } from "@/app/lists/ListDataview";
 
-export interface Item {
+export interface ItemWithListType extends ItemForPDF {
+    listType: ListName;
+}
+
+export interface ItemForPDF {
     description: string;
     quantity: string;
     notes: string;
-}
-
-export interface ItemExtendedWithListType {
-    itemOutputContent: Item;
-    listType: ListName;
 }
 
 export interface ShoppingListPdfData {
@@ -24,13 +23,13 @@ export interface ShoppingListPdfData {
     clientSummary: ClientSummary;
     householdSummary: HouseholdSummary;
     requirementSummary: RequirementSummary;
-    itemsList: Item[];
+    itemsList: ItemForPDF[];
     endNotes: string;
 }
 
 type PrepareItemsListResult =
     | {
-          data: ItemExtendedWithListType[];
+          data: ItemWithListType[];
           error: null;
       }
     | {
@@ -53,7 +52,7 @@ function numberIsValidFamilySize(value: number): value is FamilySize {
 export type GetQuantityAndNotesErrorType = "invalidFamilySize";
 export type GetQuantityAndNotesError = { type: GetQuantityAndNotesErrorType; logId: string };
 type GetQuantityAndNotesResult =
-    | { data: Pick<Item, "quantity" | "notes">; error: null }
+    | { data: Pick<ItemForPDF, "quantity" | "notes">; error: null }
     | { data: null; error: GetQuantityAndNotesError };
 
 const getQuantityAndNotes = async (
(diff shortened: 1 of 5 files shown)


src/pdf/ShoppingList/shoppingListPdfDataProps.ts
```diff
@@ -89,10 +91,11 @@ export const prepareItemsListForHousehold = async (
         if (listItemError) {
             return { data: null, error: listItemError };
         }
-        itemsList.push({
-            description: row.item_name,
-            ...listItemData,
-        });
+        row.list_type === listType &&
```
SoftJoelWire, line 94: this pattern for type checking seems a little strange in a .ts file but seems to work so no real issue. maybe a basic if statement would be better?

src/pdf/ShoppingList/shoppingListPdfDataProps.ts
```diff
@@ -89,10 +91,11 @@ export const prepareItemsListForHousehold = async (
...
+            itemsList.push({
+                description: row.item_name,
+                ...listItemData,
```
SoftJoelWire: this pattern for type checking seems a little strange in a .ts file but seems to work so no real issue. maybe a basic if statement would be better?

src/pdf/ShoppingList/shoppingListPdfDataProps.ts
```diff
@@ -89,10 +90,11 @@ export const prepareItemsListForHousehold = async (
         if (listItemError) {
             return { data: null, error: listItemError };
         }
-        itemsList.push({
-            description: row.item_name,
-            ...listItemData,
-        });
+        row.list_type === listType &&
```
Stuart-Knowles-SW: Can we move this before all the processing, so the check is the first thing in the for loop and we don't do any processing if we don't need to? We can use the "Early return pattern" (ask Joel and Leah) and a `continue` statement, so it'll look like this:
```
for (const row of listData) {
        if (row.list_type !== listType) {
            continue
        }
        const { data: listItemData, error: listItemError } = await getQuantityAndNotes(
            row,
            householdSize
        );
        if (listItemError) {
            return { data: null, error: listItemError };
        }
        itemsList.push({
            description: row.item_name,
            ...listItemData,
        });
    }
```

Stuart-Knowles-SW (commented):
Looks good, one small change :)

Stuart-Knowles-SW (approved):
Nice work :)

PR #444, opened 2024-07-30
VFB-305 add list type to expanded parcel details modal

Jessica-Al:
## What's changed
Parcel details modal includes list type

## Screenshots / Videos
Before
image
After
image

## Checklist

---

* **Implementation of List Type in getExpandedParcelDetails Functions**
The functions for gathering detailed information about parcels, specific `getExpandedParcelDetails` and `getExpandedParcelDataForDataViewer`, are now enhanced with an additional property, `listType`. This property was added to their relevant interfaces (`FetchExpandedParcelDetailsResult` and `ParcelDataForInactiveClient`), providing more comprehensive data about each parcel. Through this update, we've improved the depth of parcel information that can be fetched, thereby enhancing data accuracy and comprehension.

### DIFF
--- a/src/app/parcels/getExpandedParcelDetails.ts
+++ b/src/app/parcels/getExpandedParcelDetails.ts
@@ -8,13 +8,14 @@ import {
     formatBreakdownOfChildrenFromFamilyDetails,
     formatHouseholdFromFamilyDetails,
 } from "@/app/clients/getExpandedClientDetails";
-import { formatDateTime, formatDatetimeAsDate } from "@/common/format";
+import { capitaliseWords, formatDateTime, formatDatetimeAsDate } from "@/common/format";
 import {
     Data,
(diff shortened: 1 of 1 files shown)


ebinezer04:
From lines 118 to 170, the returned expanded parcel data appears to be the same, except for isActive being true or false. Would it be worth investigating a way to not repeat listing all of the properties, and only changing isActive depending on the condition?

rjesson-softwire:
> From lines 118 to 170, the returned expanded parcel data appears to be the same, except for isActive being true or false. Would it be worth investigating a way to not repeat listing all of the properties, and only changing isActive depending on the condition?

There are three properties that are different between the two objects, so we are leaving it as is.

Stuart-Knowles-SW (approved):
LGTM!

PR #451, opened 2024-07-31
VFB-67 phone number validation is too strict

Jessica-Al:
## What's changed
Phone number validation changes:
Three options:
starts with 0 and can be 10 to 12 digits (UK mobile/landline)
starts +44 and can be 11 to 13 (UK mobile)
or starts with + and can be 7 to 15 digits (max and min phone number with country code)

## Screenshots / Videos
Before
image

After
image

## Checklist

---

* **Enhancement of Phone Number Input**
We have improved the way that phone numbers are validated in our system. A new pattern has been implemented to check the format of the phone numbers, ensuring only valid numbers are accepted.
* **Improved Data Handling in Text Change Function**
In the process of data entry, we have changed the function that processes your text input when entering phone numbers. It will now remove any spaces, dashes, and parentheses automatically. This way, we eliminate potential input errors before the system checks for any further issues.

### DIFF
--- a/src/common/format.ts
+++ b/src/common/format.ts
@@ -11,7 +11,7 @@ export const displayNameForDeletedClient = "Deleted Client";
 export const displayNameForNullDriverName = "Unknown Driver";
 
 // Following characters excluded from regex as are removed before checking format matches: ( ) - \s
-export const phoneNumberRegex = /^([0][0-9]{9-12}|[+][0-9]{7,15})?$/;
+export const phoneNumberRegex = /^([0][0-9]{9,12}|[+][0-9]{7,15})?$/;
 // Regex source: https://ihateregex.io/expr/phone/
 
 export const formatCamelCaseKey = (objectKey: string): string => {
--- a/src/common/format.ts
+++ b/src/common/format.ts
@@ -11,7 +11,7 @@ export const displayNameForDeletedClient = "Deleted Client";
 export const displayNameForNullDriverName = "Unknown Driver";
 
 // Following characters excluded from regex as are removed before checking format matches: ( ) - \s
-export const phoneNumberRegex = /^([0][0-9]{9,12}|[+][0-9]{7,15})?$/;
+export const phoneNumberRegex = /^((0|\+44)\d{9,11}|\+(?!44)\d{7,15})?$/;
 // Regex source: https://ihateregex.io/expr/phone/
 
 export const formatCamelCaseKey = (objectKey: string): string => {
--- a/src/common/format.ts
+++ b/src/common/format.ts
@@ -10,7 +10,8 @@ export const displayNameForDeletedClient = "Deleted Client";
 
 export const displayNameForNullDriverName = "Unknown Driver";
 
-export const phoneNumberRegex = /^([+]?[(]?[0-9]{3}[)]?[-\s.]?[0-9]{3}[-\s.]?[0-9]{4,6})?$/;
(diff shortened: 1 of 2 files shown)


src/components/Form/formFunctions.ts
```diff
@@ -95,7 +95,14 @@ export const onChangeText = <SpecificFields extends Fields>(
 ): SelectChangeEventHandler => {
     return (event) => {
         const input = event.target.value;
-        const errorType = getErrorType(input, required, regex, additionalCondition);
+        const errorType = getErrorType(
+            key === "telephoneNumber" || key === "phoneNumber"
```
Stuart-Knowles-SW, line 99: Just as a note, we could also do something like `key.includes("phone")` here. I think we want to keep it explicit in this instance, but there might be situations where the more general solution is appropriate.

src/common/format.ts
```diff
@@ -10,7 +10,8 @@ export const displayNameForDeletedClient = "Deleted Client";
 
 export const displayNameForNullDriverName = "Unknown Driver";
 
-export const phoneNumberRegex = /^([+]?[(]?[0-9]{3}[)]?[-\s.]?[0-9]{3}[-\s.]?[0-9]{4,6})?$/;
+// Following characters excluded from regex as are removed before checking format matches: ( ) - \s
+export const phoneNumberRegex = /^([0][0-9]{9,12}|[+][0-9]{7,15})?$/;
```
Stuart-Knowles-SW: This is already a much simpler regex than before thanks to pre-processing, but I think there are some tweaks we can make for it to be even better.
1. Handle the +44 code explicitly. It's much more likely that a user types a +44 number than any other country code, so I think we want to handle it as precisely as possible.
2. For regex-cleanliness, we can avoid using square brackets for single characters, which can just be by themselves, and `[0-9]` can be shorthanded as `\d`.
3. As far as I can tell, UK phone numbers are always either nine or ten digits after the 0 or +44 ([source](https://en.wikipedia.org/wiki/Telephone_numbers_in_the_United_Kingdom#Format)), so I think we can tighten that range a bit.

Putting this all together, I've come up with the regex:
`/^((0|\+44)\d{9,10}|\+(?!44)\d{7,15})?$/`
Have a play around with it on [regex101.com](regex101.com) to check you understand what it's doing, and let me know if I've misunderstood anything or if you can think of other improvements.

Stuart-Knowles-SW:
Can we also follow consistent formatting of PR titles, capitalising and hyphenating "VFB-xxx"?

src/common/format.ts
```diff
@@ -10,7 +10,8 @@ export const displayNameForDeletedClient = "Deleted Client";
 
 export const displayNameForNullDriverName = "Unknown Driver";
 
-export const phoneNumberRegex = /^([+]?[(]?[0-9]{3}[)]?[-\s.]?[0-9]{3}[-\s.]?[0-9]{4,6})?$/;
+// Following characters excluded from regex as are removed before checking format matches: ( ) - \s
+export const phoneNumberRegex = /^((0|\+44)\d{9,11}|\+(?!44)\d{7,15})?$/;
 // Regex source: https://ihateregex.io/expr/phone/
```
Stuart-Knowles-SW: I don't think this second comment applies any more

Stuart-Knowles-SW (approved):
If we can just remove the out of date comment, I'm happy for this to be merged :)

PR #452, opened 2024-08-01
VFB-312 Create Batch Add Page and minimal table component

ebinezer04:
## What's changed
Added new page on path parcels/batch.
Added a minimal batch add client/parcel table (with no functionality) populated with one row of mock data. 
The row directly below the header is a template for later functionality for overriding entire columns with values.

## Screenshots / Videos
image

## Checklist
---

* **New Component Test File Added**
A new testing file named `BatchParcelDataGrid.test.tsx` has been added, aimed at testing the function and reliability of the 'BatchParcelDataGrid' component.

* **Implementation of BatchParcelDataGrid Component**
A new file `BatchParcelDataGrid.tsx` contains the complete working blueprint of the `BatchParcelDataGrid` component. This includes necessary importations, defining an interface for the display rows, styling of columns and creation of the main `BatchParcelDataGrid` component.

* **Introduction of the Batch Parcel Creation Page**
A new file `page.tsx` has been entrant, representing a new interface dedicated to batch parcel creation. Alongside necessary components, it defines a function component `Batch` to return the page layout. The `Metadata` object has also been provided to set the page title, and the `Batch` component has been exported as the default.


src/app/parcels/batch/BatchParcelDataGrid.tsx
```diff
@@ -0,0 +1,267 @@
...
+import { DataGrid, GridColDef } from "@mui/x-data-grid";
+import React from "react";
+
+interface BatchGridDisplayRow {
+    id: number;
+    rowName: number;
+    fullName: string;
+    phoneNumber: string;
+    address: string;
+    adults: number | null;
```
rjesson-softwire, line 13: when can adults be null? the client add form requires an integer between 1 and 20. same question with children - there can be no children in a family, but would this not be 0 children, rather than null?

src/app/parcels/batch/BatchParcelDataGrid.tsx
```diff
@@ -0,0 +1,267 @@
...
+    return { ...column, headerAlign: "center", align: "center" };
+});
+
+const mockData: BatchGridDisplayRow[] = Array.from({ length: 6 }, (_, index) => ({
+    id: index,
+    rowName: index,
+    fullName: "",
+    phoneNumber: "",
+    address: "",
+    adults: null,
```
rjesson-softwire, line 192: linked to the comment in the props: is there a reason you choose to have the default value as null for numeric values, rather than 0 or -1, but choose to use empty strings for strings?

rjesson-softwire (dismissed):
looks good! just one small question about types

src/app/parcels/batch/BatchParcelDataGrid.tsx
ebinezer04: because we want empty cells, this can be an empty string "" or null. for actual data we want adults to be a number, but for an empty row we don't want to display anything (instead of 0 as default for example).

this is specifically for the display of empty cells for a new empty row.

src/app/parcels/batch/page.tsx
```diff
@@ -0,0 +1,19 @@
...
+    return (
+        <main>
+            <Title>Batch parcel creation - work in progress, do not use</Title>
+            <BatchParcelDataGrid />
+        </main>
+    );
+};
+
+export const metadata: Metadata = {
+    title: "Batch",
```
Stuart-Knowles-SW, line 16: Can this be "Batch Parcels", or something else a bit more descriptive?

src/app/parcels/batch/BatchParcelDataGrid.tsx
```diff
@@ -0,0 +1,267 @@
...
+    extraInformation: "No",
+    attentionFlag: "No",
+    signpostingCall: "No",
+    notes: "No",
+    voucherNumber: "1234567890",
+    packingDate: "2022-01-01",
+    packingSlot: "AM",
+    shippingMethod: "Collection",
+    collectionInfo: "Test centre, date, slot",
+};
```
Stuart-Knowles-SW: Can we define the mock data in the test file? Long term that's the only place it's required, so we may as well put it there now

src/app/parcels/batch/BatchParcelDataGrid.tsx
```diff
@@ -0,0 +1,267 @@
+"use client";
+
+import { Button } from "@mui/material";
+import { DataGrid, GridColDef } from "@mui/x-data-grid";
+import React from "react";
```
Stuart-Knowles-SW, line 5: I don't think this has been required for a couple of versions of react now

src/app/parcels/batch/BatchParcelDataGrid.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+describe("Parcels - Batch Parcel Data Grid", () => {
+    beforeEach(() => {
+        render(<BatchParcelDataGrid />);
+    });
+    it("renders intial header view", () => {
+        expect(screen.getByText("Full Name")).toBeInTheDocument();
+        expect(screen.getByText("Phone Number")).toBeInTheDocument();
+    });
+});
```
Stuart-Knowles-SW: We can be a lot more thorough with these tests.

Firstly, we can export the columns definition from the main file (preferably with a more descriptive name than "columns"), and check that all of the column headings are rendered. This will not only provide better coverage, but make it more maintainable since we won't have to update the test if the columns change.

Secondly, you'll need a test to make sure all the data is rendered correctly. With static data this is probably best as a snapshot test

src/app/parcels/batch/BatchParcelDataGrid.tsx
```diff
@@ -0,0 +1,267 @@
...
+    voucherNumber: "1234567890",
+    packingDate: "2022-01-01",
+    packingSlot: "AM",
+    shippingMethod: "Collection",
+    collectionInfo: "Test centre, date, slot",
+};
+
+interface BatchParcelDataGridProps {
+    rows?: BatchGridDisplayRow[];
+}
```
Stuart-Knowles-SW: I might have missed something, but we probably want rows to always be defined and to just be an empty array if there's nothing to show.

src/app/parcels/batch/BatchParcelDataGrid.tsx
ebinezer04: created new mock data in the test file, this cannot be exported. we created a new default data object in the component file which displays in the table by default.

src/app/parcels/batch/BatchParcelDataGrid.tsx
ebinezer04: required for testing purposes, Jest throws an error when the test file is run without this line

src/app/parcels/batch/__snapshots__/BatchParcelDataGrid.test.tsx.snap
Stuart-Knowles-SW, line 1: This doesn't appear to have any data or column headings in it?

src/app/parcels/batch/BatchParcelDataGrid.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+describe("Data Grid Snapshot Test", () => {
+    it("Matches DOM Snapshot", () => {
+        const domTree = renderer
+            .create(<BatchParcelDataGrid rows={mockData} />)
+            .toJSON() as renderer.ReactTestRendererJSON;
+        expect(domTree).toMatchSnapshot();
+    });
+});
+
+export default mockData;
```
Stuart-Knowles-SW: I imagine this was a leftover from an old version?

Stuart-Knowles-SW (approved):
LGTM, can you make sure that one of the QA steps is to check the tests fail if something is changed in the code that changes the display

PR #453, opened 2024-08-01
VFB-269 add functionality to reorder wiki items

rjesson-softwire:
## What's changed
Add a functionality to reorder wiki items. The wiki items can be reordered when in display or in edit mode.

Error messages added to page when supabase calls return an error. Also added audit logs to all wiki actions: reorder, delete, edit.

Changed save icon to be more obviously a save icon.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| 
image
 | image |

## Checklist

    - Checkout the dev branch
    - Checkout your branch

### DIFF
--- a/src/app/info/WikiItemEdit.tsx
+++ b/src/app/info/WikiItemEdit.tsx
@@ -1,7 +1,12 @@
 "use client";
 
 import { DbWikiRow } from "@/databaseUtils";
-import { WikiEditModeButton, WikiItemAccordionSurface } from "@/app/info/StyleComponents";
+import {
+    ReorderArrowDiv,
+    WikiEditModeButton,
+    WikiItemAccordionSurface,
+    WikiUpdateDataButton,
+} from "@/app/info/StyleComponents";
 import { TextField } from "@mui/material";
 import DeleteIcon from "@mui/icons-material/Delete";
 import CancelIcon from "@mui/icons-material/Cancel";
@@ -9,6 +14,8 @@ import SaveAltIcon from "@mui/icons-material/SaveAlt";
 import supabase from "@/supabaseClient";
 import { logErrorReturnLogId } from "@/logger/logger";
 import { WikiRowQueryType } from "@/app/info/AddWikiItemButton";
+import KeyboardDoubleArrowUpIcon from "@mui/icons-material/KeyboardDoubleArrowUp";
+import KeyboardDoubleArrowDownIcon from "@mui/icons-material/KeyboardDoubleArrowDown";
 
 interface WikiItemEditProps {
     rowData: DbWikiRow;
@@ -16,6 +23,8 @@ interface WikiItemEditProps {
     setIsInEditMode: (isInEditMode: boolean) => void;
     appendNewRow: (newRow: DbWikiRow, index: number) => void;
     removeRow: (row: DbWikiRow) => number;
+    swapRows: (row1: DbWikiRow, upwards: boolean) => void;
+    setErrorMessage: (error: string | null) => void;
 }
 
 const WikiItemEdit: React.FC<WikiItemEditProps> = ({
@@ -24,6 +33,8 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
     setIsInEditMode,
     appendNewRow,
     removeRow,
+    swapRows,
+    setErrorMessage,
 }) => {
     const deleteWikiItem = async (): Promise<void> => {
         const deleteResponse = (await supabase
@@ -31,7 +42,11 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
             .delete()
             .match({ wiki_key: rowData.wiki_key })) as WikiRowQueryType;
         if (deleteResponse.error) {
-            logErrorReturnLogId("error deleting wiki row item", deleteResponse.error);
+            const logId = await logErrorReturnLogId(
+                "error deleting wiki row item",
+                deleteResponse.error
+            );
+            setErrorMessage(`Failed to delete wiki item. Log ID: ${logId}`);
         }
         removeRow(rowData);
         setRowData(undefined);
@@ -58,7 +73,11 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
                     wiki_key: rowData.wiki_key,
                 })) as WikiRowQueryType;
                 if (updateResponse.error) {
-                    logErrorReturnLogId("error updating wiki row item", updateResponse.error);
+                    const logId = await logErrorReturnLogId(
+                        "error updating wiki row item",
+                        updateResponse.error
+                    );
+                    setErrorMessage(`Failed to update wiki item. Log ID: ${logId}`);
                 } else {
                     const updatedRow: DbWikiRow = {
                         title: newTitle,
@@ -88,6 +107,23 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
 
     return (
         <>
+            <ReorderArrowDiv>
+                <WikiUpdateDataButton
+                    onClick={() => {
+                        swapRows(rowData, true);
+                    }}
+                >
+                    <KeyboardDoubleArrowUpIcon />
+                </WikiUpdateDataButton>
+                <WikiUpdateDataButton
+                    onClick={() => {
+                        swapRows(rowData, false);
+                    }}
+                >
+                    <KeyboardDoubleArrowDownIcon />
+                </WikiUpdateDataButton>
+            </ReorderArrowDiv>
+
             <WikiEditModeButton onClick={cancelWikiItemEdit}>
                 <CancelIcon />
             </WikiEditModeButton>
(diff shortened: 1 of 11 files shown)


src/app/info/WikiItems.tsx
```diff
@@ -94,13 +115,98 @@ const WikiItems: React.FC<WikiItemsProps> = ({ rows }) => {
...
-            const temp = DisplayRows.slice();
+        setDisplayRows((displayRows) => {
+            const temp = displayRows.slice();
             temp.splice(indexToRemove, 1);
             return temp;
         });
         return indexToRemove;
     };
+
+    const [isSwapping, setIsSwapping] = useState(false);
```
SoftJoelWire: maybe rename this to isSwapMode to be similar to the isEditMode state used before. Completely up to you though, if this makes more sense go for it.

SoftJoelWire (commented):
looks good to me! one purely stylistic comment

supabase/migrations/20240801131704_add_wiki_id_column_to_audit_log_table.sql
```diff
@@ -0,0 +1,7 @@
+alter table "public"."audit_log" add column "wiki_id" uuid;
+
+alter table "public"."audit_log" add constraint "public_audit_log_wiki_id_fkey" FOREIGN KEY (wiki_id) REFERENCES wiki(wiki_key) ON DELETE RESTRICT not valid;
+
```
Stuart-Knowles-SW: I think this means we can't delete wiki items that are referenced in the audit log table. I know it's the pattern used for most other columns in the audit log tables, but most of those columns are soft-delete, so that would work for them but not here.

Can you:

1. Check whether I'm correct in this: move a wiki item around so it appears in the audit log then try and delete it.
2. If so, change this to `ON DELETE SET null`

supabase/migrations/20240801131704_add_wiki_id_column_to_audit_log_table.sql
```diff
@@ -0,0 +1,7 @@
+alter table "public"."audit_log" add column "wiki_id" uuid;
+
+alter table "public"."audit_log" add constraint "public_audit_log_wiki_id_fkey" FOREIGN KEY (wiki_id) REFERENCES wiki(wiki_key) ON DELETE RESTRICT not valid;
```
Stuart-Knowles-SW: Do you know what triggers the audit log entries? Do we need to add entries for creating / updating and deleting from the wiki?

src/app/info/WikiItems.tsx
```diff
@@ -109,13 +215,20 @@ const WikiItems: React.FC<WikiItemsProps> = ({ rows }) => {
                     appendNewRow={appendNewRow}
                 />
             </AdminManagerDependentView>
+            {errorMessage && (
+                <TextValueContainer>
+                    <ErrorSecondaryText>{errorMessage}</ErrorSecondaryText>
+                </TextValueContainer>
+            )}
```
Stuart-Knowles-SW, line 227: Nice :)

src/app/info/WikiItems.tsx
```diff
@@ -94,13 +115,98 @@ const WikiItems: React.FC<WikiItemsProps> = ({ rows }) => {
...
+        });
+
+        const auditLog = {
+            action: `move a wiki item ${row1.row_order <= row2.row_order ? "down" : "up"}`,
+            wikiId: row1.wiki_key,
+            content: {
+                itemName: row1.title,
+                oldRow_order: row1.row_order,
+            },
+        } as const satisfies Partial<AuditLog>;
```
Stuart-Knowles-SW, line 179: This is stronger typing than I'd expect here, is this required?

src/app/info/WikiItems.tsx
```diff
@@ -94,13 +115,98 @@ const WikiItems: React.FC<WikiItemsProps> = ({ rows }) => {
...
-            const temp = DisplayRows.slice();
+        setDisplayRows((displayRows) => {
+            const temp = displayRows.slice();
             temp.splice(indexToRemove, 1);
             return temp;
         });
         return indexToRemove;
     };
+
+    const [isSwapMode, setIsSwapMode] = useState(false);
```
Stuart-Knowles-SW: what are we trying to prevent with this guard? State setting is not synchronous, so I'm worried that some of the invocations below aren't actually doing what we think they would.

src/app/info/WikiItems.tsx
```diff
@@ -94,13 +115,98 @@ const WikiItems: React.FC<WikiItemsProps> = ({ rows }) => {
...
+            });
+        }
+        setIsSwapMode(false);
+
+        reorderRows(row1, row2);
+    };
+
+    const reorderRows = (row1: DbWikiRow, row2: DbWikiRow): void => {
+        const wiki_keys = displayRows.map((displayRows) => displayRows.wiki_key);
+
```
Stuart-Knowles-SW: Can we rename this to make it clear that this is just doing it in the front end and to distinguish it from the db function?

src/app/info/WikiItems.tsx
```diff
@@ -94,13 +115,98 @@ const WikiItems: React.FC<WikiItemsProps> = ({ rows }) => {
...
+        row1Item.row_order = row2Order;
+        row2Item.row_order = row1Order;
+
+        const newDisplayRows = [...displayRows];
+
+        newDisplayRows[row1Index] = row2Item;
+        newDisplayRows[row2Index] = row1Item;
+
+        setDisplayRows(newDisplayRows);
+    };
```
Stuart-Knowles-SW: This is getting a bit long now, can we extract the bit of this function that doesn't need to be inside the react component (i.e, everything except `setDisplayRows(...)`) for clarity. So this would look something like
```
const reOrderRows(displayRows, index1, direction) => {
  setDisplayRows(swapRows(displayRows, index1, direction))
}
```
where `swapRows` is a pure typescript function that handles the logic of moving the rows around

src/app/info/WikiItemEdit.tsx
```diff
@@ -24,14 +33,20 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
...
         if (deleteResponse.error) {
-            logErrorReturnLogId("error deleting wiki row item", deleteResponse.error);
+            const logId = await logErrorReturnLogId(
+                "error deleting wiki row item",
+                deleteResponse.error
+            );
+            setErrorMessage(`Failed to delete wiki item. Log ID: ${logId}`);
         }
         removeRow(rowData);
         setRowData(undefined);
```
Stuart-Knowles-SW: I know this wasn't your work, but can we change this so we only remove the row if the database deletion succeeds?

src/app/info/WikiItems.tsx
```diff
@@ -94,13 +115,98 @@ const WikiItems: React.FC<WikiItemsProps> = ({ rows }) => {
...
+            const temp = displayRows.slice();
             temp.splice(indexToRemove, 1);
             return temp;
         });
         return indexToRemove;
     };
+
+    const [isSwapMode, setIsSwapMode] = useState(false);
+
+    const swapRows = async (row1: DbWikiRow, upwards: boolean): Promise<void> => {
```
Stuart-Knowles-SW: can we change this so instead of a boolean we have a direction, and since it's a funciton that is called on a single row can the name reflect that?
`const moveRow = async (row: DbWikiRow, direction: "up" | "down"): Promise<void> => {`
It'll make the intention a lot clearer where it's called

src/app/info/WikiItems.tsx
```diff
@@ -94,13 +115,98 @@ const WikiItems: React.FC<WikiItemsProps> = ({ rows }) => {
...
+            key2: row2.wiki_key,
+        });
+
+        const auditLog = {
+            action: `move a wiki item ${row1.row_order <= row2.row_order ? "down" : "up"}`,
+            wikiId: row1.wiki_key,
+            content: {
+                itemName: row1.title,
+                oldRow_order: row1.row_order,
+            },
```
Stuart-Knowles-SW: Great work with this, I wouldn't have thought of it!

Stuart-Knowles-SW (commented):
Great work! Nice job with the audit logs as well. A few refactorings and one functionality thing

src/app/info/WikiItems.tsx
rjesson-softwire: Not sure but this is used every time an audit log is made, elsewhere in the codebase.

src/app/info/WikiItems.tsx
rjesson-softwire: we've removed this. we copied this from the lists page, but that reloads the page when reordering items, so takes a lot longer.

src/app/info/WikiItems.tsx
Stuart-Knowles-SW: Cool, I think it's a bit more than we need, but let's stick to pattern

Stuart-Knowles-SW (approved):
Great work, let's get it merged 🚢

PR #464, opened 2024-08-06
VFB-311 Create data schema and reducer function to manage state

ebinezer04:
## What's changed
Designed the JSON schema and implemented types
Added the reducer function for managing the overall state of the Batch Data Grid component
Added component tests for the reducer function

## Checklist

---

* **Introduction of Batch Data Management**
A new system component named `BatchDataRow.ts` has been added. This is essentially a collection of protocols (interfaces and types in technical terms) that sets the rules on how the new data batches should be handled in the system. This file will help us with actions related to our data collection and processing. To accompany this, a reducer function is also added, which helps us consolidate different data inputs into a single, more manageable output efficiently, keeping our system tidy and well-organized.


src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,377 @@
...
+            if (!formattedPerson.birthMonth) {
+                currentYear - formattedPerson.birthYear >= 16
+                    ? adults.push(formattedPerson)
+                    : children.push(formattedPerson);
+            } else {
+                currentYear - formattedPerson.birthYear > 16 ||
+                (currentYear - formattedPerson.birthYear === 16 &&
+                    currentMonth >= formattedPerson.birthMonth)
+                    ? adults.push(formattedPerson)
+                    : children.push(formattedPerson);
```
Jessica-Al: This if-else seems a bit messy. Not sure if there is a better way but maybe give it a look over. You do the same check (year-birthyear > 16) first in both cases, can you take that out of the if statement at all?

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,377 @@
...
+
+    const adults: Person[] = [];
+    const children: Person[] = [];
+
+    const currentYear = new Date().getFullYear();
+    const currentMonth = new Date().getMonth();
+
+    if (error) {
+        logErrorReturnLogId("Error with fetch: family_members", { error: error });
+        return { adults: [], children: [] };
```
Jessica-Al: What exactly happens at this point when the empty arrays are returned? The error is logged but is there any indication that an error occurred or does the adult, child info just show up empty? Might be outside scope of ticket but thought I'd bring it up.

src/app/parcels/batch/BatchDataRow.ts
SoftJoelWire: Do you think this is cleaner? 
```
const yearsDiff = currentYear - formattedPerson.birthYear
if( yearsDiff > 16) {
    adults.push(formattedPerson);
} else {
    !formattedPerson.birthMonth ?
    yearsDiff == 16 ? adults.push(formattedPerson) : children.push(formattedPerson)
    : (yearsDiff == 16 && formattedPerson.birthMonth <= currentMonth) ? adults.push(formattedPerson) : children.push(formattedPerson);
}
```

Jessica-Al: Better, what about including a check for years diff being equal to and then less than 16? (if, else if, else) Would that work?

src/app/parcels/batch/supabaseCalls.ts
Stuart-Knowles-SW, line 1: Can we call this file `supabaseHelpers.ts` and remove the word `Call` from the function names? Conceptually, "a call" is when we use the function, not the function itself, and the naming should reflect that. We could also use "call" as an imperative verb rather than a noun, so these would read, e.g., `callGetClientSupabase` or even just `callGetClient` since there's nowhere other than supabase we could be calling

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+interface Address {
+    addressLine1: string;
+    addressLine2: string | null;
+    addressTown: string;
+    addressCounty: string | null;
+    addressPostcode: string;
+}
+
+interface AdultInfo {
+    adults: Person[];
```
Stuart-Knowles-SW: Can we move all the interfaces into a separate `types.ts` file?

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+interface OverrideClient {
+    [key: string]:
+        | string
+        | null
+        | Address
+        | AdultInfo
+        | ChildrenInfo
+        | ListType
+        | BooleanGroup
+        | boolean;
```
Stuart-Knowles-SW: nice!

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+            attentionFlag: data.flagged_for_attention ?? false,
+            signpostingCall: data.signposting_call_required ?? false,
+            notes: data.notes,
+        };
+    } else {
+        return null;
+    }
+};
+
+export const reducer = async (
```
Stuart-Knowles-SW: Can we call this `batchParcelsReducer` for clarity, make it a default export and rename the file `BatchParcelsReducer.ts` since that's really what it's for

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+
+const getNappySize = (info: string | null): string | null => {
+    if (info) {
+        const match = info.match(/Nappy Size:\s*(\d+)/);
+        if (match) {
+            return match[1];
+        }
+    }
+    return null;
+};
```
Stuart-Knowles-SW: Is this format for nappy size set elsewhere in the app?

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+        return null;
+    } else if (data) {
+        data.forEach((person) => {
+            const formattedPerson: Person = {
+                gender: person.gender,
+                birthYear: person.birth_year,
+                birthMonth: person.birth_month,
+            };
+
+            const yearsDiff = currentYear - formattedPerson.birthYear;
```
Stuart-Knowles-SW: This is just the person's age, right? I think we can just call it `personAge` 😅

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+                      : children.push(formattedPerson);
+            } else if (yearsDiff < 16) {
+                children.push(formattedPerson);
+            } else {
+                adults.push(formattedPerson);
+            }
+        });
+    } else {
+        return { adults: [], children: [] };
+    }
```
Stuart-Knowles-SW: Can we shift this conditional above the main one (early return)? We also then don't need the main function body in a conditional

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+            otherItems: createBooleanGroupFromStrings(data.other_items),
+            deliveryInstructions: data.delivery_instructions,
+            extraInformation: parseExtraInfo(data.extra_information),
+            attentionFlag: data.flagged_for_attention ?? false,
+            signpostingCall: data.signposting_call_required ?? false,
+            notes: data.notes,
+        };
+    } else {
+        return null;
+    }
```
Stuart-Knowles-SW: Same here with an early return

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+    if (info) {
+        const match = info.match(/Nappy Size:\s*\d+,\s*Extra Information:\s*(.*)/);
+        if (match) {
+            return match[1];
+        }
+    }
+    return info;
+};
+
+const getExistingClientData = async (clientId: string): Promise<BatchClient | null> => {
```
Stuart-Knowles-SW: Can we make it really clear in naming that this function is not only getting the client data but formatting it for the parcels table? If I didn't know any better, I'd think this function could also be used on the Clients page. Something like `getClientDataForBatchParcels` would work

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+
+export interface BatchTableDataState {
+    overrideDataRow: OverrideDataRow;
+    batchDataRows: BatchDataRow[];
+}
+
+export interface BatchActionType {
+    type: "update_cell" | "add_row" | "delete_row" | "override_column" | "use_existing_client";
+    payload?: BatchActionPayload;
+}
```
Stuart-Knowles-SW: Can we completely define this interface? Each `type` should have it's own payload definition which should be specific to the data required for that action

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+                const updatedRows: BatchDataRow[] = state.batchDataRows.map((row) => {
+                    return row.id === rowId ? newRow : row;
+                });
+                return {
+                    ...state,
+                    batchDataRows: updatedRows,
+                };
+            } else {
+                return state;
+            }
```
Stuart-Knowles-SW: Once we've fully defined the Action interface, we won't need these checks because we know that each action can only take valid data for it

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+    }
+};
+
+export const reducer = async (
+    state: BatchTableDataState,
+    action: BatchActionType
+): Promise<BatchTableDataState> => {
+    switch (action.type) {
+        case "update_cell": {
+            if (action.payload && action.payload.newRow) {
```
Stuart-Knowles-SW: Why do we need to pass in an entirely new row if we're only updating a cell?

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+                return {
+                    ...state,
+                    batchDataRows: updatedRows,
+                };
+            } else {
+                return state;
+            }
+        }
+        case "add_row": {
+            return state.batchDataRows.length < 99
```
Stuart-Knowles-SW: How do we surface it to the user if this limit is hit?

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+                                  Math.max(
+                                      ...state.batchDataRows.map((row) => {
+                                          return row.id;
+                                      })
+                                  ) + 1,
+                              clientId: null,
+                              data: null,
+                          },
+                      ],
+                  }
```
Stuart-Knowles-SW: This is quite messy and nested, can we define it as a `const` above the return statement?

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+                : state;
+        }
+        case "delete_row": {
+            return action.payload?.rowId !== 0
```
Stuart-Knowles-SW: What's the case where this might be 0?

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+                ? {
+                      ...state,
+                      batchDataRows: state.batchDataRows.filter(
+                          (row) => row.id !== action.payload?.rowId
+                      ),
+                  }
+                : state;
+        }
+        case "override_column": {
```
Stuart-Knowles-SW: Can we call this `set_column_override` and have a separate action for `remove_column_override`?

src/app/parcels/batch/BatchDataRow.ts
```diff
@@ -0,0 +1,384 @@
...
+            if (action.payload?.newOverrideRow) {
+                const overriddenClientFields = getOverridenFields(
+                    action.payload?.newOverrideRow.data.client
+                );
+                const overriddenParcelFields = getOverridenFields(
+                    action.payload?.newOverrideRow.data.parcel
+                );
+                const { newOverrideRow } = action.payload;
```
Stuart-Knowles-SW: This feels a bit logically tangled at the moment, hopefully making the types more precise should clarify it somewhat

src/app/parcels/batch/BatchDataRow.ts
ebinezer04: Yes, this is how it is stored in the clients table

src/app/parcels/batch/BatchDataRow.ts
ebinezer04: It will not necessarily be the person's age depending on the current month and month of birth

src/app/parcels/batch/BatchDataRow.ts
ebinezer04: 0 is always the top override row which we do not want to be able to delete

src/app/parcels/batch/BatchDataRow.ts
ebinezer04: We will check the length of the grid on the front end when the user performs an add row action and communicate to the user if the max length has been reached. This means that this check is slightly redundant, but, as this ticket is for the reducer only we thought that it would be good to keep it in anyway.

src/app/parcels/batch/supabaseHelpers.ts
```diff
@@ -0,0 +1,24 @@
...
+        .from("clients")
+        .select("*")
+        .eq("primary_key", clientId)
+        .single();
+    return { data, error };
+};
+
+export const callGetFamily = async (
+    familyId: string
+): Promise<{ data: Schema["families"][] | null; error: PostgrestError | null }> => {
```
Stuart-Knowles-SW: Sorry to be a pain about this, there are some more specific conventions we prefer to follow here. I believe Aya went through them with Rebecca and Leah in detail yesterday, do you want to talk to them about it? Partly to save me typing it out in a lot of detail but also as an interesting exercise for the two of them :)

src/app/parcels/batch/BatchParcelsReducer.ts
```diff
@@ -0,0 +1,278 @@
...
+            !formattedPerson.birthMonth
+                ? adults.push(formattedPerson)
+                : currentMonth >= formattedPerson.birthMonth
+                  ? adults.push(formattedPerson)
+                  : children.push(formattedPerson);
+        } else if (yearsDiff < 16) {
+            children.push(formattedPerson);
+        } else {
+            adults.push(formattedPerson);
+        }
```
Stuart-Knowles-SW: Excellent point that the years diffis not the same as age. I think we can simplify this a bit by just using Date objects. Specifically, use the `dayjs` library (we already have it installed), it's better than the inbuilt `Date` library, then just compare the two dates

src/app/parcels/batch/BatchDataRow.ts
Stuart-Knowles-SW: Cool, are we able to reference that directly rather than implicitly?

src/app/parcels/batch/BatchParcelsReducer.ts
```diff
@@ -0,0 +1,282 @@
...
+const getChildrenAndAdults = async (
+    familyId: string
+): Promise<{ adults: Person[]; children: Person[] } | null> => {
+    const { data, error } = await getAllPeopleFromFamily(familyId);
+
+    const adults: Person[] = [];
+    const children: Person[] = [];
+
+    //do not run this test at 23:59:59.99 on the last day of the month
+    //this will cause the test to fail
```
Stuart-Knowles-SW: This isn't in a test file? If we need to have this comment can we move it to the test file

src/app/parcels/batch/BatchParcelsReducer.ts
```diff
@@ -0,0 +1,282 @@
...
+            birthYear: person.birth_year,
+            birthMonth: person.birth_month,
+        };
+
+        const birthDate = dayjs()
+            .year(formattedPerson.birthYear)
+            .month(formattedPerson.birthMonth || 0)
+            .startOf("month");
+        const age = currentDate.diff(birthDate, "year");
+        console.log(age, birthDate, currentDate);
```
Stuart-Knowles-SW: Missed console logs (there's at least one below here as well)

src/app/parcels/batch/BatchParcelsReducer.ts
```diff
@@ -0,0 +1,282 @@
...
+        if (age < 16 || (age === 16 && currentDate.month() < birthDate.month())) {
```
Stuart-Knowles-SW: Part of the reasoning behind the `dayjs` approach is that the diff should be both years and months so we shouldn't have to check for month?

Stuart-Knowles-SW (commented):
A couple more little changes. Sorry this is taking so long and thanks for the hard work :)

Stuart-Knowles-SW (approved):
Great :)

PR #466, opened 2024-08-07
VFB-320 date range filter fix

rjesson-softwire:
## What's changed
Changed the date filter so:

When I set the Parcel table FROM date to be later than the currently-shown TO date (either using the datepicker or keyboard input)

Then the TO date should update to match the new FROM date (so the parcels that match that packing date are shown in the table)

The opposite is  true also (that setting the TO date to be earlier than the FROM date causes FROM to update accordingly).

## Screenshots / Videos
Before
image

## Checklist

---

wtd:summary

### DIFF
--- a/src/components/DateInputs/DateRangeInputs.tsx
+++ b/src/components/DateInputs/DateRangeInputs.tsx
@@ -38,29 +38,30 @@ const DateRangeInputs: React.FC<Props> = (props) => {
         }
     };
 
+    const shouldSetDatesToBeEqual = (from: Dayjs | null, to: Dayjs | null): boolean => {
+        return from !== null && to !== null && from > to && areDatesAfterMinDate(from, to);
+    };
+
     const setFromValue = (fromValue: Dayjs | null): void => {
         setLocalFromValue(fromValue);
-        if (
-            fromValue &&
-            localToValue &&
-            fromValue > localToValue &&
-            fromValue >= reasonableMinDate
-        ) {
-            setLocalToValue(fromValue);
-            setRangeIfPossible(fromValue, fromValue);
-        } else {
-            setRangeIfPossible(fromValue, localToValue);
+
+        if (!shouldSetDatesToBeEqual(fromValue, localToValue)) {
+            return setRangeIfPossible(fromValue, localToValue);
         }
+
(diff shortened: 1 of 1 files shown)


ebinezer04:
looks good to me :)

src/components/DateInputs/DateRangeInputs.tsx
```diff
@@ -40,12 +40,27 @@ const DateRangeInputs: React.FC<Props> = (props) => {
...
+            fromValue &&
+            localToValue &&
+            fromValue > localToValue &&
+            fromValue >= reasonableMinDate
+        ) {
+            setLocalToValue(fromValue);
+            setRangeIfPossible(fromValue, fromValue);
+        } else {
+            setRangeIfPossible(fromValue, localToValue);
+        }
```
AyaPK, line 50: This method isn't very readable, in what it it is trying to achieve. It feels like a prime candidate for an [Early Return Pattern](https://dev.to/arikaturika/one-concept-a-day-early-return-pattern-in-javascript-3pol)! :)

Let me know if you'd like an example of what that might look like

src/components/DateInputs/DateRangeInputs.tsx
```diff
@@ -40,12 +40,27 @@ const DateRangeInputs: React.FC<Props> = (props) => {
 
     const setFromValue = (fromValue: Dayjs | null): void => {
         setLocalFromValue(fromValue);
-        setRangeIfPossible(fromValue, localToValue);
+        if (
+            fromValue &&
+            localToValue &&
+            fromValue > localToValue &&
+            fromValue >= reasonableMinDate
+        ) {
```
Stuart-Knowles-SW: Can we extract this and the complicated boolean check below into little helper function functions (plain `.ts`, outside the component) with descriptive names so it's easy to read at a glance what the check is doing?

AyaPK:
LGTM, thanks!
Pending Stuart's approval, of course.

Stuart-Knowles-SW (approved):
Looks good! There's quite a lot of logical paths here, so Ideally we'd have really thorough component tests. But those will be handled in VFB-310 so no need for them now. Let's make sure the QA is really thorough though

Issue #391, opened 2024-06-26
Postcode Search - can't search for multiple postcodes separated by comma


dan-langridge:
### Reproduction steps

When I enter the postcode search box field on the parcels page
and enter
E1, CB
it shows There are no records to display

### Expected behaviour

When i enter multiple postcodes (partial or full) separated by a comma, it would ideally show both or multiple postcodes

e.g. search E1 3AB, E2 7YT - it would show both of those postcodes
or
search E3, E4, E5 - it would show all postcodes in E3, E4, E5

This is an important feature for us, for example for the delivery manager to filter by certain postcodes to assign them to a driver

It would also be helpful to have this on the name search as well - both on the Parcels and Clients page
e.g.
Search name: Dan, John
would show all the Dan's and all the Johns

### What browsers are you seeing the problem on?

Safari

### What OS is your computer running on?

macOS

stonelink:
## Spec
* All the `TextFilter` components will treat `,` as a delimiting character - so not just postcode and name, this would apply to Phone number, Voucher, Family size etc. (I can't see a reason for those filters to not support a delimiter)
* When applying text in the input as a filter:
  * divide the string into an array of substrings, delimited by `,`
  * trim whitespace from the start and end of each substring
  * remove any empty substrings from the list
  * apply an 'or' in the DB query to check for a match to any of the substrings

stonelink:
The filters that have had this functionality added are:
- postcode
- full name
- family size
- phone number
- voucher

All of the above appear on the Parcels table; all except Voucher are also on the Clients table. 
Note that the family size filter has only been added to Clients table as part of this ticket.

Note that each filter still has the special cases that it used to support:
- most match '-' with a deleted client
- in postcode filter, 'nfa' matches clients with no address
- in family filter, numbers are the main expectation, any substring of "single" matches Singles, any substring of "family of" matches 2 or more
- in voucher filter, '?' matches no voucher

stonelink:
## QA steps
- test each of the filters listed in the previous comment (individually and in combination), on the Parcels and Clients tables
- for each filter:
  - test that the simple behaviour still works (i.e., a simple string with no commas)
  - test that a few values separated by commas works as expected - i.e., if you enter in the one of the filters a few strings with commas between them, then any parcel that matches any one of those strings should appear in the table
  - check that whitespace before or after the commas is ignored
[...]

PR #470, opened 2024-08-11
#391: Text filter strings can now contain multiple strings separated by commas

stonelink:
## What's changed
- The text filters on the Parcels and Clients tables now accept multiple strings separated by commas
- The pre-existing special cases (like '-' for deleted client, or 'NFA' for client with no address) still work
- The Family filter now appears on the Clients table, too

## Checklist

---

wtd:summary

### DIFF
--- a/src/common/databaseFilters.test.ts
+++ b/src/common/databaseFilters.test.ts
@@ -1,0 +1,426 @@
+import { beforeEach, expect, it } from "@jest/globals";
+import { familySearch, fullNameSearch, phoneSearch, postcodeSearch } from "./databaseFilters";
+import { DbParcelRow } from "@/databaseUtils";
+import { DbQuery, ServerSideFilterMethod } from "@/components/Tables/Filters";
+
+const logID = "a2adb0ba-873e-506b-abd1-8cd1782923c8";
+
+jest.mock("@/logger/logger", () => ({
+    logErrorReturnLogId: jest.fn(() => Promise.resolve(logID)),
+}));
+
+jest.mock("@/supabaseClient", () => {
+    return { default: jest.fn() };
+});
+
+const mockQueryWithOrAppended = jest.fn();
+const mockOr: jest.Mock = jest.fn(() => mockQueryWithOrAppended);
+
+describe("test for full name filter", () => {
+    let mockDbQuery: DbQuery<DbParcelRow>;
+    let fullNameFilter: ServerSideFilterMethod<DbParcelRow, string>;
+
+    beforeEach(() => {
+        jest.clearAllMocks();
+
+        // Arrange
+        mockDbQuery = {} as DbQuery<DbParcelRow>;
+        mockDbQuery.or = mockOr;
+
+        fullNameFilter = fullNameSearch<DbParcelRow>("client_full_name", "client_is_active");
+    });
+
+    it("should leave the query unchanged when name input is empty", () => {
+        // Act
+        const result = fullNameFilter(mockDbQuery, "");
+
+        // Assert
+        expect(result).toBe(mockDbQuery);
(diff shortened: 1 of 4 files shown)


src/common/databaseFilters.ts
```diff
@@ -3,57 +3,86 @@ import { displayPostcodeForHomelessClient } from "./format";
 import { DbClientRow, DbParcelRow } from "@/databaseUtils";
 import { parcelsPageDeletedClientDisplayName } from "@/app/parcels/parcelsTable/format";
 
-export const fullNameSearch = <DbData extends DbClientRow | DbParcelRow>(
-    fullNameColumnLabel: Extract<keyof DbData, "full_name" | "client_full_name">,
-    clientIsActiveColumnLabel: Extract<keyof DbData, "is_active" | "client_is_active">
+const textFilterDelimiter = ",";
+
+export const dbFilterWithSubstringQueries = <DbData extends DbClientRow | DbParcelRow>(
```
stonelink, line 8: This is the core of this change. 
The special characters allowed in row 14 are the same for all text filters and are the superset of all the special characters that were allowed previously in different cases.

src/common/databaseFilters.test.ts
```diff
@@ -0,0 +1,426 @@
...
+}));
+
+jest.mock("@/supabaseClient", () => {
+    return { default: jest.fn() };
+});
+
+const mockQueryWithOrAppended = jest.fn();
+const mockOr: jest.Mock = jest.fn(() => mockQueryWithOrAppended);
+
+describe("test for full name filter", () => {
```
stonelink, line 19: There were enough special cases for each filter that it seemed to be worth having all the tests in this file.

src/common/databaseFilters.test.ts
```diff
@@ -0,0 +1,426 @@
...
+    let mockDbQuery: DbQuery<DbParcelRow>;
+    let fullNameFilter: ServerSideFilterMethod<DbParcelRow, string>;
+
+    beforeEach(() => {
+        jest.clearAllMocks();
+
+        // Arrange
+        mockDbQuery = {} as DbQuery<DbParcelRow>;
+        mockDbQuery.or = mockOr;
```
stonelink, line 28: Because we only define a mock function for `or`, if in future the filters change to call another operator (like `and`) at the top level, the test will error.

src/common/databaseFilters.ts
```diff
@@ -3,57 +3,86 @@ import { displayPostcodeForHomelessClient } from "./format";
...
+    return dbFilterWithSubstringQueries((substring) => {
+        if (substring === "-") {
+            return `${clientIsActiveColumnLabel}.is.false`;
         }
-        if (displayPostcodeForHomelessClient.toLowerCase().includes(state.toLowerCase())) {
-            return query
-                .or(`${postcodeColumnLabel}.ilike.%${state}%, ${postcodeColumnLabel}.is.null`)
-                .neq(clientIsActiveColumnLabel, false);
+        if (displayPostcodeForHomelessClient.toLowerCase().includes(substring.toLowerCase())) {
+            return `and(${clientIsActiveColumnLabel}.is.true, or(${postcodeColumnLabel}.ilike.%${substring}%, ${postcodeColumnLabel}.is.null))`;
```
stonelink, line 46: I'm not thrilled that the subqueries are all written as text, but I spent a while trying to find a more elegant way without success.

Stuart-Knowles-SW: I've seen this in the supabase docs somewhere, it is the "correct" way to do it. I don't love it either, but I'm pretty sure there isn't a better way

Stuart-Knowles-SW (approved):
Looks sensible to me, I haven't gone through line by line but nothing rings alarm bells for me looking at the diff

PR #471, opened 2024-08-11
VFB-313: migration to fix problematic old seed data

stonelink:
## What's changed
Replace 3 Cambridge postcodes on `dev` and `staging`.
The seed data was already fixed to have all London postcodes, so this is only for test databases that we don't want to wipe - notably `staging` has a lot of test data that the users have added.

I'm not convinced that this should be a migration, but since there's only test data atm it won't cause a problem. 
I'd be happy to just run these queries in the supabase console if we agree that that's better.

## Checklist
- [X] I have self-reviewed this PR

  - [-] I have updated the typescript definitions for the database with `npm run generate_types:local`
  - [-] I have modified the seed in `supabase/seed/seed.mts` if appropriate
  - [X] I have taken reasonable measures to ensure constraint violations do not happen when the migration occurs, e.g.:
    - Checkout the dev branch
    - Checkout your branch

---

wtd:summary

### DIFF
--- a/supabase/migrations/20240811151507_replace_some_test_postcodes.sql
+++ b/supabase/migrations/20240811151507_replace_some_test_postcodes.sql
@@ -1,0 +1,4 @@
+-- this is replacing test data from historic seed.sql
+update public.clients set address_postcode = 'SE11 5RD' where address_postcode = 'CB2 3JU';
+update public.clients set address_postcode = 'SE24 0HG' where address_postcode = 'CB8 9LJ';
+update public.clients set address_postcode = 'SW1P 4JL' where address_postcode = 'CB24 4RT';


stonelink:
I've thought a bit more about this and rather than having a migration I suggest using the Supabase SQL Editor interface for the environment to run the three update queries separately, with a rollback check:

```
START TRANSACTION;
UPDATE public.clients SET address_postcode = 'SE11 5RD' WHERE address_postcode = 'CB2 3JU' RETURNING *;
ROLLBACK;
```

then checking that the number of rows in the output (that would be changed) is reasonable, then changing the ROLLBACK to COMMIT and running the command again.

And similarly for the other two.

What do you think?

Stuart-Knowles-SW (commented):
~If this is just to replace historic data we can run it directly in supabase rather than commit it as a migration?~

Edit: just saw you'd had the same thought

Stuart-Knowles-SW:
> I've thought a bit more about this and rather than having a migration I suggest using the Supabase SQL Editor interface for the environment to run the three update queries separately, with a rollback check:
> 
> ```
> START TRANSACTION;
> UPDATE public.clients SET address_postcode = 'SE11 5RD' WHERE address_postcode = 'CB2 3JU' RETURNING *;
> ROLLBACK;
> ```
> 
> then checking that the number of rows in the output (that would be changed) is reasonable, then changing the ROLLBACK to COMMIT and running the command again.
> 
> And similarly for the other two.
> 
> What do you think?

Honestly I think this is a very low risk change. The rollback check is sensible, but I don't think we need to be worried about this.

stonelink:
Thanks for your thoughts on this. Please excuse my excess of caution.

I'll abandon this PR.

PR #475, opened 2024-08-13
VFB-311 Disable existing client reducer action, render reducer state

ebinezer04:
## What's changed
Disabled the "use existing client" reducer action as it is async while reducers can only be synchronous. This will be reimplemented synchronously in a later ticket.

Linked the reducer state directly to be rendered in the component.

## Checklist


src/app/parcels/batch/BatchParcelDataGrid.tsx
```diff
@@ -180,39 +188,212 @@ const styledBatchGridDisplayColumns: GridColDef<BatchGridDisplayRow[][number]>[]
...
-    packingSlot: "",
-    shippingMethod: "",
-    collectionInfo: "",
-}));
-
-const BatchParcelDataGrid: React.FC<BatchParcelDataGridProps> = ({ rows }) => {
+const getEmptyRow = (id: number): BatchGridDisplayRow => {
+    return {
+        id: id,
+        rowName: id,
```
Stuart-Knowles-SW: What's the purpose of having both of these?

src/app/parcels/batch/BatchParcelsReducer.ts
```diff
@@ -212,6 +211,9 @@ export const batchParcelsReducer = async (
         }
         case "override_column": {
             const { newOverrideRow } = action.overrideColumnPayload;
+            if (!newOverrideRow.data) {
+                return state;
+            }
```
Stuart-Knowles-SW, line 72: Should this be `newOverrideColumn`, or have I misunderstood what's going on here?

src/app/parcels/batch/BatchParcelsReducer.test.ts
```diff
@@ -329,65 +329,67 @@ describe("reducer", () => {
...
-        const actionAddRow: BatchActionType = {
-            type: "add_row",
-        };
+    // When the use existing client method is implemented in the reducer use this test
 
-        const lastRowEmptyState1 = await batchParcelsReducer(mockTableDataState, actionAddRow);
+    // it("should autofill for an existing client and format result correctly", async () => {
+    //     const actionAddRow: BatchActionType = {
+    //         type: "add_row",
+    //     };
```
Stuart-Knowles-SW: Rather than commenting out the test, can we [mark it as skipped ](https://jestjs.io/docs/api#testskipname-fn). That way it'll be listed in the test reports as a skipped test so we won't lose track of it accidentally

src/app/parcels/batch/BatchParcelDataGrid.tsx
```diff
@@ -180,39 +188,212 @@ const styledBatchGridDisplayColumns: GridColDef<BatchGridDisplayRow[][number]>[]
...
+    const { collectionDate, collectionSlot, collectionCentreId } = collectionInfo;
+    return `${collectionDate}, ${collectionSlot}, ${collectionCentreId}`;
+};
+
+const booleanGroupToString = (booleanGroup: BooleanGroup): string => {
+    const trueKeys: string[] = Object.keys(booleanGroup).filter((key) => booleanGroup[key]);
+    return trueKeys.join(", ");
+};
+
+const overrideDataToOverrideDisplayRow = (dataRow: OverrideDataRow): BatchGridDisplayRow => {
```
Stuart-Knowles-SW: There's a lot of helper functions in this file, can we move them to a `displayHelpers.ts` file (or something like that) to keep it readable?

src/app/parcels/batch/BatchParcelsReducer.ts
```diff
@@ -35,128 +34,128 @@ const getOverridenFieldsAndValues = (
...
-}
+// function createBooleanGroupFromStrings(strings: string[] | null): BooleanGroup {
+//     const result: BooleanGroup = {};
+//     if (strings) {
+//         strings.forEach((str) => {
+//             result[str] = true;
+//         });
+//     }
+//     return result;
+// }
```
Stuart-Knowles-SW: If these functions aren't currently required we don't need to comment them out, we can just not call them. Also, similarly to the above, I think it might be worth moving these to a `reducerHelpers.ts` file to keep things compartmentalised

Stuart-Knowles-SW (commented):
A few things to look at. One thing to keep in mind is that – because there's a lot of code here – we'll need to keep really on top of how we're splitting it between files and maintaining the structure to keep it clean and logical

src/app/parcels/batch/BatchParcelsReducer.ts
ebinezer04: this checks for existence of the override row. if it exists, then we check for the not null values in that row and apply overrides to those corresponding columns.

Stuart-Knowles-SW (approved):
LGTM

PR #477, opened 2024-08-14
VFB 300 create user form should reset after successful creation

LeaHe188:
## What's changed
Create user forms will now reset after successful creation. Fix up the UI so that it fits into one page as well, and it now supports copying and pasting. The user form also supports adding more than 3 users in a row (the limit depends on the subscription plan for supabase).

Also fix the edit user form issue in the ticket comments, and now we can edit the user details.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `UserRoleSelect` component for managing user roles.
	- Enhanced user account creation and editing forms with additional fields for email, first name, last name, and phone number.
	- Added a new optional email property for user profile updates.
	- Launched `DropDownSelect` component with versatile dropdown functionality, including `ControlledSelect` and `UncontrolledSelect` variations.

- **Bug Fixes**
	- Improved error handling and state management for text inputs.
  
- **Documentation**
	- Updated interface definitions to reflect new properties and functionality.

- **Style**
	- Added new styled components for better form layouts and user experience.

- **Refactor**
	- Streamlined component structures and props for better integration and usability.

- **Chores**
	- Updated import statements and restructured component usage.


src/app/admin/createUser/CreateUserForm.tsx
```diff
@@ -59,7 +64,7 @@ const getServerErrorMessage = (serverError: InviteUserError): string => {
     }
 };
 
-const formSections = [AccountDetails, UserRoleCard, UserDetailsCard];
+const formSections = [AccountDetails, UserDetailsCard, UserRoleCard];
 
 const CreateUserForm: React.FC = () => {
     const [fields, setFields] = useState(initialFieldValues);
```
ebinezer04, line 68: Maybe make these names a bit clearer e.g. userFields? Same for the initial field values

LeaHe188: I think there's only one field in each file, so these names won't be hard to distinguish?

src/app/admin/createUser/onChangetextDeferredError.ts
```diff
@@ -0,0 +1,29 @@
...
+    additionalCondition?: (value: string) => boolean
+): SelectChangeEventHandler => {
+    return (event) => {
+        InvitedUserSetter?.(null);
+        const input = event.target.value;
+        const errorType = getErrorType(input, required, regex, additionalCondition);
+        errorSetter({ [key]: errorType } as {
+            [key in keyof FormErrors<SpecificFields>]: Errors;
+        });
+        const newValue = formattingFunction ? formattingFunction(input) : input;
```
ebinezer04: Does this need to be typed?

LeaHe188: Just discovered that react automatically recognizes it as string :)

src/app/admin/common/UserRoleDropdownInput.tsx
```diff
@@ -11,16 +11,45 @@ const roleLabelsAndValues: [string, UserRole][] = [
...
+                    return (
+                        <MenuItem key={value} value={value}>
+                            {label}
+                        </MenuItem>
+                    );
+                })}
+            </Select>
+        </FormControl>
+    );
+};
```
Stuart-Knowles-SW: This is very similar to the existing common component. Can we make this a new common component instead? I'm thinking that the current `DropdownListInput` can be altered to take both `value` and `defaultValue`, but then we don't export it. Instead, have two wrapper components `ControlledDropdownInput` and `UncontrolledDropdownInput`, which only take the specific props for the correct type of input.

You can read a bit more about it [here](https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable) :)

src/app/admin/createUser/AccountDetails.tsx
```diff
@@ -1,18 +1,21 @@
...
     formErrors,
     errorSetter,
+    InvitedUserSetter,
 }) => {
     return (
-        <GenericFormCard
+        <UserFormCard
             title="Account Details"
             text="Please enter the email for the new user."
             required
```
Stuart-Knowles-SW: I know they pre-exist your work, but can we remove the redundant Fragment tags from here?

src/app/admin/createUser/CardFormat.tsx
Stuart-Knowles-SW, line 1: This can be a generic component, since we might want to use it for all the other form cards later. Rather than define it fresh, can we alter `GenericFormCard` to take a prop `variant: "relaxed" | "compact"` which defaults to `relaxed`. relaxed should render the original form card, and "compact" should render the new one you've written here.

As a part of this, all of the styling and components you've written here should be moved to the components folder.

src/app/admin/createUser/CreateUserForm.tsx
```diff
@@ -99,23 +104,26 @@ const CreateUserForm: React.FC = () => {
...
+                {formSections.map((Card: React.FC<UserFormProps>, index) => {
                     return (
                         <Card
                             key={index} // eslint-disable-line react/no-array-index-key
                             fields={fields}
                             fieldSetter={fieldSetter}
                             formErrors={formErrors}
                             errorSetter={errorSetter}
+                            InvitedUserSetter={setInvitedUser}
                         />
```
Stuart-Knowles-SW: I know this is outside the scope of your ticket, but this is really not a good pattern so we're going to take the opportunity to clean it up. Let's talk about it in person

src/components/Form/formStyling.tsx
```diff
@@ -36,6 +36,31 @@ export const StyledCard = styled(Paper)`
...
+    border-radius: 10px;
+    background-color: ${(props) => props.theme.main.background[0]};
+    color: ${(props) => props.theme.main.foreground[0]};
+
+    & > div {
+        width: 100%;
+        margin: 0.15em 0;
+    }
+`;
+
```
Stuart-Knowles-SW: Because these are generic, can we call them something generic, maybe swap `User` for `Compact`?

src/app/admin/createUser/onChangetextDeferredError.ts
Stuart-Knowles-SW, line 1: Since this is a generic function which we might need elsewhere, can we move it to the generic `formFunctions`? This also might be worth a comment explaining why we need it.

src/app/admin/createUser/onChangetextDeferredError.ts
```diff
@@ -0,0 +1,29 @@
...
+    errorSetter: Setter<FormErrors<SpecificFields>> | Setter<Required<FormErrors<SpecificFields>>>,
+    key: keyof SpecificFields,
+    required?: boolean,
+    regex?: RegExp,
+    InvitedUserSetter?: { (user: User | null): void; (arg0: null): void } | undefined,
+    formattingFunction?: (value: string) => SpecificFields[keyof SpecificFields],
+    additionalCondition?: (value: string) => boolean
+): SelectChangeEventHandler => {
+    return (event) => {
+        InvitedUserSetter?.(null);
```
Stuart-Knowles-SW: This is the only part which is not generic, but if we rename the prop to something like `additionalAction?: () => void` then pass in the "clear invited user" as a void function we can keep this generic.

src/app/admin/manageUser/EditUserForm.tsx
```diff
@@ -21,18 +34,65 @@ interface Props {
...
+    const initialFirstName: string =
+        props.userToEdit.firstName !== null ? props.userToEdit.firstName : "";
+
+    const initialLastName: string =
+        props.userToEdit.lastName !== null ? props.userToEdit.lastName : "";
+
+    const initialEmail: string = props.userToEdit.email !== null ? props.userToEdit.email : "";
+
+    const initialPhone: string =
+        props.userToEdit.telephoneNumber !== null ? props.userToEdit.telephoneNumber : "";
```
Stuart-Knowles-SW: All of these can be simplified with the nullish coalescing operator:
`const initialFirstName: string =  props.userToEdit.firstName ?? "";`
Though that's probably terse enough to just write inline in the initial fields

src/app/admin/manageUser/EditUserForm.tsx
```diff
@@ -56,18 +115,73 @@ const EditUserForm: React.FC<Props> = (props) => {
...
+                <EditHeader>Email Address</EditHeader>
+                <FreeFormTextInput
+                    id="edit-user-email-address"
+                    label="Email"
+                    defaultValue={fields.email}
+                    error={errorExists(formErrors.email)}
+                    helperText={errorText(formErrors.email)}
+                    onChange={onChangeText(fieldSetter, errorSetter, "email", true, emailRegex)}
+                    fullWidth={true}
+                />
```
Stuart-Knowles-SW: I don't think we want to be able to edit the email address? It's pretty much the thing that defines a user, if they need a new one we'd just delete the account and make a new one

src/app/admin/manageUser/EditUserForm.tsx
Stuart-Knowles-SW: I think we still need to address this?

Also, I realised there was a typo in the original comment, I typed a single question mark rather than a double. Have fixed now :)

src/components/DataInput/README.md
Stuart-Knowles-SW, line 1: Can we add some docs for the `ControlledSelect`?

src/components/Form/formStyling.tsx
```diff
@@ -10,30 +10,67 @@ export const CenterComponent = styled.div`
...
     display: flex;
-    flex-direction: column;
     padding: 2em;
-    width: 100%;
     height: 80%;
     border-radius: 10px;
     background-color: ${(props) => props.theme.main.background[0]};
     color: ${(props) => props.theme.main.foreground[0]};
-
+    flex-direction: ${(props) => (props.$compact === undefined ? "column" : props.$compact ? "row" : "column")};
```
Stuart-Knowles-SW: Undefined is falsy so I think we can just keep the second boolean check here: `props.$compact ? "row" : "column"`

src/components/Form/formStyling.tsx
```diff
@@ -10,30 +10,67 @@ export const CenterComponent = styled.div`
     margin-bottom: 1rem;
 `;
 
-export const StyledForm = styled.form`
+export const StyledForm = styled.form<{ $compact?: boolean }>`
+    ${(props) =>
+        props.$compact
+            ? `
+    width: 1000px;`
```
Stuart-Knowles-SW: Can width be relative here so it adapts to different screen sizes?

src/app/admin/manageUser/EditUserForm.tsx
LeaHe188: Fixed now, the code was complaining last week so didn't adopt this form, but now it's not complaining any more

src/components/DataInput/DropDownSelect.tsx
```diff
@@ -40,4 +42,26 @@ const DropdownListInput = <Value,>(props: Props<Value>): React.ReactElement => {
...
+    return (
+        <GenericSelect
+            selectLabelId={props.selectLabelId}
+            listTitle={props.listTitle}
+            defaultValue={props.defaultValue}
+            labelsAndValues={props.labelsAndValues}
+            onChange={props.onChange}
+        />
+    );
+};
```
Stuart-Knowles-SW: Can we rename `Props` to `GenericProps` then have two different sets of props for the two components based on what they can and can't accept?

While we're at it, let's rename `Value` to `ValueType` for clarity

Stuart-Knowles-SW (commented):
Great work, just a few more things to fix :)

src/components/DataInput/DropDownSelect.tsx
```diff
@@ -3,17 +3,35 @@
...
     selectLabelId: string;
     focusOnDropdown?: boolean;
 }
 
-const GenericSelect = <Value,>(props: Props<Value>): React.ReactElement => {
+interface ControlledProps<ValueType> {
+    labelsAndValues: [string, string][];
+    listTitle?: string;
+    value?: ValueType;
+    onChange?: (event: SelectChangeEvent<ValueType>) => void;
```
Stuart-Knowles-SW: I think `value` and `onChange` need to be required here?

Stuart-Knowles-SW (approved):
LGTM, thanks for the hard work on this :)

PR #484, opened 2024-08-19
Vfb 315 make address field editable and add phone number validation

SoftJoelWire:
## What's changed
Added validation for phone number field. Any phone number cell will appear as the theme error colour if a non-valid number is entered. 

Added input method for address cell using the MUI popover component which has built in validation.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image|

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced components for editing address information within a data grid, enhancing user interface and experience.
	- Added phone number validation functionality to ensure proper formatting during data entry.

- **Improvements**
	- Enhanced visibility of address-related utility functions and regex constants for better usability across components.
	- Standardized empty data structures for batch editing, promoting consistency and maintainability.
	- Improved rendering of data grid columns with centralized configuration, ensuring a more organized layout.

- **Bug Fixes**
	- Improved internal logic for handling updates in batch processing, ensuring data integrity during edits.


src/app/parcels/batch/AddressEditCell.tsx
```diff
@@ -0,0 +1,79 @@
...
+    dispatchBatchTableAction,
+}) => {
+    const id = gridRenderCellParams.id as number;
+    const apiRef = useGridApiContext();
+    const [popoverAnchorEl, setPopoverAnchorEl] = useState<null | HTMLElement>(null);
+    const gridCellReference = useRef<HTMLDivElement>(null);
+    const open = Boolean(popoverAnchorEl);
+    useEffect(() => {
+        setPopoverAnchorEl(gridCellReference.current);
+    }, []);
```
ebinezer04: Is the useEffect and the set function required? As both the popover element and the ref are already defaulted to null

src/app/parcels/batch/AddressEditCellInput.tsx
```diff
@@ -0,0 +1,214 @@
...
+            : tableState.overrideDataRow.data.client.address?.addressLine2 ?? undefined;
+    const defaultAddressTown: string | undefined =
+        id !== 0
+            ? tableState.batchDataRows[id - 1].data.client.address?.addressTown ?? undefined
+            : tableState.overrideDataRow.data.client.address?.addressTown ?? undefined;
+    const defaultAddressCounty: string | undefined =
+        id !== 0
+            ? tableState.batchDataRows[id - 1].data.client.address?.addressCounty ?? undefined
+            : tableState.overrideDataRow.data.client.address?.addressCounty ?? undefined;
+    const defatulAddressPostcode: string | undefined =
```
ebinezer04: Spelling  :(

src/app/parcels/batch/AddressEditCellInput.tsx
```diff
@@ -0,0 +1,214 @@
...
+const AddressEditCellInput: React.FC<AddressEditCellInputProps> = ({
+    tableState,
+    id,
+    dispatchBatchTableAction,
+    simulateEscapeKeyPress,
+}) => {
+    const defaultAddressLine1: string | undefined =
+        id !== 0
+            ? tableState.batchDataRows[id - 1].data.client.address?.addressLine1 ?? undefined
+            : tableState.overrideDataRow.data.client.address?.addressLine1 ?? undefined;
```
ebinezer04: Possibly could extract `tableState.batchDataRows[id - 1].data.client.address?` into its own variable, just to not keep writing the same thing? Same for override data row.

src/app/parcels/batch/AddressEditCellInput.tsx
```diff
@@ -0,0 +1,214 @@
...
+                            isPostcodeNotValid ||
+                            !addressInputFields?.addressPostcode?.match(postcodeRegex)
+                        ) {
+                            valid = false;
+                            setIsPostcodeNotValid(true);
+                        }
+                        if (valid || clientHasNoAddress) {
+                            dispatchInputFieldData();
+                        }
+                    }}
```
ebinezer04: Could extract this into its own function for readability?

ebinezer04:
For accessibility purposes, if the address is too long to be displayed on the cell, would it be worth making the cell scrollable or some other way to see the whole address?

ebinezer04:
Change file names to be consistent with naming conventions e.g. emptyData

src/app/parcels/batch/AddressEditCell.tsx
SoftJoelWire: Yes, as without it either the anchorEl will always be null or infinite re-renders will occur

src/app/parcels/batch/AddressEditCell.tsx
```diff
@@ -0,0 +1,79 @@
...
+    tableState: BatchTableDataState;
+    dispatchBatchTableAction: React.Dispatch<BatchActionType>;
+}
+
+const AddressEditCell: React.FC<AddressEditCellProps> = ({
+    gridRenderCellParams,
+    tableState,
+    dispatchBatchTableAction,
+}) => {
+    const id = gridRenderCellParams.id as number;
```
Stuart-Knowles-SW: Why do we have to assert this is a number? Is there a chance it's not?

src/app/parcels/batch/AddressEditCell.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <div ref={gridCellReference}>
+            <div>
+                {addressToString(
+                    id !== 0
+                        ? tableState.batchDataRows[id - 1].data.client.address ?? null
+                        : tableState.overrideDataRow.data.client.address ?? null
+                ) ?? ""}
+            </div>
+            <Popover
+                open={open && Boolean(popoverAnchorEl)}
```
Stuart-Knowles-SW: `open` is already defined as `Boolean(popoverAnchorEl)` above, so this is a little bit redundant. Also, we prefer explicit comparisons rather than naive castings: so something like: `open={popoverAnchorEl !== null}` would be better.

src/app/parcels/batch/AddressEditCell.tsx
```diff
@@ -0,0 +1,79 @@
...
+                onClose={() => {
+                    apiRef.current.setEditCellValue({
+                        id: id,
+                        field: gridRenderCellParams.field,
+                        value:
+                            addressToString(
+                                id !== 0
+                                    ? tableState.batchDataRows[id - 1].data.client.address ?? null
+                                    : tableState.overrideDataRow.data.client.address ?? null
+                            ) ?? "",
```
Stuart-Knowles-SW: I think these nullish checks are redundant? Also, can we reverse the order of the ternary so it's `id === 0?`? I think it's a little be easier if there's one fewer negative

src/app/parcels/batch/AddressEditCellInput.tsx
```diff
@@ -0,0 +1,219 @@
...
+        id !== 0
+            ? batchRowAddress?.addressPostcode ?? undefined
+            : overrideRowAddress?.addressPostcode ?? undefined;
+    const [addressInputFields, setAddressInputFields] = useState<AddressInputFields>({
+        addressLine1: defaultAddressLine1,
+        addressLine2: defaultAddressLine2,
+        addressTown: defaultAddressTown,
+        addressCounty: defaultAddressCounty,
+        addressPostcode: defaultAddressPostcode,
+    });
```
Stuart-Knowles-SW: I think we can clean this up a bit.

1. Typescript is smart enough to work out types, so we don't need the `: string | undefined` declarations.
2. Some of the nullish operators `??` are redundant. Specifically, for address line 1, town and postcode, the variables are already constrained to `string | undefined` so we don't need them there. The others will need a check to convert `null` to `undefined`, but we can do this with one `??` not two by putting it outside the ternary operator. However, there may be a better way of doing it:
3. We can extract this into a helper method. Specifically, something that takes `(id, tableState) => AddressInputFields`. Something like this:
```
const getInitialAddressFields = (
    id: number,
    tableState: BatchTableDataState
): AddressInputFields => {
    const currentRowAddress: Address | null =
        id === 0
            ? tableState.overrideDataRow.data.client.address
            : tableState.batchDataRows[id - 1].data.client.address;
    return {
        addressLine1: currentRowAddress?.addressLine1,
        addressLine2: currentRowAddress?.addressLine2 ?? undefined,
        addressTown: currentRowAddress?.addressTown,
        addressCounty: currentRowAddress?.addressCounty ?? undefined,
        addressPostcode: currentRowAddress?.addressPostcode,
    };
};
```

Then the top of the component is just: 
```
const [addressInputFields, setAddressInputFields] = useState<AddressInputFields>(
    getInitialAddressFields(id, tableState)
);
```

src/app/parcels/batch/BatchParcelDataGrid.tsx
```diff
@@ -34,200 +38,224 @@ export interface BatchGridDisplayRow {
...
-        field: "id",
-        headerName: "Row Number",
-        width: 150,
-        editable: false,
-        renderCell: (params) => {
-            if (params.row.id === 0 && params.field === "id") {
-                return <Button variant="contained">Apply Column</Button>;
-            }
-            return params.row.id;
+export const getCenteredBatchGridDisplayColumns = (
```
Stuart-Knowles-SW: This is a big function, can we move it into a file by itself?

src/app/parcels/batch/BatchParcelDataGrid.tsx
```diff
@@ -34,200 +38,224 @@ export interface BatchGridDisplayRow {
...
-            columns={styledBatchGridDisplayColumns}
+            columns={centeredBatchGridDisplayColumns}
             sx={{
                 "& .MuiDataGrid-cell": {
                     border: "1px solid",
-                    borderColor: "rgba(81, 81, 81, 0.5)",
+                    borderColor: `${theme.main.border}`,
+                    "& .MuiInputBase-root": {
+                        height: "100%",
+                    },
```
Stuart-Knowles-SW: Given that our usual pattern is styled components, might be worth leaving a small comment above the component explaining why we're doing it here.

src/app/parcels/batch/BatchParcelsReducer.ts
```diff
@@ -129,6 +140,28 @@ export const batchParcelsReducer = (
...
+            //     },
+            // };
+            // const updatedRows: BatchDataRow[] = state.batchDataRows.map((row) => {
+            //     return row.id === rowId ? newRow : row;
+            // });
+            // return {
+            //     ...state,
+            //     batchDataRows: updatedRows,
+            // };
+            throw Error("Not implemented"); // TODO
```
Stuart-Knowles-SW: Why's this been duplicated?

src/app/parcels/batch/FieldValidationFunctions.ts
```diff
@@ -0,0 +1,11 @@
...
+import { phoneNumberRegex } from "@/common/format";
+
+export const phoneNumberValidation = (params: GridPreProcessEditCellProps): GridEditCellProps => {
+    const unformattedInput = params.props.value.replaceAll(/[\s-()]/g, "");
+    const hasError = unformattedInput.match(phoneNumberRegex) === null;
+    return {
+        ...params.props,
+        error: hasError,
+    };
+};
```
Stuart-Knowles-SW: Can we commonise this with the phone number validators elsewhere?

Stuart-Knowles-SW (commented):
A few comments here to get started. This is a pretty chunky PR so I'm expecting it to go through a few rounds of review.

Great work though, this is complicated stuff

src/app/parcels/batch/AddressEditCell.tsx
SoftJoelWire: The built in type is of type string | number and must be used as part of the renderEditCell property. However our rows are defined explicitly with ids only being numbers so casting should be safe here.

src/app/parcels/batch/FieldValidationFunctions.ts
SoftJoelWire: Its using the same exported regex from the form validation, is this what you mean?

src/app/parcels/batch/AddressEditCell.tsx
Stuart-Knowles-SW: Cool, let's put a comment explaining that :)

src/app/parcels/batch/AddressEditCell.tsx
```diff
@@ -0,0 +1,82 @@
...
+const AddressEditCell: React.FC<AddressEditCellProps> = ({
+    gridRenderCellParams,
+    tableState,
+    dispatchBatchTableAction,
+}) => {
+    const id = gridRenderCellParams.id as number;
+    const apiRef = useGridApiContext();
+    const [popoverAnchorEl, setPopoverAnchorEl] = useState<null | HTMLElement>(null);
+    const gridCellReference = useRef<HTMLDivElement>(null);
+    const open = popoverAnchorEl !== null;
```
Stuart-Knowles-SW: If we're going to define this as a variable, let's be strict with our naming conventions: `const isEditComponentOpen` or something. However, given that it's only used in one place I'd skip defining it and just calculate it inline in the jsx code:
```
    <Popover
        open={popoverAnchorEl !== null}
```

src/app/parcels/batch/AddressEditCell.tsx
```diff
@@ -0,0 +1,82 @@
...
+        }
+    };
+    return (
+        <div ref={gridCellReference}>
+            <div>
+                {addressToString(
+                    id === 0
+                        ? tableState.overrideDataRow.data.client.address
+                        : tableState.batchDataRows[id - 1].data.client.address
+                ) ?? ""}
```
Stuart-Knowles-SW: This is calculated twice, let's do it above the return statement and reuse the result. I know the second usage is dynamic, so we might want to define a function with `useCallback` that calculates it then call that function in the return statement

src/app/parcels/batch/FieldValidationFunctions.ts
Stuart-Knowles-SW: Hmm, I was wondering if we could commonise it a bit more with an `isPhoneNumberValid` function. If that's not feasible, can we at least make the `phonNumberIgnoredCharactersRegex` an exported constant

src/app/parcels/batch/AddressEditCellInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+    margin-top: 0.5rem;
+`;
+
+interface AddressInputFields {
+    addressLine1?: string;
+    addressLine2?: string;
+    addressTown?: string;
+    addressCounty?: string;
+    addressPostcode?: string;
+}
```
Stuart-Knowles-SW: I think we can remove the need for this by just using `Partial<Address>`, which makes all of the properties optional.
`getInitialAddressFields` would then look like:
```
const getInitialAddressFields = (id: number, tableState: BatchTableDataState): Partial<Address> => {
    const currentRowAddress =
        id === 0
            ? tableState.overrideDataRow.data.client.address
            : tableState.batchDataRows[id - 1].data.client.address;
    return currentRowAddress ?? {};
};
```

Stuart-Knowles-SW (commented):
Couple of little comments and one general request: There's already quite a lot of files in this directory, can we organise them a bit more? Eeven if we just make a `displayComponents` directory where all of the new components created here can live

Stuart-Knowles-SW (approved):
LGTM, thanks!

PR #488, opened 2024-08-21
VFB 307 prevent overlapping names on shipping labels

LeaHe188:
## What's changed
The names on shipping labels can now be displayed without overlapping with other texts. Meanwhile all the information will remain on the shipping label. The format of the label has been refactored so that the delivery instructions can be as large as possible. Note that the delivery instruction has a limit of 320 characters.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
|image|image|

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced styling of the card component for better flexibility and layout.
	- Improved visual presentation of shipping labels with new styles for recipient information.
	- Added delivery instructions field to forms, enhancing error tracking and validation.
	- Introduced real-time error feedback for delivery instructions in input forms.

- **Bug Fixes**
	- Adjusted height and spacing in the card component to enhance responsiveness.
	- Streamlined layout of the shipping label components for improved clarity.

### DIFF
--- a/src/components/Form/formFunctions.ts
+++ b/src/components/Form/formFunctions.ts
@@ -66,12 +66,11 @@ export const createSetter = <SpecificFields extends Fields>(
     };
 };
 
-export const getErrorType = <SpecificFields extends Fields>(
+export const getErrorType = (
     input: string,
-    key: keyof SpecificFields,
     required?: boolean,
     regex?: RegExp,
-    additionalCondition?: (value: string) => boolean
+    additionalCondition?: (value: string, maxCharacters?: number) => boolean
 ): Errors => {
     if (input == "") {
         return required ? Errors.required : Errors.none;
@@ -80,9 +79,6 @@ export const getErrorType = <SpecificFields extends Fields>(
         (regex !== undefined && !input.match(regex)) ||
         (additionalCondition !== undefined && !additionalCondition(input))
     ) {
-        if (key === "deliveryInstructions") {
-            return Errors.deliveryInstructions;
-        }
         return Errors.invalid;
     }
 
@@ -96,7 +92,7 @@ export const onChangeText = <SpecificFields extends Fields>(
     required?: boolean,
     regex?: RegExp,
     formattingFunction?: (value: string) => SpecificFields[keyof SpecificFields],
-    additionalCondition?: (value: string) => boolean
+    additionalCondition?: (value: string, maxCharacters?: number) => boolean
 ): SelectChangeEventHandler => {
     return (event) => {
         const input = event.target.value;
@@ -104,7 +100,6 @@ export const onChangeText = <SpecificFields extends Fields>(
             key === "telephoneNumber" || key === "phoneNumber"
                 ? input.replaceAll(phoneNumberFormatSymbolsRegex, "")
                 : input,
-            key,
             required,
             regex,
             additionalCondition
@@ -134,7 +129,7 @@ export const onChangeTextDeferredError = <SpecificFields extends Fields>(
             clearInvitedUser();
         }
         const input = event.target.value;
-        const errorType = getErrorType(input, key, required, regex, additionalCondition);
+        const errorType = getErrorType(input, required, regex, additionalCondition);
         errorSetter({ [key]: errorType } as {
             [key in keyof FormErrors<SpecificFields>]: Errors;
         });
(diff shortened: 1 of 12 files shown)


Jessica-Al (commented):
All looks good to me

src/pdf/ShippingLabels/ShippingLabelsPdf.tsx
```diff
@@ -122,7 +126,9 @@ const SingleLabelCard: React.FC<LabelCardProps> = ({ data, index, quantity }) =>
                     </View>
                     <View style={styles.middleCol}>
                         <Text style={styles.headingText}>Delivery Instructions: </Text>
-                        <Text>{data.delivery_instructions}</Text>
```
Stuart-Knowles-SW, line 125: Nice catch!

Stuart-Knowles-SW (commented):
This does solve the problem, But I think we can have a better solution if we think more broadly, because text wrapping the name is still not ideal. If we always have the name below the "NAME:" label, we'll both have more room by default and if it overflows it'll slot below the contact and date information.

You also mentioned that if the delivery instructions are too long they overflow. Are there any knock-on problems if we fix that by unconstraining the height of the label?

src/components/Form/formFunctions.ts
```diff
@@ -23,6 +23,7 @@ export enum Errors {
     invalidPackingSlot = "The previous packing slot is no longer available, please select a new packing slot.",
     invalidCollectionCentre = "The previous collection centre is no longer available, please select a new collection centre.",
     noCollectionSlotsSet = "There are no collection slots set for this collection centre, please select a different collection centre or contact admin.",
+    deliveryInstructions = "Please enter delivery instructions below 320 words.",
```
Stuart-Knowles-SW: Should this be 320 characters?

src/components/Form/formFunctions.ts
```diff
@@ -69,12 +70,15 @@ export const getErrorType = (
...
+    additionalCondition?: (value: string) => boolean,
+    deliveryInstructionsOverflow?: boolean
 ): Errors => {
     if (input == "") {
         return required ? Errors.required : Errors.none;
     }
-
+    if (deliveryInstructionsOverflow) {
+        return Errors.deliveryInstructions;
+    }
```
Stuart-Knowles-SW: Is there a way we can do this generically?

src/pdf/ShippingLabels/ShippingLabelsPdf.tsx
```diff
@@ -21,12 +21,15 @@ export interface ShippingLabelData {
     delivery_instructions?: string;
 }
 
-const mmToPixels = (mm: number): number => {
-    const pixelsPerMmAt72Dpi = 72 / 25.4;
-    return mm * pixelsPerMmAt72Dpi;
-};
+// This is a hard requirement because of the foodbank's printer settings, note that height is adjustable but preferably set at 62mm
```
Stuart-Knowles-SW: Looks like the change talked about wasn't committed. The height is a hard requirement, fixed at 62mm, the width is technically adjustable but preferred at 150

Stuart-Knowles-SW (commented):
Great work. I'd like to generalise the error process, I think we might be able to use the `additionalCondition` argument, but other than that we're nearly there.

src/components/Form/formFunctions.ts
LeaHe188: I've moved the character limit condition into additionalConditions, but also adding a argument "key" when getting error type, so that I can specify the error message

src/components/Form/formFunctions.ts
```diff
@@ -65,20 +66,23 @@ export const createSetter = <SpecificFields extends Fields>(
...
-
     if (
         (regex !== undefined && !input.match(regex)) ||
         (additionalCondition !== undefined && !additionalCondition(input))
     ) {
+        if (key === "deliveryInstructions") {
+            return Errors.deliveryInstructions;
+        }
         return Errors.invalid;
     }
```
Stuart-Knowles-SW: We might want to implement character limits elsewhere. Can this just be a super generic character limit error? rather than passing `key` as a prop we can just pass `maxCharacters?: number` then check the length inside

Stuart-Knowles-SW (approved):
Great work, thanks!

PR #490, opened 2024-08-22
Vfb 315 make remaining table cells editable

SoftJoelWire:
## What's changed
All cells are now editable in the batch add table with custom UI components with in built verification. Previously, only up to the Address Cell were editable.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image| image|

## Checklist


src/app/clients/form/formSections/NumberAdultsCard.tsx
```diff
@@ -14,9 +14,7 @@ import { FormText, StyledCard } from "@/components/Form/formStyling";
 import { ClientCardProps, ClientSetter } from "@/app/clients/form/ClientForm";
 import { UncontrolledSelect } from "@/components/DataInput/DropDownSelect";
 import { getAdultBirthYears } from "@/app/clients/form/birthYearDropdown";
-
-const MINIMUM_NUMBER_OF_ADULTS = 1;
-const MAXIMUM_NUMBER_OF_ADULTS = 20;
+import { MAXIMUM_NUMBER_OF_ADULTS, MINIMUM_NUMBER_OF_ADULTS } from "../bounds";
```
rjesson-softwire: fix the import format to be "@/..."

src/app/parcels/batch/getCenteredBatchGridDisplayColumns.tsx
```diff
@@ -55,81 +104,217 @@ const getCenteredBatchGridDisplayColumns = (
...
+                        booleanGroupLabelAndKeys={dietaryReqsLabelsAndKeys}
+                    />
+                );
+            },
         },
         {
             field: "feminineProducts",
             headerName: "Feminine Products",
-            width: 150,
+            width: FEMINIE_PRODUCTS_WIDTH,
```
rjesson-softwire: **feminine (spelling mistake)

src/app/parcels/batch/getCenteredBatchGridDisplayColumns.tsx
```diff
@@ -140,26 +325,65 @@ const getCenteredBatchGridDisplayColumns = (
         {
             field: "packingDate",
             headerName: "Packing Date",
-            width: 150,
+            width: 200,
```
rjesson-softwire: is there a reason this width hasn't been changed to a constant like other cells, and is still a number?

src/app/parcels/batch/fieldValidationFunctions.ts
```diff
@@ -1,11 +1,7 @@
-import { GridEditCellProps, GridPreProcessEditCellProps } from "@mui/x-data-grid";
+import { GridPreProcessEditCellProps } from "@mui/x-data-grid";
 import { phoneNumberRegex, phoneNumberFormatSymbolsRegex } from "@/common/format";
 
-export const phoneNumberValidation = (params: GridPreProcessEditCellProps): GridEditCellProps => {
+export const phoneNumberValidation = (params: GridPreProcessEditCellProps): boolean => {
```
rjesson-softwire: could change function name to reflect it just performs a check, and doesn't do any validation - maybe something like isPhoneNumberValid

src/app/parcels/batch/displayHelpers.ts
```diff
@@ -48,12 +50,18 @@ export const addressToString = (address: Address | null): string | null => {
     return addressArray.join(", ");
 };
 
-const collectionInfoToString = (collectionInfo: CollectionInfo): string => {
-    const { collectionDate, collectionSlot, collectionCentreId } = collectionInfo;
-    return `${collectionDate}, ${collectionSlot}, ${collectionCentreId}`;
+export const collectionInfoToString = (collectionInfo: CollectionInfo | null): string | null => {
+    if (!collectionInfo) {
+        return null;
```
rjesson-softwire: in collectionInfoToString, we return null if !collectionInfo, while in the function booleanGroupToString, we return an empty string if !booleanGroup. Is there a reason for this inconsistency? If so, maybe collectionInfoToString should be renamed as it could return null.

rjesson-softwire (commented):
first round of comments 😅

src/app/parcels/batch/getCenteredBatchGridDisplayColumns.tsx
SoftJoelWire: The others are constants are they are used in multiple files. This one is only configured here. Do you think it would be clearer to have all column widths, regardless of multiple file use, in the columnWidths file?

src/app/parcels/batch/displayComponents/AddressEditCellInput.tsx
```diff
@@ -68,9 +59,13 @@ const AddressEditCellInput: React.FC<AddressEditCellInputProps> = ({
                         fieldName: "address",
                         newValue: {
                             addressLine1: addressInputFields.addressLine1 ?? "",
-                            addressLine2: addressInputFields.addressLine2 ?? null,
+                            addressLine2: addressInputFields.addressLine2
+                                ? addressInputFields.addressLine2
+                                : null,
```
rjesson-softwire: the removed line and new changes seem to do the same thing, and the old version looks a bit cleaner

same thing for below

src/app/clients/form/formSections/DietaryRequirementCard.tsx
```diff
@@ -4,26 +4,28 @@ import { checkboxGroupToArray, onChangeCheckbox } from "@/components/Form/formFu
 import GenericFormCard from "@/components/Form/GenericFormCard";
 import { ClientCardProps } from "../ClientForm";
 
+export const dietaryReqsLabelsAndKeys: [string, string][] = [
```
Stuart-Knowles-SW: Given this is now a top level constant can with capital case it (`DIETARY_REQS_LABELS_AND_KEYS`)?

Edit: this applies in a few places.

src/app/parcels/batch/batchTypes.ts
```diff
@@ -24,8 +24,12 @@ export interface CollectionInfo {
     collectionDate: string;
     collectionSlot: string;
     collectionCentreId: string;
+    collectionCentreName: string;
+    collectionCentreAcronymn: string;
 }
 
+type BabyProducts = "Yes" | "No" | "Don't Know";
```
Stuart-Knowles-SW: Purely as convention, can we call this `BabyProductsType`?

src/app/parcels/batch/displayComponents/AddressEditCellInput.tsx
Stuart-Knowles-SW: The functional difference would be that in the old implementation empty strings `""` would not be converted to null, but they are in the new implementation.

We can keep it clean by doing `addressLine2: addressInputFields.addressLine2 || null`, which would convert empty strings

src/app/parcels/batch/displayComponents/BooleanGroupEditCellInput.tsx
```diff
@@ -0,0 +1,63 @@
...
+    currentBooleanGroup,
+    clientField,
+    fieldWidth,
+    booleanGroupLabelAndKeys,
+}) => {
+    return (
+        <BooleanGroupCheckboxesDiv width={fieldWidth}>
+            {booleanGroupLabelAndKeys.map((bGLK: [string, string]) => {
+                const bGLabel: string = bGLK[0];
+                const bGKey: string = bGLK[1];
```
Stuart-Knowles-SW: Can we just spell out these variable names?

src/app/parcels/batch/displayComponents/AddressEditCellInput.tsx
SoftJoelWire: Previous version does not check for all falsey values, only null/undefined. new version assigns any falsey value to null

src/app/parcels/batch/displayComponents/PersonEditCellInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+    personField,
+    personArray,
+}) => {
+    const [people, setPeople] = useState<Person[]>([]);
+    useEffect(() => {
+        setPeople(personArray);
+    }, [personArray]);
+    const MIN_NUMBER_PEOPLE =
+        personField === "childrenInfo" || id === 0 ? 0 : MINIMUM_NUMBER_OF_ADULTS;
+    const MAX_NUMBER_PEOPLE = MAXIMUM_NUMBER_OF_ADULTS;
```
Stuart-Knowles-SW: non-top level constants should be camelCase

src/app/parcels/batch/displayComponents/PersonEditCellInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+                          type: "client",
+                          fieldName: personField,
+                          newValue: {
+                              children: newPoeple,
+                              numberOfChildren: newPoeple.length,
+                          },
+                      },
+                  },
+              });
+    };
```
Stuart-Knowles-SW: Firstly, typo in `newPeople`. Secondly, I think we can simplify this by defining `newValue` conditionally then having oly one dispatch call:
```
    const dispatchPerson = (newPoeple: Person[]): void => {
        const newValue =  personField === "adultInfo"
            ? {
                 adults: newPoeple,
                 numberOfAdults: newPoeple.length,
             }
            : {
                 children: newPoeple,
                 numberOfChildren: newPoeple.length,
             },
        dispatchBatchTableAction({
                  type: "update_cell",
                  updateCellPayload: {
                      rowId: id,
                      newValueAndFieldName: {
                          type: "client",
                          fieldName: personField,
                          newValue
                      },
                  },
              })
    };
```

src/app/parcels/batch/displayComponents/PersonInput.tsx
```diff
@@ -0,0 +1,109 @@
...
+            ? getAdultBirthYears().map((year) => [`${year}`, `${year}`] as [string, string])
+            : getChildBirthYears().map((year) => [`${year}`, `${year}`] as [string, string]);
+    const genderLabelsAndValues: [string, string][] = [
+        ["Male", "male"],
+        ["Female", "female"],
+        ["Prefer Not To Say", "other"],
+    ];
+    const defaultMonthString = people[index].birthMonth
+        ? dayjs(people[index].birthMonth).month() + 1
+        : "";
```
Stuart-Knowles-SW: This feels weird. It's called "string" but it looks like it can be a number. I also don't fully understand what the `dayjs` call is for, does that convert between e.g., `"April"` and `3`?

Stuart-Knowles-SW (commented):
I've left a few comments. To be honest, this is far too large to review properly. We should have kept a closer eye on that earlier, but there's nothing we can do about it now. The high level code structure looks decent, and any smaller issues can be found and fixed in QA.

I don't love taking the "just merge then see if it's broken" approach, but given this is still essentially a feature in alpha I think we can get away with it.

Stuart-Knowles-SW (approved):

PR #495, opened 2024-08-27
Vfb 318 Create Records in DB on submit

ebinezer04:
## What's changed
Added a submit button at the bottom of the data grid component that submits the batch data into the database. 
On submit, a success or error dialog will display.
Added error checks for invalid/incomplete data, a further PR will correctly convey this information to the user.
Implement an add button to add new rows.
Implement a button on the navigation bar to switch to the batch creation page.

## Checklist


src/app/parcels/batch/displayComponents/BatchParcelDataGrid.tsx
```diff
@@ -52,50 +59,173 @@ export const defaultTableState: BatchTableDataState = {
...
+
     const theme: DefaultTheme = useTheme();
+
     const centeredBatchGridDisplayColumns: GridColDef[] = getCenteredBatchGridDisplayColumns(
         tableState,
         dispatch
     );
+
+    const handleSubmit = async (): Promise<void> => {
+        console.log(tableState.batchDataRows);
```
rjesson-softwire: are these console.logs supposed to be here?

src/app/parcels/batch/submitTableData.ts
```diff
@@ -0,0 +1,315 @@
...
+    return hasRequiredFields && validCollectionOrDelivery;
+};
+
+const checkForEmptyParcel = (parcel: ParcelData): boolean => {
+    return Object.values(parcel).every((value) => value == null);
+};
+
+const checkForEmptyRow = (data: BatchEditData): boolean => {
+    return data === emptyBatchEditData;
+};
```
rjesson-softwire: these functions could be renamed to be clearer in exactly what theyre checking - taking the last one as an example, maybe a name like checkRowIsNotEmpty might be more appropriate? the name checkForEmptyRow sounds to me like it's taking an array of rows and checking if it contains an empty row.

src/app/parcels/batch/submitTableData.ts
```diff
@@ -0,0 +1,315 @@
...
+        const logId = await logErrorReturnLogId(
+            "Error with inserting new client and their family",
+            {
+                error,
+            }
+        );
+        return { clientId: null, error: { type: "failedToInsertClientAndFamily", logId } };
+    }
+
+    console.log(`Row ${rowId}: Client ${client.fullName}, ${clientId} added successfully`);
```
rjesson-softwire: should this console log be here?

rjesson-softwire (commented):
generally looks really good! just a few little comments

src/app/parcels/batch/displayComponents/BatchParcelDataGrid.tsx
```diff
@@ -37,65 +45,277 @@ export interface BatchGridDisplayRow {
...
 
+const countErrorTypes = (errors: AddBatchRowError[]): Record<string, number> => {
+    return errors.reduce(
+        (acc, rowError) => {
+            acc[rowError.displayMessage] = (acc[rowError.displayMessage] || 0) + 1;
+            return acc;
+        },
+        {} as Record<string, number>
+    );
+};
```
Stuart-Knowles-SW: I really like this function, but I think there's a couple of tweaks we can make.
Firstly, lets use `??` rather than `||`, I think it makes the intention a bit clearer.

Secondly, let's declare the initial state above the `errors.reduce` call, to improve type safety:
`const initialState: Record<string, number> = {};`. Although this looks basically the same, it's more type safe because `as` overrides type errors, whereas declaring the type of a `const` doesn't. So `{'foo': 'bar'} as Record<string, number>` doesn't throw a type error but `const initialState: Record<string, number> = { 'foo': 'bar' }` does. Although either works in this case, we prefer the second pattern because it's safer in general. Using `as` is usually a red flag so we should avoid it when we can.

src/app/parcels/batch/displayComponents/BatchParcelDataGrid.tsx
```diff
@@ -37,65 +45,277 @@ export interface BatchGridDisplayRow {
...
+
 const BatchParcelDataGrid: React.FC = () => {
     const [tableState, dispatch] = useLocalStorage(batchParcelsReducer, defaultTableState);
+    const [isConfirmationDialogVisible, setIsConfirmationDialogVisible] = useState(false);
+    const [isSuccessDialogVisible, setIsSuccessDialogVisible] = useState(false);
+    const [dialogMessage, setDialogMessage] = useState("");
```
Stuart-Knowles-SW: Can we combine these two into one state? i.e., `dialogMessage` is `string | null` and the dialog is open based on `dialogMessage !== null`

src/app/parcels/batch/displayComponents/BatchParcelDataGrid.tsx
```diff
@@ -37,65 +45,277 @@ export interface BatchGridDisplayRow {
...
+            handleReset();
+        } else {
+            // setDialogMessage(`There were ${submitErrors.length} error(s) encountered during submission.`);
+            setDialogMessage("Errors encountered during submission.");
+            handleUnsubmittedRows(submitErrors);
+        }
+
+        setIsConfirmationDialogVisible(false);
+        setIsSuccessDialogVisible(true);
+    };
```
Stuart-Knowles-SW: For these comments, rather than talk about the next PR, can you create a Jira ticket to do these changes then reference that ticket number in a todo here. Then we can delete the second/third comments. Something like:
```
        // VFB-XXX improve messaging on batch form submission
        if (submitErrors.length == 0) {
            setDialogMessage("Successfully submitted these rows");
            handleReset();
        } else {
            setDialogMessage("Errors encountered during submission.");
            handleUnsubmittedRows(submitErrors);
        }
```

src/app/parcels/batch/displayComponents/BatchParcelDataGrid.tsx
```diff
@@ -37,65 +45,277 @@ export interface BatchGridDisplayRow {
...
+                open={isConfirmationDialogVisible}
+                sx={{
+                    "& .MuiDialog-paper": {
+                        padding: "1rem",
+                        borderRadius: "15px",
+                        boxShadow: "0 3px 5px rgba(0,0,0,0.8)",
+                        margin: "auto",
+                        borderColor: confirmationErrors.length == 0 ? "#a8d49c" : "#990f0f",
+                        borderWidth: "1px",
+                        borderStyle: "solid",
```
Stuart-Knowles-SW: Is `sx` required for a Dialog as it is for `DataGrid`, or can we do this with a styled component? Same for all the ones below where we use sx.

This is a low priority change, so probably leave as is but tell me whether we could move the styling into styled components and I can do it later.

src/app/parcels/batch/displayComponents/BatchParcelDataGrid.tsx
```diff
@@ -37,65 +45,277 @@ export interface BatchGridDisplayRow {
...
+                            Close
+                        </Button>
+                        {confirmationErrors.length == 0 && (
+                            <Button
+                                variant="contained"
+                                onClick={handleSubmit}
+                                sx={{ margin: "1rem", minWidth: "120px" }}
+                            >
+                                Submit
+                            </Button>
```
Stuart-Knowles-SW: Rather than having this conditionally rendered, could we `disable` it based on the errors?

src/app/parcels/batch/displayComponents/BatchParcelDataGrid.tsx
```diff
@@ -37,65 +45,277 @@ export interface BatchGridDisplayRow {
...
+                            <Button
+                                variant="contained"
+                                sx={{ margin: "1rem", minWidth: "120px" }}
+                                onClick={handleReturnToParcels}
+                            >
+                                Return to the Parcels Page
+                            </Button>
+                        </>
+                    )}
```
Stuart-Knowles-SW: I'd ideally like these two options split out into their own components. Something like:
```
                <DialogActions sx={{ justifyContent: "center" }}>
                    {submitErrors.length > 0 ? 
                        <BatchParcelsPostSumbitErrorButtons ... />
                     : <BatchParcelsPostSumbitSuccessButtons ... />
                    )}
```

src/app/parcels/batch/submitTableData.ts
```diff
@@ -0,0 +1,250 @@
...
+        flagged_for_attention: client.attentionFlag,
+        notes: client.notes,
+    };
+};
+
+const batchClientToInsertRecords = (
+    client: BatchClient
+): { clientRecord: ClientDatabaseInsertRecord; familyMembers: FamilyDatabaseInsertRecord[] } => {
+    const adults: Person[] = client.adultInfo?.adults || [];
+    const children: Person[] = client.childrenInfo?.children || [];
```
Stuart-Knowles-SW: same here, let's use `??` not `||`

Stuart-Knowles-SW (commented):
Few changes but overall this looks really good, really nice work on the submission and verification functions

src/app/parcels/batch/displayComponents/BatchParcelDataGrid.tsx
ebinezer04: sx is not required for the Dialog or the Button so styled components can be used

src/app/parcels/batch/displayComponents/BatchParcelDataGrid.tsx
ebinezer04: Implemented this. It might be worth doing the same for the separate dialogs for confirmation and post submission.

Stuart-Knowles-SW:
I've ended up bundling a lot of stuff into this PR which didn't really belong here, but for the sake of expediency we're going to push ahead with this

PR #500, opened 2024-08-28
VFB-314 fix pagination in packing manager view and fix areDatesIdenti…

rjesson-softwire:
## What's changed
Fixed the pagination on the packing manager view table, and disabled the row breaks. Fixed the areDaysIdentical function to check if two days are the same, rather than checking if two date times are the same.

## Screenshots / Videos
Before
image
After
image

## Checklist

### DIFF
--- a/src/app/parcels/parcelsTable/ParcelsTable.tsx
+++ b/src/app/parcels/parcelsTable/ParcelsTable.tsx
@@ -27,7 +27,7 @@ import { parcelTableColumnStyleOptions } from "@/app/parcels/parcelsTable/styles
 import parcelsSortableColumns, {
     defaultParcelsSortConfig,
 } from "@/app/parcels/parcelsTable/sortableColumns";
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import dayjs from "dayjs";
 import {
     shouldFilterBeDisabled,
@@ -190,7 +190,6 @@ const ParcelsTable: React.FC<ParcelsTableProps> = ({
                 }
             } else {
                 setParcelsDataPortion(data.parcelTableRows);
-                setFilteredParcelCount(data.count);
                 if (sortState.sortEnabled && sortState.column.headerKey) {
                     setParcelRowBreakPointConfig(
                         searchForBreakPoints(sortState.column.headerKey, data.parcelTableRows)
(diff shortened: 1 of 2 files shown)


ebinezer04 (commented):
LGTM

src/app/parcels/parcelsTable/ParcelsTable.tsx
```diff
@@ -226,15 +225,22 @@ const ParcelsTable: React.FC<ParcelsTableProps> = ({
...
+            });
+        },
+        [today, yesterday]
+    );
+
+    useEffect(() => {
+        isPackingManagerView
+            ? setFilteredParcelCount(packingManagerViewDataPortion(parcelsDataPortion).length)
+            : setFilteredParcelCount(parcelsDataPortion.length);
+    }, [isPackingManagerView, parcelsDataPortion, packingManagerViewDataPortion]);
```
Stuart-Knowles-SW: Is there a way we can do this without an effect? This use case is pretty much one of the examples of when you shouldn't use an effect in the [React docs](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state)

src/app/parcels/parcelsTable/ParcelsTable.tsx
```diff
@@ -226,15 +225,22 @@ const ParcelsTable: React.FC<ParcelsTableProps> = ({
...
-    const packingManagerViewDataPortion = (
-        parcelsDataPortion: ParcelsTableRow[]
-    ): ParcelsTableRow[] => {
-        return parcelsDataPortion.filter((parcel) => {
-            if (shouldBeInPackingManagerView(parcel, today, yesterday)) {
-                return parcel;
-            }
-        });
-    };
+    const packingManagerViewDataPortion = React.useCallback(
```
Stuart-Knowles-SW: Very minor point: we're already importing `useCallback`, so we don't have to do `React.` and can avoid importing `React`

Stuart-Knowles-SW: If this is a function could we name it as such: `getPackingManagerParcelView`. Though I think we can just memoise the calculation and avoid using a function altogether. This way does solve the problem of calculating the Packing Manager Parcel view when we're on the normal view, at the cost of having to calculate it every render when we're in the Packing Manager view. I think a larger refactor would be required to actually solve both problems at once, but I think I prefer memoisation to this approach.

src/app/parcels/parcelsTable/ParcelsTable.tsx
```diff
@@ -225,22 +224,17 @@ const ParcelsTable: React.FC<ParcelsTableProps> = ({
...
-        },
-        [today, yesterday]
-    );
+    const packingManagerViewDataPortion = useMemo(() => {
+        return parcelsDataPortion.filter((parcel) => {
+            if (shouldBeInPackingManagerView(parcel, today, yesterday)) {
+                return parcel;
+            }
+        });
+    }, [parcelsDataPortion, today, yesterday]);
```
Stuart-Knowles-SW: `Array.filter` works on boolean return types from the filter function, and if there's only one statement in a function we can remove the curly brackets and return statement. Overall, this means we can simplify this a bit more:
```
const packingManagerViewDataPortion = useMemo(() =>
    parcelsDataPortion.filter((parcel) => shouldBeInPackingManagerView(parcel, today, yesterday))
    );
}, [parcelsDataPortion, today, yesterday]);
```

Stuart-Knowles-SW (approved):

PR #502, opened 2024-08-29
VFB 325 pdf tweaks

LeaHe188:
## What's changed
Cosmetic changes to shopping lists, shipping labels and driver overviews.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |
| image | image |
| image | image |

## Checklist


src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -248,27 +274,38 @@ const DriverOverviewCard: React.FC<DriverOverviewCardProps> = ({ data }) => {
...
     return (
         <Document>
-            <Page size="A4" style={[styles.container, styles.flexColumn]}>
-                <View style={[styles.infoAndLogoContainer, styles.flexRow]}>
-                    <View style={styles.informationContainer}>
-                        <Text style={styles.h1text}>Driver Overview</Text>
-                        <Text style={styles.h2text}>
+            <Page size="A4" orientation="landscape" style={[styles.container, styles.flexColumn]}>
+                <View style={styles.DriverOverviewBoard}>
+                    <View style={{ flexDirection: "column", fontFamily: "Helvetica-Bold" }}>
```
rjesson-softwire, line 279: there are quite a few views where you add the fontFamily: "Helvetica-Bold" styling. If this is used everywhere on the PDF, is it possible to move this so it is the default font for the page, to reduce the number of times you need to define the font?

rjesson-softwire (commented):
LGTM!

src/pdf/DriverOverview/DriverOverviewPdf.tsx
LeaHe188: There's still a place where I used helvetica (not bold), so maybe not setting the bold one as default is better?

package.json
```diff
@@ -136,6 +137,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
+    "lodash": "^4.17.21",
```
Stuart-Knowles-SW: Lodash is a very heavyweight module, have we checked what this does to the bundle size? Even if it's negligible, we've already got a `capitaliseWords` function in `format.ts` so we can just use that

src/app/clients/form/formSections/DeliveryInstructionsCard.tsx
```diff
@@ -21,7 +21,7 @@ const DeliveryInstructionsCard: React.FC<ClientCardProps> = ({
         <GenericFormCard
             title="Delivery Instructions"
             required={false}
-            text="Please input not more than 320 characters."
+            text="Please input not more than 380 characters."
```
Stuart-Knowles-SW: Given that we define `MAX_CHARACTERS` above, can we format it into this string rather than hard coding.

src/pdf/ShippingLabels/ShippingLabelsPdf.tsx
```diff
@@ -82,70 +89,90 @@ interface LabelCardProps {
...
-                            {data.full_name}
-                        </Text>
-                    </View>
-                    <View style={styles.thirdRow}>
-                        {data.address_postcode && (
-                            <>
-                                <Text>
-                                    {data.address_1}
-                                    <br />
+        data.full_name !== "Deleted Client" && (
```
Stuart-Knowles-SW: It's best practice to not have conditional rendering at the top level of a component. Ideally what we'd do is move this check outside, to wherever this component is called.

src/pdf/ShippingLabels/ShippingLabelsPdf.tsx
```diff
@@ -82,70 +89,90 @@ interface LabelCardProps {
     quantity: number;
 }
 
+const convertDateFormat = (date: string): string => {
+    const dateArray = date.split("-");
+    return `${dateArray[2]}/${dateArray[1]}/${dateArray[0]}`;
+};
```
Stuart-Knowles-SW: This is a little implicit for my liking. Could we do it with `dayjs(date).format(...)` instead?

src/pdf/ShoppingList/ShoppingListPdf.tsx
```diff
@@ -107,9 +117,25 @@ interface OneLineProps {
...
+    const displayValue =
+        header === "LIST TYPE"
+            ? capitalize(value)
+            : header === "HOUSEHOLD SIZE" && value === "1 (1 Adult 0 Child)"
+              ? "Single (1 Adult 0 Child)"
+              : header === "AGE AND GENDER OF ADULTS" || header === "AGE AND GENDER OF CHILDREN"
+                ? value.replace(/M/g, "Male").replace(/F/g, "Female").replace(/O/g, "Other")
+                : header === "BABY PRODUCTS REQUIRED" && value.includes("No")
+                  ? "None"
+                  : value;
```
Stuart-Knowles-SW: This many nested ternary operators is a red flag, and adding these extra checks makes the purpose of this function a lot less clear.

Can we move the first boolean check into a `shouldDisplay` function, then filter the list of inputs where OneLine is called.

And then move the ternary operators into a `getDisplayValue` function, probably using a `switch...case` instead of the ternaries

src/pdf/ShippingLabels/ShippingLabelsPdf.tsx
```diff
@@ -89,14 +90,11 @@ interface LabelCardProps {
...
-    return `${dateArray[2]}/${dateArray[1]}/${dateArray[0]}`;
-};
-
 const SingleLabelCard: React.FC<LabelCardProps> = ({ data, index, quantity }) => {
-    return (
-        data.full_name !== "Deleted Client" && (
+    if (data.full_name === "Deleted Client") {
+        return null;
+    } else {
+        return (
```
Stuart-Knowles-SW: Sorry, I might not have been clear. `SingleLabelCard` should always return JSX, no conditionals or null returns. Where it's called (`ShippingLabelForSingleParcel`) we should filter the array we map across using this conditional.

Stuart-Knowles-SW (approved):

PR #506, opened 2024-09-03
VFB-none internship cleanup

Stuart-Knowles-SW:
## What's Changed

A few little cleanup tasks from the internship. Mainly places where the interns made minimal changes but actually slightly larger refactors were better.

- Fixed the logic when mapping and filtering the parcels to prepare them for the shipping label pdfs
- Changed the logic for max input length on an input field to give better validation.
- Fixed the delete confirmation modals to properly reset when multiple are used in quick succession
- Fixed the delete client modal to properly redirect back to the clients page.


Stuart-Knowles-SW:
I'm bypassing merge restrictions because these changes are blocking some of the other in-flight work I need to merge. I've QA'd these myself and I think they're pretty low risk but I'll QA again when they're in dev

PR #507, opened 2024-09-05
VFB-None Improvements to styling that were found while investigating VFB-88

Stuart-Knowles-SW:
## What's changed

Whilst investigating [VFB-88](https://softwiretech.atlassian.net/browse/VFB-88), a few chore-level improvements to styling were found.

- Remove `@types/styled-components` - it's not required for `styled-components` v6 and above, and is in fact not up to date with v6 changes.
- standardised import pattern for `styled` (named vs default import)
- Fixed a bug on the `FreeFormTextInput` Component where an invalid prop was being passed to the DOM
- updated the `DefaultTheme` definition to include the text property

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,6 @@
         "@types/react-dom": "18.2.6",
         "@types/react-fontawesome": "^1.6.5",
         "@types/react-modal": "^3.16.0",
-        "@types/styled-components": "^5.1.26",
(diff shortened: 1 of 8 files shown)


Stuart-Knowles-SW:
Closed because I found a way to actually solve VFB-88 so I'm bundling these changes with that

PR #508, opened 2024-09-06
VFB-88 Fix styling warnings

Stuart-Knowles-SW:
## What's changed
Fix the warnings in dev about props and styled components, as well as a few other styling pattern improvements

The fix for VFB-88 was that we were only using a `StyleSheetManager` (which handles prop forwarding) for server rendering, and it needed to be added for client rendering as well.

In addition:

- Remove @types/styled-components - it's not required for styled-components v6 and above, and is in fact not up to date with v6 changes.
- standardised import pattern for styled (named vs default import)
- Fixed a bug on the FreeFormTextInput Component where an invalid prop was being passed to the DOM
- updated the DefaultTheme definition to include the text property

## Notes:
- This does mean some styles that were previously not being properly applied are now handled correctly. Whilst this may make a few things visually different, it overall means that what we're seeing now lines up with what the CSS says, so any changes are positive. 
  - The only changes I noticed are subtle ones to the nav bar, but I can't guarantee there aren't others.
  - Those changes have been fixed in this PR by altering the CSS

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -43,6 +43,10 @@ export default defineConfig({
         setupNodeEvents(on: Cypress.PluginEvents, config: Cypress.PluginConfigOptions) {
             on("task", {
                 readPdf,
+                log: (message: unknown) => {
+                    console.log(message);
+                    return null;
+                },
             });
             return config;
         },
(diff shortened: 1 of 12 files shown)


stonelink (approved):
Thank you for fixing this! I'm relieved about the StyleSheetManager fix especially.

Issue #401, opened 2024-07-11
Client DOB – need an option to keep blank


dan-langridge:
### Describe the solution / feature you'd like

When entering a client DOB, need an option to leave it blank if we don't know

stonelink:
That is, the Year for both Adults and Children should be optional (and the default)

stonelink:
Blank year of birth 
= "Don't Know" or "Unknown" in UI, or no need to select from a droplist
= null in Database (in Families table)

Plan:
- update DB schema
- update form field (see precedent of optional child birth month)
- update client display in modal (perhaps "unknown age" for both adults & children)
- update client info on shopping list
- update batch client display & creation (see precedent of optional child birth month)
- ~update seed data to include some with unknown age~ (not changing seed data because of Snaplet shutdown)

~Note regarding Adults vs Children: because of the logic that defines a child as being < 16 years old, the Edit Client form still requires that the birth year be set for Children.~

stonelink:
Hi @dan-langridge, there are a few options here and I'd like to get your opinion on how important the ideal fix is if it would delay the release.

There is an initial solution in review that makes the adult birth year nullable. Because of the logic that states that defines a child as being any person under 16, each child's birth year is still required. To make that optional might push the next release back by a day or two. 

The other question I have is whether _birth year_ is the appropriate UI for the callers to use. I wonder whether there was a misunderstanding of the requirements on our side early on... Before I took extended leave last year I discussed storing ages in the database using the year and potentially month - however that doesn't need to be the way it's presented in the UI. Changing the UI to have simple inputs for age in years and/or months (but still storing an inferred birth year & month) would probably push the release back by 3 days.

What are your thoughts on how this should work?

dan-langridge:
I think being able to nullify for adult and child is a useful feature, not a problem if that pushes it back a few days.

I think the UI of adding a birth year (and month for child) is fine and gives us what we need. I don’t think it will be confusing. Plus entering the birth year gives us accurate info in the future, where as just entering how old they are at the time of entering the info will mean that info would go out of date right, and agree not worth adding the logic to work it out when we can just ask for birth year.

But yes, option to nullify both is going to be needed, thanks!

stonelink:
## QA Notes
* Tested as Admin & Volunteer
* Tested on current Chrome and Safari 13.1 (via Browserstack)

QA steps:
* creation of client record with multiple adults and children, all with no information about birth year or month
[...]

PR #529, opened 2024-10-22
#401 Make client birth years optional

stonelink:
## What's changed
The request here was that client family members' dates of birth should be optional. To distinguish between a child and an adult without birth year, there is a new optional flag 'recorded_as_child', which indicates that that is the field in which the person details were entered. Note that if a child has a birth year recorded, that takes precedence so when they turn 16 they will be listed as an adult.

## Checklist

  - [X] I have updated the typescript definitions for the database with `npm run generate_types:local`
    - Checkout the dev branch
    - Checkout your branch

### DIFF
--- a/src/app/clients/form/submitFormHelpers.ts
+++ b/src/app/clients/form/submitFormHelpers.ts
@@ -24,17 +24,7 @@ export const getFamilyMembersForDatabase = (
     adults: Person[],
     children: Person[]
 ): FamilyDatabaseInsertRecord[] => {
-    const peopleToInsert = children
-        .map((child) => {
-            child.recordedAsChild = true;
-            return child;
-        })
-        .concat(
-            adults.map((adult) => {
-                adult.recordedAsChild = false;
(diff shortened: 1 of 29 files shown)


src/app/clients/form/submitFormHelpers.ts
```diff
@@ -16,14 +16,25 @@ const personToFamilyRecordWithoutFamilyId = (person: Person): FamilyDatabaseInse
...
     };
 };
 
-export const getFamilyMembers = (
+export const getFamilyMembersForDatabase = (
     adults: Person[],
     children: Person[]
 ): FamilyDatabaseInsertRecord[] => {
-    const peopleToInsert = children.concat(adults);
+    const peopleToInsert = children
```
stonelink: This isn't necessary for new family members, because they're created with the new field in ClientForm.tsx.
We're not worried about updating pre-existing adults & children to have the flag, so let's remove this change to simplify.

Issue #519, opened 2024-09-23
Allow Admin to edit "Dietary Requirements" and "Other Items" form responses


dan-langridge:
### Describe the solution / feature you'd like

The admin should be able to edit the responses to the Client Form sections Dietary Requirements and Other Items, as these items change seasonally (for example at Christmas we offer Christmas foods such as Yams, and we remove Hot Water bottles in the summer)

stonelink:
## Notes for this ticket
See the changes under #518 for how the groups of options are currently defined in code (and that has been made more generalised & null-friendly). It should now be possible to move those lists of options (dietary, feminine, pet, other) into the database.

stonelink:
@dan-langridge 
I have a couple of questions about the behaviour of this feature.

* When seasonal items change:
  * Because the dietary & other requirements are associated with the client (rather than the parcel), they will be in the database after the seasonal items change. 
  * I assume that the Client Details & Shopping List should only show those things that are currently in the Admin list - is that right?

* Defining the list of items
  * Should the Admins' interface for editing the options have a set of free text fields where they can edit, add, delete and reorder?
 
And a couple of notes:
* I think this change will probably be made for all checkbox group questions in the Client form at the same time: dietary requirements, feminine products, pet food, other items and cooking facilities
* Implementing this change will probably take 4 days, depending on the time I have available
* I'm not sure about the priority of this vs. the signposting report

dan-langridge:
@stonelink 
I assume that the Client Details & Shopping List should only show those things that are currently in the Admin list - is that right?
- Yes that is correct. So for example, if in November we offer "Melted Cheese", and in December the product runs out, when I delete "Melted Cheese", it should also delete this from previous clients who had "Melted Cheese" selected.

Should the Admins' interface for editing the options have a set of free text fields where they can edit, add, delete and reorder? 
- Yes that sounds good. Similar to the Shopping List, without the comments and quantity box, if thats what you are thinking? Google Form UI  for example for multiple choice answers is fine for this.

I think this change will probably be made for all checkbox group questions in the Client form at the same time: dietary requirements, feminine products, pet food, other items and cooking facilities
- Perfect - the  ability to change all responses to these would be ideal

Implementing this change will probably take 4 days, depending on the time I have available
-No problem

I'm not sure about the priority of this vs. the signposting report
This is more important for go live than signposting. However signposting is probably next?

stonelink:
## Possible approaches

### Predefined questions
[...]

Issue #518, opened 2024-09-23
Amend Dietary Form Responses


dan-langridge:
### Describe the solution / feature you'd like

Amend the client dietary form responses the below:

**Dietary Requirements**

Don’t Know
Fresh Fruit
Fresh Veg
Garlic
Ginger
Chillies
Spices
Bread
Tea
Coffee
Pasta
Rice
Meat (No Pork)
Meat & Pork
Gluten Free
Dairy Free
Vegetarian
Vegan
Pescatarian
Halal
Diabetic
Nut Allergy
Seafood Allergy

**Other Items**

Hot Water Bottle
Blanket

stonelink:
### Notes on "Don't Know"
- "Don't Know" should be represented by null entry in the db
- in the UI, if "Don't Know" is ticked then the remaining checkboxes should be disabled
- in the Shopping List, "Don't Know" should be written for null entry in db

### Notes on test data
- the existing data should ideally undergo migration, that specifically:
  - updates clients with Garlic/Ginger/Chillies/Spices under "Other Items" so that is under "Dietary Requirements"
  - the "No" options removed (i.e., No Bread, No Pasta etc.), with the inverse set in their absence (i.e., Bread, Pasta, ...)
  - most test clients should have "Fresh Fruit" and "Fresh Veg" ticked; perhaps the migration should be that all clients have these and then some specific test cases where:
    - the client has "Don't Know" for dietary
    - the client has some combination of no fruit/veg
  - "Hot Water Bottle" and "Blanket" are now singular
- the seed data will need to be updated - this will be covered by VFB-273

stonelink:
### Next steps
* update the insert/update sprocs to handle null dietaryRequirements
* migration for existing data

stonelink:
## QA Notes
* Tested as Admin & Volunteer
* Tested on current Chrome and Safari 13.1 (via Browserstack)

QA steps:
* Check a random selection of existing client records, and verify that the "Dietary Requirements" and "Other Requirements" in both the Client Details dialog and Shopping List are consistent with the list on this ticket - that is:
  * no old entries such as "No Beef" or the plural of "Blankets"
  * the order of the entries is consistent with the above
* create a new client record and leave "Don't Know" ticked under Dietary Requirements
  * verify that this shows as expected in the Client Details dialog and Shopping List
* create a new client record with a selection of Dietary and Other Requirements
  * verify that this shows as expected in the Client Details dialog and Shopping List
[...]

PR #533, opened 2024-11-04
#518 Change dietary requirements responses

stonelink:
## What's changed
This ticket specified changes for the responses for "dietary requirements" and "other items" in client records. They're defined in the code (although that would change when we implement #519).
Notable other changes are the addition of a "Don't Know" response for dietary requirements (represented by NULL in the DB) and ordering the entries for all the similar questions (dietary, feminine, pet, other) based on a given canonical ordering.

Note that because snaplet doesn't work, I haven't regenerated the seed data or done the checks that involve regenerating the db.

## Checklist

  - [ ] I have updated the typescript definitions for the database with `npm run generate_types:local`
    - Checkout the dev branch
    - Checkout your branch

### DIFF
--- a/src/app/clients/edit/[id]/autofill.ts
+++ b/src/app/clients/edit/[id]/autofill.ts
@@ -38,7 +38,10 @@ const autofill = (
         numberOfChildren: children.length,
         children: children,
         listType: clientData.default_list,
-        dietaryRequirements: arrayToBooleanGroup(clientData.dietary_requirements ?? []),
+        dietaryRequirements:
+            clientData.dietary_requirements !== null
+                ? arrayToBooleanGroup(clientData.dietary_requirements)
+                : null,
         feminineProducts: arrayToBooleanGroup(clientData.feminine_products ?? []),
         babyProducts: clientData.baby_food,
         nappySize: nappySize.replace("Nappy Size: ", ""),
--- a/src/app/clients/form/submitFormHelpers.ts
+++ b/src/app/clients/form/submitFormHelpers.ts
@@ -6,10 +6,6 @@ import { ClientFields } from "./ClientForm";
 import { AuditLog, sendAuditLog } from "@/server/auditLog";
 import { ListType } from "@/common/databaseListTypes";
 import { EXTRA_INFORMATION_LABEL, NAPPY_SIZE_LABEL } from "@/app/clients/form/labels";
-import { dietaryRequirementOptions } from "./formSections/DietaryRequirementCard";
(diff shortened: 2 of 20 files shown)


src/app/clients/edit/[id]/autofill.ts
```diff
@@ -38,7 +38,10 @@ const autofill = (
         numberOfChildren: children.length,
         children: children,
         listType: clientData.default_list,
-        dietaryRequirements: arrayToBooleanGroup(clientData.dietary_requirements ?? []),
+        dietaryRequirements:
+            clientData.dietary_requirements !== null
+                ? arrayToBooleanGroup(clientData.dietary_requirements)
+                : null,
         feminineProducts: arrayToBooleanGroup(clientData.feminine_products ?? []),
```
Comment (stonelink) [inline src/app/clients/edit/[id]/autofill.ts:45]: It would be better for all these fields that are nullable in the DB to be handled properly here - feminine, pet, other

src/app/clients/form/submitFormHelpers.ts
```diff
@@ -46,11 +50,14 @@ export const formatClientRecord = (
         address_county: fields.addressCounty,
         address_postcode: fields.addressPostcode,
         default_list: fields.listType as ListType,
-        dietary_requirements: checkboxGroupToArray(fields.dietaryRequirements),
-        feminine_products: checkboxGroupToArray(fields.feminineProducts),
+        dietary_requirements:
+            fields.dietaryRequirements !== null
+                ? checkboxGroupToArray(fields.dietaryRequirements, dietaryRequirementOptions)
```
stonelink: I don't think that this needs to sort the fields because it's just what's going into the database - better for checkboxGroupToArray() to just convert & not sort. Leave the canonical sorting to be only for UI

src/components/Form/formFunctions.ts
```diff
@@ -256,8 +256,42 @@ export const getErrorText = (errorType: Errors, maxCharacters?: number): string
...
+            }
+        } else if (firstIndex !== -1 && secondIndex === -1) {
+            return -1;
+        }
+
+        return firstIndex - secondIndex;
+    });
+};
+
+export const checkboxGroupToArray = (
```
stonelink: Comment from earlier - I don't think this needs to do the canonical sorting

src/app/clients/edit/[id]/autofill.ts
Comment (stonelink) [reply src/app/clients/edit/[id]/autofill.ts:45]: Making that change would require more logic in presenting and handling the null / Don't Know case in the UI for each thing, so have decided to leave as-is

Issue #576, opened 2025-02-07
Changes to Client Form Questions


dan-langridge:
### Is your feature request related to a problem? If so, please describe.

A few changes to wording and answers on the client form

### Describe the solution / feature you'd like

**Feminine Hygiene Products**

1. Rename “Feminine Hygiene Products” to “Hygiene Products”

2. New Answers:
- Tampons
    - If Yes, “How Many”
- Pads
    - If Yes, “How Many”
- Female Incontinence Pads
- Male Incontinence Pads

**Baby Products**

3. New Answers:

- Nappies
		- If Yes, “What Size”
- Baby Formula 
		- If yes, “Specify brand and stage”
- Baby Food
		- If Yes, “More Info”
- Baby Wipes
- Baby Toiletries

stonelink:
## Preparatory notes:
- The Hygiene Products section is currently a list of options (like dietary) in the field `feminine_products`.
  - This change to optional "how many" fields will complicate the plan to have all these fields edited by an admin; I should consider how to keep admin editability.
- Baby needs are currently spread between `baby_food` and `extra_information`
  - The nappy size value is currently stored in the `extra information` field in the database, which is an historic anomaly; therefore part of this ticket will rectify that and move this data into a baby field or fields
  - Also these optional detail fields should be solved in the same way as the how many fields under Hygiene Products

### JSON/JSONB field solution
Each of these fields (and eventually the comparable fields for dietary requirements etc.) would be a JSON/JSONB array field such that each element of the array has the form `{ optionName: boolean | string }`, wherein:
* each key (`optionName`) corresponds to a user-facing string such as "Baby Formula" or "Male Incontinence Pads"
* if the value of the option is a boolean then it corresponds to a checkbox, and `{ "this option": false }` is equivalent to `"this option"` being absent
* if the value of the option is a string then it corresponds to `true` with extra detail

There is a question over how complex it is worth making this structure. 
The advantages of a JSON field are:
* it encapsulates the detail from a set of questions, which may change over time, into a single DB field
* it theoretically makes it possible for these questions to be edited by admins in future
But it isn't clear how reasonable the editing process will be. If the name of an option changes then we don't want to be changing every client record, which implies that the client record needs to have a code for each option - and does the admin choose the codes?

### Separate fields
[...]

PR #582, opened 2025-02-09
#576 Client form changes for hygiene & baby products

stonelink:
## What's changed
The client field `feminine products`, which was an array of strings mapping to checkboxes, has been replaced by 4 `hygiene` fields. Two are booleans, and two are optional string fields.

The ticket also covers a new set of fields for baby needs, but they have not been implemented yet.

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/databaseTypesFile.ts
```diff
@@ -598,15 +607,7 @@ export type Database = {
...
-        Relationships: [
-          {
-            foreignKeyName: "profiles_user_id_fkey"
-            columns: ["user_id"]
-            isOneToOne: true
-            referencedRelation: "users"
-            referencedColumns: ["id"]
-          },
-        ]
+        Relationships: []
```
stonelink, line 616: It isn't clear why the types generation script removed the user foreign key from `profiles` and `audit_log`. 
From inspection in Supabase, I can't see a difference between my local db and the Dev db.

src/databaseTypesFile.ts
```diff
@@ -1070,3 +1064,18 @@ export type Enums<
     ? PublicSchema["Enums"][PublicEnumNameOrOptions]
     : never
 
+export type CompositeTypes<
```
stonelink, line 1073: This extra definition is probably because a recent commit upgraded Supabase CLI.

src/databaseTypesFile.ts
stonelink: Confirmed. When the GitHub checks was using CLI v1.174.x to generate the types, it failed the comparison with this one generated by v2.9.6 when processing the same db schema.

src/databaseTypesFile.ts
stonelink: Confimed

src/app/batch-create/_tests/batchParcelsReducer.test.ts
```diff
@@ -94,7 +94,6 @@ describe("reducer", () => {
                         listType: null,
                         cookingFacilities: null,
                         dietaryRequirements: null,
-                        feminineProducts: null,
                         babyProducts: null,
```
stonelink: Need to remove baby & nappy fields from all of batch-create. VFB-382 notes that they will need to be added when Batch is re-animated.

src/app/clients/edit/[id]/page.tsx
```diff
@@ -65,7 +65,7 @@ const EditClients: ({ params }: EditClientsParameters) => React.ReactElement = (
         addressPostcode: Errors.none,
         numberOfAdults: Errors.none,
         numberOfChildren: Errors.none,
-        nappySize: Errors.none,
+        babyNappies: Errors.none,
```
Comment (stonelink) [inline src/app/clients/edit/[id]/page.tsx]: `babyNappies` not needed here because not required

src/app/batch-create/_tests/batchParcelsReducer.test.ts
stonelink: Resolved

src/app/clients/edit/[id]/page.tsx
Comment (stonelink) [reply src/app/clients/edit/[id]/page.tsx]: Resolved

PR #701, opened 2025-06-26
NONE: Add workflows for intern-dev env

AlexandraIrinaMardale:


.github/workflows/deploy-dry-intern-dev-database.yml
```diff
@@ -0,0 +1,24 @@
+name: Dry-run deployment to intern-dev database
```
stonelink: Do you need a dry-run for the Intern Dev db? We have that for Staging and Production so there's a sanity check, but there isn't one for the Dev environment.

Btw, is this the Supabase project `test-foodbankapp-dev`?

AlexandraIrinaMardale: Probably not, I though I copy pasted only from the dev ones, but apparently not 🙃 
And yeah, it should be for ``test-foodbankapp-dev``

stonelink (approved):

stonelink (approved):

Issue #697, opened 2025-06-20
Searching parcel select box error


dan-langridge:
### Reproduction steps

I would search for a postcode, and the search wouldn’t work and would still show all of the clients. I would backspace one character, into the character again and it would then show my single client. 
Or
I search a response, and it shows the correct single reponse,

However, straight away:
- the “tick all clients” check box would be ticked
- my client would not be ticked
- I am unable to tick my client
- I would have to refresh the page and then all would be fine.
- I would print a shipping label for that client
- Clear the postcode box
- Search for the next client by postcode
- 
And repeat from above, tick all clients check box is already ticked, I can’t tick my client…..

Video below.

Happens with name search too

### Expected behaviour

When I search a post code or client name, the result should appear, the tick-all box should be de-selected and I should be able to select the client

### What browsers are you seeing the problem on?

Safari

### What OS is your computer running on?

macOS

dan-langridge:
https://github.com/user-attachments/assets/c99bc833-555b-453e-8a43-5cd755643b7f

stonelink:
I've reproduced this behaviour, but not consistently. It's timing-related and when a partial update is done when the filter shows no results, with the view not being properly updated when the parcel list changes.

I believe that it happens because: 
* the "Select All' checbox (top of the column) gets ticked when there are no items in the table; as well as being vital to this bug, the checkbox shouldn't behave like this anyway
* sometimes when the table is updated to contain items, the Select All checkbox isn't being updated appropriately

The two items above need to be fixed.

stonelink:
## QA Results

This is not reliably reproducible, so I've repeatedly run tests on the Parcels page using the Postcode filter in ways that previously showed the bug sometimes - and have not seen the bug.

Tested on:
* Latest Chrome on Windows 11
* Safari 13.1.2 on MacOs High Sierra

QA Passed

dan-langridge:
@stonelink I'm still nothing some errors when searching Postcode. Screen recording below, note 2 situations:
1. (30 seconds on video) Type a postcode, then remove the postcode = the search remains - it should clear and show all postcodes
2. (5seconds on video) Search a postcode CB2 = it shows more like a 'fuzzy' search, shows CB8 for example

dan-langridge:
https://github.com/user-attachments/assets/9868cf1a-3a80-41bc-a6d8-3cc772506543

stonelink:
Hi @dan-langridge 
[...]

PR #705, opened 2025-07-13
#697: Table 'Select All' checkbox is only ticked if nonempty

stonelink:
## What's changed
The Parcels Table previously ticked the Select All box by comparing the length of the selected IDs list to the number of parcels in the table, without checking whether there actually were any in the table.

## Checklist

### DIFF
--- a/src/components/Tables/_tests/WrappedTable.tsx
+++ b/src/components/Tables/_tests/WrappedTable.tsx
@@ -49,7 +49,8 @@ const WrappedTableForTest: React.FC<MockTableProps> = ({
     const [isAllCheckBoxSelected, setAllCheckBoxSelected] = useState(false);
 
     useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/components/Tables/_tests/WrappedTable.tsx
```diff
@@ -49,7 +49,8 @@ const WrappedTableForTest: React.FC<MockTableProps> = ({
     const [isAllCheckBoxSelected, setAllCheckBoxSelected] = useState(false);
 
     useEffect(() => {
-        const allChecked = checkedRowIds.length === testDataPortion.length;
+        const allChecked =
```
stonelink, line 52: This component only exists for tests, and has replicated the checkbox logic from the Parcels table - so no point writing tests of test code.
Ideally the checkbox logic should be filtered out of the Parcels table, so it can be shared and can be properly tested.

PR #711, opened 2025-07-15
decreased spacing for clients, parcel pages

anamariasuditu:
## What's changed
Decreased spacing for rows in client and parcel pages

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| 466395647-99134cc6-52af-4144-ba30-4f3cec5ae63c| after|

## Checklist


src/components/Tables/Table.tsx
```diff
@@ -196,7 +197,7 @@ const CustomCell = <Data,>({
         <>
             {columnDisplayFunctions[headerKey]
                 ? columnDisplayFunctions[headerKey]?.(row.data[headerKey])
-                : row.data[headerKey]}
+                : String(row.data[headerKey])}
```
AlexandraIrinaMardale: I think we don't need this String part, right?

src/components/Tables/Table.tsx
```diff
@@ -429,6 +431,15 @@ const Table = <
 
     const rows = dataPortion.map((data, index) => ({ rowId: index, data }));
 
+    const selectedRows = [
+        {
+            when: () => compressRows == true,
+            style: {
+                padding: "0rem 0rem!important",
```
AlexandraIrinaMardale: Can you set the height instead of changing the padding please? And if there's a way to not use `!important` that would be great 😄

src/components/Tables/Table.tsx
```diff
@@ -429,6 +431,15 @@ const Table = <
 
     const rows = dataPortion.map((data, index) => ({ rowId: index, data }));
 
+    const selectedRows = [
+        {
+            when: () => compressRows == true,
```
AlexandraIrinaMardale: Can we do `===` here please?

src/components/Tables/Table.tsx
```diff
@@ -429,6 +431,15 @@ const Table = <
 
     const rows = dataPortion.map((data, index) => ({ rowId: index, data }));
 
+    const selectedRows = [
```
AlexandraIrinaMardale: Can we rename this please? You can call it `conditionalRowStyles`

src/components/Tables/Table.tsx
```diff
@@ -239,6 +240,7 @@ const Table = <
     paginationConfig,
     editableConfig,
     pointerOnHover,
+    compressRows,
```
AlexandraIrinaMardale: I also think we should rename this to something like `reduceRowHeight` but I'm not sure if this is the best naming option, so if you have any other ideas, feel free to use them:))

PR #712, opened 2025-07-15
VFB-386: Renaming PDF's

mihaicutaiulia:
## What's changed
- change default `formatName` value
- set `formatName` for the next buttons:
    - `DayOverviewPdfButton`
    - `DriverOverviewPdfButton`
    - `ShoppingListPdfButton`
    - `ShippingLabelsPdfButton`

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image|image |
|image|image
|image|image|
|image|image|

## Checklist

### DIFF
--- a/src/components/FileGenerationButtons/PdfButton.tsx
+++ b/src/components/FileGenerationButtons/PdfButton.tsx
@@ -21,7 +21,7 @@ const PdfButton = <Data, ErrorType extends string>({
     pdfComponent: PdfComponent,
     onFileCreationCompleted: onFileCreationCompleted = () => undefined,
     onFileCreationFailed: onFileCreationFailed,
-    formatName = false,
+    formatName = true,
     disabled = false,
     focusOnButton = false,
     formSubmitButton = false,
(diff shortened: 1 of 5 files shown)


src/components/FileGenerationButtons/PdfButton.tsx
```diff
@@ -21,7 +21,7 @@ const PdfButton = <Data, ErrorType extends string>({
     pdfComponent: PdfComponent,
     onFileCreationCompleted: onFileCreationCompleted = () => undefined,
     onFileCreationFailed: onFileCreationFailed,
-    formatName = true,
+    formatName = false,
```
AlexandraIrinaMardale: Can you leave the default as true? In this case you also don't need to set the formatName for DayOverview

mihaicutaiulia: 👍

mihaicutaiulia: done 👍

AlexandraIrinaMardale (commented):
If you generate multiple reports what do the names look like?

mihaicutaiulia:
> If you generate multiple reports what do the names look like?

currently, on windows:
image

are there any changes necessary?

AlexandraIrinaMardale (approved):
Looks good 😄

PR #725, opened 2025-07-17
VFB-132 Filter Icon

claudia-golaes-dev:
[VFB-132: Add an icon to indicate when filters are applied](https://softwiretech.atlassian.net/browse/VFB-https://softwiretech.atlassian.net/browse/VFB-132)

## What's changed
Changes apply to `Parcels`, `Clients` and `Lists` pages
- The filter icon in the top left corner of the tables changes from outlined to full when any primary filter is active
- The `Clear` button becomes green when any filter is active and the icon inside turns from outlined to full
- For the `Parcels` page: 
    - The `More/Less` button turns green if there are any additional fields active
    - The filter icon in front of the additional filters changes from outlined to full when any additional filter is active

### `Parcels`  Page
https://github.com/user-attachments/assets/60ec6c87-d349-46b5-a572-03900ce46ec4

### `Clients`  Page
https://github.com/user-attachments/assets/c126eb16-649e-4f0c-bfef-c938639d33f6

### `Lists`  Page
https://github.com/user-attachments/assets/a8910700-852b-4315-aa8a-fbb8d01b902c

## Checklist

### DIFF
--- a/src/components/Tables/TableFiltersBar.tsx
+++ b/src/components/Tables/TableFiltersBar.tsx
@@ -120,16 +120,16 @@ function TableFiltersBar<Data, Filter extends FilterBase<Data, State>, State>(
     }
 
     const hasActiveFilters = (filters: Filter[] | undefined): boolean => {
-        let activeFilters: Filter[] = [];
         if (filters) {
-            activeFilters = filters.filter(
+            return filters.some(
                 (filter) =>
                     !filter.isDisabled &&
                     !filter.areStatesIdentical(filter.state, filter.initialState) &&
                     filter.key !== "listType"
             );
+        } else {
+            return false;
         }
-        return activeFilters.length > 0;
     };
 
     const hasActiveAdditionalFilters: boolean = hasActiveFilters(props.additionalFilters);
--- a/src/components/Tables/TableFiltersBar.tsx
+++ b/src/components/Tables/TableFiltersBar.tsx
@@ -121,36 +121,46 @@ function TableFiltersBar<Data, Filter extends FilterBase<Data, State>, State>(
 
     const hasActiveAdditionalFillters = (): boolean => {
         let hasAdditionalFilters = false;
-
-        for (let i = 0; i < props.additionalFilters?.length; i++) {
-            if (
-                props.additionalFilters[i].state != props.additionalFilters[i].initialState &&
-                !props.additionalFilters[i].isDisabled
+        if (props.additionalFilters) {
+            for (
+                let indexFilter = 0;
+                indexFilter < props.additionalFilters?.length;
+                indexFilter++
             ) {
-                hasAdditionalFilters = true;
-                break;
+                if (
+                    props.additionalFilters[indexFilter].state !=
+                        props.additionalFilters[indexFilter].initialState &&
+                    !props.additionalFilters[indexFilter].isDisabled
+                ) {
+                    hasAdditionalFilters = true;
+                    break;
+                }
             }
         }
-
         return hasAdditionalFilters;
     };
 
     const hasActivePrimaryFilters = (): boolean => {
         let hasPrimaryFilters = false;
-
-        for (let i = 0; i < props.primaryFilters?.length; i++) {
-            if (props.primaryFilters[i].key === "view") {
-                continue;
-            }
-            if (
-                props.primaryFilters[i].state != props.primaryFilters[i].initialState &&
-                !props.primaryFilters[i].isDisabled
-            ) {
-                hasPrimaryFilters = true;
-                break;
+        if (props.primaryFilters) {
+            for (let indexFilter = 0; indexFilter < props.primaryFilters?.length; indexFilter++) {
+                if (
+                    props.primaryFilters[indexFilter].key === "view" ||
+                    props.primaryFilters[indexFilter].key === "packingDate" ||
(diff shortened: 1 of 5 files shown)


AlexandraIrinaMardale:
You added the link the the plan for 18-22 July instead of the ticket link :)))

src/components/Tables/TableFiltersBar.tsx
```diff
@@ -114,10 +119,59 @@ function TableFiltersBar<Data, Filter extends FilterBase<Data, State>, State>(
         return <></>;
     }
 
+    const hasActiveAdditionalFillters = (): boolean => {
```
AlexandraIrinaMardale: There's a typo in the word filters

claudia-golaes-dev:
> You added the link the the plan for 18-22 July instead of the ticket link :)))

we're sorry :(

src/components/Tables/TableFiltersBar.tsx
```diff
@@ -114,10 +119,59 @@ function TableFiltersBar<Data, Filter extends FilterBase<Data, State>, State>(
...
+    };
+
+    const hasActivePrimaryFilters = (): boolean => {
+        let hasPrimaryFilters = false;
+        if (props.primaryFilters) {
+            for (let indexFilter = 0; indexFilter < props.primaryFilters?.length; indexFilter++) {
+                if (
+                    props.primaryFilters[indexFilter].key === "view" ||
+                    props.primaryFilters[indexFilter].key === "packingDate" ||
+                    props.primaryFilters[indexFilter].key === "listType"
```
AlexandraIrinaMardale: Why are you specifically checking for listType? 🤔

src/components/Tables/TableFiltersBar.tsx
```diff
@@ -114,10 +119,59 @@ function TableFiltersBar<Data, Filter extends FilterBase<Data, State>, State>(
...
+                    !props.primaryFilters[indexFilter].isDisabled
+                ) {
+                    hasPrimaryFilters = true;
+                    break;
+                }
+            }
+        }
+        return hasPrimaryFilters;
+    };
+
```
AlexandraIrinaMardale: There's some similar logic in these 2 functions, could you maybe extract this into a helper function? And you could try to use some array methods like `some()` or `.filter()` to reduce the amount of code and simplify it?

Also I think we might already be checking if the filter list is undefined or empty in `hasPrimaryFilters` and `hasAdditionalFilters` 🤔

AlexandraIrinaMardale (changes requested):

src/components/Tables/TableFiltersBar.tsx
stefanghenescu-work: In the Lists Page, the Regular/Hotel buttons should not trigger the Clear button to turn green. This filter is named listType in the DB

src/components/Tables/TableFiltersBar.tsx
```diff
@@ -119,55 +119,27 @@ function TableFiltersBar<Data, Filter extends FilterBase<Data, State>, State>(
...
-                }
-            }
+    const hasActiveFilters = (filters: Filter[] | undefined): boolean => {
+        let activeFilters: Filter[] = [];
+        if (filters) {
+            activeFilters = filters.filter(
+                (filter) =>
+                    !filter.isDisabled &&
+                    !filter.areStatesIdentical(filter.state, filter.initialState) &&
+                    filter.key !== "listType"
```
claudia-golaes-dev: We still need the "listType" condition for filters, on the `Lists Page`. The "view" and "packingDate" conditions from `Parcels Page` can be removed after using the filter.areStatesIdentical() function.

src/components/Tables/TableFiltersBar.tsx
AlexandraIrinaMardale: Nice one, thank 😄

src/components/Tables/TableFiltersBar.tsx
```diff
@@ -119,55 +119,27 @@ function TableFiltersBar<Data, Filter extends FilterBase<Data, State>, State>(
...
+            );
         }
-        return hasAdditionalFilters;
+        return activeFilters.length > 0;
```
AlexandraIrinaMardale: One more small things 😄 Instead of using filter and then checking the length, you can use [`.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

AlexandraIrinaMardale (commented):
Just one more comment 😄

src/components/Tables/TableFiltersBar.tsx
```diff
@@ -114,10 +119,31 @@ function TableFiltersBar<Data, Filter extends FilterBase<Data, State>, State>(
         return <></>;
     }
 
+    const hasActiveFilters = (filters: Filter[] | undefined): boolean => {
+        if (filters) {
```
AlexandraIrinaMardale: One final change 🙃 
Can you first check 
`if (!filters) {
return false;
}`

And then the do the `filters.some(..)` outside any if else, please?

AlexandraIrinaMardale: Oh and you can leave the lint commented out:))

I think that's the only conflict you have

AlexandraIrinaMardale (approved):
👍

PR #732, opened 2025-07-21
VFB-419: Common Delivery Instructions

alexandrutrifu-urzica:
## What's changed
Client Delivery Instructions can now be edited from the _Edit Parcel_ Form.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image| image|

## Checklist


src/app/parcels/form/formSections/DeliveryInstructionsCard.tsx
```diff
@@ -0,0 +1,40 @@
...
+import FreeFormTextInput from "@/components/DataInput/FreeFormTextInput";
+import {
+    errorExists,
+    getDefaultTextValue,
+    getErrorText,
+    onChangeText,
+} from "@/components/Form/formFunctions";
+import GenericFormCard from "@/components/Form/GenericFormCard";
+
+export const MAX_CHARACTERS = 380;
```
LeonardCiobanu-Softwire: Why do you need a limit for the number of characters?

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -44,46 +45,67 @@ export const insertParcel: InsertParcel = async (parcelRecord) => {
...
+            const logId = await logWarningReturnLogId("Concurrent editing of parcel");
+            await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
+            return { parcelId: null, error: { type: "concurrentUpdateConflict", logId } };
+        }
 
-    await sendAuditLog({ ...auditLog, wasSuccess: true });
-    return { parcelId: primaryKey, error: null };
-};
+        const { error: updateDeliveryInstructionsError } = await updateClientDeliveryInstructions(
+            parcelRecord.client_id == undefined ? "" : parcelRecord.client_id,
```
LeonardCiobanu-Softwire: Why would the client_id be undefined? Maybe handle it differently than just send an empty string.

LeonardCiobanu-Softwire (dismissed):

src/app/parcels/form/formSections/DeliveryInstructionsCard.tsx
alexandrutrifu-urzica: The limit had already been imposed.

src/app/parcels/form/submitFormHelpers.ts
alexandrutrifu-urzica: This is the way it was originally defined.

src/app/clients/form/formSections/DeliveryInstructionsCard.tsx
AlexandraIrinaMardale, line 1: Can you revert the changes in this file please?

src/app/parcels/form/formSections/DeliveryInstructionsCard.tsx
AlexandraIrinaMardale, line 1: Can you find a way to not duplicate this code, please?

src/app/clients/form/formSections/DeliveryInstructionsCard.tsx
alexandrutrifu-urzica: I removed this client-specific card and kept the one in the _/common/formSections_ folder as a common component.
When rebasing the branch, I'll make sure to drop this initial commit as well.

src/app/parcels/form/formSections/DeliveryInstructionsCard.tsx
alexandrutrifu-urzica: The new common Card component uses a _CommonCardProps_ type, which seems to have solved the issue.

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -69,6 +70,7 @@ export interface ParcelFields extends Fields {
     collectionSlot: string | null;
     collectionCentre: string | null;
     lastUpdated: string | undefined;
+    deliveryInstructions: string | null | undefined;
```
LeonardCiobanu-Softwire: I believe type `string | null` would be enough.

LeonardCiobanu-Softwire (commented):
👍

src/app/parcels/form/ParcelForm.tsx
alexandrutrifu-urzica: We updated the type definition and added an extra _null check_ in the `ParcelForm` to remove the possibility of the field becoming undefined👍.

src/app/clients/form/ClientForm.tsx
```diff
@@ -76,7 +76,7 @@ export interface ClientFields extends Fields {
     babyOtherItems: BooleanGroup;
     petFood: BooleanGroup;
     otherItems: BooleanGroup;
-    deliveryInstructions: string;
+    deliveryInstructions: string | null | undefined;
```
mihaicutaiulia: can we apply the same changes here as we did for the parcel form so that `undefined` is no longer needed?

src/app/clients/form/ClientForm.tsx
```diff
@@ -64,7 +64,7 @@ export interface ClientFields extends Fields {
     numberOfAdults: number;
     children: Person[];
     numberOfChildren: number;
-    listType: ListType | null;
+    listType?: ListType | null;
```
mihaicutaiulia: in what scenarios do we need `listType?` instead of a required `listType` ?

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -17,46 +21,43 @@ import {
...
+import Modal from "@/components/Modal/Modal";
+import { Schema } from "@/databaseUtils";
 import supabase from "@/supabaseClient";
 import ListTypeCard from "./formSections/ListTypeCard";
 import ParcelNotesCard from "@/app/parcels/form/formSections/ParcelNotes";
 
 export interface ParcelFields extends Fields {
     clientId: string | null;
-    listType?: ListType;
+    listType?: ListType | null;
```
mihaicutaiulia: same here

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -285,21 +297,25 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
 
         const parcelRecord = {
             client_id: clientId || fields.clientId || "",
-            list_type: fields.listType,
+            list_type: fields.listType ?? undefined,
```
mihaicutaiulia, line 325: also here

src/app/clients/form/ClientForm.tsx
anamariasuditu: > can we apply the same changes here as we did for the parcel form so that `undefined` is no longer needed?

Yes, you are right, I applied the same changes as in the parcel form

src/app/parcels/form/ParcelForm.tsx
anamariasuditu: > also here

Because in `databaseTypeFIles`, list_type is declared like this: `list_type?: "regular" | "hotel" | undefined` so 
type "regular" | "hotel" | null is not assignable to type "regular" | "hotel" | undefined

src/app/parcels/form/ParcelForm.tsx
anamariasuditu: > same here

i modified it to requiered `listType`

src/app/clients/form/ClientForm.tsx
anamariasuditu: > in what scenarios do we need `listType?` instead of a required `listType` ?

i modified it to a required `listType`

src/app/parcels/form/submitFormHelpers.ts
AlexandraIrinaMardale: And what happens if we do end up having an undefined client id and we pass on the empty string?

src/app/clients/updateDeliveryInstructions.ts
```diff
@@ -0,0 +1,34 @@
...
+    const { error } = await supabase
+        .from("clients")
+        .update({ delivery_instructions: delivery_instructions })
+        .eq("primary_key", clientId);
+
+    if (error) {
+        const logId = await logErrorReturnLogId("update client delivery instructions failed", {
+            error,
+        });
+        await sendAuditLog({ ...baseAuditLogProps, wasSuccess: false, logId });
```
AlexandraIrinaMardale: I think you're sending the audit log twice? Here and in `submitFormHelper`

src/app/parcels/form/ParcelForm.tsx
AlexandraIrinaMardale: Why did you need to make any changes to the list type? 🤔

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -102,6 +109,7 @@ export const initialParcelFields: ParcelFields = {
     collectionSlot: null,
     collectionCentre: null,
     lastUpdated: undefined,
+    deliveryInstructions: null, // this should be the client's field
```
AlexandraIrinaMardale: Can you remove this comment please?

src/common/fetch.ts
```diff
@@ -72,6 +75,10 @@ export const fetchParcel = async (
                 name,
                 primary_key,
                 is_shown
+            ),
+            client: clients(
+                primary_key,
```
AlexandraIrinaMardale: Do you need the primary key here?

AlexandraIrinaMardale (commented):
Left some comments 😄

alexandrutrifu-urzica:
**FIXED**: Another thing I noticed was that the `Delivery Instructions Card` in the `Add Parcel Form` has not been synchronized with the client's fields - this means that the Parcel Form field is initially empty.

I will look into this and come back with a fix.

src/app/clients/updateDeliveryInstructions.ts
alexandrutrifu-urzica: Correct, I've deleted this call and only kept the one in the `submitFormHelper`, thank you!

src/app/parcels/form/submitFormHelpers.ts
alexandrutrifu-urzica: The `UpdateSchema` generated for the _parcels_ table automatically labels the `client_id` field as optional, hence the possibility of it being passed forward as `undefined`.

This `client_id` is used to identify the client whose instructions should be modified, so either it being null or an empty string would result in the same error, which is then logged below.

I've noticed the associated `parcelRecord` object had already defined the client_id field as `clientId || fields.clientId || ""`, with the empty string alternative included, so we decided to keep it this way. Again, the behaviour would remain the same if we opted for passing it on as null.

src/common/fetch.ts
alexandrutrifu-urzica: I removed it, I had forgotten it wasn't necessary for identifying the client object, thank you!

src/app/parcels/form/ParcelForm.tsx
alexandrutrifu-urzica: There were some compatibility issues at first, but I fixed everything on a new rebase - now `listType` should be required and initialized with a default value.

alexandrutrifu-urzica: **_Update_**: @AlexandraIrinaMardale I've since noticed that the additional null type is **indeed required** for our ticket. 

Currently, the `ClientFields` interface on `intern-dev` defines `listType` as `ListType | null`. To successfully refactor the `DeliveryInstructionsCard` to be common for both parcel and client forms, we needed an extra interface for the common `CardProps` it needed to receive. Therefore, the `ParcelFields` interface needed to be compatible with `ClientFields` - hence the refactoring of the `listType` parcel-field as possibly null as well.

src/app/parcels/form/submitFormHelpers.ts
AlexandraIrinaMardale: Ok, but when you pass the empty string, that results in a bad request. Can we check for empty before we send the request please?

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -40,50 +44,85 @@ export const insertParcel: InsertParcel = async (parcelRecord) => {
...
+        parcelRecord.client_id == undefined ? "" : parcelRecord.client_id,
+        deliveryInstructions
+    );
+
+    if (updateDeliveryInstructionsError) {
+        const logId = await logErrorReturnLogId(
+            "Error with update: parcel data",
+            updateDeliveryInstructionsError
+        );
+        await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
```
AlexandraIrinaMardale, line 47: You need to update the audit log to also include the delivery instructions

src/common/fetch.ts
```diff
@@ -72,6 +75,9 @@ export const fetchParcel = async (
                 name,
                 primary_key,
                 is_shown
+            ),
+            client: clients(
```
AlexandraIrinaMardale: Nitcpick: can this be `client:clients (` and same for packing slot

src/common/fetch.ts
```diff
@@ -12,6 +12,8 @@ type CollectionCentre = Pick<
 
 type PackingSlot = Pick<Schema["packing_slots"], "primary_key" | "is_shown" | "name">;
 
+type Client = Pick<Schema["clients"], "delivery_instructions">;
```
AlexandraIrinaMardale: If we are just picking the delivery instructions, the type should reflect that

alexandrutrifu-urzica: We renamed it to `ClientWithDeliveryInstructions`, following the naming convention for Parcel types (e.g. `ParcelWithCollectionCentreAndPackingSlot`).

src/app/parcels/form/submitFormHelpers.ts
alexandrutrifu-urzica: I added some checks to both the insert and update methods.

src/common/fetch.ts
```diff
@@ -68,10 +72,13 @@ export const fetchParcel = async (
                 primary_key,
                 is_shown
             ),
-            packing_slot: packing_slots (
+            packing_slot:packing_slots (
                 name,
                 primary_key,
                 is_shown
+            ),
+            client:clients(
```
AlexandraIrinaMardale: Can you add a space before the parenthesis? 😁

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -40,50 +44,110 @@ export const insertParcel: InsertParcel = async (parcelRecord) => {
...
+    deliveryInstructions: string | undefined
+) => Promise<UpdateParcelReturnType>;
 
-export const updateParcel: UpdateParcelWithPrimaryKey = (primaryKey) => async (parcelRecord) => {
-    const { error, count } = await supabase
-        .from("parcels")
-        .update(parcelRecord, { count: "exact" })
-        .eq("primary_key", primaryKey)
-        .eq("last_updated", parcelRecord.last_updated);
+export const updateParcel: UpdateParcelWithPrimaryKey =
```
AlexandraIrinaMardale, line 77: We need this update to be atomic, i.e when we update the parcel, we need to make sure that either the entire update succeeds or fails. Currently, it might update the parcel details but fail on the delivery instructions or the other way around.

alexandrutrifu-urzica: I added 2 new Supabase functions and now the parcel insert/update operations also handle the client's delivery instructions.

src/app/clients/form/ClientForm.tsx
```diff
@@ -78,7 +78,7 @@ export interface ClientFields extends Fields {
     babyOtherItems: BooleanGroup;
     petFood: BooleanGroup;
     otherItems: BooleanGroup;
-    deliveryInstructions: string;
+    deliveryInstructions: string | null;
```
DanielaNegrut, line 81: Why is this allowed a `null` value now?

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -32,29 +51,13 @@ import {
...
-import Modal from "@/components/Modal/Modal";
-import { Schema } from "@/databaseUtils";
 import supabase from "@/supabaseClient";
 import ListTypeCard from "./formSections/ListTypeCard";
+import ParcelNotesCard from "@/app/parcels/form/formSections/ParcelNotes";
 
 export interface ParcelFields extends Fields {
     clientId: string | null;
-    listType?: ListType;
+    listType: ListType | null;
```
DanielaNegrut, line 60: Why was this changed from `undefined` to `null`?

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -177,7 +191,7 @@ const databaseErrorMessageFromErrorType = (
 const ParcelForm: React.FC<ParcelFormProps> = ({
     initialFields,
     initialFormErrors,
-    clientId,
+    clientId = null,
```
DanielaNegrut: Why is this default null, is there any case where there is a parcel without a client id?

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -79,15 +83,20 @@ export interface ParcelErrors extends FormErrors<ParcelFields> {
...
+export type CommonFields = ParcelFields | ClientFields;
+export type CommonErrors = ParcelErrors | ClientErrors;
+
 export type ParcelCardProps = CardProps<ParcelFields, ParcelErrors>;
+export type CommonCardProps = CardProps<CommonFields, CommonErrors>;
 
 export const initialParcelFields: ParcelFields = {
     clientId: null,
-    listType: undefined,
+    listType: "regular",
```
DanielaNegrut, line 99: How come this has been initialized now and wasn't before?

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -1,89 +1,117 @@
...
         await sendAuditLog({ ...auditLog, wasSuccess: false, logId });
         return { parcelId: null, error: { type: "failedToInsertParcel", logId } };
     }
 
-    await sendAuditLog({ ...auditLog, wasSuccess: true, parcelId: data.primary_key });
-    return { parcelId: data.primary_key, error: null };
+    await sendAuditLog({
+        ...auditLog,
+        wasSuccess: true,
+        parcelId: parcelDataWithCount.parcel_primary_key || undefined,
```
DanielaNegrut: In what case will the primary key be undefined?

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -1,89 +1,117 @@
...
+    });
+    return { parcelId: parcelDataWithCount.parcel_primary_key, error: null };
 };
 
-type UpdateParcelErrorType = "failedToUpdateParcel" | "concurrentUpdateConflict";
+type UpdateParcelErrorType =
+    | "failedToUpdateParcel"
+    | "concurrentUpdateConflict"
+    | "failedToUpdateDeliveryInstructions";
```
DanielaNegrut: Let's order the error types so the ones starting with `failed to update ...` are next to each other.

src/common/fetch.ts
```diff
@@ -37,6 +39,7 @@ export interface ParcelWithCollectionCentreAndPackingSlot {
     primary_key: string;
     voucher_number: string | null;
     last_updated: string | undefined;
+    client: ClientWithDeliveryInstructions | null;
```
DanielaNegrut: The naming is a bit weird.

Having a whole field/property called `client` makes me think there is more information there, not just `delivery instructions`

src/databaseTypesFile.ts
```diff
@@ -982,10 +997,18 @@ export type Database = {
         time: string | null
         is_active: boolean | null
       }
+      insert_parcel_result: {
+        parcel_primary_key: string | null
+        rows_inserted: number | null
+      }
```
DanielaNegrut: Does it make sense to have such a small difference in the insert and update interfaces? Are there other situations where this is made?

supabase/migrations/20250812092433_add_update_parcel_with_delivery_instructions_function.sql
```diff
@@ -0,0 +1,45 @@
...
+
+CREATE OR REPLACE FUNCTION update_parcel_with_delivery_instructions(
+    parcel_record JSONB,
+    delivery_instructions TEXT,
+    parcel_primary_key UUID
+) RETURNS update_parcel_result AS
+$$
+DECLARE
+    rows_updated INTEGER;
+    client_id UUID := (parcel_record->>'client_id')::UUID;
```
DanielaNegrut: Aren't there cases where this can be null?

supabase/migrations/20250812101258_add_insert_parcel_with_delivery_instructions_function.sql
```diff
@@ -0,0 +1,56 @@
...
+    rows_inserted INTEGER
+    );
+
+CREATE OR REPLACE FUNCTION insert_parcel_with_delivery_instructions(
+    parcel_record JSONB,
+    delivery_instructions TEXT
+) RETURNS insert_parcel_result AS
+$$
+DECLARE
+rows_inserted INTEGER;
```
DanielaNegrut: This isn't aligned the same as the other lines in the variable definition of the declare block

supabase/migrations/20250812092433_add_update_parcel_with_delivery_instructions_function.sql
stefanghenescu-work, line 1: SQL syntax should be written in lower case. (e.g. `create type`).

supabase/migrations/20250812101258_add_insert_parcel_with_delivery_instructions_function.sql
stefanghenescu-work, line 1: Same for lower case here

supabase/migrations/20250812092433_add_update_parcel_with_delivery_instructions_function.sql
anamariasuditu: > SQL syntax should be written in lower case. (e.g. `create type`).

From what I've seen, most migrations are written in upper case.

src/app/parcels/form/ParcelForm.tsx
anamariasuditu: > How come this has been initialized now and wasn't before?

Because the `listType` can be `Regular `or `Hotel`, and it should be set on `Regular `by default in the form, unless the user changes it.

src/app/parcels/form/submitFormHelpers.ts
anamariasuditu: > Let's order the error types so the ones starting with `failed to update ...` are next to each other.

Yes, I updated it

src/app/parcels/form/ParcelForm.tsx
anamariasuditu: > Why was this changed from `undefined` to `null`?

Because we now changed the default value to `Regular`, so it's no longer undefined by default

src/app/clients/form/ClientForm.tsx
anamariasuditu: > Why is this allowed a `null` value now?

Because `deliveryInstructions` can be `null`, if another value is not provided

src/common/fetch.ts
anamariasuditu: > The naming is a bit weird.
> 
> Having a whole field/property called `client` makes me think there is more information there, not just `delivery instructions`

We considered it would be more logic to name it client and have `client.delivery_instructions` when we access the field, instead of naming it `deliveryInstructions `and later call it `deliveryInstructions.delivery_instructions`

supabase/migrations/20250812092433_add_update_parcel_with_delivery_instructions_function.sql
anamariasuditu: > Aren't there cases where this can be null?

The` delivery_instructions` field belongs to `client`, so the `client_id` can't be null if we try to pass the field to the `parcel`

src/app/parcels/form/ParcelForm.tsx
anamariasuditu: > Why is this default null, is there any case where there is a parcel without a client id?

You are right, there shouldn't be a parcel without a client id. I updated it, thank you!

supabase/migrations/20250812101258_add_insert_parcel_with_delivery_instructions_function.sql
anamariasuditu: > This isn't aligned the same as the other lines in the variable definition of the declare block

I updated it, thank you!

src/app/parcels/form/submitFormHelpers.ts
alexandrutrifu-urzica: There was an issue with the previous Supabase functions that resulted in the `primary key` being possibly null. I've since fixed this and the `undefined` type is no longer needed.

src/databaseTypesFile.ts
alexandrutrifu-urzica: After refactoring the migrations, these composite types should no longer be there.

src/common/fetch.ts
alexandrutrifu-urzica: I renamed it to `clientWithDeliveryInstructions` to match the type for now. If we decide to reference more fields from the `clients` table later on, we can switch back to the old name.

src/app/parcels/form/ParcelForm.tsx
DanielaNegrut: But it sill is null by default?

AlexandraIrinaMardale: @anamariasuditu what was the conclusion on this?

anamariasuditu: > @anamariasuditu what was the conclusion on this?

`ListType` is `Regular` by default, and the `listType` field can be either of `ListType` type, or `null`, making the field optional. The `| null` makes the field nullable, while the default value in `initialParcelFields `ensures it always has a sensible starting value.

supabase/migrations/20250812124033_add_update_parcel_with_delivery_instructions_function.sql
```diff
@@ -0,0 +1,41 @@
...
+        last_updated        = (parcel_record->>'last_updated')::timestamp,
+        list_type           = CAST(parcel_record->>'list_type' AS list_type),
+        notes               = parcel_record->>'notes',
+        packing_date        = (parcel_record->>'packing_date')::date,
+        packing_slot        = (parcel_record->>'packing_slot')::uuid,
+        referral_agency     = parcel_record->>'referral_agency',
+        referrer_email      = parcel_record->>'referrer_email',
+        referrer_name       = parcel_record->>'referrer_name',
+        referrer_phone      = parcel_record->>'referrer_phone',
+        voucher_number      = parcel_record->>'voucher_number'
```
AlexandraIrinaMardale: Can you double check that these are all the fields that need updating, i.e we haven't added anything else in parcels since this PR was added

supabase/migrations/20250812124058_add_insert_parcel_with_delivery_instructions_function.sql
```diff
@@ -0,0 +1,52 @@
...
+    collection_datetime,
+    list_type,
+    notes,
+    packing_date,
+    packing_slot,
+    referral_agency,
+    referrer_email,
+    referrer_name,
+    referrer_phone,
+    voucher_number
```
AlexandraIrinaMardale: Same here, please

supabase/migrations/20250812124033_add_update_parcel_with_delivery_instructions_function.sql
AlexandraIrinaMardale, line 1: Can you rename this to be after the last migration on intern-dev please?
Check with @mihaicutaiulia - she'll add another migration in a bit

AlexandraIrinaMardale (commented):
😄

AlexandraIrinaMardale (approved):

PR #739, opened 2025-07-22
VFB-404: Display Collection Centres' Availability

alexandrutrifu-urzica:
## What's changed
The _Collection Centre_ selector card now displays the centres' availability in the dropdown menu options.

This PR now contains all changes from tickets [VFB-405](https://softwiretech.atlassian.net/browse/VFB-405) and [VFB-403](https://softwiretech.atlassian.net/browse/VFB-403).

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image| image|

## Checklist


claudia-golaes-dev (approved):
looks nice!

src/app/parcels/form/formSections/CollectionCentreCard.tsx
```diff
@@ -1,21 +1,42 @@
...
+        if (!availableDaysObject) {
+            return "";
+        }
+        return availableDaysObject.available_days
+            ?.map((days) => (days.day == undefined ? "" : days.day))
+            .join(", ");
+    });
+
+    const collectionCentresLabelsAndValuesWithDays: CollectionCentresLabelsAndValues =
+        useMemo(() => {
```
LeonardCiobanu-Softwire: Why did you use useMemo here?

LeonardCiobanu-Softwire (commented):
👍

src/app/parcels/form/formSections/CollectionCentreCard.tsx
alexandrutrifu-urzica: `useMemo()` only updates the state of `collectionCentresLabelsAndValuesWithDays` once, when the CollectionCentreCard is first rendered, then saves this state until further modifications are applied to its referenced variables (`[collectionCentresLabelsAndValues, availableDaysNamesArray]`).

Without this, the available days would be appended to the initial labels over and over again, each time we closed and reopened the Card. Since the labels always remain the same while the user is filling in the form, the available days are only appended once, on the initial page render.

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -223,25 +222,20 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
-                    return;
+            const { data, error } = await getAvailableDaysForCollectionCentres(supabase);
+
+            if (error) {
+                let errorMessage;
+                switch (error.type) {
+                    case "collectionAvailableDaysFetchFailed":
+                        errorMessage = "Failed to fetch collection available days";
+                        break;
                 }
```
mihaicutaiulia: we might want to replace the switch here

src/app/parcels/form/formSections/CollectionCentreCard.tsx
```diff
@@ -1,21 +1,42 @@
...
     errorSetter,
     formErrors,
     fields,
     collectionCentresLabelsAndValues,
+    availableDays,
 }) => {
+    const availableDaysNamesArray = availableDays?.map((availableDaysObject) => {
+        if (!availableDaysObject) {
+            return "";
+        }
```
mihaicutaiulia, line 24: we could display a message indicating that no collection days are available

src/common/fetch.ts
```diff
@@ -234,13 +237,19 @@ export const getAvailableDaysForCollectionCentre = async (
...
 
-    if (!data.available_days) {
+    if (!data) {
         return { data: [], error: null };
     }
 
-    const availableDays: DbCollectionCentreAvailableDaysType = data.available_days.filter(
-        (availableDay) => availableDay.is_active
-    );
+    const availableDays = data.map((availableDaysObject) => {
```
mihaicutaiulia: perhaps we should rename the `availableDays` variable - it wasn't immediately clear that it contains the available days for all collection centres

src/app/parcels/form/formSections/CollectionCentreCard.tsx
mihaicutaiulia: we could display a message indicating that no collection days are available

src/common/fetch.ts
anamariasuditu: > perhaps we should rename the `availableDays` variable - it wasn't immediately clear that it contains the available days for all collection centres

Fixed it!

src/app/parcels/form/ParcelForm.tsx
anamariasuditu: > we might want to replace the switch here

done

src/app/parcels/form/formSections/CollectionCentreCard.tsx
anamariasuditu: > we could display a message indicating that no collection days are available

okay

src/app/parcels/form/formSections/CollectionCentreCard.tsx
```diff
@@ -1,21 +1,45 @@
...
     fieldSetter,
     errorSetter,
     formErrors,
     fields,
     collectionCentresLabelsAndValues,
+    availableDays,
 }) => {
+    const availableDaysNamesArray = availableDays?.map((availableDaysObject) => {
+        if (!availableDaysObject) {
+            return "No days available";
```
AlexandraIrinaMardale: Can we say something along the lines of "This centre is all days" or something that sounds a bit better? So that it's more clear to the user what this means

supabase/migrations/20250805101109_add_column_available_days_collection_centre_table.sql
AlexandraIrinaMardale, line 1: Can you rename this to `add_column_available_days_to_collection_centre_table`, please?

src/common/databaseDaysOfWeek.ts
```diff
@@ -0,0 +1,16 @@
...
+    "Monday",
+    "Tuesday",
+    "Wednesday",
+    "Thursday",
+    "Friday",
+    "Saturday",
+    "",
+] as const;
+
+export type DaysOfWeekLabelsAndValues = [string, string, string, string, string, string, string][];
```
AlexandraIrinaMardale: Is this used anywhere?

src/app/parcels/form/formSections/CollectionDateCard.tsx
```diff
@@ -1,17 +1,36 @@
...
 }) => {
+    const daysForSelectedCentre = availableDays.find(
+        (availableDaysObject) => availableDaysObject?.primary_key === fields.collectionCentre
+    );
+    const availableDaysIndices =
+        daysForSelectedCentre?.available_days === undefined
+            ? []
+            : getAvailableDaysIndices(daysForSelectedCentre?.available_days);
+
+    const isUnavailable = (availableDay: Dayjs): boolean => {
```
AlexandraIrinaMardale: Can you rename this to something along the lines of `isCentreClosedOnDay`?

src/app/parcels/form/formSections/CollectionDateCard.tsx
```diff
@@ -1,17 +1,36 @@
...
+        return !availableDaysIndices?.includes(dayjs(availableDay).day()) as boolean;
+    };
+
```
AlexandraIrinaMardale: This is overcomplicating the problem. You should be able to use the `isActive` field from the list of days

src/app/parcels/form/formSections/CollectionDateCard.tsx
```diff
@@ -1,17 +1,36 @@
...
+import { getAvailableDaysIndices } from "@/common/format";
+import { getErrorText, onChangeDate } from "@/components/Form/formFunctions";
 import { ErrorText } from "@/components/Form/formStyling";
-import dayjs from "dayjs";
+import GenericFormCard from "@/components/Form/GenericFormCard";
 import { ParcelCardProps } from "../ParcelForm";
 
-const CollectionDateCard: React.FC<ParcelCardProps> = ({
+interface DateCardProps extends ParcelCardProps {
+    availableDays: DbCollectionCentreType[];
```
AlexandraIrinaMardale: The name doesn't match the data type. Can we pass just the available days instead of this `DbCollectionCentreType` type?

src/app/parcels/form/formSections/CollectionCentreCard.tsx
```diff
@@ -1,21 +1,45 @@
...
+            .join(", ");
+    });
+
+    const collectionCentresLabelsAndValuesWithDays: CollectionCentresLabelsAndValues =
+        useMemo(() => {
+            return collectionCentresLabelsAndValues.map((centre, index) => {
+                const [label, value] = centre;
+                const daysString =
+                    availableDaysNamesArray?.[index] == ""
+                        ? "No days available"
```
AlexandraIrinaMardale: Same as before

src/app/parcels/form/formSections/CollectionCentreCard.tsx
AlexandraIrinaMardale: If there are no hidden dependencies, a `useMemo` should result in the same behaviour as not using it. It's just for optimising performance

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -237,6 +240,26 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
+
+                const errorMessage = errorMessages[error.type] || "An unexpected error occurred";
+                setSubmitErrorMessage(`${errorMessage}. Log ID: ${error.logId}`);
+                return;
+            }
+
+            setAvailableDays(data);
+        };
+
+        void getAvailableDays();
```
AlexandraIrinaMardale, line 325: What does this mean?

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -237,6 +240,26 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
         void getTimeSlots();
     }, [fields.collectionCentre, initialFormErrors]);
 
+    useEffect(() => {
+        const getAvailableDays = async (): Promise<void> => {
+            const { data, error } = await getAvailableDaysForCollectionCentres(supabase);
+
+            if (error) {
+                const errorMessages = {
+                    collectionAvailableDaysFetchFailed: "Failed to fetch collection available days",
```
AlexandraIrinaMardale: collection centre maybe?

src/app/admin/collectionCentresTable/CollectionCentresTable.tsx
```diff
@@ -275,6 +279,31 @@ const CollectionCentresTable: React.FC = () => {
...
+                    setAvailableDaysModalIsOpen(true);
+                };
+
+                return (
+                    <Button
+                        variant="outlined"
+                        size="small"
+                        onClick={handleEditCollectionCentreAvailableDays}
+                        disabled={params.row.isNew || params.row.isDelivery}
+                        aria-label={`Edit collection available days for ${params.row.name}`}
```
AlexandraIrinaMardale: Maybe `edit available collection` days?

AlexandraIrinaMardale (dismissed):

src/app/parcels/form/formSections/CollectionDateCard.tsx
alexandrutrifu-urzica: `DbCollectionCentreType` contains both a centre's primary key and its available days. This is needed to correctly append the right availability to each collection centre in the Card's dropdown menu.

We could rename this variable to `collectionCentreWithAvailableDays` to better match its type.

src/app/parcels/form/ParcelForm.tsx
alexandrutrifu-urzica: This is a regular call of the `getAvailableDays()` async method. The `void` keyword is there for the linter to know the returned `Promise<void>` should not be awaited or handled (the method returns nothing).

src/common/databaseDaysOfWeek.ts
alexandrutrifu-urzica: It was not, I removed it, thank you!

src/app/parcels/form/formSections/CollectionDateCard.tsx
alexandrutrifu-urzica: The collection centres' availability is initially filtered through the `isActive` field. This is because the current `available_days` database type contains a list of all possible`(day: boolean)` tuples, including the ones for which the centre is closed.

The callback above is not related to the `isActive` field - it is used to filter the _calendar days_ in the `Date Picker` as follows:
- the `dayjs(availableDay).day()` gets the week-index (0-6) of the calendar date
- we check if this index is included in the list of indices associated with the collection centre's own available days (the ones that we already know are active)

AlexandraIrinaMardale: Yes, I understand we're using it for the Date Picker, but there is a one line solution equivalent to your code

src/app/parcels/form/formSections/CollectionCentreCard.tsx
alexandrutrifu-urzica: We removed it, thank you!

src/common/databaseDaysOfWeek.ts
```diff
@@ -0,0 +1,14 @@
+import { Database } from "@/databaseTypesFile";
+
+export type DaysOfWeekType = Database["public"]["Enums"]["day_of_week"] | "";
+
+export const DAYSOFWEEK_ARRAY: DaysOfWeekType[] = [
```
AlexandraIrinaMardale: Is this used?

src/common/databaseDaysOfWeek.ts
```diff
@@ -0,0 +1,14 @@
+import { Database } from "@/databaseTypesFile";
+
+export type DaysOfWeekType = Database["public"]["Enums"]["day_of_week"] | "";
```
AlexandraIrinaMardale: Why do you need the empty string?

src/app/parcels/form/formSections/CollectionDateCard.tsx
```diff
@@ -1,17 +1,31 @@
...
     formErrors,
     fields,
+    availableDaysForSelectedCentre,
 }) => {
+    const isCentreClosedOnDay = (day: Dayjs): boolean => {
+        if (!availableDaysForSelectedCentre) {
+            return true;
+        }
+
+        return !availableDaysForSelectedCentre[dayjs(day).day()].is_active as boolean;
```
AlexandraIrinaMardale: The week for the date picker starts on Sunday so this will end up disabling the wrong days

And `is_active` is already a boolean, you don't need to cast it

alexandrutrifu-urzica: I refactored it and fixed the day order.

mihaicutaiulia:
@alexandrutrifu-urzica  When a collection centre slot is **no longer available**, it isn’t shown during parcel editing, but the form **can still be submitted** without selecting a new, valid time slot.

In contrast, when a **packing slot** is no longer available, a message appears (`The previous packing slot is no longer available, please select a new packing slot.`) and the user is required to choose a new one.

The same behavior should be applied to collection centre slots for consistency.

alexandrutrifu-urzica:
> @alexandrutrifu-urzica When a collection centre slot is **no longer available**, it isn’t shown during parcel editing, but the form **can still be submitted** without selecting a new, valid time slot.
> 
> In contrast, when a **packing slot** is no longer available, a message appears (`The previous packing slot is no longer available, please select a new packing slot.`) and the user is required to choose a new one.
> 
> The same behavior should be applied to collection centre slots for consistency.

I updated the components, thanks!

src/common/databaseDaysOfWeek.ts
alexandrutrifu-urzica: I updated the migration to make the `available_days` type not nullable - the empty string is now gone.

src/app/admin/collectionCentresTable/CollectionCentreActions.ts
```diff
@@ -89,6 +135,12 @@ const formatNewRowToDBCollectionCentre = (
         is_shown: newRow.isShown,
         is_delivery: newRow.isDelivery,
         time_slots: newRow.timeSlots,
+        available_days: initialCollectionAvailableDays.map((collectionAvailableDay) => {
+            return {
+                day: collectionAvailableDay.day,
+                is_active: collectionAvailableDay.isActive,
+            };
+        }),
```
DanielaNegrut: If this is the only place where we use `initialCollectionAvailableDays` why don't we keep it exactly in the needed format so we won't need to map it?

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+): FormattedAvailableDaysWithPrimaryKey => {
+    let formattedAvailableDays: FormattedAvailableDayType[];
+
+    if (row.availableDays === null || row.availableDays === undefined) {
+        formattedAvailableDays = [];
+    } else {
+        formattedAvailableDays = row.availableDays.map((availableDays) => {
+            return {
+                day: availableDays.day,
+                isActive: availableDays.is_active !== null ? availableDays.is_active : false,
```
DanielaNegrut: Isn't `availableDays.is_active ?? false` the same thing?

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+const formatCollectionCentreAvailableDaysDbData = (
+    row: CollectionCentresTableRow
+): FormattedAvailableDaysWithPrimaryKey => {
+    let formattedAvailableDays: FormattedAvailableDayType[];
+
+    if (row.availableDays === null || row.availableDays === undefined) {
+        formattedAvailableDays = [];
+    } else {
+        formattedAvailableDays = row.availableDays.map((availableDays) => {
```
DanielaNegrut: The name of the iterator variable is a bit confusing since it is plural

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+            return;
+        }
+        const payload: FormattedAvailableDaysWithPrimaryKey = {
+            ...availableDaysModalData,
+        };
+        props.onSave(payload);
+        props.onClose();
+    };
+
+    const toggleAvailableDaysInModalData = (dayLabel: string): void => {
```
DanielaNegrut: I think a better name would be `dayToToggle` isntead of `dayLabel`

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        availableDay.isActive = !availableDay.isActive;
+
+        const updatedAvailableDaysData: FormattedAvailableDaysWithPrimaryKey = {
+            ...availableDaysModalData,
+            availableDays: availableDaysModalData.availableDays,
+        };
+
+        setAvailableDaysModalData(updatedAvailableDaysData);
```
DanielaNegrut: We can move all of these into the setter for available days modal.

Also, we can defined the updated available days a bit better.

We can map the previous available days and if the day is the one to toglle then we flip `isActive` directly using `!`

src/app/admin/collectionCentresTable/CollectionCentreTimeSlotsModal.tsx
```diff
@@ -34,25 +40,6 @@ interface Props {
...
-const ColumnContainer = styled.div`
-    display: flex;
-    flex-direction: column;
-`;
-
-const RowContainer = styled.div`
-    display: flex;
-    flex-direction: row;
-`;
-
```
DanielaNegrut, line 55: Why where these removed?

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -235,25 +245,65 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
                 collectionCentre:
-                    fields.collectionCentre == null || fields.collectionCentre == deliveryPrimaryKey
-                        ? Errors.initial
-                        : Errors.none,
-                collectionDate: fields.collectionDate == null ? Errors.initial : Errors.none,
-                collectionSlot:
-                    fields.collectionSlot == null || fields.collectionSlot == "-"
-                        ? Errors.initial
-                        : Errors.none,
+                    fields.collectionCentre == deliveryPrimaryKey ? Errors.initial : Errors.none,
```
DanielaNegrut: Can you tell me a bit what this means?

src/app/parcels/form/formSections/CollectionCentreCard.tsx
```diff
@@ -1,21 +1,45 @@
...
+    collectionAvailableDays,
 }) => {
+    const availableDaysNamesArray = collectionAvailableDays?.map((availableDaysObject) => {
+        if (!availableDaysObject) {
+            return "This centre is closed all week";
+        }
+        return availableDaysObject.available_days
+            ?.filter((collectionDay) => collectionDay.is_active)
+            ?.map((collectionDay) => (collectionDay.day == undefined ? "" : collectionDay.day))
+            .join(", ");
```
DanielaNegrut: If we are mapping something to empty strings and then having a join, is there a posibility we can end up with a result like

`Monday, , , Sunday`

since the empty strings are still used in the join?

src/app/parcels/form/formSections/CollectionDateCard.tsx
```diff
@@ -1,17 +1,41 @@
...
     fields,
+    deliveryPrimaryKey,
+    availableDaysForSelectedCentre,
 }) => {
+    const isCentreClosedOnDay = (day: Dayjs): boolean => {
+        if (!availableDaysForSelectedCentre || !availableDaysForSelectedCentre.length) {
+            return true;
+        }
+
+        const dayIndex = day.day() != 0 ? day.day() - 1 : 6;
```
DanielaNegrut: We should always check for strict inequalities - `!== 0`

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -17,6 +24,35 @@ export type InsertParcelReturnType = {
...
+    const collectionDateDayIndex =
+        dayjs(fields.collectionDate).day() != 0 ? dayjs(fields.collectionDate).day() - 1 : 6;
+
+    return fields.collectionDate == null
+        ? Errors.initial
+        : !availableDaysForCentre ||
+            !availableDaysForCentre.length ||
+            availableDaysForCentre[collectionDateDayIndex].is_active
+          ? Errors.none
+          : Errors.invalidCollectionDate;
```
DanielaNegrut: Could you explain a bit what this does?

src/common/fetch.ts
```diff
@@ -166,6 +167,30 @@ type DbCollectionTimeSlotType = {
...
+    logId: string;
+};
+
+export type DbCollectionCentreAvailableDaysType = Schema["collection_centres"]["available_days"];
+
+export type DbCollectionCentreType = {
+    available_days: Schema["collection_centres"]["available_days"];
+    primary_key: Schema["collection_centres"]["primary_key"];
+} | null;
+
```
DanielaNegrut: A collection center has just available days and a primary key?

src/app/admin/collectionCentresTable/CollectionCentreActions.ts
alexandrutrifu-urzica: I switched the format and removed the map, thank you!

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
alexandrutrifu-urzica: I switched it to `availableDayObject`

src/app/admin/collectionCentresTable/CollectionCentreTimeSlotsModal.tsx
alexandrutrifu-urzica: I moved them to `admin/common/modalStyles.tsx`, since they are similar to the other modals' CSS components.

After extracting a generic Modal component for them to use, the style variables will need to be refactored as well.

src/app/parcels/form/ParcelForm.tsx
alexandrutrifu-urzica: This is a follow-up to the issue that was partially fixed in  [VFB-431](https://softwiretech.atlassian.net/browse/VFB-431).
- For `collectionCentre`
  - Currently, when adding a new parcel with its shipping method set to `Delivery`, its collection-related fields are not set to null or an empty string - they have default `Delivery` values assigned to them (the `collection_centres` Supabase table has a separate entry called `DLVR`, with a `deliveryPrimaryKey` that's common between all parcels set to `Delivery`).
  - Not having empty collection fields raises a couple of issues. Mainly, when editing a parcel and switching its shipping method from `Delivery` to `Collection`, although the `CollectionCentre` card displays an empty value, the Parcel Form can be submitted, assigning that specific `deliveryPrimaryKey` to the parcel object.
  - We considered refactoring the code to set the `collection` fields' values to null, but (as a temporary fix) decided to check for this `deliveryPrimaryKey` and switch the field's error type to `initial` whenever necessary, forcing the user to select a valid option.

- For `collectionDate`/`collectionSlot`:
  - When switching between collection centres, date & timeslots options that were previously picked don't always align with the new centre's schedule. In this case, the fields need to be emptied and an `invalid` error should be raised.

src/app/parcels/form/formSections/CollectionCentreCard.tsx
alexandrutrifu-urzica: I updated the `availableDays` type and undefined days are no longer a possibility, so I kept the `collectionDay.day` as is.

src/common/fetch.ts
alexandrutrifu-urzica: I renamed it to `DbCollectionCentreWithAvailableDaysType`, as we only need those two fields from it.

src/app/parcels/form/submitFormHelpers.ts
alexandrutrifu-urzica: This is used to decide which error is assigned to the `collectionDate` field.
- if the user has not selected a collection date, the field must be filled in and is assigned an `initial` error
- if the chosen centre has not had its availability set (_this state has an empty list as its initial value, so there are some problematic access attempts before the full page render_) **OR** 
- if the chosen centre is unavailable on the selected date, then the field is assigned an `invalid` error
- if the chosen centre is available on the selected date, no error is required

I've removed the null check since it isn't a possibility anymore.

cypress/e2e/admin/collectionCentres.cy.ts
```diff
@@ -1,31 +1,27 @@
 import { v4 as uuidv4 } from "uuid";
+import {
+    addNewCollectionCentre,
+    addNewTimeSlotInModal,
+    clickEditButtonForCentre,
+    saveAvailableDaysForCentre,
+    saveTimeSlotsForCentre,
+    startEditingCollectionCentreRow,
+    //tickAvailabilityCheckbox,
```
AlexandraIrinaMardale: Missed a comment here

supabase/migrations/20250812124009_add_column_available_days_to_collection_centre_table.sql
```diff
@@ -0,0 +1,6 @@
+create type "public"."day_of_week" as enum ('Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday');
+
+create type "public"."collection_availability_days" as ("day" day_of_week, "is_active" boolean);
+
+alter table "public"."collection_centres" add column "available_days" collection_availability_days[];
+alter table "public"."collection_centres" alter column "available_days" set not null;
```
AlexandraIrinaMardale: New line 😅 And can you please check that this is after all migrations already merged?

src/components/DataInput/DropDownSelect.tsx
```diff
@@ -22,6 +23,7 @@ interface ControlledProps<ValueType> {
     selectLabelId: string;
     focusOnDropdown?: boolean;
     error?: boolean;
+    disabled?: boolean;
 }
 
 interface UncontrolledProps<ValueType> {
```
AlexandraIrinaMardale, line 39: Would it be worth adding disabled to the uncontrolled props as well?

src/common/fetch.ts
```diff
@@ -202,6 +227,37 @@ export const getActiveTimeSlotsForCollectionCentre = async (
...
+        );
+        return { data: null, error: { type: "collectionAvailableDaysFetchFailed", logId: logId } };
+    }
+
+    if (!data) {
+        return { data: [], error: null };
+    }
+
+    const availableDaysForAllCollectionCentres = data.map((availableDaysObject) => {
+        if (availableDaysObject.available_days == null) {
```
AlexandraIrinaMardale: Can the available days actually be null?

src/common/databaseDaysOfWeek.ts
```diff
@@ -0,0 +1,3 @@
+import { Database } from "@/databaseTypesFile";
+
+export type DaysOfWeekType = Database["public"]["Enums"]["day_of_week"] | null;
```
AlexandraIrinaMardale: I think we don't need null here and if we remove it we'll need to update `formatCollectionCentreAvailableDaysDbData` - I'll leave some comments there as well

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -17,6 +21,33 @@ export type InsertParcelReturnType = {
...
+    const collectionDateDayIndex =
+        dayjs(fields.collectionDate).day() !== 0 ? dayjs(fields.collectionDate).day() - 1 : 6;
+
+    return fields.collectionDate == null
+        ? Errors.initial
+        : !availableDaysForCentre.length || availableDaysForCentre[collectionDateDayIndex].is_active
+          ? Errors.none
+          : Errors.invalidCollectionDate;
+}
+
```
AlexandraIrinaMardale: Do you think it might be easier to follow if we use a few if cases witch comments instead of the ternary operator?

src/app/parcels/form/submitFormHelpers.ts
```diff
@@ -17,6 +21,33 @@ export type InsertParcelReturnType = {
...
+): Errors {
+    return fields.collectionSlot == "-" || !fields.collectionSlot
+        ? Errors.initial
+        : collectionSlotsLabelsAndValues.some(
+                (slotLabelAndValue) => slotLabelAndValue[1] === fields.collectionSlot
+            )
+          ? Errors.none
+          : Errors.invalidCollectionSlot;
+}
+
```
AlexandraIrinaMardale: Same here, we could first check if no slot is selected yet, then if it is valid.

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -194,6 +201,10 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
     const [clientDetails, setClientDetails] = useState<ExpandedClientData | null>(null);
     const [collectionSlotsLabelsAndValues, setCollectionSlotsLabelsAndValues] =
         useState<CollectionTimeSlotsLabelsAndValues>([]);
+    const [collectionAvailableDays, setAvailableDays] = useState<
+        DbCollectionCentreWithAvailableDaysType[]
+    >([]);
+    const [availableDaysForCentre, setAvailableDaysForCentre] = useState<DbAvailableDaysType>([]);
```
AlexandraIrinaMardale, line 240: These seem very similar and they are a bit difficult to follow. Can you explain what they do please?

cypress/e2e/admin/commonActions/collectionCentres.tsx
```diff
@@ -0,0 +1,167 @@
...
+        .find('[data-field="name"]') // eslint-disable-line quotes
+        .find('input[type="text"]') // eslint-disable-line quotes
+        .type(newCollectionCentreName);
+
+    cy.get("@newRow")
+        .find('[data-field="acronym"]') // eslint-disable-line quotes
+        .find('input[type="text"]') // eslint-disable-line quotes
+        .type(newCollectionCentreName);
+
+    cy.intercept("POST", "**/rest/v1/collection_centres?select=*").as("saveCollectionCentre");
```
AlexandraIrinaMardale: Can you move all intercepts in the before section, please? Same as in [this](https://github.com/LambethFoodbank/foodbankapp/pull/800) PR

supabase/migrations/20250812124009_add_column_available_days_to_collection_centre_table.sql
```diff
@@ -0,0 +1,6 @@
+create type "public"."day_of_week" as enum ('Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday');
+
+create type "public"."collection_availability_days" as ("day" day_of_week, "is_active" boolean);
```
AlexandraIrinaMardale: Can you rename this to `collection_availability_day` - I think it's just meant to represent something like Monday-false right?

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+    row: CollectionCentresTableRow
+): FormattedAvailableDaysWithPrimaryKey => {
+    let formattedAvailableDays: FormattedAvailableDayType[];
+
+    if (row.availableDays === null || row.availableDays === undefined) {
+        formattedAvailableDays = [];
+    } else {
+        formattedAvailableDays = row.availableDays.map((availableDayObject) => {
+            return {
+                day: availableDayObject.day,
```
AlexandraIrinaMardale: If we make the change in the days of week type, we'll need a non-null assertion here.

Can you please confirm if my understanding of what can and cannot be null is correct?

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+    onClose: () => void;
+    onSave: (updated: FormattedAvailableDaysWithPrimaryKey) => void;
+}
+
+const formatCollectionCentreAvailableDaysDbData = (
+    row: CollectionCentresTableRow
+): FormattedAvailableDaysWithPrimaryKey => {
+    let formattedAvailableDays: FormattedAvailableDayType[];
+
+    if (row.availableDays === null || row.availableDays === undefined) {
```
AlexandraIrinaMardale: I don't think this can be null/undefined, right?

src/app/parcels/form/formSections/CollectionDateCard.tsx
```diff
@@ -25,6 +49,8 @@ const CollectionDateCard: React.FC<ParcelCardProps> = ({
                     }}
                     label="Date"
                     value={fields.collectionDate ? dayjs(fields.collectionDate) : null}
+                    shouldDisableDate={isCentreClosedOnDay}
+                    disabled={isDisabledFormInput}
```
AlexandraIrinaMardale, line 56: So in which case is this disabled?

src/app/parcels/form/formSections/CollectionSlotCard.tsx
```diff
@@ -30,6 +40,7 @@ const CollectionSlotCard: React.FC<CollectionSlotsCardProps> = ({
                     listTitle="Collection Slot"
                     value={fields.collectionSlot ?? ""}
                     onChange={valueOnChangeDropdownList(fieldSetter, errorSetter, "collectionSlot")}
+                    disabled={isDisabledFormInput}
```
AlexandraIrinaMardale, line 46: Same here?

supabase/migrations/20250812124009_add_column_available_days_to_collection_centre_table.sql
AlexandraIrinaMardale: I mean the timestamp:)

cypress/e2e/admin/commonActions/collectionCentres.tsx
```diff
@@ -0,0 +1,167 @@
...
+        .click();
+    cy.get('div[data-testid="CollectionCentreTimeSlotsModal"]').should("not.exist"); // eslint-disable-line quotes
+
+    cy.get('div[aria-label="Collection Centres Table"]') // eslint-disable-line quotes
+        .find(".MuiDataGrid-row--editing")
+        .find('[data-testid="SaveIcon"]') // eslint-disable-line quotes
+        .click();
+
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(5000);
```
AlexandraIrinaMardale: This wait is a bit too long - as the disable warning says we shouldn't have unnecessary waiting

cypress/e2e/admin/commonActions/collectionCentres.tsx
```diff
@@ -0,0 +1,167 @@
...
+        .click();
+    cy.get('div[data-testid="CollectionCentreAvailableDaysModal"]').should("not.exist"); // eslint-disable-line quotes
+
+    cy.get('div[aria-label="Collection Centres Table"]') // eslint-disable-line quotes
+        .find(".MuiDataGrid-row--editing")
+        .find('[data-testid="SaveIcon"]') // eslint-disable-line quotes
+        .click();
+
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(5000);
```
AlexandraIrinaMardale: Same here

cypress/e2e/admin/commonActions/collectionCentres.tsx
AlexandraIrinaMardale: Actually this should be removed

cypress/e2e/admin/collectionCentres.cy.ts
```diff
@@ -15,35 +27,12 @@ describe("Edit a collection centre on admins page", () => {
         );
 
         cy.visit("/admin");
+
+        newCollectionCentreName = `${uuidv4()}`;
     });
 
     it("Adds a collection centre", () => {
```
AlexandraIrinaMardale: Can you rename this to `Adds a collection centre and marks it as shown successfully`, please?

cypress/e2e/admin/commonActions/collectionCentres.tsx
```diff
@@ -0,0 +1,167 @@
...
+        .find(newCollectionCentreName)
+        .should("not.exist"); // If this fails then the random UUID is already there
+
+    startAddingNewCollectionCentre();
+    cy.get('div[aria-label="Collection Centres Table"]') // eslint-disable-line quotes
+        .find(".MuiDataGrid-row--editing", { timeout: 5000 })
+        .should("exist");
+
+    fillOutNewCollectionCentreRowAndSave(newCollectionCentreName);
+    cy.get('div[aria-label="Collection Centres Table"]') // eslint-disable-line quotes
```
AlexandraIrinaMardale, line 134: Can you keep the comment here please?

> // Wait for the data to reload and ensure the new item appears with proper data

cypress/e2e/admin/commonActions/collectionCentres.tsx
```diff
@@ -0,0 +1,167 @@
...
+//         const index = Math.floor(Math.random() * 6);
+//         return index >= checkboxIndex ? index + 1 : index;
+//     })();
+//
+//     cy.get("@availableDays").eq(checkboxIndex).find('input[type="checkbox"]').should("be.checked"); // eslint-disable-line quotes
+//     cy.get("@availableDays")
+//         .eq(otherCheckboxIndex)
+//         .find('input[type="checkbox"]') // eslint-disable-line quotes
+//         .should("not.be.checked"); // eslint-disable-line quotes
+// };
```
AlexandraIrinaMardale: What's up with this?

cypress/e2e/admin/commonActions/collectionCentres.tsx
```diff
@@ -0,0 +1,167 @@
...
+): void => {
+    cy.get('div[aria-label="Collection Centres Table"]') // eslint-disable-line quotes
+        .find('[aria-label="' + ariaLabel + " " + collectionCentreName + '"]', { timeout: 6000 }) // eslint-disable-line quotes
+        .click();
+    cy.get('div[data-testid="' + modalName + '"]') // eslint-disable-line quotes
+        .should("be.visible");
+};
+
+export const saveTimeSlotsForCentre = (collectionCentreName: string): void => {
+    void collectionCentreName;
```
AlexandraIrinaMardale: Can you remove the parameter if it's not being used please?

cypress/e2e/admin/commonActions/collectionCentres.tsx
```diff
@@ -0,0 +1,167 @@
...
+        .click();
+};
+
+export const clickEditButtonForCentre = (
+    ariaLabel: string,
+    collectionCentreName: string,
+    modalName: string
+): void => {
+    cy.get('div[aria-label="Collection Centres Table"]') // eslint-disable-line quotes
+        .find('[aria-label="' + ariaLabel + " " + collectionCentreName + '"]', { timeout: 6000 }) // eslint-disable-line quotes
```
AlexandraIrinaMardale: Can we use template literals please? There's quite a few places where this could be updated
 e.g `` .find(`[aria-label="${ariaLabel} ${collectionCentreName}"]`)``

AlexandraIrinaMardale (commented):
Left a few more comments on the tests

cypress/e2e/admin/collectionCentres.cy.ts
anamariasuditu: > Missed a comment here

Yes, I deleted it, thank you!

cypress/e2e/admin/collectionCentres.cy.ts
anamariasuditu: > Can you rename this to `Adds a collection centre and marks it as shown successfully`, please?

Yes, I renamed it

cypress/e2e/admin/commonActions/collectionCentres.tsx
anamariasuditu: > What's up with this?

I removed the comment

cypress/e2e/admin/commonActions/collectionCentres.tsx
anamariasuditu: > Can you keep the comment here please?
> 
> > // Wait for the data to reload and ensure the new item appears with proper data

Yes, I added it

cypress/e2e/admin/commonActions/collectionCentres.tsx
anamariasuditu: > This wait is a bit too long - as the disable warning says we shouldn't have unnecessary waiting

I decreased the waiting time to 500

cypress/e2e/admin/commonActions/collectionCentres.tsx
anamariasuditu: > Same here

I decreased the waiting time to 500

cypress/e2e/admin/commonActions/collectionCentres.tsx
anamariasuditu: > Can you remove the parameter if it's not being used please?

Yes, I removed it

cypress/e2e/admin/commonActions/collectionCentres.tsx
anamariasuditu: > .find(`[aria-label="${ariaLabel} ${collectionCentreName}"]`)

Yes, I modified it

src/components/DataInput/DropDownSelect.tsx
alexandrutrifu-urzica: From my understanding, the only difference between the two is that the `UncontrolledSelect` has a default value that doesn't depend on the current options/state of the app, whereas the `ControlledSelect` first displays the actual value of the selected element.

This is why the `ControlledSelect` would need to be disabled (if its current value is no longer available). Something similar couldn't really happen for the `UncontrolledSelect`, but I'll add the option just in case.

src/common/fetch.ts
alexandrutrifu-urzica: Not anymore, I removed the check, thank you!

cypress/e2e/admin/commonActions/collectionCentres.tsx
anamariasuditu: > Actually this should be removed

I removed it

src/app/parcels/form/ParcelForm.tsx
alexandrutrifu-urzica: `collectionAvailableDays` contains every centre's availability and is needed for the centre selector card, while `availableDaysForCentre` is set only for the selected centre and is used by the date/slot cards.

src/app/parcels/form/formSections/CollectionDateCard.tsx
alexandrutrifu-urzica: `shouldDisableDate` targets the selectable days in the calendar.
`disabled` refers to the entire Date Card, which requires the user to select a collection centre first.

src/app/parcels/form/formSections/CollectionSlotCard.tsx
alexandrutrifu-urzica: The `Timeslots Card` is disabled until the user selects a collection centre.

supabase/migrations/20250812124009_add_column_available_days_to_collection_centre_table.sql
anamariasuditu: > Can you rename this to `collection_availability_day` - I think it's just meant to represent something like Monday-false right?

Yes, I renamed it

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
anamariasuditu: > If we make the change in the days of week type, we'll need a non-null assertion here.
> 
> Can you please confirm if my understanding of what can and cannot be null is correct?

The `availableDays `array contains objects with a `day` property that can be `null` in Supabase.
The `FormattedAvailableDayType `interface expects day to be a non-null day of the week, since we removed ` | null` from `DaysOfWeekType`, so we added this filter:
`.filter((dayObj) => dayObj.day !== null)`

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
anamariasuditu: > I don't think this can be null/undefined, right?

Yes, only `day `can be null, not `availableDays `. It also can't be undefined. The array `availableDays ` can contain null values, but not be null itself.

anamariasuditu: > > I don't think this can be null/undefined, right?
> 
> Yes, only `day `can be null, not `availableDays `. It also can't be undefined. The array `availableDays ` can contain null values, but not be null itself.

cypress/e2e/admin/commonActions/collectionCentres.tsx
AlexandraIrinaMardale: Do we need to wait here though?

src/app/parcels/form/formSections/CollectionSlotCard.tsx
AlexandraIrinaMardale: Great, thanks 😄

src/components/DataInput/DropDownSelect.tsx
AlexandraIrinaMardale: Ah that makes sense, let's leave it as is then. Thank you

src/app/parcels/form/formSections/CollectionCentreCard.tsx
```diff
@@ -1,21 +1,67 @@
...
+                availabilityString: formattedAvailabilityString,
+            };
+        });
+
+    const collectionCentresLabelsAndValuesWithDays: CollectionCentresLabelsAndValues =
+        collectionCentresLabelsAndValues.map((centre) => {
+            const [label, value] = centre;
+            const availabilityString =
+                collectionCentresWithAvailabilityStrings.find(
+                    (availabilityObject) => availabilityObject?.primary_key === value
```
alexandrutrifu-urzica, line 61: I updated this not to use array indices anymore. Centres have their availability matched by primary keys now.

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -288,12 +297,18 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
             setFormErrors((prevErrors) => ({
                 ...prevErrors,
-                collectionCentre:
-                    fields.collectionCentre == deliveryPrimaryKey ? Errors.initial : Errors.none,
-                collectionDate: switchErrorForCollectionDate(fields, availableDaysForCentre),
+                collectionCentre: switchErrorForCollectionCentre(
+                    fields,
+                    collectionCentreIsActive,
+                    deliveryPrimaryKey
+                ),
```
alexandrutrifu-urzica, line 337: Update: If a parcel's selected collection centre is hidden, this now triggers an error for all other collection cards and disables their input boxes.

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -235,25 +255,71 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
             }));
         }
     }, [
+        availableDaysForCentre,
+        collectionSlotsLabelsAndValues,
         deliveryPrimaryKey,
+        fields,
         fields.collectionCentre,
         fields.collectionDate,
         fields.collectionSlot,
```
AlexandraIrinaMardale: There's a missing dependency here 👀 - is it intentional?

supabase/migrations/20250823124009_add_column_available_days_to_collection_centre_table.sql
AlexandraIrinaMardale, line 1: Don't forget to update the name 😄

AlexandraIrinaMardale (commented):
There's some conflicts with intern-dev 😢

src/app/parcels/form/ParcelForm.tsx
alexandrutrifu-urzica: It wasn't, I made sure to add it, thank you!

src/app/parcels/form/ParcelForm.tsx
```diff
@@ -249,25 +266,74 @@ const ParcelForm: React.FC<ParcelFormProps> = ({
...
+        };
+
+        void getAvailableDays();
+    }, []);
+
     useEffect(() => {
         // If the Shipping Method changes, errors for collection date and slot should be reset
         if (fields.shippingMethod == "Collection") {
-            // The collection centre fields initially have 'Delivery' default values
+            console.log(availableDaysForCentre);
```
AlexandraIrinaMardale: Missed a console log here 👀

supabase/migrations/20250901124009_add_column_available_days_to_collection_centre_table.sql
```diff
@@ -0,0 +1,6 @@
+create type "public"."day_of_week" as enum ('Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday');
+
+create type "public"."collection_availability_day" as ("day" day_of_week, "is_active" boolean);
+
+alter table "public"."collection_centres" add column "available_days" collection_availability_day[];
+alter table "public"."collection_centres" alter column "available_days" set not null;
```
AlexandraIrinaMardale: We should either give it a default value or make sure that we backfill the existing data

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    <Centerer>
+                        <Heading>{props.selectedCollectionCentreInfo?.name}</Heading>
+                    </Centerer>
+                    <Centerer>
+                        <ModalAvailableDaysContainer aria-label="List of defined available days">
+                            <FormGroup>
+                                {availableDaysModalData &&
+                                    availableDaysModalData.availableDays.map((availableDay) => {
+                                        return (
+                                            <>
```
AlexandraIrinaMardale: I don't think we need to wrap this 🤔

src/app/admin/collectionCentresTable/CollectionCentreAvailableDaysModal.tsx
```diff
@@ -0,0 +1,169 @@
...
+                                                {availableDay.day != null && (
+                                                    <ModalAvailableDaysRow key={availableDay.day}>
+                                                        <SpaceBetween>
+                                                            <CheckboxInput
+                                                                label={availableDay.day}
+                                                                checked={availableDay.isActive}
+                                                                onChange={() =>
+                                                                    !!availableDay.day &&
```
AlexandraIrinaMardale: We already check that day is not null so this is not necessary

src/app/admin/common/modalStyles.tsx
AlexandraIrinaMardale, line 1: The time slots/available days containers and rows have the same style, can we keep one of each and have a generic name?

supabase/migrations/20250901124009_add_column_available_days_to_collection_centre_table.sql
alexandrutrifu-urzica: I updated this so every centre is closed on all days by default.

AlexandraIrinaMardale: I think it might be better to go with open all days by default 😅

src/app/admin/collectionCentresTable/CollectionCentreActions.ts
```diff
@@ -44,6 +58,37 @@ type FetchCollectionCentresResult =
           error: { type: "failedToFetchCollectionCentres"; logId: string };
       };
 
+export const initialCollectionAvailableDays: DbAvailableDaysType = [
```
AlexandraIrinaMardale, line 61: And we should probably go with all true here as well

src/app/admin/common/modalStyles.tsx
AlexandraIrinaMardale, line 1: Nice 😄

src/app/parcels/form/submitFormHelpers.ts
AlexandraIrinaMardale, line 1: Great 🥳

supabase/migrations/20250901124009_add_column_available_days_to_collection_centre_table.sql
AlexandraIrinaMardale: Can you set not null and default at the same time please?

AlexandraIrinaMardale (approved):

PR #721, opened 2025-07-16
VFB-396: Referral fields in Parcel Form

mihaicutaiulia:
[VFB-396:](https://softwiretech.atlassian.net/jira/software/c/projects/VFB/boards/1130?selectedIssue=VFB-396) Parcel Form - add Referral Agency Details

## What's changed
Parcel details now include information about their referrer & referral agency.
Empty referral fields will not be shown in the Parcel Details Card View.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image|
| image | image|

## Checklist
- [x] The ticket is up-to-date - Please document any deviations from the original approach if there is any.
- [x] I have documented the testing steps for QA
- [x] I have self-reviewed this PR
- [x] Make sure you've verified it works via `npm run dev`
- [x] Make sure you've verified it works via `npm run build` and `npm run start`
- [x] Make sure you've fixed all linting problems with `npm run lint_fix`
- [x] Make sure you've tested via `npm run test`

- [x] The migration files are up-to-date with my final set up (`npx supabase db diff -f <name_of_migration>` should create nothing at this point)
- [x] I have updated the typescript definitions for the database with `npm run generate_types:local`
- [x] I have modified the seed in `supabase/seed/seed.ts` if appropriate
- [x] If I have modified the seed, I have also generated the seed with `npm run db:generate_seed` 
- [x] With my final set up, I can run `npm run reset_supabase` without any errors.
- [ ] (Just before merging the ticket) I have updated the timestamps of the new migration files so that they are after all existing migration files.
- [x] I have taken reasonable measures to ensure constraint violations do not happen when the migration occurs, e.g.:
  - audit_log FK constraints that could be violated
  - any constraints that I have added, e.g. `set not null`
- [x] I have checked that migration can happen successfully without resetting the database, doing the following.
  - Make sure you have rebased your branch onto dev or merged dev into your branch.
  - Checkout the dev branch
  - Run `npm run post_checkout` to reset the database, including the seed data.
  - Checkout your branch
  - Run `npx supabase migrations up --include-all --local` to run all outstanding migration files
  - Check that the resulting database is what you expect, bar any seed data changes.

PR #747, opened 2025-07-23
VFB-396: Referral fields

alexandrutrifu-urzica:
[VFB-396:](https://softwiretech.atlassian.net/jira/software/c/projects/VFB/boards/1130?selectedIssue=VFB-396) Parcel Form - add Referral Agency Details

## What's changed
Parcel details now include information about their referrer & referral agency.
Empty referral fields will not be shown in the Parcel Details Card View.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image|
| image | image|

## Checklist

  - Checkout the dev branch
  - Checkout your branch


alexandrutrifu-urzica:
This is an updated, fixed version of the original ticket PR [(here)](https://github.com/LambethFoodbank/foodbankapp/pull/721)

src/app/parcels/ActionBar/ActionAndStatusBar.test.tsx
```diff
@@ -67,6 +67,11 @@ export const mockData: ParcelsTableRow[] = [
             flaggedForAttention: false,
         },
         voucherNumber: "123456789",
+        listType: "regular",
```
AlexandraIrinaMardale, line 70: Why does list type still appear as changed?

src/app/parcels/edit/EditParcelForm.tsx
```diff
@@ -37,7 +37,7 @@ const prepareParcelDataForForm = (
     return {
         clientId: parcelData.client_id,
         listType: parcelData.list_type,
-        voucherNumber: parcelData.voucher_number,
+        voucherNumber: parcelData.voucher_number ?? "",
```
AlexandraIrinaMardale, line 40: This too

src/components/Form/formFunctions.ts
```diff
@@ -68,6 +68,7 @@ const getErrorType = (
     additionalCondition?: (value: string) => boolean,
     maxCharacters?: number
 ): Errors => {
+    console.log(input);
```
AlexandraIrinaMardale: Can you remove this please?

supabase/migrations/20250716085151_add_referral_columns_to_parcels.sql
```diff
@@ -0,0 +1,9 @@
+alter table "public"."parcels" add column "referral_agency" text;
+
+alter table "public"."parcels" add column "referrer_email" text;
+
+alter table "public"."parcels" add column "referrer_name" text;
+
+alter table "public"."parcels" add column "referrer_phone" text;
+
+
```
AlexandraIrinaMardale: some extra lines here

supabase/migrations/20250724110901_change-types-of-referral-fields.sql
```diff
@@ -0,0 +1,11 @@
...
+
+alter table "public"."parcels" alter column "referrer_email" set default ''::text;
+
+alter table "public"."parcels" alter column "referrer_name" set default ''::text;
+
+alter table "public"."parcels" alter column "referrer_phone" set default ''::text;
+
+alter table "public"."parcels" alter column "voucher_number" set default ''::text;
+
+
```
AlexandraIrinaMardale: same here

supabase/seed/.snaplet/dataModel.json
```diff
@@ -4427,7 +4427,7 @@
           "isList": false,
           "isGenerated": false,
           "sequence": false,
-          "hasDefaultValue": false,
+          "hasDefaultValue": true,
```
AlexandraIrinaMardale, line 4430: Why did you change this?

src/app/parcels/edit/EditParcelForm.tsx
mihaicutaiulia, line 40: We changed this to fix another error related to this field

src/components/DataViewer/DataViewer.tsx
```diff
@@ -123,7 +123,13 @@ const DataViewer: React.FC<DataViewerProps> = ({ data }) => {
...
         <DataViewerContainer>
             {Object.entries(data).map(([key, value]) => {
-                if (!value.hide) {
+                const isAHiddenReferralField = (key: string, value: ValueConfig): boolean => {
+                    return (
+                        (key.includes("referral") || key.includes("referrer")) && value.value === ""
+                    );
+                };
+
+                if (!value.hide && !isAHiddenReferralField(key, value)) {
```
AlexandraIrinaMardale: Can you move the check for displaying the referral fields to `getExpandedParcelDetails.ts`, please?

src/common/format.ts
```diff
@@ -18,6 +18,7 @@ export const displayNameForNullDriverName = "Unknown Driver";
 export const phoneNumberFormatSymbolsRegex = /[\s-()]/g;
 // Following characters excluded from regex as are removed before checking format matches: ( ) - \s
 export const phoneNumberRegex = /^((0|\+44)\d{9,11}|\+(?!44)\d{7,15})?$/;
+export const emailRegex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/;
```
AlexandraIrinaMardale: I'm not sure this regex is correct, can we stick with the more lenient one in `AccountDetailsCard.tsx`?

AlexandraIrinaMardale (changes requested):

src/app/parcels/edit/EditParcelForm.tsx
AlexandraIrinaMardale: Which error?

alexandrutrifu-urzica: We noticed that most text fields that had input boxes had the same Supabase definition and an empty string default value - the `voucher number` was an exception, and this interfered with how the input box treated it. More precisely, the user couldn't delete their input completely, therefore triggering an error.

We added a new migration file to change the `voucher number`'s field definition and modified the input box props accordingly. Now, instead of the input being null, it's seen as an empty string and doesn't trigger an error anymore.

supabase/seed/.snaplet/dataModel.json
alexandrutrifu-urzica: > We noticed that most text fields that had input boxes had the same Supabase definition and an empty string default value - the `voucher number` was an exception, and this interfered with how the input box treated it. More precisely, the user couldn't delete their input completely, therefore triggering an error.
> 
> We added a new migration file to change the `voucher number`'s field definition and modified the input box props accordingly. Now, instead of the input being null, it's seen as an empty string and doesn't trigger an error anymore.

Again, most input fields were defined as such, so we thought keeping the format uniform would be a good idea🤞

src/common/format.ts
mihaicutaiulia: i made the change, and while testing with different inputs, i noticed that the updated version validates string like `a@a`, whereas the previous version only accepted formats like `a@a.aa`. 

let me know what version would you like us to keep

AlexandraIrinaMardale: Yes, you're right, this version is more lenient and accepts strings that are not actually correct email addresses. But the other one is a bit too restrictive (and I don't think it fully conforms to the standard). If we go with the stricter one than we might end up not accepting valid email addresses
But really good spot, thank you! 😄

src/app/parcels/edit/EditParcelForm.tsx
AlexandraIrinaMardale: That makes sense, thank you 😄

AlexandraIrinaMardale (approved):
Looks good 😄 

Can you please double check that there are no new migrations on intern-dev since your last rebase before merging?

PR #752, opened 2025-07-28
Vfb 425 concurrent edits admin packing slots

anamariasuditu:
## What's changed
Concurrent edits in the admin page for updating packing slots are now blocked.

### Before 

https://github.com/user-attachments/assets/eda13f7e-2bbf-4f57-9228-e153bb5d4c76

### After 

https://github.com/user-attachments/assets/b948c2c6-df07-4e7c-be60-8cdcc689cbb8

 

## Checklist

    - Checkout the dev branch
    - Checkout your branch


alexandrutrifu-urzica (approved):

src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
```diff
@@ -156,56 +163,82 @@ const PackingSlotsTable: React.FC = () => {
...
+                } else {
+                    void sendAuditLog({
+                        ...baseAuditLog,
+                        wasSuccess: true,
+                    });
+                }
+            }
+
+            return { ...newRow, isNew: false };
+        } finally {
```
LeonardCiobanu-Softwire, line 239: You should also have the `catch` part of this `try`.

LeonardCiobanu-Softwire (commented):
👍

src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
```diff
@@ -46,6 +46,8 @@ export interface PackingSlotRow {
     isShown: boolean;
     order: number;
     isNew: boolean;
+    lastUpdated: string;
+    originalLastUpdated: string;
```
mihaicutaiulia, line 50: could you explain the difference between the variables `lastUpdated` and `originalLastUpdated`? what is the role of each one?

mihaicutaiulia (approved):

src/app/admin/packingSlotsTable/PackingSlotActions.ts
```diff
@@ -22,6 +23,8 @@ export const fetchPackingSlots = async (): Promise<PackingSlotRow[]> => {
             isShown: row.is_shown,
             order: row.order,
             isNew: false,
+            lastUpdated: row.last_updated,
+            originalLastUpdated: row.last_updated,
         })
```
claudia-golaes-dev, line 27: same here with the fields' name, if you can change them

src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
anamariasuditu: > could you explain the difference between the variables `lastUpdated` and `originalLastUpdated`? what is the role of each one?

When a concurrent edit occurs, our goal is to check if the timestamps of the `last_updated` field in Supabase is different from the one saved in `originalLastUpdated`, and if so, it means a new edit took place and the concurrency appears. If they are the same, we can edit our fields.
Since the value of the field `last_updated` in the database changes automatically when the first user completes the edit, we need to store in a new variable the initial timestamp of the `last_updated` when we first opened the edit form, in order to later compare it to the new one in the database. If we don't to that, the last_updated field will always compare with itself and will allow concurrent edits.

src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
anamariasuditu: > You should also have the `catch` part of this `try`.

Yes, I added it

src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
```diff
@@ -433,7 +472,20 @@ const PackingSlotsTable: React.FC = () => {
                     columns={packingSlotsColumns}
                     editMode="row"
                     rowModesModel={rowModesModel}
-                    onRowModesModelChange={setRowModesModel}
+                    onProcessRowUpdateError={(error) => {
+                        console.error("Error while updating row:", error);
```
AlexandraIrinaMardale: We don't want to log the error in the console, in which case we also don't really want to throw the error on line 240

supabase/migrations/20250725112801_add_last_updated_packing_slots.sql
AlexandraIrinaMardale, line 1: Please make sure to change the timestamps before merging

src/app/admin/packingSlotsTable/PackingSlotActions.ts
```diff
@@ -88,18 +93,37 @@ export const updateDbPackingSlot = async (
...
+        await sendAuditLog({ ...baseAuditLogProps, wasSuccess: false, logId });
+        return { error: { type: "ConcurrentEdit", logId } };
+    } else if (error) {
         const logId = await logErrorReturnLogId("Failed to update packing slot", {
             error,
             newPackingSlotData: processedData,
         });
 
-        return { error: { dbError: error, logId } };
+        return { error: { type: "UpdatePackingSlotsFailed", logId } };
```
AlexandraIrinaMardale: Can you keep this part consistent with VFB-424?

src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
```diff
@@ -156,56 +163,84 @@ const PackingSlotsTable: React.FC = () => {
...
-            }
-        }
 
-        setIsLoading(false);
+                if (updatePackingSlotError) {
+                    let message = `Failed to update the packing slot. Log ID: ${updatePackingSlotError.logId}`;
+
+                    if (updatePackingSlotError.type === "ConcurrentEdit") {
+                        message =
+                            "This packing slot was modified by someone else.\n" +
```
AlexandraIrinaMardale: Can you make sure to use the same messages as the other concurrent edit tickets? And can you double check this for VFB-424 as well?

src/app/admin/packingSlotsTable/PackingSlotActions.ts
```diff
@@ -88,18 +93,37 @@ export const updateDbPackingSlot = async (
...
+            },
+            { count: "exact" }
+        )
+        .eq("primary_key", processedData.primary_key)
+        .eq("last_updated", lastUpdated)
+        .select();
+
+    if (count === 0) {
+        const logId = await logWarningReturnLogId("Concurrent editing of packing slots");
+        await sendAuditLog({ ...baseAuditLogProps, wasSuccess: false, logId });
```
AlexandraIrinaMardale: I think you're adding a log for the same thing twice

src/app/admin/packingSlotsTable/PackingSlotActions.ts
anamariasuditu: > same here with the fields' name, if you can change them

This is the name used in all the others concurrency tickets to ensure consistency.

src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
anamariasuditu: > Can you make sure to use the same messages as the other concurrent edit tickets? And can you double check this for VFB-424 as well?

Yes, I changed it to be consistent

src/app/admin/packingSlotsTable/PackingSlotActions.ts
anamariasuditu: > Can you keep this part consistent with VFB-424?

Yes, I updated it

src/app/admin/packingSlotsTable/PackingSlotActions.ts
anamariasuditu: > I think you're adding a log for the same thing twice

I removed it, thank you!

src/app/admin/packingSlotsTable/PackingSlotsTable.tsx
anamariasuditu: > We don't want to log the error in the console, in which case we also don't really want to throw the error on line 240

I changed it to `setErrorMessage(error.message);`

AlexandraIrinaMardale (commented):
You'll need to update the migration names one more time, but make sure it's dated 20250812-something later than the previous one, don't increase the day, please

AlexandraIrinaMardale (approved):

PR #753, opened 2025-07-29
VFB-442: Pending More Info Report Undefined Query Result

alexandrutrifu-urzica:
## What's changed
[VFB-442](https://softwiretech.atlassian.net/jira/software/c/projects/VFB/boards/1130?selectedIssue=VFB-442): Trying to download a Pending More Info Report for an interval with no available parcels now displays an error.

(Bug fix for [VFB-400](https://softwiretech.atlassian.net/browse/VFB-400))

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image| image|

## Checklist

### DIFF
--- a/src/app/parcels/ActionBar/ActionButtons/PendingMoreInfoReportCsvButton.tsx
+++ b/src/app/parcels/ActionBar/ActionButtons/PendingMoreInfoReportCsvButton.tsx
@@ -177,10 +177,7 @@ const getPendingMoreInfoReportData = async (
 
     if (!rawParcelList || rawParcelList.length === 0) {
         const logId = await logErrorReturnLogId(
-            "No parcels with specified status to create Pending More Info report",
-            {
-                error: parcelFetchError,
-            }
+            "No parcels with specified status to create Pending More Info report"
         );
         return {
             data: null,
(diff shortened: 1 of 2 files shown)


src/app/parcels/ActionBar/ActionButtons/PendingMoreInfoReportCsvButton.tsx
```diff
@@ -175,6 +175,22 @@ const getPendingMoreInfoReportData = async (
         };
     }
 
+    if (!rawParcelList || rawParcelList.length === 0) {
+        const logId = await logErrorReturnLogId(
+            "No parcels with specified status to create Pending More Info report",
+            {
+                error: parcelFetchError,
```
mihaicutaiulia: as I understand it, when `parcelFetchError` is not null, the if statement above will be triggered. In that case, isn’t `parcelFetchError` here always null?

anamariasuditu (approved):

src/app/parcels/ActionBar/ActionButtons/PendingMoreInfoReportCsvButton.tsx
alexandrutrifu-urzica: I removed it as it was only useful for returning database errors - since this statement only handles the case of an empty result, an error message should be enough.

LeonardCiobanu-Softwire (commented):
👍

claudia-golaes-dev (approved):

AlexandraIrinaMardale (approved):

AlexandraIrinaMardale (commented):
Just want to double check that we only send one log per error, right?

alexandrutrifu-urzica:
> Just want to double check that we only send one log per error, right?

Yes, all errors are parsed by the same `onFileCreationFailed` handler, which only sends one log (failed to fetch/no parcels).

PR #754, opened 2025-07-29
VFB-435: Called no response - Extra question

stefanghenescu-work:
## Jira ticket
https://softwiretech.atlassian.net/jira/software/c/projects/VFB/boards/1130?selectedIssue=VFB-435

## What's changed
- In `StatusesModal.tsx`
  - Add selected statuses to props
  - If selected status is `Call and No Response`, a checkbox will appear
  - Add local state for selected tick boxes
  - Pass selected options to `onSubmit` method
 
- In `Statuses.tsx`
  - Modify `submitStatus` to handle selected statuses
  - Append selected statuses to the shown string as in the example format

## Screenshots
| Before     | After      |
|------------|------------|
| image | image |
| image | image |

## Checklist

### DIFF
--- a/src/app/parcels/ActionBar/Statuses.tsx
+++ b/src/app/parcels/ActionBar/Statuses.tsx
@@ -77,7 +77,7 @@ const Statuses: React.FC<Props> = ({
         void getParcelStatuses();
     }, [setModalError]);
 
-    const submitStatus = async (date: Dayjs): Promise<void> => {
+    const submitStatus = async (date: Dayjs, callNoResponseStatuses: string[]): Promise<void> => {
         setServerErrorMessage(null);
         if (selectedStatus === null) {
             setServerErrorMessage("Chosen status was not found.");
@@ -133,6 +133,7 @@ const Statuses: React.FC<Props> = ({
                 headerId="status-modal-header"
                 onSubmit={submitStatus}
                 errorText={serverErrorMessage}
+                selectedStatus={selectedStatus}
             >
                 <></>
             </StatusesModal>
--- a/src/app/parcels/ActionBar/Statuses.tsx
+++ b/src/app/parcels/ActionBar/Statuses.tsx
@@ -83,10 +83,13 @@ const Statuses: React.FC<Props> = ({
             setServerErrorMessage("Chosen status was not found.");
             return;
         }
+
+        const callNoResponseStatusesList: string = callNoResponseStatuses.join(" / ");
+
         const { error } = await saveParcelTableRowsStatus(
             selectedParcels,
             selectedStatus,
-            null,
+            callNoResponseStatusesList,
             undefined,
             date
         );
--- a/src/app/parcels/ActionBar/StatusesModal.test.tsx
+++ b/src/app/parcels/ActionBar/StatusesModal.test.tsx
@@ -131,4 +131,47 @@ describe("StatusesModal component", () => {
 
         expect(mockOnSubmit).toHaveBeenCalledWith(mockDate, []);
     });
+
+    it("selects 'Called and No Response' checkboxes and verifies the selection", () => {
+        cleanup();
+        render(
+            <Localization>
+                <StyleManager>
+                    <StatusesModal
+                        isOpen={true}
+                        onClose={mockOnClose}
+                        selectedParcels={mockSelectedParcels}
+                        onSubmit={mockOnSubmit}
+                        errorText={null}
+                        headerId="status-modal-header"
+                        header="Apply Status: Called and No Response"
+                        selectedStatus="Called and No Response"
+                    >
+                        children={null}
+                    </StatusesModal>
+                </StyleManager>
+            </Localization>
+        );
+
+        const voicemailCheckbox = screen.getByLabelText("Voicemail");
+        const textCheckbox = screen.getByLabelText("Text");
+        const emailCheckbox = screen.getByLabelText("Email");
+
+        expect(voicemailCheckbox).not.toBeChecked();
+        expect(textCheckbox).not.toBeChecked();
+        expect(emailCheckbox).not.toBeChecked();
+
+        fireEvent.click(voicemailCheckbox);
+        fireEvent.click(emailCheckbox);
+
+        expect(voicemailCheckbox).toBeChecked();
+        expect(textCheckbox).not.toBeChecked();
+        expect(emailCheckbox).toBeChecked();
+
+        fireEvent.click(screen.getByText("Submit"));
+
+        const mockDate = dayjs("2024-01-01 12:00:00");
+
+        expect(mockOnSubmit).toHaveBeenCalledWith(mockDate, ["Voicemail", "Email"]);
+    });
 });
--- a/src/app/parcels/ActionBar/StatusesModal.test.tsx
+++ b/src/app/parcels/ActionBar/StatusesModal.test.tsx
@@ -132,7 +132,7 @@ describe("StatusesModal component", () => {
         expect(mockOnSubmit).toHaveBeenCalledWith(mockDate, []);
     });
 
-    it("selects 'Called and No Response' checkboxes and verifies the selection", () => {
+    it("saves the checked extra question fields when 'Called and No Response' status is selected", () => {
         cleanup();
         render(
             <Localization>
--- a/src/app/parcels/ActionBar/StatusesModal.test.tsx
+++ b/src/app/parcels/ActionBar/StatusesModal.test.tsx
@@ -129,6 +129,6 @@ describe("StatusesModal component", () => {
         const mockDate = dayjs("2024-01-01 12:00:00");
         fireEvent.click(screen.getByText("Submit"));
 
-        expect(mockOnSubmit).toHaveBeenCalledWith(mockDate);
+        expect(mockOnSubmit).toHaveBeenCalledWith(mockDate, []);
     });
 });
--- a/src/app/parcels/ActionBar/StatusesModal.tsx
+++ b/src/app/parcels/ActionBar/StatusesModal.tsx
@@ -15,6 +15,11 @@ const NO_RESPONSE_FOLLOW_UP = ["Voicemail", "Text", "Email"];
 const NO_RESPONSE_STATUS = "Called and No Response";
 const MAX_PARCELS_TO_SHOW = 5;
 
+const FOLLOW_UP_LABELS_AND_KEYS: [string, string][] = NO_RESPONSE_FOLLOW_UP.map((followUp) => [
+    followUp,
+    followUp,
+]);
+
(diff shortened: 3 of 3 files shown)


src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -52,6 +54,24 @@ const StatusesModal: React.FC<StatusesModalProps> = (props) => {
 
     const maxParcelsToShow = 5;
 
+    const [callNoResponseStatuses, setCallNoResponseStatuses] = useState<string[]>([]);
+
```
claudia-golaes-dev: Can you move this line below `const [date, setDate] = useState(dayjs(new Date()));` ?

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -73,8 +93,54 @@ const StatusesModal: React.FC<StatusesModalProps> = (props) => {
...
+                        />
+                        <FormControlLabel
+                            control={
+                                <Checkbox
+                                    checked={callNoResponseStatuses.includes("Email")}
+                                    onChange={() => toggleCallNoResponseStatuses("Email")}
+                                />
+                            }
+                            label="Email"
+                        />
```
claudia-golaes-dev, line 110: Is there any way you can refactor the `Form Group`?

src/app/parcels/ActionBar/StatusesModal.test.tsx
DanielaNegrut, line 1: Is it relevant for us to add a test case to see if the statuses list is sent correctly?

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -30,6 +32,8 @@ const ModalInner = styled.div`
 
 const StatusesModal: React.FC<StatusesModalProps> = (props) => {
     const [date, setDate] = useState(dayjs(new Date()));
+    const [callNoResponseStatuses, setCallNoResponseStatuses] = useState<string[]>([]);
+    const noResponseStatuses = ["Voicemail", "Text", "Email"];
```
DanielaNegrut: Are these constants for the whole file/program? Will they ever be modified?

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -52,6 +56,22 @@ const StatusesModal: React.FC<StatusesModalProps> = (props) => {
 
     const maxParcelsToShow = 5;
 
+    useEffect(() => {
+        if (props.isOpen) {
+            setCallNoResponseStatuses([]);
+        }
+    }, [props.isOpen]);
```
DanielaNegrut: I see there is another `useEffect` with  the same dependency. What is the best approach here, keeping these separate or having them both in the same use effect?

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -52,6 +56,22 @@ const StatusesModal: React.FC<StatusesModalProps> = (props) => {
...
+
+    const toggleCallNoResponseStatuses = (newNoResponseStatus: string): void => {
+        setCallNoResponseStatuses((prevCallNoResponseStatuses) =>
+            prevCallNoResponseStatuses.includes(newNoResponseStatus)
+                ? prevCallNoResponseStatuses.filter(
+                      (prevStatus) => prevStatus !== newNoResponseStatus
+                  )
+                : [...prevCallNoResponseStatuses, newNoResponseStatus]
+        );
```
DanielaNegrut: This seems like a pretty generic functionality. Adding or removing something from an array depending on its state. Do you think it is relevant for other functions in the codebase?

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -73,8 +93,39 @@ const StatusesModal: React.FC<StatusesModalProps> = (props) => {
                     parcels={props.selectedParcels}
                     maxParcelsToShow={maxParcelsToShow}
                 />
+                {props.selectedStatus === "Called and No Response" && (
```
DanielaNegrut: I don't like that we are using a "magic string" 🤔

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -73,8 +93,39 @@ const StatusesModal: React.FC<StatusesModalProps> = (props) => {
...
+                    <FormGroup>
+                        <FormLabel
+                            component="p"
+                            style={{
+                                fontSize: 18,
+                                marginBottom: 8,
+                                marginTop: 8,
+                                color: "inherit",
+                            }}
```
DanielaNegrut: Is it up to project standards to provide styling like this?

src/app/parcels/ActionBar/Statuses.tsx
```diff
@@ -77,16 +77,19 @@ const Statuses: React.FC<Props> = ({
...
 
-    const submitStatus = async (date: Dayjs): Promise<void> => {
+    const submitStatus = async (date: Dayjs, callNoResponseStatuses: string[]): Promise<void> => {
         setServerErrorMessage(null);
         if (selectedStatus === null) {
             setServerErrorMessage("Chosen status was not found.");
             return;
         }
+
+        const callNoResponseStatusesList: string = callNoResponseStatuses.join(" / ");
```
DanielaNegrut, line 87: Why are they kept like this?

DanielaNegrut (commented):
Thank you @stefanghenescu-work, I've left some comments. Let me know if you would like to go through them together!

src/app/parcels/ActionBar/StatusesModal.test.tsx
stefanghenescu-work: Yes, I am planning to add a new test to explicitly verify the selection and update of the status list.

src/app/parcels/ActionBar/StatusesModal.tsx
stefanghenescu-work: The `noResponseStatuses` list is only used to refactor the `Form Group`. Based on how the ticket is defined, these three (Voicemail, Text, Email) are the only statuses that can currently be selected. Since users cannot add new statuses at this point, I don’t see the need to fetch them dynamically.

If, in the future, the admin will be able to add new statuses from the database, then it would make sense to fetch them from there. For now, keeping them as a constant is sufficient.

src/app/parcels/ActionBar/StatusesModal.tsx
stefanghenescu-work: For clarity I would keep them separate. Each effect is responsible for a different state (date vs status list), and keeping them isolated makes the code easier to read and understand.

src/app/parcels/ActionBar/StatusesModal.tsx
DanielaNegrut: The usual practice of constants in React/TS is to keep them at the top of the file / component named in all upper case.

src/app/parcels/ActionBar/StatusesModal.tsx
stefanghenescu-work: I will look into this to try and find a checkbox component in the codebase!

src/app/parcels/ActionBar/Statuses.tsx
stefanghenescu-work: This is the way the ticket specifies the statuses should be displayed when multiple checkboxes are selected.

src/app/parcels/ActionBar/StatusesModal.tsx
stefanghenescu-work: I did not find any constant in the codebase with the name of the statuses so I will make one just for this status in the file.

src/app/parcels/ActionBar/StatusesModal.tsx
stefanghenescu-work: In this case, this functionality is specific to how I manage the checkbox state in `StatusesModal`. There are no other functions in the codebase that use the same toggle-for-array logic, so it’s not relevant elsewhere.

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -8,10 +8,17 @@ import { ParcelsTableRow } from "../parcelsTable/types";
 import SelectedParcelsOverview from "./SelectedParcelsOverview";
 import { ErrorSecondaryText } from "@/app/errorStylingandMessages";
 import { Centerer } from "@/components/Modal/ModalFormStyles";
+import { FormElementWithSpacing } from "@/components/Form/formStyling";
+import CheckboxGroupInput from "@/components/DataInput/CheckboxGroupInput";
+
+const NO_RESPONSE_FOLLOW_UP = ["Voicemail", "Text", "Email"];
+const NO_RESPONSE_STATUS = "Called and No Response";
+const MAX_PARCELS_TO_SHOW = 5;
```
DanielaNegrut, line 16: Awesome!

src/app/parcels/ActionBar/StatusesModal.test.tsx
DanielaNegrut: Was this added?

src/app/parcels/ActionBar/StatusesModal.test.tsx
```diff
@@ -129,6 +129,49 @@ describe("StatusesModal component", () => {
         const mockDate = dayjs("2024-01-01 12:00:00");
         fireEvent.click(screen.getByText("Submit"));
 
-        expect(mockOnSubmit).toHaveBeenCalledWith(mockDate);
+        expect(mockOnSubmit).toHaveBeenCalledWith(mockDate, []);
+    });
+
+    it("selects 'Called and No Response' checkboxes and verifies the selection", () => {
```
DanielaNegrut: This test name doesn't say much to me.

As a general rule that I think makes tests easier to follow is:
- define what should happen early on
- present when the thing should happen

So something like: "when user selects 'Called and no response' status the checked fields are saved "

src/app/parcels/ActionBar/StatusesModal.tsx
```diff
@@ -71,10 +93,27 @@ const StatusesModal: React.FC<StatusesModalProps> = (props) => {
...
+                    maxParcelsToShow={MAX_PARCELS_TO_SHOW}
                 />
+                {props.selectedStatus === NO_RESPONSE_STATUS && (
+                    <FormElementWithSpacing>
+                        <CheckboxGroupInput
+                            groupLabel="Did you send any of the following?"
+                            labelsAndKeys={NO_RESPONSE_FOLLOW_UP.map((followUp) => [
+                                followUp,
+                                followUp,
+                            ])}
```
AlexandraIrinaMardale: Can you do this like it's done in the `BabyProductsCard`, please?

AlexandraIrinaMardale (approved):
👍

PR #756, opened 2025-07-29
VFB-436 More than 1 phone number per client

claudia-golaes-dev:
### Jira Ticket
- [VFB-436: More than 1 phone number per client](https://softwiretech.atlassian.net/browse/VFB-436)

## What's changed
- Added additional_phone_numbers array to the clients table and modified the logic for insert, update and delete a client
- Generated seeds for the additional_phone_number column from the clients_table
- Displayed additional phone numbers to `Clients/Parcels` details card
- Added migrations for parcels_plus and clients_plus to have additional phone numbers
- Concatenated additional phone numbers to the `Phone` column in the Parcels/Clients Table
- Implemented checks for: an existing phone number, not being able to add a phone number if the user didn't enter a primary phone number
- Added additional phone numbers to the following documents: Shipping Labels (only the first three phone numbers), Driver Overview, Shopping Lists, and all reports (Signposting, Missing Voucher, Pending More Info, Selected Parcels)
- Modified tests and added tests for filling in the additional phone numbers in `clientForm.cy.ts`

## Screenshots 
| Before     | After      |
|------------|------------|
|image|image|
|image|image|
|image|image|
|image|image|
|image|image|
|image|image|
|image|image|

## Videos
add1
add2
add3
add4

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/clients/form/formSections/AddressCard.tsx
```diff
@@ -60,42 +62,67 @@ const AddressCard: React.FC<ClientCardProps> = ({
                             defaultValue={getDefaultTextValue(fields, "addressLine1")}
                             error={errorExists(formErrors.addressLine1)}
                             helperText={getErrorText(formErrors.addressLine1)}
-                            onChange={onChangeText(fieldSetter, errorSetter, "addressLine1", {
-                                required: true,
-                            })}
+                            onChange={onChangeText(
+                                fieldSetter,
+                                errorSetter as Setter<FormErrors<ClientFields>>,
```
claudia-golaes-dev, line 67: This is a temporary fix, I needed this cast because of the following error: 

>  TS2345: Argument of type Setter<ParcelErrors> is not assignable to parameter of type
Setter<FormErrors<ParcelFields>> | Setter<Required<FormErrors<ParcelFields>>>
Type Setter<ParcelErrors> is not assignable to type Setter<FormErrors<ParcelFields>>
Types of parameters fieldValuesToUpdate and fieldValuesToUpdate are incompatible.
Type Partial<FormErrors<ParcelFields>> is not assignable to type Partial<ParcelErrors>
Types of property listType are incompatible.
Type Errors | Errors[] | undefined is not assignable to type Errors | undefined
Type Errors[] is not assignable to type Errors | undefined

src/app/admin/manageUser/EditUserForm.tsx
```diff
@@ -112,8 +112,8 @@ const EditUserForm: React.FC<Props> = (props) => {
                     id="edit-user-first-name"
                     label="First Name"
                     defaultValue={fields.firstName}
-                    error={errorExists(formErrors.firstName)}
-                    helperText={getErrorText(formErrors.firstName)}
+                    error={errorExists(formErrors.firstName as Errors)}
+                    helperText={getErrorText(formErrors.firstName as Errors)}
                     onChange={onChangeText(fieldSetter, errorSetter, "firstName", {
```
claudia-golaes-dev, line 117: This change made this cast mandatory 🥀 
```
export type FormErrors<SpecificFields extends Fields> = {
    [errorKey in keyof SpecificFields]?: Errors;
    [errorKey in keyof SpecificFields]?: Errors | Errors[];
};
```

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -122,7 +123,10 @@ export const rawDataToExpandedClientDetails = (client: RawClientDetails): Expand
         fullName: client.full_name ?? "",
         address: formatAddressFromClientDetails(client),
         deliveryInstructions: client.delivery_instructions ?? "",
-        phoneNumber: client.phone_number ?? "",
+        phoneNumber: formatAdditionalPhoneNumbers(
+            client.phone_number,
```
claudia-golaes-dev, line 120: Now, the additional numbers are displayed in the same column as the primary phone number

src/app/parcels/ActionBar/ActionButtons/DriverOverview/getDriverOverviewData.ts
```diff
@@ -138,7 +139,13 @@ const transformRowToDriverOverviewTableData = (
             client?.address_postcode,
             false
         ),
-        contact: clientIsActive ? client?.phone_number ?? "" : "-",
+        contact: clientIsActive
+            ? formatAdditionalPhoneNumbers(
+                  client.phone_number,
+                  client.additional_phone_numbers,
+                  true
+              )
```
claudia-golaes-dev, line 147: I added two optional checks for the additional phone numbers formatting:
- if the phone numbers are displayed in a document, and the phone numbers are empty, instead of showing "None" (for the `Clients/Parcel Details` modal), it should be an empty string, similarly to other fields' behaviour when added in a document
- if the file downloaded is `Shipping Labels`, because of its width and length constraints, **only** the first 3 phone numbers are displayed

supabase/migrations/20250813163651_add_additional_phone_numbers_to_parcels_plus.sql
```diff
@@ -0,0 +1,42 @@
...
+    parcels_events.last_event_is_successfully_completed as last_status_is_successfully_completed,
+    parcels_events.all_events,
+    clients.delivery_instructions as client_delivery_instructions,
+    parcels.list_type,
+    parcels.referral_agency,
+    parcels.referrer_name,
+    parcels.referrer_email,
+    parcels.referrer_phone,
+    clients.email as client_email,
+    array_to_string(clients.additional_phone_numbers, ', ') as client_additional_phone_numbers_text
```
claudia-golaes-dev: It is stored like this because of the filtering problems I encountered + it's used just as a string

supabase/migrations/20250813123651_add_additional_phone_numbers_to_clients.sql
alexandrutrifu-urzica, line 1: Newline at the EOF😄 (this applies to all migrations + clientsSeed)

src/app/clients/form/ClientForm.tsx
```diff
@@ -100,7 +102,6 @@ export interface ClientErrors extends FormErrors<ClientFields> {
 }
 
 export type ClientSetter = Setter<ClientFields>;
-export type ClientErrorSetter = Setter<ClientErrors>;
```
alexandrutrifu-urzica, line 96: Why did you remove this?

src/common/format.ts
```diff
@@ -159,6 +159,32 @@ export const getParcelOverviewString = (
...
+export const formatAdditionalPhoneNumbers = (
+    primaryPhoneNumber: string | null,
+    additionalPhoneNumbers: string[] | null,
+    inDocument?: boolean,
+    inShippingLabels?: boolean
+): string => {
+    if (primaryPhoneNumber === null || primaryPhoneNumber.length === 0) {
+        if (inDocument) {
+            return "";
+        } else {
```
alexandrutrifu-urzica: The `else` can be dropped

src/common/format.ts
```diff
@@ -159,6 +159,32 @@ export const getParcelOverviewString = (
...
+            return "None";
+        }
+    }
+
+    if (additionalPhoneNumbers) {
+        const allPhoneNumbers: string[] = [primaryPhoneNumber, ...additionalPhoneNumbers];
+        if (inShippingLabels) {
+            const firstThreeNumbers = allPhoneNumbers.slice(0, 3);
```
alexandrutrifu-urzica: The variable name gets a bit confusing when the client has less than 3 phone numbers listed, so you could return the result inline (`return allPhoneNumbers.slice(0, 3).join(", ");`)

src/common/format.ts
```diff
@@ -159,6 +159,32 @@ export const getParcelOverviewString = (
...
+            return firstThreeNumbers.join(", ");
+        }
+        return allPhoneNumbers.join(", ");
+    } else {
```
alexandrutrifu-urzica: > The `else` can be dropped

Same here

src/app/parcels/ActionBar/ActionButtons/ShippingLabelsPdfButton.tsx
```diff
@@ -134,7 +134,14 @@ const getRequiredData = async (
             collection_datetime: formatDatetime(parcel.collection_datetime),
             voucher_number: parcel.voucher_number ?? "",
             full_name: clientIsActive ? client.full_name ?? "" : displayNameForDeletedClient,
-            phone_number: clientIsActive ? client.phone_number ?? "" : "-",
+            phone_number: clientIsActive
+                ? formatAdditionalPhoneNumbers(
+                      client.phone_number,
+                      client.additional_phone_numbers,
+                      true,
+                      true
```
alexandrutrifu-urzica: `inDocument` and `inShippingLabels` not being mutually exclusive is a bit strange, maybe there is some way of refactoring these props 🤔

src/app/parcels/parcelsTable/convertParcelDBtoParcelRow.ts
```diff
@@ -48,7 +47,14 @@ const convertParcelDbtoParcelRow = async (
                     ? familyCountToFamilyCategory(parcel.family_count ?? 0)
                     : "-",
                 addressPostcode: clientActive ? parcel.client_address_postcode : "-",
-                phoneNumber: clientActive ? parcel.client_phone_number ?? "" : "-",
+                phoneNumber: clientActive
+                    ? parcel?.client_additional_phone_numbers_text !== null &&
```
alexandrutrifu-urzica: Why is `parcel` marked as undefined?

src/common/databaseFilters.ts
```diff
@@ -54,13 +54,21 @@ export const postcodeSearch = <DbData extends DbClientRow | DbParcelRow>(
...
+        "additional_phone_numbers_text" | "client_additional_phone_numbers_text"
+    >,
     clientIsActiveColumnLabel: Extract<keyof DbData, "is_active" | "client_is_active">
 ): ServerSideFilterMethod<DbData, string> => {
     return dbFilterWithSubstringQueries((substring) => {
-        if ("-".includes(substring.toLowerCase())) {
-            return `or(${clientIsActiveColumnLabel}.is.false, ${phoneColumnLabel}.ilike.%${substring}%)`;
+        if (substring === "-") {
+            const phoneColumnQueryInactiveClient = `or(${clientIsActiveColumnLabel}.is.false, ${phoneColumnLabel}.ilike.%${substring}%)`;
+            const additionalPhoneColumnQueryInactiveClient = `or(${clientIsActiveColumnLabel}.is.false, ${additionalPhoneColumnLabel}.ilike.%${substring}%)`;
```
alexandrutrifu-urzica: Why is the `additionalPhoneColumn` check necessary if empty fields aren't displayed anyway?

src/components/Form/formFunctions.ts
```diff
@@ -61,16 +63,29 @@ export const createSetter = <SpecificFields extends Fields>(
...
-    maxCharacters?: number
+    maxCharacters?: number,
+    currentAdditionalPhoneNumbers?: string[] | null,
+    primaryPhoneNumber?: string | null,
+    index?: number
 ): Errors => {
+    if (
+        currentAdditionalPhoneNumbers &&
+        currentAdditionalPhoneNumbers.length > 0 &&
+        input === "" &&
```
alexandrutrifu-urzica: `input` is a string containing all phone numbers, including the primary one, right?

docs/e2e-testing.md
alexandrutrifu-urzica, line 1: Newline at EOF 😒

cypress/e2e/clients/clientForm.cy.ts
```diff
@@ -38,13 +39,28 @@ describe("Add client form", () => {
 
         assertPostcodeContentNotDisplayed();
     });
+
+    it("Try to add a client with the same additional phone number", () => {
```
alexandrutrifu-urzica: These test names should suggest that the behaviour should not work, e.g. `it("Tries to add a client with the same additional phone number **and fails**")`

cypress/e2e/clients/clientForm.cy.ts
```diff
@@ -58,6 +74,19 @@ function fillPhoneNumber(value: string): void {
...
 
+function fillAdditionalPhoneNumbers(values: string[], test?: boolean): void {
+    cy.contains("Add another phone number").click();
+    fillTextboxWithId("client-additional-phone-number-0", values[0]);
+    cy.contains("Add another phone number").click();
+    fillTextboxWithId("client-additional-phone-number-1", values[1]);
+    if (test) {
+        cy.contains(
+            "This phone number already exists, please add a different phone number."
+        ).should("exist");
```
alexandrutrifu-urzica: Can you change this to `.should("be.visible")`? It's more effective😸

src/app/clients/clientsTable/getClientsData.ts
```diff
@@ -65,7 +65,11 @@ const getClientsDataAndCount = async (
             fullName: client.full_name ?? "",
             familyCategory: familyCountToFamilyCategory(client.family_count ?? 0),
             addressPostcode: client.address_postcode,
-            phoneNumber: client.phone_number,
+            phoneNumber:
+                client?.additional_phone_numbers_text !== null &&
```
alexandrutrifu-urzica: Again, I don't think the client can be undefined/null

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +28,167 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+
+    useEffect(() => {
+        const indexToRemove = fields.additionalPhoneNumbers?.indexOf("");
+        if (indexToRemove != -1 && indexToRemove !== undefined) {
+            const timer = setTimeout(() => {
+                phoneNumbers.splice(indexToRemove, 1);
+                const updatedAdditionalPhones = [...(fields.additionalPhoneNumbers || [])];
+                updatedAdditionalPhones.splice(indexToRemove, 1);
+                fieldSetter({ additionalPhoneNumbers: updatedAdditionalPhones });
+            }, 3000);
```
alexandrutrifu-urzica: The empty input box disappearing after a certain timeout is a bit confusing to me, mostly because it only works after deleting your own input.
That means I can still add a new phone number and not type anything in the form, without anything happening. However, if I keep this empty field and start deleting a previous phone number, both its input box and the one after it will disappear at once.

Is this mechanic necessary in the first place?

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +28,167 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+        }
+    }, [errorMessage]);
+
+    const handleAddPhoneNumber = (): void => {
+        const newId = phoneNumbers.length - 1;
+        if (newId > 0) {
+            if (
+                fields?.additionalPhoneNumbers &&
+                fields?.additionalPhoneNumbers[newId - 1] !== undefined &&
+                fields.additionalPhoneNumbers[newId - 1].length > 1
```
alexandrutrifu-urzica: Again, I don't think `fields` can be undefined.

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +28,167 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+            ) {
+                setPhoneNumbers([...phoneNumbers, { id: newId }]);
+            } else {
+                setErrorMessage(
+                    "The previous phone number should be filled before adding another phone number."
+                );
+            }
+        } else {
+            if (fields.phoneNumber && fields.phoneNumber?.length > 0) {
```
alexandrutrifu-urzica: If the `phoneNumber` is defined, the `?` in the second condition is not necessary.

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +28,167 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+            text="Primary Phone Number"
+        >
+            <GappedDiv>
+                {phoneNumbers.map((phone, index) => (
+                    // eslint-disable-next-line react/no-array-index-key
+                    <GappedDiv key={index}>
+                        {index === 1 && (
+                            <Divider aria-hidden="true" orientation="horizontal" flexItem />
+                        )}
+                        {index === 1 && <FormText>Additional Phone Numbers</FormText>}
```
alexandrutrifu-urzica: The 2 above could be merged into one

src/components/Form/formFunctions.ts
```diff
@@ -61,16 +63,29 @@ export const createSetter = <SpecificFields extends Fields>(
     };
 };
 
-const getErrorType = (
+export const getErrorType = (
```
alexandrutrifu-urzica: I think this too generic of a method to explicitly receive specific fields such as `currentAdditionalPhoneNumbers`. Maybe you could bundle them inside a new prop type and use that instead?

e.g.
```typescript
phoneNumberValidation?: {
        currentAdditionalPhoneNumbers?: string[] | null;
        primaryPhoneNumber?: string | null;
        index?: number;
    };
```

src/components/Form/formFunctions.ts
```diff
@@ -93,11 +120,67 @@ interface OnChangeTextOptions<SpecificFields> {
...
+            const updatedArray = [...(currentAdditionalPhoneNumbers || [])];
+            updatedArray[index] = newValue as string;
+            fieldSetter({ [key]: updatedArray } as {
+                [key in keyof SpecificFields]: SpecificFields[key];
+            });
+        }
+    };
+};
+
+export const onChangeText = <SpecificFields extends Fields>(
```
alexandrutrifu-urzica, line 224: Same as above, think about refactoring this to use a new prop for phone number information.

src/components/Form/formStyling.tsx
alexandrutrifu-urzica, line 1: Newline at EOF

alexandrutrifu-urzica (dismissed):

src/app/clients/form/ClientForm.tsx
claudia-golaes-dev: It wasn't used

src/app/clients/form/formSections/PhoneNumberCard.tsx
claudia-golaes-dev: It is indeed unnecessary. I will find a way so that the empty entries aren't displayed in the modals or stored in the database

src/app/clients/form/formSections/PhoneNumberCard.tsx
claudia-golaes-dev: Fixed

src/components/Form/formFunctions.ts
claudia-golaes-dev: It contains only the value that the user is currently writing

src/app/parcels/ActionBar/ActionButtons/ShippingLabelsPdfButton.tsx
claudia-golaes-dev: I remove the `inDocument` and `inShippingLabels` and added a `documentName?: string`
so instead of 
```
if (inShippingLabels) {
    const firstThreeNumbers = allPhoneNumbers.slice(0, 3);
    return firstThreeNumbers.join(", ");
}
```
I added this:
```
if (documentName === "ShippingLabels") {
     return allPhoneNumbers.slice(0, 3).join(", ");
}
```

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +27,152 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+            required={phoneNumberIsRequired}
+            text="Primary Phone Number"
+        >
+            <GappedDiv>
+                {phoneNumbers.map((phone, index) => (
+                    // eslint-disable-next-line react/no-array-index-key
+                    <GappedDiv key={index}>
+                        {index === 1 && (
+                                <Divider aria-hidden="true" orientation="horizontal" flexItem />
+                            ) && <FormText>Additional Phone Numbers</FormText>}
```
AlexandraIrinaMardale: This can actually be even simpler:
```
{index === 1 && (
    <>
        <Divider aria-hidden="true" orientation="horizontal" flexItem />
        <FormText>Additional Phone Numbers</FormText>
    </>
)}
```

src/components/Form/formFunctions.ts
```diff
@@ -62,8 +64,35 @@ export const createSetter = <SpecificFields extends Fields>(
         setFields({ ...fieldValues, ...fieldValuesToUpdate });
     };
 };
+export const getPhoneNumbersErrorType = (
```
AlexandraIrinaMardale, line 69: Can we add a new line before and after this please?

AlexandraIrinaMardale, line 69: This function is a bit hard to follow, can we refactor it please? 

We can have
 ```const formattedInput = formatPhoneNumber(input);
    const additionalNumbers = currentAdditionalPhoneNumbers || [];
    const isEditingPrimaryPhone = index === undefined;
    const isInputEmpty = input === "";
```

And then we can have an if condition checking that the primary phone number is filled in
And another if condition that checks for duplicates if `!isInputEmpty`

And checking for duplicates can be a separate function that again treats the 2 cases - editing the primary phone numbers and the additional ones.
For primary we'll have almost the same check (a bit easier to read with the new variables)
And for secondary we could again do things a bit separate. So we'd create a variable to check if the number is a duplicate of the primary one, and another to check if it's among the additional ones. And if either of this is true, we have a duplicate, otherwise all good

File: src/components/Form/formFunctions.ts [hunk 25]
```diff
@@ -62,8 +64,35 @@ export const createSetter = <SpecificFields extends Fields>(
...
+        currentAdditionalPhoneNumbers &&
+        currentAdditionalPhoneNumbers.length > 0 &&
+        input === "" &&
+        index === undefined
+    ) {
+        return Errors.emptyPrimaryPhoneNumber;
+    }
+    if (currentAdditionalPhoneNumbers && index !== undefined) {
+        return (currentAdditionalPhoneNumbers.includes(formatPhoneNumber(input)) &&
+            currentAdditionalPhoneNumbers.indexOf(formatPhoneNumber(input)) !== index) ||
```
AlexandraIrinaMardale: I don't think this actually works. Index of returns the first occurrence of that input and I think we might be missing some duplicate cases this way. Something like this should do the trick:
``` 
const isDuplicateOfOtherAdditional = additionalNumbers.some(
    (phoneNumber, i) => phoneNumber === formattedInput && i !== index
);

```

src/components/Form/formFunctions.ts
```diff
@@ -95,32 +124,112 @@ interface OnChangeTextOptions<SpecificFields> {
...
     fieldSetter: Setter<SpecificFields>,
     errorSetter: Setter<FormErrors<SpecificFields>> | Setter<Required<FormErrors<SpecificFields>>>,
     key: keyof SpecificFields,
+    errorType: Errors,
+    input: string,
     options?: OnChangeTextOptions<SpecificFields>
+): void => {
+    errorSetter({ [key]: errorType } as {
+        [key in keyof FormErrors<SpecificFields>]: Errors | Errors[];
+    });
```
AlexandraIrinaMardale: Is this the same as `errorSetter({ [key]: errorType } as Partial<FormErrors<SpecificFields>>);`?

src/components/Form/formFunctions.ts
```diff
@@ -95,32 +124,112 @@ interface OnChangeTextOptions<SpecificFields> {
     maxCharacters?: number;
 }
 
-export const onChangeText = <SpecificFields extends Fields>(
+const errorAndFieldSetters = <SpecificFields extends Fields>(
```
alexandrutrifu-urzica: Maybe you can rename this to `callErrorAndFieldSetters`?

src/components/Form/formFunctions.ts
```diff
@@ -95,32 +124,112 @@ interface OnChangeTextOptions<SpecificFields> {
...
+    input: string,
     options?: OnChangeTextOptions<SpecificFields>
+): void => {
+    errorSetter({ [key]: errorType } as {
+        [key in keyof FormErrors<SpecificFields>]: Errors | Errors[];
+    });
+    if (errorType === Errors.none) {
+        const newValue = options?.formattingFunction ? options.formattingFunction(input) : input;
+        fieldSetter({ [key]: newValue } as {
+            [key in keyof SpecificFields]: SpecificFields[key];
```
AlexandraIrinaMardale, line 146: And this could be `as Partial<SpecificFields>`

AlexandraIrinaMardale: And this and the previous comment apply to quite a few lines, if you could go through them 😄

alexandrutrifu-urzica (approved):

src/components/Form/formFunctions.ts
AlexandraIrinaMardale: I counted around 8 field setters and 7 error setters 😅

src/components/Form/formFunctions.ts
```diff
@@ -305,19 +416,36 @@ export const checkErrorOnSubmit = <
 ): boolean => {
     let errorExists = false;
     let amendedErrorTypes = { ...errorType };
+
     for (const [errorKey, error] of Object.entries(errorType)) {
         if (!keysToCheck || keysToCheck.includes(errorKey)) {
-            if (error !== Errors.none) {
+            // Handle array of errors for additionalPhoneNumbers
+            if (errorKey === "additionalPhoneNumbers" && Array.isArray(error)) {
```
AlexandraIrinaMardale: Do we actually need to check the key, or is the fact that it's an array enough?

src/components/Form/formFunctions.ts
```diff
@@ -305,19 +416,36 @@ export const checkErrorOnSubmit = <
...
+                if (error === Errors.initial) {
+                    amendedErrorTypes = {
+                        ...amendedErrorTypes,
+                        [errorKey]: Errors.required,
+                    };
+                }
             }
         }
     }
+
```
AlexandraIrinaMardale: I'm not sure I'm following what's happening here. Can you explain it please?

src/components/Form/formStyling.tsx
```diff
@@ -108,3 +108,17 @@ export const GappedDiv = styled.div`
...
     gap: 1em;
 `;
+
+export const GappedRowDiv = styled.div`
+    display: flex;
+    flex-direction: row;
+    gap: 0.5em;
+
+    > :first-child {
+        flex: 0 0 97%;
```
AlexandraIrinaMardale: Very specific 🤔 

Would it work if we did
```
    > :first-child {
        flex: 1; /* Grows to fill available space */
    }
    
    > :not(:first-child) {
        flex: 0 0 auto; /* Only as big as content needs */
    }
```

src/components/Form/formFunctions.ts
```diff
@@ -95,32 +124,112 @@ interface OnChangeTextOptions<SpecificFields> {
...
+                : getPhoneNumbersErrorType(
+                      cleanedInput,
+                      currentAdditionalPhoneNumbers,
+                      primaryPhoneNumber
+                  );
         }
+
+        if (key === "additionalPhoneNumbers" && index !== undefined) {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const currentErrors = Array.isArray((errorSetter as any).additionalPhoneNumbers) // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
AlexandraIrinaMardale: Isn't error setter a function?

AlexandraIrinaMardale: So we'd need the actual errors

src/components/Form/formFunctions.ts
claudia-golaes-dev: The fact that it's an array should be enough, deleted the errorKey check

src/components/Form/formFunctions.ts
claudia-golaes-dev: There are two separate cases that appeared when trying to fix the problem I encountered with having a shared error for all the additional fields. 
→ If the error is an array, it checks if the array contains any error other than `Errors.none`. If yes, it flags that some error exists
→ It then turns any Errors.initial entries to Errors.required on submit
→ To avoid unnecessary updates, it only writes back to amendedErrorTypes[errorKey] if the array actually changed

src/components/Form/formFunctions.ts
```diff
@@ -63,7 +65,37 @@ export const createSetter = <SpecificFields extends Fields>(
...
+        if (additionalNumbers.length && isInputEmpty) {
+            return Errors.emptyPrimaryPhoneNumber;
+        }
+        return isDuplicateOfOtherAdditional ? Errors.phoneNumberAlreadyExists : Errors.none;
+    }
+
+    return isDuplicateOfPrimaryPhone || isDuplicateOfOtherAdditional
+        ? Errors.phoneNumberAlreadyExists
+        : Errors.none;
+};
```
AlexandraIrinaMardale: Nice, just a few more notes. We can add some early returns:
```
    // Primary phone must be filled before any additional phones
    if (isEditingPrimaryPhone && isInputEmpty && additionalNumbers.length > 0) {
        return Errors.emptyPrimaryPhoneNumber;
    }

    // Don't check duplicates for empty input
    if (isInputEmpty) {
        return Errors.none;
    }
```

and we can also do
```
       // Only check primary phone duplicate if primary phone exists
    const isDuplicateOfPrimaryPhone = primaryPhoneNumber && 
        formattedInput === formatPhoneNumber(primaryPhoneNumber);

    if (isEditingPrimaryPhone) {
        return isDuplicateOfOtherAdditional ? Errors.phoneNumberAlreadyExists : Errors.none;
    }

```

supabase/migrations/20250905161949_add_additional_phone_numbers_to_clients_plus.sql
```diff
@@ -0,0 +1,16 @@
+create or replace view
+  public.clients_plus as
```
AlexandraIrinaMardale: Can you add the `with (security_invoker = true)` here as well please?

supabase/seed/seed.ts
```diff
@@ -72,6 +73,12 @@ const main = async (): Promise<never> => {
                 delivery_instructions: () => copycat.oneOf(ctx.seed, defaultDeliveryInstructions),
                 family_id: () => copycat.uuid(ctx.seed),
                 default_list: () => copycat.oneOf(ctx.seed, possibleListTypesWeighted),
+                additional_phone_numbers: (ctx) =>
+                    copycat.someOf(
+                        ctx.seed,
+                        [0, possiblePhoneNumbers.length],
+                        possiblePhoneNumbers
+                    ),
```
AlexandraIrinaMardale: Will this add duplicate phone numbers? 🤔

src/components/Form/formFunctions.ts
```diff
@@ -305,19 +410,36 @@ export const checkErrorOnSubmit = <
...
+                }
+                const updatedErrors = error.map((err) =>
+                    err === Errors.initial ? Errors.required : err
+                );
+                if (JSON.stringify(updatedErrors) !== JSON.stringify(error)) {
+                    amendedErrorTypes = {
+                        ...amendedErrorTypes,
+                        [errorKey]: updatedErrors,
+                    };
+                }
```
AlexandraIrinaMardale: Do we need them to also have the same order? 🤔 

If yes, we could do:
```
const needsUpdate = !error.every((err, index) => err === updatedErrors[index]);

```

src/common/format.ts
```diff
@@ -166,6 +166,28 @@ export const getParcelOverviewString = (
     );
 };
 
+export const formatAdditionalPhoneNumbers = (
```
AlexandraIrinaMardale, line 169: Nice 😄

src/app/parcels/getExpandedParcelDetails.ts
```diff
@@ -11,14 +11,20 @@ import {
...
+    formatAdditionalPhoneNumbers,
+} from "@/common/format";
 import {
     Data,
     DataForDataViewer,
     convertDataToDataForDataViewer,
 } from "@/components/DataViewer/DataViewer";
 import { formatEventName } from "@/app/parcels/format";
 import { ListType } from "@/common/databaseListTypes";
+
```
AlexandraIrinaMardale: Can you remove this?

src/app/parcels/parcelsTable/convertParcelDBtoParcelRow.ts
```diff
@@ -48,7 +47,14 @@ const convertParcelDbtoParcelRow = async (
...
                 addressPostcode: clientActive ? parcel.client_address_postcode : "-",
-                phoneNumber: clientActive ? parcel.client_phone_number ?? "" : "-",
+                phoneNumber: clientActive
+                    ? parcel.client_additional_phone_numbers_text !== null &&
+                      parcel.client_additional_phone_numbers_text?.length > 0
+                        ? parcel.client_phone_number +
+                          ", " +
+                          parcel.client_additional_phone_numbers_text
+                        : parcel.client_phone_number
+                    : "-",
```
AlexandraIrinaMardale: I think we can simplify this

Something like this should work:
```
clientActive 
    ? [parcel.client_phone_number, parcel.client_additional_phone_numbers_text]
        .filter(phone => phone && phone.length > 0)
        .join(", ")
    : "-"
```

src/app/clients/clientsTable/getClientsData.ts
```diff
@@ -65,7 +65,11 @@ const getClientsDataAndCount = async (
             fullName: client.full_name ?? "",
             familyCategory: familyCountToFamilyCategory(client.family_count ?? 0),
             addressPostcode: client.address_postcode,
-            phoneNumber: client.phone_number,
+            phoneNumber:
```
AlexandraIrinaMardale: Same simplification here

src/app/batch-create/helpers/submitTableData.ts
AlexandraIrinaMardale, line 1: Why is this here?

docs/e2e-testing.md
```diff
@@ -4,3 +4,13 @@
 - Run `npm run dev -- -p 3200` to run the dev server on port 3200, where Cypress will run the tests on
 - Use `npx cypress open` to open Cypress client to run specific tests and allow hot-reloads.
 - You can also do `npm run build` and then `npm run test:e2e --spec=<your_test_file_path>` to run tests in the specific file
+
+
+## Common problems
+### Accessibility.cy.ts
```
AlexandraIrinaMardale, line 10: Great!

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +27,157 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+
+    useEffect(() => {
+        if (errorMessage) {
+            const timer = setTimeout(() => {
+                setErrorMessage(null);
+            }, 3000);
+
+            return () => clearTimeout(timer);
+        }
+    }, [errorMessage]);
```
AlexandraIrinaMardale, line 54: What does this do?

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +27,157 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+
+    const handleAddPhoneNumber = (): void => {
+        const newId = phoneNumbers.length - 1;
```
AlexandraIrinaMardale, line 78: Why is this -1?

AlexandraIrinaMardale: Would extracting something like this do the same thing?
```
const validatePhoneSequence = (): string | null => {
    if (!fields.phoneNumber || fields.phoneNumber.length === 0) {
        return Errors.emptyPrimaryPhoneNumber;
    }
    
    const lastAdditionalIndex = phoneNumbers.length - 2;
    if (lastAdditionalIndex >= 0) {
        const lastPhone = fields.additionalPhoneNumbers?.[lastAdditionalIndex];
        if (!lastPhone || lastPhone.length === 0) {
            return "The previous phone number should be filled before adding another phone number.";
        }
    }
    
    return null;
};
```

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +27,157 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
-            />
+        <GenericFormCard
+            title="Phone Number"
+            required={phoneNumberIsRequired}
+            text="Primary Phone Number"
+        >
+            <GappedDiv>
+                {phoneNumbers.map((phone, index) => (
+                    // eslint-disable-next-line react/no-array-index-key
+                    <GappedDiv key={index}>
```
AlexandraIrinaMardale: Can we use the phone id as a key  here instead of the array index?

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +27,157 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+                                        formattingFunction: formatPhoneNumber,
+                                    }
+                                )}
+                            />
+                        )}
+                        {index >= 1 && (
+                            <GappedRowDiv>
+                                <FreeFormTextInput
+                                    disabled={errorExists(formErrors.phoneNumber)}
+                                    key={phone.id}
```
AlexandraIrinaMardale: And I don't think we need this one

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +27,157 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+                                    id={`client-additional-phone-number-${phone.id}`}
+                                    label={`Phone Number ${index > 0 ? index + 1 : ""}`}
```
AlexandraIrinaMardale: You already check that index >= 1 on line 129, so we can just have ``label={`Phone Number ${index + 1}`}``

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +27,157 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+                                    defaultValue={fields.additionalPhoneNumbers?.[index - 1] || ""}
+                                    error={
+                                        formErrors.additionalPhoneNumbers?.[index - 1] !==
+                                            undefined &&
+                                        errorExists(formErrors.additionalPhoneNumbers?.[index - 1])
+                                    }
```
AlexandraIrinaMardale: Can we extract this into 
```
const getPhoneNumberError = (index: number) => {
    const errorIndex = index - 1;
    const error = formErrors.additionalPhoneNumbers?.[errorIndex];
    return error !== undefined && errorExists(error);
};
```

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +27,157 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
-                    formattingFunction: formatPhoneNumber,
-                })}
-            />
+        <GenericFormCard
+            title="Phone Number"
+            required={phoneNumberIsRequired}
+            text="Primary Phone Number"
+        >
+            <GappedDiv>
+                {phoneNumbers.map((phone, index) => (
```
AlexandraIrinaMardale: It might be helpful to add 
```
            const isAdditionalPhone = index >= 1;
            const additionalPhoneIndex = index - 1;
```
and use them where needed

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +27,157 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+                                    formErrors,
+                                    {
+                                        required: phoneNumberIsRequired,
+                                        regex: phoneNumberRegex,
+                                        formattingFunction: formatPhoneNumber,
+                                    }
+                                )}
+                            />
+                        )}
+                        {index >= 1 && (
```
AlexandraIrinaMardale: This can also be
```
{index === 0 && (...) : (...)}
```

AlexandraIrinaMardale (commented):
Nice work 😄

src/app/clients/form/formSections/PhoneNumberCard.tsx
claudia-golaes-dev: I added a timeout for the following warnings, so that they are only displayed for three seconds:
`The previous phone number should be filled before adding another phone number.`
`The primary phone number should be filled before adding another phone number.`
If you think this is not necessary I can remove it 😄

supabase/seed/seed.ts
claudia-golaes-dev: No, it won't

> copycat.someOf(input, range, values)
Takes in an [input](https://www.npmjs.com/package/@snaplet/copycat#input) value and an array of values, repeatedly picks items from that array a number of times within the given range that corresponds to that input. Each item will be picked no more than once.

This is the resource: https://www.npmjs.com/package/@snaplet/copycat

src/app/clients/form/formSections/PhoneNumberCard.tsx
claudia-golaes-dev: The `phoneNumbers` array's initial length is 1, because even if the primary phone number is empty, this entry is stored inside the phoneNumber, so to keep the ids starting from 0, the `newId` is set to `phoneNumbers.length - 1;`

src/app/clients/form/formSections/PhoneNumberCard.tsx
```diff
@@ -18,20 +28,167 @@ const PhoneNumberCard: React.FC<ClientCardProps> = ({
...
+            const timer = setTimeout(() => {
+                setErrorMessage(null);
+            }, 3000);
+
+            return () => clearTimeout(timer);
+        }
+    }, [errorMessage]);
+
+    const getPhoneNumberError = (index: number): boolean => {
+        const error = formErrors.additionalPhoneNumbers?.[index];
```
AlexandraIrinaMardale, line 57: Doesn't this need to be index - 1?

AlexandraIrinaMardale: Nevermind 😅

claudia-golaes-dev: I added the `const additionalPhoneIndex = index - 1;` that you suggested above and now it doesn;t nedd to be `index - 1` anymore.

AlexandraIrinaMardale (approved):
Great, thanks 😄 Let's make sure we test this thoroughly

PR #758, opened 2025-07-29
Vfb 422 concurrent edits info page

anamariasuditu:
## What's changed
Concurrent edits in the Info Page are now blocked.

### Before

https://github.com/user-attachments/assets/f033aae6-b193-4199-a368-2432ac4859be

 ### After 

https://github.com/user-attachments/assets/0bd9c442-c078-4961-961a-635032b5a653

 

## Checklist

    - Checkout the dev branch
    - Checkout your branch


supabase/migrations/20250729134042_last_updated_column_automatically_updates_wiki_table.sql
```diff
@@ -0,0 +1,4 @@
+create extension if not exists "moddatetime" with schema "extensions";
+
+CREATE TRIGGER handle_updated_at BEFORE UPDATE ON public.wiki FOR EACH ROW EXECUTE FUNCTION moddatetime('last_updated');
+
```
claudia-golaes-dev: extra space 🥹

src/app/info/WikiItemEdit.tsx
```diff
@@ -73,8 +77,19 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
         }
     };
 
-    const cancelWikiItemEdit = (): void => {
+    const cancelWikiItemEdit = async (): Promise<void> => {
+        setErrorMessage("");
+        setIsInEditMode(false);
         !rowData.title && !rowData.content ? deleteWikiItem() : setIsInEditMode(false);
```
LeonardCiobanu-Softwire, line 84: I think this should be updated because you already do the 

> setIsInEditMode(false);

in the line before

src/app/info/WikiItemEdit.tsx
```diff
@@ -85,53 +100,66 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
...
+        const auditLog = {
+            action: "edit a wiki item",
+            wikiId: rowData.wiki_key,
+            content: {
+                itemTitle: newTitle,
+                itemContent: newContent,
+                rowOrder: rowData.row_order,
+                lastUpdated: rowData.last_updated,
+            },
+        } as const satisfies Partial<AuditLog>;
```
LeonardCiobanu-Softwire, line 122: Why did you add the 

> as const

in the type?

LeonardCiobanu-Softwire (commented):
Left some comments 😄

src/app/info/WikiItemEdit.tsx
anamariasuditu: > I think this should be updated because you already do the
> 
> > setIsInEditMode(false);
> 
> in the line before

I updated it, thank you!

src/app/info/WikiItemEdit.tsx
anamariasuditu: > Why did you add the
> 
> > as const
> 
> in the type?

I didn't add it, it was originally like this in the code, `as const` ensures the object is read-only and the values become literals instead of wide types. In my case, `action `is not just of type string, but exactly _"edit a wiki item"_

src/app/info/WikiItemEdit.tsx
```diff
@@ -85,53 +101,66 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
...
+        if (updateError) {
+            if (updateError.message.includes("edited recently")) {
+                setErrorMessage(updateError.message);
+                void sendAuditLog({
+                    ...auditLog,
+                    wasSuccess: false,
+                });
+                return;
+            }
+            const logId = await logErrorReturnLogId("error updating wiki row item", updateError);
```
AlexandraIrinaMardale: This should probably mention that it's a concurrent edit, and it should be a warning not an error - check the other concurrent edit solutions. And we want the log id in the audit log as well so this needs to go before that

AlexandraIrinaMardale: Update: I just saw the comment on [VFB-421](https://softwiretech.atlassian.net/browse/VFB-421) and we don't want to send a failed save to the audit log or log the error

src/app/info/WikiItemEdit.tsx
```diff
@@ -163,7 +164,6 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
 
         setIsInEditMode(false);
     };
-
```
AlexandraIrinaMardale, line 136: Why was this removed?

src/app/info/WikiItemEdit.tsx
```diff
@@ -131,14 +131,15 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
         if (updateError) {
             if (updateError.message.includes("edited recently")) {
                 setErrorMessage(updateError.message);
-                void sendAuditLog({
-                    ...auditLog,
-                    wasSuccess: false,
-                });
                 return;
+            } else {
```
AlexandraIrinaMardale: We don't need the `else`

src/app/info/WikiItemEdit.tsx
```diff
@@ -131,14 +131,15 @@ const WikiItemEdit: React.FC<WikiItemEditProps> = ({
...
+                const logId = await logErrorReturnLogId(
+                    "error updating wiki row item",
+                    updateError
+                );
+                setErrorMessage(`Failed to update wiki item. Log ID: ${logId}`);
             }
-            const logId = await logErrorReturnLogId("error updating wiki row item", updateError);
-            setErrorMessage(`Failed to update wiki item. Log ID: ${logId}`);
+
```
AlexandraIrinaMardale: Can you remove this line?

src/app/info/WikiItemEdit.tsx
anamariasuditu: > Why was this removed?

The `);` that was removed was the closing for the `confirm()` function call, but I deleted it and refactored the code.

AlexandraIrinaMardale: Ah sorry, it was showing up as an empty line to me 🙃

src/app/info/EditModeDependentItem.tsx
```diff
@@ -21,31 +23,59 @@ const EditModeDependentItem: React.FC<EditProps> = ({
...
+        setRowsInEditMode((prev) => {
+            const newSet = new Set(prev);
+            if (val) {
+                row?.wiki_key && newSet.add(row.wiki_key);
+            } else {
+                row?.wiki_key && newSet.delete(row.wiki_key);
+            }
+            return newSet;
+        });
+    };
```
AlexandraIrinaMardale, line 49: We can use useCallback here and we can also check that `row` exists before proceeding with the checks, e.g:
```
const handleSetIsInEditMode = useCallback((val: boolean): void => {
    setIsInEditMode(val);
    
    if (!row?.wiki_key) {
        return; // Early return if no valid wiki_key
    }
    
    setRowsInEditMode((prev) => {
        const newSet = new Set(prev);
        if (val) {
            newSet.add(row.wiki_key);
        } else {
            newSet.delete(row.wiki_key);
        }
        return newSet;
    });
}, [row?.wiki_key, setRowsInEditMode]);

```

src/app/info/EditModeDependentItem.tsx
```diff
@@ -21,31 +23,59 @@ const EditModeDependentItem: React.FC<EditProps> = ({
...
+
+    const openEditMode = (): void => {
+        handleSetIsInEditMode(true);
+    };
```
AlexandraIrinaMardale, line 53: This isn't necessary, we can just pass handleSetIsInEditMode as needed, so:
```
<WikiItemDisplay
    openEditMode={() => handleSetIsInEditMode(true)}
    // other props...
/>
```

supabase/migrations/20250912132618_add_last_updated_column_to_wiki_table.sql
AlexandraIrinaMardale, line 1: These need to be updated again 🙃

Issue #684, opened 2025-06-11
Reports: not showing correct numbers of clients


dan-langridge:
### Reproduction steps

On the Reports page, the table is not showing the correct totals for each family size

For example, the week commencing the 2025-04-21, the numbers are very low. compared to other weeks.
For the 21st-28th April, there are 108 singles with the status Fulfilled with Trussel, but the table only shows 7.

To confirm, the table should take into account parcels  that have a positive status of delivery/collection, for example  the current status:
Fulfilled With Trussell
Delivered
Parcel Collected

Is this issue because we renamed the Status Fulfilled with Trussel (emitting Trust?)

### Expected behaviour

The table should take into account parcels  that have a positive status of delivery/collection, for example  the current status:

Fulfilled With Trussell
Delivered
Parcel Collected

Is this issue because we renamed the Status Fulfilled with Trussel (emitting Trust?)

### What browsers are you seeing the problem on?

Safari

### What OS is your computer running on?

macOS

stonelink:
Hi Dan, where does this one come in the priority list?

stonelink:
### Investigation notes: 
* Parcels are included in the `reports` view if they have a status that has `is_succesfully_completed_event` in the database table `status_order`
* The three statuses with this flag are the three that Dan lists above; changing the text of "Fulfilled with Trussell Trust" didn't change that

For w/c 21st April:
* The `completed_parcels` view has 13 parcels; completed_parcels is built from the events table
* In the `events` table for that week there are the following:
  *  0 events that set a parcel status to "Delivered" in that week, although lots of parcels with their last status event as "Out for Delivery"
  * 12 events that set parcel status to "Parcel Collected" - across 10 distinct parcels
  * 12 events that set parcel status to "Fulfilled with Trussell" - across 12 distinct parcels
  * There are no parcel IDs that have both events for Collected & Fulfilled
* The `completed_parcels` table contains all of the "Fulfilled with Trussell" parcels, plus 1 of the "Parcel Collected" parcels

### Next questions:
* Do the parcels with "Parcel Collected" have another event following that, which is why they're not in `completed_parcels`?
* Why are there no "Delivered" events in that week - was that a temporary bug?

### Update:
* Of the parcels with the "Parcel Collected" events in the `events` table in that week, all but 1 had "Parcel Collected" as the event with the last timestamp, so they _should_ have been included in `completed_parcels` but they weren't - this is a bug
[...]

PR #760, opened 2025-07-29
#684 Build reports by packing date

stonelink:
## What's changed
The reports view is now built based on the packing date of the parcels, not the timestamp of the completion event. That has simplified the view structure, including the removal of the view completed_parcels.

## Checklist

  - [X] I have updated the typescript definitions for the database with `npm run generate_types:local`
  - [-] I have modified the seed in `supabase/seed/seed.ts` if appropriate
  - [X] With my final set up, I can run `npm run reset_supabase` without any errors.
  - [X] I have taken reasonable measures to ensure constraint violations do not happen when the migration occurs, e.g.:
  - [X] I have checked that migration can happen successfully without resetting the database, doing the following.
    - Checkout the dev branch
    - Checkout your branch

### DIFF
--- a/src/app/parcels/parcelsTable/ParcelsPage.test.tsx
+++ b/src/app/parcels/parcelsTable/ParcelsPage.test.tsx
@@ -45,6 +45,7 @@ const sampleProcessingData: DbParcelRow[] = [
         last_status_event_data: "SOME_RELATED_DATA",
         last_status_timestamp: "2023-08-04T13:30:00+00:00",
         last_status_workflow_order: 1,
+        last_status_is_successfully_completed: false,
         all_events: ["LAST_EVENT"],
         created_at: "2023-12-31T12:00:00+00:00",
         client_is_active: true,
--- a/src/databaseTypesFile.ts
+++ b/src/databaseTypesFile.ts
@@ -805,37 +805,6 @@ export type Database = {
         }
         Relationships: []
       }
-      completed_parcels: {
-        Row: {
-          completed_timestamp: string | null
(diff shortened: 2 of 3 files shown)


stonelink:
Hi @AlexandraIrinaMardale 

To fix the bug in the `reports` view, I've had to:
* rebuild the `parcels_events` and `parcels_plus` views - each now includes a boolean field indicating whether the parcel's last event was a "completed" event
* remove the `completed_parcels` view 

Do any of the migrations created by the interns redefine or use any of those 3 views?
If they do, then I think we'll need to update those migrations to be based on the new versions.

Also, the date of the migration in this PR is 20250713 - I think that should be before any migrations written by the interns, but could you please check? The date of the last migration on `dev` is 20250609.

Thanks,
Lincoln

AlexandraIrinaMardale:
Hey @stonelink! We have a few migrations on `parcels_plus`. I'll update them and let you know when I've added the PR so you can double check them 😄 

And yeah, I'm pretty sure all of our migrations are after yours, so that should be fine

AlexandraIrinaMardale:
I think the best way to go about this is wait for you to merge this PR, rebase intern-dev on dev and then (hopefully) also add a PR for merging intern-dev into dev. Does this make sense? 🤔

stonelink:
> I think the best way to go about this is wait for you to merge this PR, rebase intern-dev on dev and then (hopefully) also add a PR for merging intern-dev into dev. Does this make sense? 🤔

That sounds good, thanks.

I'll merge now, and do QA on dev later today.

PR #762, opened 2025-07-30
VFB-426: Concurrent edits website data

stefanghenescu-work:
## Jira ticket
https://softwiretech.atlassian.net/jira/software/c/projects/VFB/boards/1130?selectedIssue=VFB-426

## What's changed
- Concurrent edits in the admin page for editing website data are now blocked
- Add new parameter with the local timestamp for updating function
- Adding a new column for "last_updated " to the website_data table
- Making the "last_updated" column to change with a trigger when a row is modified

### Before
https://github.com/user-attachments/assets/82701131-b49d-4f78-9bdd-80464c5523ac

### After
https://github.com/user-attachments/assets/60d6e324-54bf-429f-9f26-a6d09536341f

## Checklist

    - Checkout the dev branch
    - Checkout your branch


supabase/migrations/20250729151124_add_last_updated_column_website_data_table.sql
```diff
@@ -0,0 +1,3 @@
+alter table "public"."website_data" add column "last_updated" timestamp with time zone not null default CURRENT_TIMESTAMP;
+
```
claudia-golaes-dev: extra spaces 🥹

src/app/admin/websiteDataTable/fetchWebsiteData.ts
```diff
@@ -37,24 +37,28 @@ export const fetchWebsiteData = async (): Promise<FetchWebsiteDataErrorReturn> =
...
+        value: newRow.value,
+        last_updated: newRow.lastUpdated,
     };
 
     const { data: updatedWebsiteData, error } = await supabase
         .from("website_data")
-        .update(processedData)
+        .update({ value: processedData.value })
         .eq("name", processedData.name)
+        .eq("last_updated", oldTimestamp)
```
LeonardCiobanu-Softwire: I think it might be good to either:
- use `count` to verify if the entry was updated, and so we would know it is a concurrency issue
- verify if the data returned is an empty list, in which case the same would be true
In this way, we would send an error and a log addressing the concurrency specifically, instead of just calling it an `update error`

LeonardCiobanu-Softwire: Just like @anamariasuditu did in [VFB-425](https://github.com/LambethFoodbank/foodbankapp/pull/752/files) in the src/app/admin/packingSlotsTable/PackingSlotActions.ts file at line 118

src/app/admin/websiteDataTable/fetchWebsiteData.ts
```diff
@@ -37,26 +37,32 @@ export const fetchWebsiteData = async (): Promise<FetchWebsiteDataErrorReturn> =
...
+        error,
+        count,
+    } = await supabase
         .from("website_data")
-        .update(processedData)
+        .update({ value: processedData.value }, { count: "exact" })
         .eq("name", processedData.name)
-        .select()
-        .single();
+        .eq("last_updated", oldTimestamp);
```
LeonardCiobanu-Softwire: There is no select query, so the data will be null no matter what. This is important for the audit log.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -90,55 +90,126 @@ const WebsiteDataTable: React.FC = () => {
...
+                `Failed to refresh row ${id}: ${error?.message}`
+            );
+            setErrorMessage(`Failed to refresh data. Log ID: ${logId}`);
+            return null;
+        }
+
+        const refreshedRow: WebsiteDataRow = {
+            id: latestData.name,
+            dbName: latestData.name,
+            readableName: getReadableWebsiteDataName(latestData.name),
```
stefanghenescu-work, line 111: `latestData.name` is fetched from the DB and is snake case. In the table we want the name to be shown with spaces and upper case.

LeonardCiobanu-Softwire (approved):
👍

src/app/admin/websiteDataTable/fetchWebsiteData.ts
```diff
@@ -69,6 +76,13 @@ export const updateDbWebsiteData = async (
         void sendAuditLog({ ...auditLog, wasSuccess: false, logId });
         return { error: { type: "failedToUpdateWebsiteData", logId } };
     }
+
+    if (count == 0) {
```
LeonardCiobanu-Softwire: You should also verify the case in which the data returned is null/empty. It is the same case as the count === 0.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -90,55 +90,126 @@ const WebsiteDataTable: React.FC = () => {
...
+
+                setRowErrors((prev) => ({ ...prev, [newRow.id]: message }));
+                setErrorMessage(message);
+                throw new Error(message);
             }
+
+            // Clear any previous errors for this row
+            setRowErrors((prev) => {
+                const newErrors = { ...prev };
+                delete newErrors[newRow.id];
```
LeonardCiobanu-Softwire, line 147: I'm not sure that const is a good choice here

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -90,55 +90,126 @@ const WebsiteDataTable: React.FC = () => {
...
             ...currentValue,
             [id]: { mode: GridRowModes.Edit },
         }));
     };
 
-    const handleCancelClick = (id: GridRowId) => () => {
+    const handleCancelClick = (id: GridRowId) => async () => {
+        setRowErrors((prev) => {
+            const newErrors = { ...prev };
+            delete newErrors[id];
```
LeonardCiobanu-Softwire, line 191: Same here.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -90,55 +90,126 @@ const WebsiteDataTable: React.FC = () => {
...
+            return newErrors;
+        });
+
+        setBlockedSaveRows((prev) => {
+            const newSet = new Set(prev);
+            newSet.delete(id);
```
LeonardCiobanu-Softwire, line 197: Once again

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -90,55 +90,126 @@ const WebsiteDataTable: React.FC = () => {
...
+            // Clear any previous errors for this row
+            setRowErrors((prev) => {
+                const newErrors = { ...prev };
+                delete newErrors[newRow.id];
+                return newErrors;
+            });
+
+            setBlockedSaveRows((prev) => {
+                const newSet = new Set(prev);
+                newSet.delete(newRow.id);
```
LeonardCiobanu-Softwire, line 153: Here too

LeonardCiobanu-Softwire (commented):
LGTM

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
stefanghenescu-work: As `newErrors` is used just for update and never reassigned is better to use `const`.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -90,55 +90,126 @@ const WebsiteDataTable: React.FC = () => {
...
-            switch (error.type) {
-                case "failedToUpdateWebsiteData":
-                    setErrorMessage(`Failed to update website data. Log ID ${error.logId}`);
-                    break;
+                if (error.type === "concurrentEditWebsiteData") {
+                    message =
+                        "Record has been edited recently - please refresh the page. " +
+                        `Log ID: ${error.logId}`;
+                    setBlockedSaveRows((prev) => new Set(prev).add(newRow.id));
+                    await refreshRow(newRow.id);
```
AlexandraIrinaMardale: Why do we refresh the row here?

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -90,55 +90,126 @@ const WebsiteDataTable: React.FC = () => {
...
+            id: latestData.name,
+            dbName: latestData.name,
+            readableName: getReadableWebsiteDataName(latestData.name),
+            value: latestData.value,
+            lastUpdated: latestData.last_updated,
+        };
+
+        setRows((prevRows) => prevRows.map((row) => (row.id === id ? refreshedRow : row)));
+
+        return refreshedRow;
```
AlexandraIrinaMardale, line 118: Should we also reset the error message when we successfully refresh the row?

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -90,55 +90,126 @@ const WebsiteDataTable: React.FC = () => {
...
-            {
-                void logErrorReturnLogId(
-                    "Edited row in website data admin table is undefined onCancelClick"
-                );
-                setErrorMessage("Table error, please try again");
-            }
+        const refreshedRow = await refreshRow(id.toString());
+        if (!refreshedRow) {
+            const logId = await logErrorReturnLogId("Failed to refresh row data on cancel");
+            setErrorMessage(`Failed to refresh data. Please try again. Log ID: ${logId}`);
```
AlexandraIrinaMardale: Don't we already do this in refresh row?

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
```diff
@@ -239,7 +318,15 @@ const WebsiteDataTable: React.FC = () => {
                     editMode="row"
                     rowModesModel={rowModesModel}
                     onRowModesModelChange={setRowModesModel}
+                    onRowEditStart={handleRowEditStart}
                     onRowEditStop={handleRowEditStop}
+                    onProcessRowUpdateError={(error) => {
+                        setErrorMessage(error.message);
+                    }}
+                    onCellDoubleClick={(params, event) => {
```
AlexandraIrinaMardale, line 320: Do we need both `onRowEditStart` and `onCellDoubleClick`?

src/app/admin/websiteDataTable/fetchWebsiteData.ts
```diff
@@ -37,26 +37,33 @@ export const fetchWebsiteData = async (): Promise<FetchWebsiteDataErrorReturn> =
...
         })
     );
 
     return { data: websiteData, error: null };
 };
 
 export const updateDbWebsiteData = async (
-    row: WebsiteDataRow
+    newRow: WebsiteDataRow,
+    oldTimestamp: string | undefined
```
AlexandraIrinaMardale: Can we call this original timestamp? Or something similar

src/app/admin/websiteDataTable/fetchWebsiteData.ts
```diff
@@ -69,6 +76,13 @@ export const updateDbWebsiteData = async (
         void sendAuditLog({ ...auditLog, wasSuccess: false, logId });
         return { error: { type: "failedToUpdateWebsiteData", logId } };
     }
+
+    if (count == 0 || !processedData || processedData.value.length === 0) {
```
AlexandraIrinaMardale: 1. Can we use strict equality here, i.e `===`?
2. I'm not sure I'm following the other 2 checks. The count one definitely indicates concurrency, but why would the other 2 happen?

AlexandraIrinaMardale (commented):
Can you please rebase on intern-dev?

I like that you block the save on error - can you make sure to add this for collection centres/packing slots once you rebase?

Also, let's double check that we only send one audit log per action - it looks correct but just to make sure

And finally, I noticed that if I have an error on a row, that doesn't go away until I refresh the page. If we refresh the data on cancel then we should also clear the error. Also I think we should also clear the error when we successfully submit another row - and then if we click save again on the row with the error, we should get the message back

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
stefanghenescu-work: I will delete it, as there is no point of refreshing on error.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
stefanghenescu-work: If we do this, when a row has an error shown and we refresh the same row on a different tab, then the warning message on the first one will disappear.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
stefanghenescu-work: Yes, you are right.

src/app/admin/websiteDataTable/fetchWebsiteData.ts
stefanghenescu-work: For point 2.  you are right, there is no need in verifying that,.

src/app/admin/websiteDataTable/WebsiteDataTable.tsx
stefanghenescu-work: Yes, the original timestamp field (`originalTimestampsRef`) must be updated if the edit mode is reached by clicking the edit button (this is done by `onRowEditStart`) or by double clicking the row (this is done by `onCellDoubleClick`).

AlexandraIrinaMardale (commented):
Looks good, but let's wait for Claudia to merge 414 - you'll need to update the timestamps after that but other than this it's fine 😄

AlexandraIrinaMardale (approved):

PR #772, opened 2025-08-04
VFB-414: Delivery Area Admin Panel 

claudia-golaes-dev:
### Jira Tickets
-  [**VFB-411: Add delivery area section to admin panel**](https://softwiretech.atlassian.net/browse/VFB-411)
-  [**VFB-412: Add filter for postcodes outside of delivery area**](https://softwiretech.atlassian.net/browse/VFB-412)
-  [**VFB-413: Flag client postcodes outside of delivery area**](https://softwiretech.atlassian.net/browse/VFB-413)
-  [**VFB-414: Only staff/manager/admin should be able to open client/parcel outside delivery area**]( https://softwiretech.atlassian.net/browse/VFB-414 )
-  [**VFB-439: NFA postcode should be inside the delivery area**]( https://softwiretech.atlassian.net/browse/VFB-439 )

## What's changed

- Added delivery Area panel to the Admin page
- Implemented sorting and editing
- Made appropriate seeding and migrations (also made it so it synchronizes the delivery_areas table with the postcodes in the clients table)
- Added a filter similar to the Packing Slot checkboxes, except only one option or none can be selected at once, on the `Clients` and `Parcels` pages, with the following filters:
  - Inside of the delivery area 
  - Outside of the delivery area
- Modified the ParcelsTableRow and ClientsTableRow interfaces to contain isDeliverable
- Added a warning icon and showed it in the "Postcode" column
- Volunteers won't be able to access a client and any of the client's parcels if it is listed as outside of the delivery area
- Changed the migrations to count NULL entries in client address_postcode as deliverable
- When adding a new client and the user presses "No Address", the postcode will be set to NULL, not to an empty string

## Screenshots
| Before     | After      |
|------------|------------|
| image | image |
| image | image |
| image|image|
|image|image|
|image|image|

#### Delivery Area Admin Panel
delivery_area_panel

#### Delivery Area Filter and Icons
delivery_area_filterandicons

#### Delivery Areas Volunteer/Admin
volunteer
admin

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/admin/AdminPage.tsx
```diff
@@ -66,6 +69,11 @@ const AdminPage: React.FC = () => {
             panelIcon: auditLogIcon,
             panelContent: <AuditLogTable />,
         },
+        {
+            panelTitle: "Edit Delivery Areas",
+            panelIcon: faTruck,
+            panelContent: <DeliveryAreasTable />,
+        },
```
mihaicutaiulia: perhaps we could place `View Audit Logs` as the last table, and move this one a bit higher

src/components/Icons/ClientsOutsideDeliveryAreaIcon.tsx
claudia-golaes-dev, line 1: [**VFB-413: Flag client postcodes outside of delivery area**](https://softwiretech.atlassian.net/browse/VFB-413)
Icon for this ticket

src/components/Tables/ChecklistFilter.tsx
```diff
@@ -19,6 +19,7 @@ interface ChecklistFilterProps<
     isDisabled?: boolean;
     isHidden?: boolean;
     isHiddenInUrl?: boolean;
+    isRadio?: boolean;
 }
```
claudia-golaes-dev, line 24: isRadio check for  [**VFB-412: Add filter for postcodes outside of delivery area**](https://softwiretech.atlassian.net/browse/VFB-412), so only one option (Inside/Outside) can be checked at once or none

src/app/parcels/parcelsTable/ParcelsTable.tsx
```diff
@@ -215,8 +216,12 @@ const ParcelsTable: React.FC<ParcelsTableProps> = ({
         }
     }, [filteredParcelCount, checkedParcelIds, isAllCheckBoxSelected]);
 
+    const { role } = useContext(RoleUpdateContext);
+
     const onParcelTableRowClick = (row: Row<ParcelsTableRow>): void => {
-        openParcelModal(row.data.parcelId);
+        if (row.data.addressColumn.isDeliverable || (role !== null && role !== "volunteer")) {
+            openParcelModal(row.data.parcelId);
+        }
```
claudia-golaes-dev, line 223: Check for -  [**VFB-414: Only staff/manager/admin should be able to open client/parcel outside delivery area**]( https://softwiretech.atlassian.net/browse/VFB-414 )

src/app/parcels/parcelsTable/format.tsx
```diff
@@ -79,16 +80,36 @@ const rowToLastStatusColumn = (data: ParcelsTableRow["lastStatus"] | null): stri
...
 
+const rowToAddressColumn = ({
+    addressPostcode,
+    isDeliverable,
+}: ParcelsTableRow["addressColumn"]): React.ReactElement => {
+    const postcodeRow: React.ReactNode[] = [];
+    postcodeRow.push(formatNullPostcode(addressPostcode));
+
+    if (!isDeliverable && !addressPostcode?.includes("-")) {
+        postcodeRow.push(
```
claudia-golaes-dev: `!addressPostcode?.includes("-")` checks for the Deleted Clients (so they don't have the icon in the postcode column)

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,350 @@
...
+}
+
+export interface DeliveryAreasRow {
+    id: string;
+    postcode: string;
+    isNew: boolean;
+}
+
+const formatPostcode = (value: string): string => {
+    value = value.toUpperCase().slice(0, 4);
```
mihaicutaiulia: can you confirm that this works for postcodes where the region part has 2, 3, or 4 characters - for example, N3, E20, and SW11?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,350 @@
...
+        value = value.slice(0, value.length - 1);
+    }
+    return value;
+};
+
+function EditToolbar(props: EditToolbarProps): React.JSX.Element {
+    const { setRows, setRowModesModel, rows } = props;
+
+    const handleClick = (): void => {
+        const id = rows.length + 1;
```
mihaicutaiulia: it’s not clear to me what the id variable is meant to represent. If it’s supposed to be the `id` from the database row, shouldn’t it be generated randomly when adding the data? Could you give me a bit more context?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,350 @@
...
+        </GridToolbarContainer>
+    );
+}
+
+function getBaseAuditLogForDeliveryAreasAction(
+    action: string,
+    deliveryAreasRow: DeliveryAreasRow,
+    options?: {
+        excludeDeliveryAreasId?: boolean;
+    }
```
mihaicutaiulia: is `options` necessary?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,350 @@
...
+                "add a new delivery area",
+                newRow,
+                { excludeDeliveryAreasId: true }
+            );
+
+            if (insertDeliveryAreasError) {
+                setErrorMessage(
+                    `Failed to add the delivery area. Log ID: ${insertDeliveryAreasError.logId}`
+                );
+                setRows((rows) => rows.slice(0, -1));
```
mihaicutaiulia: this removes the row from the UI if the insertion fails, right?

src/app/clients/clientsTable/ClientsPage.tsx
```diff
@@ -174,6 +176,38 @@ const ClientsPage: React.FC = () => {
         }
     };
 
+    const { role } = useContext(RoleUpdateContext);
+
+    const onRowClick = (row: Row<ClientsTableRow>): void => {
+        if (row.data.addressColumn.isDeliverable || (role !== null && role !== "volunteer")) {
```
mihaicutaiulia: in what scenario can the role be null?

src/app/clients/clientsTable/ClientsPage.tsx
```diff
@@ -174,6 +176,38 @@ const ClientsPage: React.FC = () => {
...
+        }
+    };
+
+    const RowToIconsColumn = ({
+        addressPostcode,
+        isDeliverable,
+    }: ClientsTableRow["addressColumn"]): React.ReactElement => {
+        const postcodeRow: React.ReactNode[] = [];
+        postcodeRow.push(formatNullPostcode(addressPostcode));
+        if (!isDeliverable && !addressPostcode?.includes("-")) {
```
mihaicutaiulia: just to confirm, does `-` represent the postcode for deleted clients?

src/app/clients/clientsTable/getClientsData.ts
```diff
@@ -64,9 +64,13 @@ const getClientsDataAndCount = async (
             clientId: client.client_id ?? "",
             fullName: client.full_name ?? "",
             familyCategory: familyCountToFamilyCategory(client.family_count ?? 0),
-            addressPostcode: client.address_postcode,
+            addressColumn: {
```
mihaicutaiulia: just a suggestion, i think we can keep `addressPostcode` as the name of the original field and rename the new `addressPostcode` simply to `postcode`

supabase/migrations/20250804135852_add_delivery_area_to_parcels_plus.sql
```diff
@@ -0,0 +1,45 @@
...
+  clients.delivery_instructions as client_delivery_instructions,
+  parcels.list_type,
+  parcels.referral_agency,
+  parcels.referrer_name,
+  parcels.referrer_email,
+  parcels.referrer_phone,
+  clients.email as client_email,
+  ((delivery_areas.postcode is not null and clients.is_active) is true or clients.address_postcode is null) as is_deliverable
+from
+   ((((((parcels
```
mihaicutaiulia: you could remove the parenthesis

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
claudia-golaes-dev: yes, it works, this function constrains the input to:
- start with a letter
- only contain letters and numbers
- not have more than 4 characters

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
claudia-golaes-dev: sure, it's not the id from the database, it is only for the `GridRowsProp` to insert it in the table with a certain id. i think it's used by the table from the library for sorting the rows

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
claudia-golaes-dev: no, thank you

src/app/clients/clientsTable/ClientsPage.tsx
claudia-golaes-dev: if the user is not logged, just an edgecase

mihaicutaiulia: are we able to access the page without being logged in?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,346 @@
...
+    postcode: string;
+    isNew: boolean;
+}
+
+const formatPostcode = (value: string): string => {
+    value = value.toUpperCase().slice(0, 4);
+    if (!value.charAt(0).match(/[A-Z]/)) {
+        value = value.slice(0, 0);
+    }
+    if (!value.charAt(value.length - 1).match(/[A-Z0-9]/)) {
```
alexandrutrifu-urzica: This check doesn't seem to cover the case where the user modifies characters _in the middle of the input_ (_e.g._ if they delete the 2nd character of the string, then replace it with an invalid one, they will still be able to submit the new postcode).

The first if-statement should definitely stay as is, but maybe you could try refactoring the second one to check the format of the whole `(1, length - 1)` substring instead. This would ensure the entire input is valid while the user is typing.

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,346 @@
...
+    const handleRemoveClick = (id: GridRowId) => () => {
+        const editedRow = rows.find((row) => row.id === id);
+        if (editedRow != undefined) {
+            deleteDbDeliveryAreas(editedRow).then(() =>
+                setRows((oldRows) => oldRows.filter((row) => row.id !== id))
+            );
+        }
+    };
+
+    const deliveryAreassColumns: GridColDef[] = [
```
alexandrutrifu-urzica: Small typo here😄

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,346 @@
...
+    }, []);
+
+    const handleSaveClick = (id: GridRowId) => () => {
+        setRowModesModel((currentValue) => ({
+            ...currentValue,
+            [id]: { mode: GridRowModes.View },
+        }));
+    };
+
+    const processRowUpdate = async (newRow: DeliveryAreasRow): Promise<DeliveryAreasRow> => {
```
alexandrutrifu-urzica: Can you change the parameter name to a simple `row`? 😶‍🌫️

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,346 @@
...
+    };
+
+    const handleCancelClick = (id: GridRowId) => () => {
+        setRowModesModel((currentValue) => ({
+            ...currentValue,
+            [id]: { mode: GridRowModes.View, ignoreModifications: true },
+        }));
+
+        const editedRow = rows.find((row) => row.id === id);
+        if (editedRow === undefined) {
```
alexandrutrifu-urzica: When can the `editedRow` be undefined? Does this have to do with a case of concurrent access (i.e. a second user deletes the same row before the first user saves their changes) or is it something else?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,346 @@
...
+}
+
+const formatPostcode = (value: string): string => {
+    value = value.toUpperCase().slice(0, 4);
+    if (!value.charAt(0).match(/[A-Z]/)) {
+        value = value.slice(0, 0);
+    }
+    if (!value.charAt(value.length - 1).match(/[A-Z0-9]/)) {
+        value = value.slice(0, value.length - 1);
+    }
```
anamariasuditu: I think you could have a stricter restriction since it shouldn't allow you to type a postcode which contains only letters for example, since it's invalid.
And does it also cover these cases of postcodes ending with a letter, with a number in the middle: EC1A, W1A?

src/app/clients/clientsTable/ClientsPage.tsx
```diff
@@ -174,6 +176,38 @@ const ClientsPage: React.FC = () => {
...
+
+    const onRowClick = (row: Row<ClientsTableRow>): void => {
+        if (row.data.addressPostcode.isDeliverable || (role !== null && role !== "volunteer")) {
+            router.push(`/clients?${clientIdParam}=${row.data.clientId}`);
+        } else {
+            router.push("/clients");
+        }
+    };
+
+    const RowToIconsColumn = ({
```
anamariasuditu: could you rename this to make it more clear it's for the outside delivery area icon?

src/app/clients/clientsTable/ClientsPage.tsx
```diff
@@ -148,7 +150,7 @@ const ClientsPage: React.FC = () => {
         })();
     }, [clientId]);
 
-    const formatNullPostcode = (postcodeData: ClientsTableRow["addressPostcode"]): string => {
+    const formatNullPostcode = (postcodeData: string | null): string => {
```
anamariasuditu: why did you change `formatNullPostcode `?

src/app/clients/clientsTable/ClientsPage.tsx
```diff
@@ -174,6 +176,38 @@ const ClientsPage: React.FC = () => {
...
+        }
+    };
+
+    const RowToIconsColumn = ({
+        postcode,
+        isDeliverable,
+    }: ClientsTableRow["addressPostcode"]): React.ReactElement => {
+        const postcodeRow: React.ReactNode[] = [];
+        postcodeRow.push(formatNullPostcode(postcode));
+        if (!isDeliverable && postcode !== "-") {
```
anamariasuditu: What about postcodes marked as NULL in Supabase?

claudia-golaes-dev: The isDeliverable column also contains a check for null postcodes

src/app/clients/clientsTable/ClientsPage.tsx
claudia-golaes-dev: It was used only here and I needed it to be used on any string

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
claudia-golaes-dev: We wanted to allow multiple variants for the postcode, so the functions constraints the input to:
- start with a letter
- only contain letters and numbers
- not have more than 4 characters

I will change it so it has to end with a number, thank you.

src/app/clients/clientsTable/filters.ts
```diff
@@ -54,6 +55,7 @@ const clientsFilters: ClientsFilter[] = [
         label: "Email",
         method: clientsEmailSearch,
     }),
+    deliveryAreaFilter("is_deliverable", "is_active") as unknown as ClientsFilter,
```
anamariasuditu: could you use something more robust instead of `as unknown as`?

src/app/clients/form/formSections/AddressCard.tsx
```diff
@@ -1,16 +1,16 @@
+import { Checkbox, FormControlLabel } from "@mui/material";
 import React, { useState } from "react";
 import FreeFormTextInput from "@/components/DataInput/FreeFormTextInput";
 import {
-    Errors,
```
anamariasuditu: could you leave here the initial order of imports since nothing changed?

src/app/clients/form/submitFormHelpers.ts
```diff
@@ -1,14 +1,15 @@
-import { InsertSchema, UpdateSchema } from "@/databaseUtils";
+import { ListType } from "@/common/databaseListTypes";
 import { checkboxGroupToArray, Person } from "@/components/Form/formFunctions";
-import supabase from "@/supabaseClient";
+import { InsertSchema, UpdateSchema } from "@/databaseUtils";
```
anamariasuditu: same here with the imports order

src/app/parcels/parcelsTable/format.tsx
```diff
@@ -79,16 +80,36 @@ const rowToLastStatusColumn = (data: ParcelsTableRow["lastStatus"] | null): stri
     );
 };
 
-const formatNullPostcode = (postcodeData: ParcelsTableRow["addressPostcode"]): string => {
+const formatNullPostcode = (postcodeData: string | null): string => {
     return postcodeData ?? displayPostcodeForHomelessClient;
 };
 
+const rowToAddressColumn = ({
```
anamariasuditu: Since this has the same logic as `rowToIconsColumn`  in `ClientsPage`, but is called `rowToAddressColumn`, and we have another `RowToIconsColumn` here in `ParcelsTable`, I think it could be quite confusing. What are the differences between them and could you maybe rename the variables to something easier to understand?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
anamariasuditu: I think you have to check also cases with numbers in the middle for delivery area, since they are valid postcodes that already exists in Supabase (EC1A), so it isn't required to end with a number. The correct format for a delivery area postcode would be: A0, A00, AA0, A0A, AA00, AA0A, where A - any letter and 0 - any digit.

supabase/seed/postcodeSeed.ts
anamariasuditu, line 1: maybe include in the seed delivery areas with all possible formats to ensure at testing it works for all cases (check my other comment for the other formats), since in Supabase on intern-dev the table is populated with all types of possible accepted postcodes already.

supabase/migrations/20250804135852_add_delivery_area_to_parcels_plus.sql
```diff
@@ -0,0 +1,45 @@
+create or replace view
+  "public"."parcels_plus" with (security_invoker = true) as
+select
+  parcels.primary_key as parcel_id,
+  parcels.collection_datetime,
+  parcels.packing_date,
+  parcels.created_at,
+  packing_slots.name as packing_slot_name,
+  packing_slots."order" as packing_slot_order,
```
anamariasuditu: why does `order` have "" ?

src/app/admin/deliveryAreasTable/DeliveryAreasActions.ts
```diff
@@ -0,0 +1,120 @@
...
+const formatExistingRowToDbDeliveryAreas = (row: DeliveryAreasRow): DbDeliveryAreas => {
+    return {
+        id: row.id,
+        postcode: row.postcode,
+    };
+};
+
+const formatNewRowToDbDeliveryAreas = (newRow: DeliveryAreasRow): NewDbDeliveryAreas => {
+    return { postcode: newRow.postcode };
+};
```
alexandrutrifu-urzica: Instead of keeping both these methods, maybe you could try refactoring them into a single, generic one.

For example, the `DbDeliveryAreas` type would be enough on its own - `NewDbDeliveryAreas` is only used by its associated formatter function, so it could be omitted.
Then, you could write a single formatter function like so:

```
const formatRowToDbDeliveryAreas = (row: DeliveryAreasRow): DbDeliveryAreas | Omit<DbDeliveryAreas, "id"> => {
    const baseData = { postcode: row.postcode };
    return row.isNew ? baseData : { ...baseData, id: row.id };
};
```

This should be able to replace the other calls without complications, but you can take another look.

supabase/migrations/20250804115057_add_delivery_areas_table.sql
alexandrutrifu-urzica, line 1: Why did you choose to grant full permissions to every role? Is it necessary for the policy to work?
If not, in what context should anonymous/unprivileged users be allowed to make changes to the `delivery_areas` table?

src/app/parcels/parcelsTable/format.tsx
alexandrutrifu-urzica: I agree, since both return an addressPostcode, maybe they could be refactored into common components for the parcels/clients.
The formatNullPostcode method above could be extracted elsewhere as well (since it is used in `ClientsPage` too)

src/app/parcels/parcelsTable/filters.ts
```diff
@@ -259,6 +260,7 @@ export const buildParcelFilters = async (
         await buildDeliveryCollectionFilter(),
         await buildPackingSlotFilter(),
         await buildLastStatusFilter(),
+        deliveryAreaFilter("is_deliverable", "client_is_active"),
```
alexandrutrifu-urzica: Why is this the only synchronous filter among them?

src/app/clients/form/submitFormHelpers.ts
```diff
@@ -79,6 +94,32 @@ export type addClientResult =
           error: { type: addClientErrors; logId: string };
       };
 
+type InsertDeliveryAreaErrorType = "failedToInsertDeliveryArea";
+export type InsertDeliveryAreaReturnType = {
+    error: { type: InsertDeliveryAreaErrorType; logId: string } | null;
+};
+type InsertDeliveryArea = (
```
alexandrutrifu-urzica: Can you rename this to `InsertDeliveryAreaType` so it's not confused with the method below?

src/app/clients/form/submitFormHelpers.ts
```diff
@@ -137,6 +181,8 @@ export const submitEditClientForm = async (
             clientid: primaryKey,
         }
     );
+    const deliveryAreaRecord = formatDeliveryAreaRecord(fields);
+    await insertDeliveryArea(deliveryAreaRecord);
```
alexandrutrifu-urzica: Shouldn't the `Edit Client Form` have a separate `updateDeliveryArea()` method instead?

alexandrutrifu-urzica (changes requested):
Since this ticket involves quite some significant changes, including new visual elements and possible user interactions, I think some new tests would be necessary.

I've noticed the updated checks for the address-postcode format, however it would be nice to add a few tests for correct rendering and behaviour as well.

src/app/clients/form/submitFormHelpers.ts
claudia-golaes-dev: The `updateDeliveryArea()` method doesn't exist. In the `Edit Client Form` we used:
`await insertDeliveryArea(deliveryAreaRecord);` because if you modify the postcode of a client, the prefix would be by default added (if it isn't already in the delivery_areas table). The table has two functionalities:
1. Adding a postcode prefix
2. Removing an existent one

An admin can't edit a postcode prefix.

src/app/parcels/parcelsTable/filters.ts
claudia-golaes-dev: The `deliveryAreaFilter` function is synchronous because it returns a filter object directly - It calls `serverSideChecklistFilter()` and returns the result immediately

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
claudia-golaes-dev: It's similar to /src/app/admin/packingSlotsTable/PackingSlotsTable.tsx, and yes, that is exactly one of the reasons that we have this check

src/app/admin/deliveryAreasTable/DeliveryAreasActions.ts
LeonardCiobanu-Softwire: I believe this would be too complicated, as this function is only used in a single place. 
Replacing the usage of this function with just its output would be a lot simpler. The change should look something like this:
>  const data = formatNewRowToDbDeliveryAreas(newRow);

=> `const data = { postcode: newRow.postcode };`



Alongside this, the `formatNewRowToDbDeliveryAreas` function should be removed

src/app/admin/deliveryAreasTable/DeliveryAreasActions.ts
```diff
@@ -0,0 +1,120 @@
...
+    row: DeliveryAreasRow
+): Promise<UpdateDeliveryAreasResult> => {
+    const processedData = formatExistingRowToDbDeliveryAreas(row);
+    const { error } = await supabase
+        .from("delivery_areas")
+        .delete()
+        .eq("postcode", processedData.postcode);
+
+    if (error) {
+        const logId = await logErrorReturnLogId("Failed to update delivery area", {
```
LeonardCiobanu-Softwire: I think the error message should say: `"Failed to delete delivery area"`

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
LeonardCiobanu-Softwire: I think renaming it `rowId` should clear up the confusion for any future devs coming across this code.

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,363 @@
...
+                "postgres_changes",
+                { event: "*", schema: "public", table: "delivery_areas" },
+                async () => {
+                    setErrorMessage(null);
+                    try {
+                        const deliveryAreas = await fetchDeliveryAreas();
+                        setRows(deliveryAreas);
+                    } catch (error) {
+                        setRows([]);
+                        setErrorMessage("Error fetching data, please reload");
```
LeonardCiobanu-Softwire, line 125: The `setErrorMessage` should be beneath the `logErrorReturnLogId`, just for code consistency.

src/app/clients/clientsTable/ClientsPage.tsx
LeonardCiobanu-Softwire: > What about postcodes marked as NULL in Supabase?

I believe the `formatNullPostcode`  handles those postcodes by replacing them with `NFA`, in which case the `postcode !== "-"` check might not handle the case it should

supabase/migrations/20250804135852_add_delivery_area_to_parcels_plus.sql
claudia-golaes-dev: the quotation marks were there before

src/app/parcels/parcelsTable/format.tsx
claudia-golaes-dev: Refactored, thank you!

supabase/migrations/20250804115057_add_delivery_areas_table.sql
claudia-golaes-dev: Those are generated by supabase, you can also check `20240312105515_create_packing_slot_table.sql`.

src/app/clients/clientsTable/ClientsPage.tsx
```diff
@@ -174,6 +171,20 @@ const ClientsPage: React.FC = () => {
         }
     };
 
+    const { role } = useContext(RoleUpdateContext);
+
+    const onRowClick = (row: Row<ClientsTableRow>): void => {
+        if (row.data.addressPostcode.isDeliverable || (role !== null && role !== "volunteer")) {
```
stefanghenescu-work: It might be a good idea to extract this verification into a utility function in `roles.tsx`, as I’ve noticed the same check is repeated in several places and role name already exist in a constant.

src/app/parcels/parcelsTable/ParcelsTable.tsx
```diff
@@ -215,8 +216,12 @@ const ParcelsTable: React.FC<ParcelsTableProps> = ({
         }
     }, [filteredParcelCount, checkedParcelIds, isAllCheckBoxSelected]);
 
+    const { role } = useContext(RoleUpdateContext);
+
     const onParcelTableRowClick = (row: Row<ParcelsTableRow>): void => {
-        openParcelModal(row.data.parcelId);
+        if (row.data.addressPostcode.isDeliverable || (role !== null && role !== "volunteer")) {
```
stefanghenescu-work: Same utility function for this verification.

src/app/parcels/ActionBar/ActionModals/GenerateMapModal.tsx
```diff
@@ -63,8 +63,8 @@ const GenerateMapModal: React.FC<ActionModalProps> = (props) => {
 
     const formattedPostcodes = props.selectedParcels.reduce<string[]>(
         (formattedPostcodes, parcel) => {
-            if (parcel.addressPostcode && parcel.addressPostcode !== "-") {
-                formattedPostcodes.push(parcel.addressPostcode.replaceAll(" ", ""));
+            if (parcel.addressPostcode.postcode && parcel.addressPostcode.postcode !== "-") {
+                formattedPostcodes.push(parcel.addressPostcode.postcode.replaceAll(" ", ""));
```
stefanghenescu-work: Could you extract `parcel.addressPostcode.postcode` into a local variable  to make the code easier to read.😄

src/app/admin/deliveryAreasTable/DeliveryAreasActions.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (!data || data.length === 0) {
+        const logId = await logErrorReturnLogId("Delete did not match any records", {
+            attemptedId: row.id,
+        });
+        return {
+            error: {
+                dbError: {
+                    message: "No rows deleted",
+                    details: "ID did not match any rows",
+                } as unknown as PostgrestError,
```
stefanghenescu-work: Same issue as @anamariasuditu raised, if you could use something more robust rather than `as unknown as`.

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,351 @@
...
+            {rows && (
+                <StyledDataGrid
+                    rows={rows}
+                    aria-label="Delivery Areas Table"
+                    initialState={{
+                        sorting: {
+                            sortModel: [{ field: "postcode", sort: "asc" }],
+                        },
+                    }}
+                    sortingOrder={["asc", "desc"]}
```
stefanghenescu-work, line 297: Why don't we want to have 3 sorting states (ascending, descending, unsorted) as the `StyledDataGrid` component provides by default?

claudia-golaes-dev: By default, the sortingOrder is set to none, which led to unexpected behaviour when adding a new delivery area, so a workaround that I found was to set it to "asc" intially. I will remove this line: `sortingOrder={["asc", "desc"]}`

src/components/Tables/ChecklistFilter.tsx
AlexandraIrinaMardale: I remember this was discussed at the first demo, did we decide to keep radio buttons or checkboxes that allow multiple select?

src/app/roles.tsx
```diff
@@ -50,6 +50,13 @@ export const roleCanAccessPage = (role: UserRole | null, url: string): boolean =
     return accessList.some((page) => url.startsWith(page));
 };
 
+export const roleCanAccessModal = (
```
AlexandraIrinaMardale: Can you rename this, please? The name suggests it could be used for any modal, when this refers just to the parcel modal, right?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,351 @@
...
+    isNew: boolean;
+}
+
+const isValidPostcode = (value: string): boolean => prefixPostcodeRegex.test(value);
+
+function EditToolbar(props: EditToolbarProps): React.JSX.Element {
+    const { setRows, setRowModesModel } = props;
+
+    const handleClick = (): void => {
+        const id = `new-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
```
AlexandraIrinaMardale, line 54: What does this do?

src/app/clients/clientsTable/filters.ts
AlexandraIrinaMardale: What was the conclusion on this?

src/app/clients/form/submitFormHelpers.ts
```diff
@@ -79,6 +94,32 @@ export type addClientResult =
...
+type InsertDeliveryAreaErrorType = "failedToInsertDeliveryArea";
+export type InsertDeliveryAreaReturnType = {
+    error: { type: InsertDeliveryAreaErrorType; logId: string } | null;
+};
+type InsertDeliveryAreaType = (
+    deliveryAreaRecord: DeliveryAreaDatabaseInsertRecord
+) => Promise<InsertDeliveryAreaReturnType>;
+
+export const insertDeliveryArea: InsertDeliveryAreaType = async (deliveryAreaRecord) => {
+    const { error } = await supabase.from("delivery_areas").insert(deliveryAreaRecord).single();
```
AlexandraIrinaMardale: What happens when we get a postcode that already exists in the database?

src/app/roles.tsx
claudia-golaes-dev: Is for the parcels and clients modal, i will rename it

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
claudia-golaes-dev: In the first version, I used the id as `the number of rows + 1`, which raised some issues and now I generate the id randomly like this.

src/app/clients/form/submitFormHelpers.ts
claudia-golaes-dev: The `delivery_areas` table has the following constraint applied:
```
alter table "public"."delivery_areas" add constraint "delivery_areas_postcode_check" CHECK ((length(postcode) > 2)) not valid;
alter table "public"."delivery_areas" validate constraint "delivery_areas_postcode_check";
```
The error is thrown in the `insertDeliveryArea` method and handled in the `DeliveryAreasTable.tsx`:

```
if (row.isNew) {
            const { data: createdDeliveryAreas, error: insertDeliveryAreasError } =
                await insertNewDeliveryAreas(row);
            const baseAuditLog = getBaseAuditLogForDeliveryAreasAction(
                "add a new delivery area",
                row
            );

            if (insertDeliveryAreasError) {
                setErrorMessage(
                    `The specified delivery area already exists. Log ID: ${insertDeliveryAreasError.logId}`
                );
                setRows((prevRows) => prevRows.filter((prevRow) => prevRow.id !== row.id));
                void sendAuditLog({
                    ...baseAuditLog,
                    wasSuccess: false,
                    logId: insertDeliveryAreasError.logId,
                });
            } 
```
Should I remove the Log ID?

src/components/Tables/ChecklistFilter.tsx
claudia-golaes-dev: It looks like a checkbox and doesn't function exactly like a radio button, since it also allows an unchecked state. The conclusion from the demo is that this behaviour might be a bit confusing for users who want to build on it, but for now, it should remain as is.

src/app/clients/form/submitFormHelpers.ts
AlexandraIrinaMardale: I think those are 2 different functions - you have one in `DeliveryAreasAction` and the one in `submitFormHelpers`. This second one is responsible for adding a new delivery area when we add/update a client. And in that case we shouldn't really be showing an error to the user if already exists

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
AlexandraIrinaMardale: Can we add a comment explaining this please?

supabase/migrations/20250905142044_add_delivery_areas_table.sql
AlexandraIrinaMardale, line 1: Migrations look good, just don't forget to update the timestamps and add the fields from 436 after we merge it 😄

src/pdf/DriverOverview/DriverOverviewPdf.tsx
```diff
@@ -118,7 +118,7 @@ const styles = StyleSheet.create({
     nameColumnWidth: {
         width: "15%",
     },
-    addressColumnWidth: {
+    addressPostcodeColumnWidth: {
```
AlexandraIrinaMardale: Is this just a rename?

src/components/Tables/ChecklistFilter.tsx
```diff
@@ -58,6 +60,9 @@ export const serverSideChecklistFilter = <
             const onChangeCheckbox = (event: React.ChangeEvent<HTMLInputElement>): void => {
                 const checkboxKey = event.target.name as string;
                 if (event.target.checked) {
+                    if (isRadio && anySelected()) {
+                        state.pop();
```
AlexandraIrinaMardale: Can we do ``setState([checkboxKey]);`` instead of popping please?

src/common/tableRows.tsx
```diff
@@ -0,0 +1,23 @@
...
+import ClientOutsideDeliveryAreaIcon from "@/components/Icons/ClientsOutsideDeliveryAreaIcon";
+import { formatNullPostcode } from "@/common/format";
+
+export const rowToAddressColumn = ({
+    postcode,
+    isDeliverable,
+}: ClientsTableRow["addressPostcode"] | ParcelsTableRow["addressPostcode"]): React.ReactElement => {
+    const postcodeRow: React.ReactNode[] = [];
+    postcodeRow.push(formatNullPostcode(postcode));
+    if (!isDeliverable && postcode !== "-") {
```
AlexandraIrinaMardale: Do we need to check for "-"? Is deliverable should be true only for clients with a null postcode or clients whose postcode is one that exists in the delivery areas, right? So "-" wouldn't match a deliverable postcode 🤔

src/common/format.ts
```diff
@@ -50,6 +52,10 @@ export const displayList = (data: string[]): string => {
     return data.length === 0 ? "None" : data.join(", ");
 };
 
+export const formatNullPostcode = (postcodeData: string | null): string => {
```
AlexandraIrinaMardale: Can we rename this to `getDisplayPostcode` since it doesn't handle just the null cases?

src/common/tableRows.tsx
AlexandraIrinaMardale, line 1: The name is a bit generic and doesn't really indicate what the file contains. Could we move it to `/components/Tables/` and rename it to `AddressColumnFormatter.tsx` please?

src/common/databaseFilters.ts
```diff
@@ -101,3 +103,35 @@ export const familySearch = <DbData extends DbClientRow | DbParcelRow>(
         return `and(${clientIsActiveColumnLabel}.is.true, ${familyCountColumnLabel}.eq.${substringAsNumber})`;
     });
 };
+
+export function deliveryAreaFilter(
```
AlexandraIrinaMardale, line 113: Shouldn't this be in `src/app/parcels/parcelsTable/filters.ts`?

AlexandraIrinaMardale: Ah nevermind, I noticed we use it in the clientsTable as well

src/common/databaseFilters.ts
```diff
@@ -101,3 +103,35 @@ export const familySearch = <DbData extends DbClientRow | DbParcelRow>(
...
+    deliverableColumnLabel: string,
+    clientIsActiveColumnLabel: string
+): ServerSideFilter<ParcelsTableRow, string[], DbParcelRow> {
+    const deliveryAreasSearch: ParcelsFilterMethod<string[]> = (query, state) => {
+        if (state.length === 0) {
+            return query;
+        }
+        return query.eq(clientIsActiveColumnLabel, true).in(deliverableColumnLabel, state);
```
AlexandraIrinaMardale: Do we need to check that client is active? 🤔

src/app/parcels/parcelsTable/convertParcelDBtoParcelRow.ts
```diff
@@ -73,8 +76,8 @@ const convertParcelDbtoParcelRow = async (
                     flaggedForAttention: parcel.flagged_for_attention ?? false,
                     requiresFollowUpPhoneCall: parcel.signposting_call_required ?? false,
                 },
-                createdAt: parcel.created_at ? new Date(parcel.created_at) : null,
                 clientIsActive: parcel.client_is_active ?? false,
+                createdAt: parcel.created_at ? new Date(parcel.created_at) : null,
```
AlexandraIrinaMardale: Why reorder these?

src/app/clients/form/submitFormHelpers.ts
```diff
@@ -1,14 +1,15 @@
...
-import { ClientFields } from "./ClientForm";
 import { AuditLog, sendAuditLog } from "@/server/auditLog";
-import { ListType } from "@/common/databaseListTypes";
+import supabase from "@/supabaseClient";
+import { ClientFields } from "./ClientForm";
 
 export type FamilyDatabaseInsertRecord = Omit<InsertSchema["families"], "family_id">;
 export type ClientDatabaseInsertRecord = InsertSchema["clients"];
 export type ClientDatabaseUpdateRecord = UpdateSchema["clients"];
+export type DeliveryAreaDatabaseInsertRecord = InsertSchema["delivery_areas"];
```
AlexandraIrinaMardale: You're not using this anymore and I think the whole file can be reverted to its original version

src/app/clients/form/formSections/AddressCard.tsx
```diff
@@ -1,19 +1,19 @@
...
 } from "@/components/Form/formFunctions";
-import GenericFormCard from "@/components/Form/GenericFormCard";
 import { GappedDiv } from "@/components/Form/formStyling";
+import GenericFormCard from "@/components/Form/GenericFormCard";
 import { ClientCardProps } from "../ClientForm";
-import { Checkbox, FormControlLabel } from "@mui/material";
 
 export const postcodeRegex =
-    /^([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?)))) [0-9][A-Za-z]{2})$/;
+    /^(([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?)))) [0-9][A-Za-z]{2}))$/;
```
AlexandraIrinaMardale: Can we move this one next to the prefix one in commmon/format.ts please?

src/app/clients/clientsTable/types.ts
```diff
@@ -38,10 +38,14 @@ export type GetClientsCountReturnType =
...
     familyCategory: string;
-    addressPostcode: string | null;
-    phoneNumber: string | null;
+    addressPostcode: {
+        postcode: Schema["clients"]["address_postcode"];
+        isDeliverable: boolean | null;
+    };
+    phoneNumber: Schema["clients"]["phone_number"];
     email: string | null;
+    isActive: boolean | null;
```
AlexandraIrinaMardale: Why did you add isActive?

src/app/clients/clientsTable/ClientsPage.tsx
```diff
@@ -220,7 +229,7 @@ const ClientsPage: React.FC = () => {
                             editableConfig={{ editable: false }}
                             isLoading={isLoading}
                             pointerOnHover={true}
-                            columnDisplayFunctions={{ addressPostcode: formatNullPostcode }}
+                            columnDisplayFunctions={clientTableColumnDisplayFunctions}
```
AlexandraIrinaMardale: Can we just do `columnDisplayFunctions={{ addressPostcode: rowToAddressColumn }}` and remove the `clientTableColumnDisplayFunctions`?

src/pdf/DriverOverview/DriverOverviewPdf.tsx
claudia-golaes-dev: Changed it to its previous name, sorry

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,351 @@
...
+                        setRows((prevRows) => {
+                            const unsavedLocalRows = prevRows.filter((row) => row.isNew);
+                            return [
+                                ...deliveryAreas,
+                                ...unsavedLocalRows.filter(
+                                    (localRow) =>
+                                        !deliveryAreas.some((dbRow) => dbRow.id === localRow.id)
+                                ),
+                            ];
+                        });
```
AlexandraIrinaMardale, line 122: What does this do?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,351 @@
...
+                    } catch (error) {
+                        setRows([]);
+                        if (error instanceof Error) {
+                            void logErrorReturnLogId(
+                                "Error with fetch: Delivery areas subscription",
+                                {},
+                                error
+                            );
+                            setErrorMessage("Error fetching data, please reload");
```
AlexandraIrinaMardale: Can you move this right after setRows?

src/common/tableRows.tsx
claudia-golaes-dev: This check is needed for displaying the warning icons, the postcode is "-" when a client is deleted

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,351 @@
...
+            const baseAuditLog = getBaseAuditLogForDeliveryAreasAction(
+                "add a new delivery area",
+                row
+            );
+
+            if (insertDeliveryAreasError) {
+                setErrorMessage(
+                    `The specified delivery area already exists. Log ID: ${insertDeliveryAreasError.logId}`
+                );
+                setRows((prevRows) => prevRows.filter((prevRow) => prevRow.id !== row.id));
```
AlexandraIrinaMardale: Let's not do this please - this causes the error to remain with no way to clear it aside from refreshing or adding another delivery area

We should also clear all errors when we click `Add new area` and when we click cancel when adding a row

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,351 @@
...
+                "Edited row in delivery areas admin table is undefined onCancelClick"
+            );
+            setErrorMessage("Table error, please try again");
+        } else if (editedRow.isNew) {
+            setRows((currentValue) => currentValue.filter((row) => row.id !== id));
+        }
+    };
+
+    const handleRemoveClick = (id: GridRowId) => () => {
+        const editedRow = rows.find((row) => row.id === id);
```
AlexandraIrinaMardale: Can we call this just row?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+    const deliveryAreasColumns: GridColDef[] = [
+        {
+            field: "postcode",
+            headerName: "Postcode",
+            headerAlign: "center",
+            flex: 1,
+            editable: true,
+            align: "center",
+            valueParser: (value) => value,
```
AlexandraIrinaMardale: Do we need this?

src/app/admin/deliveryAreasTable/DeliveryAreasActions.ts
```diff
@@ -0,0 +1,114 @@
...
+            error,
+            newDeliveryAreasData: data,
+        });
+        return { data: null, error: { dbError: error, logId } };
+    }
+
+    return { data: { deliveryAreasId: deliveryAreas.id }, error: null };
+};
+
+type UpdateDeliveryAreasResult = {
```
AlexandraIrinaMardale: Should this be called DeleteDeliveryAreasResult?

src/app/admin/deliveryAreasTable/DeliveryAreasActions.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (auditUpdateError) {
+        const logId = await logErrorReturnLogId(
+            "Failed to clear audit_log references before delete",
+            {
+                error: auditUpdateError,
+                rowData: row,
+            }
+        );
+        return { error: { dbError: auditUpdateError, logId } };
+    }
```
AlexandraIrinaMardale: I don't think we need to do this if you add on delete set null to the delivery areas column in the audit logs tale, see `20240801131704_add_wiki_id_column_to_audit_log_table.sql`

src/app/admin/deliveryAreasTable/DeliveryAreasActions.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    const { data, error } = await supabase
+        .from("delivery_areas")
+        .delete()
+        .eq("id", row.id)
+        .select();
```
AlexandraIrinaMardale: Delete won't return any data so the check for data below will fail, we should probably do it as it's done in `ListDataview.tsx`

cypress/e2e/admin/deliveryAreas.cy.ts
```diff
@@ -0,0 +1,219 @@
+describe("Delivery Areas on admins page", () => {
```
AlexandraIrinaMardale: Can we say "Edit delivery areas on admins page"?

cypress/e2e/admin/deliveryAreas.cy.ts
```diff
@@ -0,0 +1,219 @@
...
+};
+
+const orderDeliveryAreas = (): void => {
+    cy.get('div[aria-label="Delivery Areas Table"]') // eslint-disable-line quotes
+        .find('[data-testid="ArrowUpwardIcon"]') // eslint-disable-line quotes
+        .click({ force: true });
+};
+
+const deliveryAreas = (descending: boolean): void => {
+    Math.abs(0 - 15 * Number(descending));
```
AlexandraIrinaMardale: This line doesn't do anything 😅

cypress/e2e/admin/deliveryAreas.cy.ts
```diff
@@ -0,0 +1,219 @@
...
+    beforeEach(() => {
+        cy.login();
+        cy.intercept({ method: "GET", url: "**/rest/v1/delivery_areas*" }).as("getDeliveryAreas");
+        cy.visit("/admin");
+    });
+
+    it("Initial delivery areas", () => {
+        toggleDeliveryAreaNameSection();
+        deliveryAreas(false);
+    });
```
AlexandraIrinaMardale: It's not really clear what's happening here (and in a few other tests as well). Could you rewrite this to something like:
```
it("displays delivery areas in default ascending order", () => {
        toggleDeliveryAreaNameSection();
        verifyDeliveryAreasOrder("ascending");
    });
```

And I'll leave another comment on `deliveryAreas`

cypress/e2e/admin/deliveryAreas.cy.ts
```diff
@@ -0,0 +1,219 @@
...
+        .click();
+};
+
+const orderDeliveryAreas = (): void => {
+    cy.get('div[aria-label="Delivery Areas Table"]') // eslint-disable-line quotes
+        .find('[data-testid="ArrowUpwardIcon"]') // eslint-disable-line quotes
+        .click({ force: true });
+};
+
+const deliveryAreas = (descending: boolean): void => {
```
AlexandraIrinaMardale: As I said above, the name should reflect what it's doing. And there's a better way to check this than using magic numbers :)
```
function verifyDeliveryAreasOrder(order: "ascending" | "descending"): void {
    const expectedPostcodes = [
        "CR0", "CR7", .....
    ];
    
    const orderedPostcodes = order === "descending" 
        ? [...expectedPostcodes].reverse() 
        : expectedPostcodes;

    orderedPostcodes.forEach((postcode, index) => {
        assertDeliveryAreaAtRow({ rowIndex: index, deliveryAreaName: postcode });
    });
}

function assertDeliveryAreaAtRow({ rowIndex, deliveryAreaName }: { rowIndex: number; deliveryAreaName: string }): void {
    cy.contains(editDeliveryAreaNameText, { matchCase: false })
        .parents(".MuiPaper-root")
        .find(`[data-rowindex="${rowIndex}"]`)
        .find('[data-field="postcode"]')
        .should("contain.text", deliveryAreaName);
}
```

or 
```
function verifyDeliveryAreasAreSorted(order: "ascending" | "descending"): void {
    cy.get('div[aria-label="Delivery Areas Table"]')
        .find('[data-field="postcode"] .MuiDataGrid-cellContent')
        .then($postcodes => {
            const actualPostcodes = Array.from($postcodes).map(el => el.textContent?.trim());
            const sortedPostcodes = [...actualPostcodes].sort();
            
            if (order === "descending") {
                sortedPostcodes.reverse();
            }
            
            expect(actualPostcodes).to.deep.equal(sortedPostcodes);
        });
}
```

AlexandraIrinaMardale (approved):

src/app/parcels/parcelsTable/convertParcelDBtoParcelRow.ts
claudia-golaes-dev: My bad

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
claudia-golaes-dev: It's used to prevent concurrency issues. If two users try to add delivery areas at the same time, and the first user presses save, if I would have used `setRows(deliveryAreas)`, the second user would have lost his unsaved rows

src/app/clients/clientsTable/types.ts
claudia-golaes-dev: Forgot it there, we used it at first for the NFA's, but we don't need it anymore, deleted it

src/common/format.ts
```diff
@@ -22,6 +22,12 @@ export const phoneNumberRegex = /^((0|\+44)\d{9,11}|\+(?!44)\d{7,15})?$/;
 
 export const emailFormatSymbolsRegex = /[\s]/g;
 export const emailRegex = /^\S+@\S+$/;
+export const prefixPostcodeRegex =
+    /^(([Gg][Ii][Rr])|(([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?)))))$/;
+export const postcodeRegex =
+    /^(([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?)))) [0-9][A-Za-z]{2}))$/;
+// Regex source: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/488478/Bulk_Data_Transfer_-_additional_validation_valid_from_12_November_2015.pdf
```
AlexandraIrinaMardale, line 26: Can you move the comment before the 2 regexes?

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
AlexandraIrinaMardale: Ok that makes sense, thank you

src/app/admin/deliveryAreasTable/DeliveryAreasTable.tsx
```diff
@@ -0,0 +1,327 @@
...
+    postcode: string;
+    isNew: boolean;
+}
+
+const isValidPostcode = (value: string): boolean => prefixPostcodeRegex.test(value);
+
+function EditToolbar(props: EditToolbarProps): React.JSX.Element {
+    const { setRows, setRowModesModel } = props;
+    // The id was initialized as `number of rows+1`, which raised some issues and now it is generated randomly below
+    const handleClick = (): void => {
```
AlexandraIrinaMardale, line 53: We should clear errors when we click add a new area

supabase/migrations/20250905161958_add_delivery_areas_id_as_fk_to_audit_log.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."audit_log" add column "delivery_areas_id" uuid;
+
+alter table "public"."audit_log" add constraint "audit_log_delivery_areas_id_fkey" FOREIGN KEY (delivery_areas_id) REFERENCES delivery_areas(id) ON DELETE SET NULL not valid;
+
+alter table "public"."audit_log" validate constraint "audit_log_delivery_areas_id_fkey";
```
AlexandraIrinaMardale: New line

src/app/admin/deliveryAreasTable/DeliveryAreasActions.ts
```diff
@@ -0,0 +1,94 @@
...
+    row: DeliveryAreasRow
+): Promise<DeleteDeliveryAreasResult> => {
+    const auditLog = {
+        action: "delete a delivery area",
+        content: {
+            deliveryArea: row.postcode,
+        },
+    } as const satisfies Partial<AuditLog>;
+
+    const { error } = await supabase.from("delivery_areas").delete().eq("id", row.id).select();
```
AlexandraIrinaMardale: You don't need the `select`

alexandrutrifu-urzica (approved):

src/common/format.ts
```diff
@@ -23,6 +23,13 @@ export const phoneNumberRegex = /^((0|\+44)\d{9,11}|\+(?!44)\d{7,15})?$/;
 export const emailFormatSymbolsRegex = /[\s]/g;
 export const emailRegex = /^\S+@\S+$/;
 
+// Regex source: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/488478/Bulk_Data_Transfer_-_additional_validation_valid_from_12_November_2015.pdf
+export const prefixPostcodeRegex =
+    /^(([Gg][Ii][Rr])|(([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?)))))$/;
+export const postcodeRegex =
+    /^(([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?)))) [0-9][A-Za-z]{2}))$/;
+// The regex has been updated to have a mandatory whitespace in between the outward and inward code
```
AlexandraIrinaMardale: This one too

PR #773, opened 2025-08-04
VFB-446: Update style for checkbox icon

AlexandraIrinaMardale:
- if we set the height then the row won't grow to fit the text
- updated the checkbox to reduce the padding so that the rows can be 48px when the cell content fits on one line


claudia-golaes-dev (commented):
Looks good to me, but can the "Select all rows" checkbox have the same padding?

claudia-golaes-dev:
> Looks good to me, but can the "Select all rows" checkbox have the same padding?

I'm asking this, because if it doesn't have the same padding set, the "Select all rows" checkbox is not aligned with the others

AlexandraIrinaMardale:
image

claudia-golaes-dev (approved):
looks good!

PR #776, opened 2025-08-06
Vfb 395 fix bugs

LeonardCiobanu-Softwire:
## What's changed

- Added custom regex for filters and fixed the email search functionality
- Added security invoker to the clients_plus view

## Checklist

### DIFF
--- a/src/common/databaseFilters.ts
+++ b/src/common/databaseFilters.ts
@@ -4,13 +4,14 @@ import { DbClientRow, DbParcelRow } from "@/databaseUtils";
 import { parcelsPageDeletedClientDisplayName } from "@/app/parcels/parcelsTable/format";
 
 const textFilterDelimiter = ",";
+const defaultQueryFilterRegex = /[^a-zA-Z0-9 '\-+?]/g;
+const emailQueryFilterRegex = /[^a-zA-Z0-9 '\-@.+?]/g;
 
 export const dbFilterWithSubstringQueries = <DbData extends DbClientRow | DbParcelRow>(
     substringToSubqueryMap: (value: string) => string,
-    customRegex?: RegExp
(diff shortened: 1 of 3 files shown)


supabase/migrations/20250729082705_add_email_field_to_clients_table_and_view.sql
```diff
@@ -1,6 +1,6 @@
 alter table "public"."clients" add column "email" text;
-create or replace view
-  public.clients_plus as
+create or replace view 
+  public.clients_plus with (security_invoker = true) as
```
AlexandraIrinaMardale: Can you keep these 2 on the same line please?

src/common/databaseFilters.ts
```diff
@@ -70,7 +72,7 @@ export const emailSearch = <DbData extends DbClientRow | DbParcelRow>(
             return `${clientIsActiveColumnLabel}.is.false`;
         }
         return `and(${clientIsActiveColumnLabel}.is.true, ${emailColumnLabel}.ilike.%${substring.toLowerCase()}%)`;
-    });
+    }, /[^a-zA-Z0-9 '\-@.+?]/g);
```
AlexandraIrinaMardale: If we just update the original regex to this, will it have a negative effect on any of the searches?

LeonardCiobanu-Softwire: Yes, the dot would create some problems, as it is a reserved character.

AlexandraIrinaMardale: Ah ok that makes sense. Then can you maybe have `defaultRegex` and `emailRegex` at the top of the file? And use `defaultRegex` as the default for `customRegex` in `dbFilterWithSubstringQueries` please? 😁

AlexandraIrinaMardale (approved):

PR #777, opened 2025-08-06
VFB-450: Include new fields for lists

mihaicutaiulia:
# [VFB-450 - Include new fields for item category and availability in lists](https://softwiretech.atlassian.net/jira/software/c/projects/VFB/boards/1130?assignee=712020%3A131ca6b3-6194-4848-a5db-d4e4a405e93e&selectedIssue=VFB-450)

## What's changed
- Added new columns `is_available` (boolean) and `item_type` (enum) to the lists table via database migration.
- Made `Description` and `Item Type` required fields
- Defined `item_type` enum in the database and corresponding TypeScript types and labels
- Updated `EditModal` to allow editing `item_type` (dropdown) and `is_available` (checkbox) when editing or creating a list item.
- Extended `ListDataview` and `ListsPage` to display and filter by `item_type` and `is_available`, with new columns and filters.
- Refactored filter types to support array/string filter states, including a new `clientSideChecklistFilter` and `clientSideDropdownFilter`

❗ The functionality to edit availability for more items has not been implemented yet due to difficulties caused by the table structure

### Before
Screenshot 2025-08-14 170835

### After
https://github.com/user-attachments/assets/44300ff2-e0e3-4fa3-983b-4af74fd5ab47

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/lists/ListDataview.tsx
```diff
@@ -96,15 +103,28 @@ export const listRowToListDB = (listRow: ListRow): Schema["lists"] => ({
...
+};
+
 const listDataViewColumnDisplayFunctions = {
+    is_available: (value: boolean) => displayBoolean(value),
+    item_type: (value: ItemType) => itemTypeLabels[value] ?? value,
     ...Object.fromEntries(
-        listsHeaderKeysAndLabels.slice(1).map(([key]) => [key, displayQuantityAndNotes])
+        listsHeaderKeysAndLabels
+            .slice(1)
+            .filter(([key]) => !["is_available", "item_type"].includes(key))
```
LeonardCiobanu-Softwire: If I'm not mistaken, just using .slice(3) would have the same effect as the slice and filter. Can you clarify what these lines do, please?

src/app/lists/ListDataview.tsx
```diff
@@ -262,10 +282,20 @@ const ListsDataView: React.FC<ListDataViewProps> = ({
...
+        filter: ListFilter
+    ): filter is DistributeClientFilter<ListRow, string[]> => Array.isArray(filter.state);
+
     useEffect(() => {
         setListData(
             listOfIngredients.filter((row) => {
                 return primaryFilters.every((filter) => {
+                    // without this if we have the following error:
+                    // Argument of type string | string[] is not assignable to parameter of type string & string[]
+                    if (isArrayStateFilter(filter)) {
```
LeonardCiobanu-Softwire, line 292: I think the problem might come from the filter being on the client side... I'm not sure.
Either way the intersection type string & string[] is an impossible one to have, as nothing can be string and string[] at the same time.
I think this should be looked into as it may produce some problems/bugs down the line.

src/app/lists/ListsPage.tsx
```diff
@@ -83,6 +88,25 @@ const formatListData = (listsData: Schema["lists"][]): ListRow[] => {
     );
 };
 
+const filterRowByAvailability = <Data,>(row: Data, state: string, rowKey?: keyof Data): boolean => {
+    if (state === "all" || !rowKey) {
+        return true;
+    }
+    const desired = state === "true";
+    return (row as never)[rowKey] === desired;
```
LeonardCiobanu-Softwire: What does this line do?

src/app/lists/ListsPage.tsx
```diff
@@ -83,6 +88,25 @@ const formatListData = (listsData: Schema["lists"][]): ListRow[] => {
...
+};
+
+const filterRowByItemTypes = <Data,>(row: Data, state: string[], rowKey?: keyof Data): boolean => {
+    if (!rowKey || state.length === 0) {
+        return true;
+    }
+    const value = (row as never)[rowKey] as string | undefined;
```
LeonardCiobanu-Softwire: Same question as above

src/app/lists/ListsPage.tsx
```diff
@@ -99,6 +123,30 @@ const filters: ListFilter[] = [
...
+        initialSelected: "all",
+        method: filterRowByAvailability,
+    }),
+    clientSideChecklistFilter({
+        key: "itemType",
+        rowKey: "item_type",
+        filterLabel: "Item Type",
+        itemLabelsAndKeys: [
+            ...Object.entries(itemTypeLabels).map(
+                ([key, label]) => [label, key] as [string, string]
```
LeonardCiobanu-Softwire: Why is this necessary?

src/components/Tables/ChecklistFilter.tsx
```diff
@@ -21,6 +24,89 @@ interface ChecklistFilterProps<
...
+    initialCheckedKeys: string[];
+    method: ClientSideFilterMethod<Data, string[]>;
+    shouldPersistOnClear?: boolean;
+    isDisabled?: boolean;
+    isHidden?: boolean;
+    isHiddenInUrl?: boolean;
+}
+
+const areStringArrayStatesIdentical = (stateA: string[], stateB: string[]): boolean =>
+    stateA.length === stateB.length && stateA.every((optionA) => stateB.includes(optionA));
```
LeonardCiobanu-Softwire: I don't think this works that well, Example:
stateA = [a, b, c, d, e, e, e]
stateB = [a, b, c, d, e, f , g]
In this case, stateA.length === stateB.length, and stateA.every(optionA) => stateB.includes(optionA)

I think you should check stateA.length === stateB.length, stateA.every(optionA) => stateB.includes(optionA), and stateB.every(optionB) => stateA.includes(optionB).

src/components/Tables/DropdownFilter.tsx
```diff
@@ -0,0 +1,85 @@
...
+    isHidden = false,
+    isHiddenInUrl = false,
+}: DropdownFilterProps<Data>): ClientSideFilter<Data, string> => {
+    return {
+        key: key,
+        rowKey: rowKey,
+        state: initialSelected,
+        initialState: initialSelected,
+        method: method,
+        areStatesIdentical: (stateA, stateB) => stateA === stateB,
```
LeonardCiobanu-Softwire, line 40: I think you forgot to use the function above (I left a comment on it)

LeonardCiobanu-Softwire (commented):
Looks good. Left some comments 😄

src/app/lists/ListDataview.tsx
mihaicutaiulia: The table used on the list page is a `ClientPaginatedTable`, so all filtering needs to happen on the client side.

src/app/lists/ListsPage.tsx
mihaicutaiulia: I've specified the return type, so the `as [string, string]` is no longer necessary.

src/components/Tables/ChecklistFilter.tsx
mihaicutaiulia: You're right — the original condition can return false positives in cases like this. I've updated the function to avoid that.

src/components/Tables/DropdownFilter.tsx
mihaicutaiulia: No, dropdown filters can only hold a single state at a time. The function you're referring to is used for filters like checkboxes, where multiple states are allowed.

src/app/lists/ListsPage.tsx
mihaicutaiulia: We return `true` if the current row has the expected value in the `rowKey` column — in our case, the availability column; otherwise, we return `false`. i removed the `as never`

src/app/lists/ListsPage.tsx
mihaicutaiulia: We retrieve the value from the row at the column specified by `rowKey`.

LeonardCiobanu-Softwire (approved):
👍

PR #780, opened 2025-08-11
VFB-396: Add Referral Columns To Parcels Page

mihaicutaiulia:
## What's changed
- Added referral fields as optional columns in the table
- Introduced "Referral Details" as a group toggle header
- Updated `toggleableHeaders` type from `readonly (keyof Data)[]` to `readonly (keyof Data | string)[]`

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image| image |

## Checklist

### DIFF
--- a/src/components/Tables/ColumnTogglePopup.tsx
+++ b/src/components/Tables/ColumnTogglePopup.tsx
@@ -31,13 +31,13 @@ const ColumnTogglePopup = <Data,>({
         "referrerName",
         "referrerEmail",
         "referrerPhone",
-    ] as const;
+    ];
 
     const onChangeCheckbox = (event: React.ChangeEvent<HTMLInputElement>): void => {
         const checkboxKey = event.target.name;
         const isReferralDetails = checkboxKey === "Referral Details";
         const keysToUpdate = isReferralDetails
-            ? (referralDetails as unknown as (keyof Data)[])
+            ? (referralDetails as (keyof Data)[])
             : [checkboxKey as keyof Data];
 
         const newKeys = event.target.checked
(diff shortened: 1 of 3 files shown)


src/components/Tables/ColumnTogglePopup.tsx
```diff
@@ -39,7 +53,7 @@ const ColumnTogglePopup = <Data,>({
                 labelsAndKeys={(toggleableHeaders ?? []).map((key) => {
                     const headerLabel =
                         headers.find(([headerKey]) => headerKey === key)?.[1] ?? key.toString();
-                    return [headerLabel, key as string];
+                    return [headerLabel, key.toString()];
```
claudia-golaes-dev: Why did you change here? Is there any possibility that the key is not a string?

src/components/Tables/ColumnTogglePopup.tsx
```diff
@@ -24,13 +24,27 @@ const ColumnTogglePopup = <Data,>({
...
-        const checkboxKey = event.target.name as keyof Data;
-        if (event.target.checked) {
-            setShownHeaderKeys([...shownHeaderKeys, checkboxKey]);
-        } else {
-            setShownHeaderKeys(shownHeaderKeys.filter((shownKey) => shownKey !== checkboxKey));
-        }
+        const checkboxKey = event.target.name;
+        const isReferralDetails = checkboxKey === "Referral Details";
+        const keysToUpdate = isReferralDetails
+            ? (referralDetails as unknown as (keyof Data)[])
```
claudia-golaes-dev: Is there any way to avoid the `unknown` method? 🥹

claudia-golaes-dev (commented):
it looks very good! 👍

src/components/Tables/ColumnTogglePopup.tsx
mihaicutaiulia: I initially used `key.toString()` because `as string` wasn't supported in an intermediate implementation. I’ve now reverted it since I see that `key as string` works correctly in the final version. Thank you!

src/components/Tables/ColumnTogglePopup.tsx
mihaicutaiulia: Yes, I made the change. Thank you😄

AlexandraIrinaMardale (approved):
👍

PR #786, opened 2025-08-14
Vfb 395 email filter regex fix

LeonardCiobanu-Softwire:
## What's changed

- Added all characters that can be in an email as searchable, except '_' and '%', which can cause query issues

- Only permitted characters can now be entered in an email address

## Checklist

### DIFF
--- a/src/common/databaseFilters.ts
+++ b/src/common/databaseFilters.ts
@@ -5,7 +5,7 @@ import { parcelsPageDeletedClientDisplayName } from "@/app/parcels/parcelsTable/
 
 const textFilterDelimiter = ",";
 const defaultQueryFilterRegex = /[^a-zA-Z0-9 '\-+?]/g;
-const emailQueryFilterRegex = /[^a-zA-Z0-9 '\-@.+?]/g;
+const emailQueryFilterRegex = /[^a-zA-Z0-9+-_~!#$%&`./=^'{}|@]/g;
 
 export const dbFilterWithSubstringQueries = <DbData extends DbClientRow | DbParcelRow>(
     substringToSubqueryMap: (value: string) => string,
--- a/src/common/databaseFilters.ts
+++ b/src/common/databaseFilters.ts
@@ -5,7 +5,7 @@ import { parcelsPageDeletedClientDisplayName } from "@/app/parcels/parcelsTable/
 
 const textFilterDelimiter = ",";
 const defaultQueryFilterRegex = /[^a-zA-Z0-9 '\-+?]/g;
-const emailQueryFilterRegex = /[^a-zA-Z0-9+-_~!#$%&`./=^'{}|@]/g;
+const emailQueryFilterRegex = /[^a-zA-Z0-9+-~!#$&`./=^'{}|@]/g;
 
 export const dbFilterWithSubstringQueries = <DbData extends DbClientRow | DbParcelRow>(
     substringToSubqueryMap: (value: string) => string,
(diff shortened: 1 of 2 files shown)


DanielaNegrut (commented):
What is the context for these changes? What is the before and after usage?

LeonardCiobanu-Softwire:
> What is the context for these changes? What is the before and after usage?

There were some characters that the email regex would replace with an empty string. Example:

- Say someone had the email "example+please@example.com"

- If you would want to search for the person that had that email, until entering "example+", you would see the result, but if you typed "example+p", all results would disappear.

- This is because the code would replace all characters except the ones in the 'emailQueryFilterRegex' with an empty string. And so when searching "example+p", what would end up being searched is "examplep".

- My change makes it so any characters that would be allowed in an email address would not be replaced with an empty string. In the above mentioned example, the search would now be "example+p", which is what it should be.

Alongside this, I also added the rule that only one '@' symbol may exist in an email address. This is one of the most basic rules of an email address's format and I believed it necessary.

src/common/databaseFilters.ts
```diff
@@ -5,7 +5,7 @@ import { parcelsPageDeletedClientDisplayName } from "@/app/parcels/parcelsTable/
 
 const textFilterDelimiter = ",";
 const defaultQueryFilterRegex = /[^a-zA-Z0-9 '\-+?]/g;
-const emailQueryFilterRegex = /[^a-zA-Z0-9 '\-@.+?]/g;
+const emailQueryFilterRegex = /[^a-zA-Z0-9+-_~!#$%&`./=^'{}|@]/g;
```
mihaicutaiulia: Since we now allow so many characters in the query, is filtering still necessary?

LeonardCiobanu-Softwire: I believe it is, as some characters that can affect supabase via query are still not allowed

DanielaNegrut (commented):
I am curios why there are 2 different regex that to me seem to do pretty much the same thing? :?

LeonardCiobanu-Softwire:
> I am curios why there are 2 different regex that to me seem to do pretty much the same thing? :?

They don't really do the same thing, `emailQueryFilterRegex` filters what characters are permitted in a search, that will be queried to supabase; Meanwhile `emailRegex` dictates what is considered a correct email address.

src/common/format.ts
```diff
@@ -20,7 +20,7 @@ export const phoneNumberFormatSymbolsRegex = /[\s-()]/g;
 export const phoneNumberRegex = /^((0|\+44)\d{9,11}|\+(?!44)\d{7,15})?$/;
 
 export const emailFormatSymbolsRegex = /[\s]/g;
-export const emailRegex = /^\S+@\S+$/;
+export const emailRegex = /^[a-zA-Z0-9+\-_~!#$%&`./=^'{}|]+@[a-zA-Z0-9+\-_~!#$%&`./=^'{}|]+$/;
```
AlexandraIrinaMardale: Let's keep this as it was before. The official RFC 5322 email regex is much more complicated and we risk rejecting valid emails.

src/common/databaseFilters.ts
AlexandraIrinaMardale: Which characters are these?

LeonardCiobanu-Softwire:
I have removed the '_' and '%' symbols from the search, as they may cause problems in the query; And reverted the formatting to the original regex.

AlexandraIrinaMardale (approved):
Looks good,  thank you!

Next time, can you please re-request a review when you're done solving markups? 😁

PR #788, opened 2025-08-18
Vfb 427 postcode sorting order

anamariasuditu:
## What's changed
Postcode sorting should now respect numerical ordering

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| Screenshot 2025-08-19 102637 | Screenshot 2025-08-19 102835 |
| Screenshot 2025-08-19 102144 | Screenshot 2025-08-18 182427 |

## Checklist

    - Checkout the dev branch
    - Checkout your branch


supabase/migrations/20250818135126_make_function_sort_postcodes.sql
mihaicutaiulia, line 1: maybe it would be worth updating this migration to follow the same structure/format as the existing ones.

supabase/migrations/20250818141321_add_sorted_postcodes_column_to_clients_plus.sql
```diff
@@ -0,0 +1,16 @@
...
+select
+    clients.primary_key as client_id,
+    clients.full_name,
+    clients.address_postcode,
+    clients.phone_number,
+    clients.is_active,
+    family_count.family_count,
+    clients.email,
+    sort_address_postcode(clients.address_postcode) AS sorted_address_postcode
+from
```
mihaicutaiulia: Are all of these columns necessary, or could some of them be removed?

mihaicutaiulia (approved):
LGTM 👍

supabase/migrations/20250818141321_add_sorted_postcodes_column_to_clients_plus.sql
anamariasuditu: > Are all of these columns necessary, or could some of them be removed?

When adding a new column to a view, I have to use `create or replace view`, so I just added my new column to all the already existing ones.

supabase/migrations/20250818135126_make_function_sort_postcodes.sql
anamariasuditu: > maybe it would be worth updating this migration to follow the same structure/format as the existing ones.

From what I've seen, most migrations are wrote in upper case

supabase/migrations/20250818135126_make_function_sort_postcodes.sql
```diff
@@ -0,0 +1,38 @@
+set check_function_bodies = off;
+
+CREATE OR REPLACE FUNCTION public.sort_address_postcode(pc text)
```
AlexandraIrinaMardale: This doesn't actually sort the postcodes, it creates a sortable key from a postcode - we should rename this to something like `postcode_to_sort_key` or `create_postcode_sort_key`

AlexandraIrinaMardale: Make sure to update the name of this migration (+ the timestamp) and use the correct function in the next migration (+ update its timestamp)

supabase/migrations/20250903092842_add_sorted_postcodes_column_to_clients_plus.sql
```diff
@@ -0,0 +1,16 @@
+create or replace view
+  public.clients_plus as
```
AlexandraIrinaMardale: Can you add with (security_invoker = true) here as well please?

AlexandraIrinaMardale (commented):
You'll need to update the timestamps and the views after Claudia's merges (436, 414)

AlexandraIrinaMardale (approved):

PR #791, opened 2025-08-19
VFB-449 Fix inconsistent behaviour

mihaicutaiulia:
### [VFB-449: Fix inconsistent behaviour and errors when selecting deleted client parcels](https://softwiretech.atlassian.net/browse/VFB-449)

## What's changed
- For **Download Shopping Lists**, **Download Shipping Labels**, **Generate Map**, and **Download Driver Overview**:
  - Parcels associated with inactive (deleted) clients are now consistently excluded from the output.
  - An info/warning message is displayed to the user indicating how many selected parcels are associated with deleted clients and will be excluded.
  - The action proceeds with active client parcels only (does not block or error if some selected parcels are deleted).

- For **Download Day Overview**, **Download Signposting Report**, **Download Missing Voucher Report**, **Download Selected Parcels Report**, and **Download Pending More Info Report**:
  - These reports now include parcels from both active and deleted clients (no exclusion based on client status).

- Refactored and unified the logic for counting and handling deleted client parcels, using a new utility and warning component for relevant actions.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |
|  | image |
| image | image |
| image | image |

## Checklist

### DIFF
--- a/src/app/parcels/ActionBar/ActionButtons/DriverOverview/getDriverOverviewData.ts
+++ b/src/app/parcels/ActionBar/ActionButtons/DriverOverview/getDriverOverviewData.ts
@@ -86,7 +86,6 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
 
     const dataWithNonNullClients: ParcelForDelivery[] = [];
     for (const parcel of data) {
-        // do we still need to check for null clients?
         if (parcel.client === null) {
             const logId = await logErrorReturnLogId(
                 "Error with fetch: Parcels. No matching client found"
--- a/src/app/parcels/ActionBar/ActionButtons/DriverOverview/getDriverOverviewData.ts
+++ b/src/app/parcels/ActionBar/ActionButtons/DriverOverview/getDriverOverviewData.ts
@@ -45,7 +45,11 @@ type ParcelsForDeliveryResponse =
           error: { type: ParcelsForDeliveryErrorType; logId: string };
       };
 
-type ParcelsForDeliveryErrorType = "parcelFetchFailed" | "noMatchingClient" | "noCollectionCentre";
+type ParcelsForDeliveryErrorType =
+    | "parcelFetchFailed"
+    | "noMatchingClient"
+    | "noCollectionCentre"
+    | "noActiveParcels";
 
 type DriverPdfResponse =
     | {
@@ -82,6 +86,7 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
 
     const dataWithNonNullClients: ParcelForDelivery[] = [];
     for (const parcel of data) {
+        // do we still need to check for null clients?
         if (parcel.client === null) {
             const logId = await logErrorReturnLogId(
                 "Error with fetch: Parcels. No matching client found"
@@ -94,6 +99,10 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
                 "Error with fetch: Parcels. No collection centre found"
             );
             return { data: null, error: { type: "noCollectionCentre", logId: logId } };
+        }
+
+        if (!parcel.client.is_active) {
+            continue;
         }
 
         const sortedParcelEvents = parcel.events.sort((event1, event2) => {
@@ -118,6 +127,11 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
             collection_centre: parcel.collection_centre,
             labelCount: Number.isNaN(labelCount) ? null : labelCount,
         });
+    }
+
+    if (dataWithNonNullClients.length === 0) {
+        const logId = await logErrorReturnLogId("All selected parcels belong to deleted clients.");
+        return { data: null, error: { type: "noActiveParcels", logId: logId } };
     }
 
     return { data: dataWithNonNullClients, error: null };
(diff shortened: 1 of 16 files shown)


src/app/parcels/ActionBar/ActionButtons/ShoppingList/getShoppingListData.ts
```diff
@@ -71,6 +71,10 @@ type FetchShoppingListForPdfResponse =
     | {
           data: null;
           error: ShoppingListPdfError;
+      }
+    | {
+          data: null;
+          error: null;
```
stefanghenescu-work, line 77: Does it make sense to have this new type just for null? Can you try and refactor this. It is possible to have `| null` for error as an example.

src/app/parcels/ActionBar/ActionModals/ShippingLabelModal.tsx
```diff
@@ -191,6 +204,10 @@ const ShippingLabelModal: React.FC<ActionModalProps> = (props) => {
         );
     }, [parcelIds]);
 
+    useEffect(() => {
```
stefanghenescu-work: I see there is another `useEffect` that has the same dependency. Do you think it would be better to have them separate or in a single use effect?

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
```diff
@@ -128,14 +138,18 @@ const ShoppingListModal: React.FC<ActionModalProps> = (props) => {
     };
 
     useEffect(() => {
-        getDuplicateDownloadedPostcodes(
+        void getDuplicateDownloadedPostcodes(
```
stefanghenescu-work, line 141: I see that you modified this as a warning was raised otherwise. Same thing happens at line 199 in `ShippingLabelModal.tsx` if you want to take a look.

src/app/parcels/ActionBar/ActionModals/ShoppingListModal.tsx
```diff
@@ -128,14 +138,18 @@ const ShoppingListModal: React.FC<ActionModalProps> = (props) => {
...
             "Shopping List Downloaded",
             setDuplicateDownloadedPostcodes,
             setErrorMessage
         );
     }, [parcelIds]);
 
+    useEffect(() => {
+        void getDeletedClientParcelsCount(parcelIds, setDeletedClientParcelsCount, setErrorMessage);
+    }, [parcelIds]);
+
```
stefanghenescu-work: Same thing with the `useEffect` dependency.

src/app/parcels/ActionBar/ActionModals/getNumberOfDeletedClientParcels.ts
stefanghenescu-work, line 1: Why is it necessary to have a new file just for this? Would it be better to have a general utility file?

src/app/parcels/ActionBar/ActionButtons/ShoppingList/getShoppingListData.ts
mihaicutaiulia: I chose this approach because it more clearly indicates that returning `{ data: null, error: null }` represents a special case. Let me know if you'd still like me to change it

src/app/parcels/ActionBar/ActionModals/ShippingLabelModal.tsx
mihaicutaiulia: I'll merge them into one `useEffect` hook

src/app/parcels/ActionBar/ActionModals/getNumberOfDeletedClientParcels.ts
mihaicutaiulia: I wanted to maintain the same format as `getDuplicateDownloadedPostcodes.ts`, but I can combine both constants into a single file if preferred.

stefanghenescu-work: If there is already a format, we can keep it like this.

src/app/parcels/ActionBar/ActionButtons/ShoppingList/getShoppingListData.ts
stefanghenescu-work: Okay, let's keep it separate so the special case can be viewed more clearly.

src/app/parcels/ActionBar/ActionButtons/DriverOverview/getDriverOverviewData.ts
```diff
@@ -82,6 +86,7 @@ const getParcelsForDelivery = async (parcelIds: string[]): Promise<ParcelsForDel
 
     const dataWithNonNullClients: ParcelForDelivery[] = [];
     for (const parcel of data) {
+        // do we still need to check for null clients?
```
AlexandraIrinaMardale: What's up with this? 😄

AlexandraIrinaMardale (commented):
There's some conflicts with intern-dev

src/app/parcels/ActionBar/ActionButtons/DriverOverview/getDriverOverviewData.ts
alexandrutrifu-urzica: I removed the comment, technically it shouldn't be necessary to check for non-null clients (because they are never entirely deleted from the database), but I left the if-statement there just to be safe.

PR #792, opened 2025-08-19
VFB-484: Fix autofill issue for family members

anamariasuditu:
## What's changed
Now, auto-fill saves properly the Number of Adults and Number of Children fields input when adding a client.
Updated the Family Size display logic on the Parcels/Clients page.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| https://github.com/user-attachments/assets/07796b90-7efb-4339-a6a4-177b299290d9 <br> Screenshot 2025-08-19 170236 Screenshot 2025-08-19 172328 | https://github.com/user-attachments/assets/853780d7-68ff-44ff-ae72-f0402817e4f3 <br> Screenshot 2025-08-19 170346 Screenshot 2025-08-19 172300 |

## Checklist

### DIFF
--- a/src/app/clients/form/formSections/NumberAdultsCard.tsx
+++ b/src/app/clients/form/formSections/NumberAdultsCard.tsx
@@ -26,6 +26,16 @@ const numberOfAdultsRange = (value: string): boolean => {
     );
 };
 
+export const resizePersonsArray = (current: Person[], target: number): Person[] => {
+    if (current.length === target) {
+        return current;
+    }
+
+    return current.length < target
+        ? [...current, ...Array(target - current.length).fill({} as Person)]
+        : current.slice(0, target);
+};
+
 const setAdultsFields = (
     fieldSetter: ClientSetter,
     adults: Person[],
@@ -55,16 +65,12 @@ const NumberAdultsCard: React.FC<ClientCardProps> = ({
     fields,
 }) => {
     useEffect(() => {
-        if (fields.numberOfAdults > fields.adults.length) {
-            const newAdults = [...fields.adults];
-            while (newAdults.length < fields.numberOfAdults) {
-                newAdults.push({});
-            }
-            fieldSetter({ adults: newAdults });
-        } else if (fields.numberOfAdults < fields.adults.length) {
-            fieldSetter({ adults: fields.adults.slice(0, fields.numberOfAdults) });
+        const resized = resizePersonsArray(fields.adults, fields.numberOfAdults);
+        if (resized !== fields.adults) {
+            fieldSetter({ adults: resized });
         }
     }, [fields.numberOfAdults, fields.adults, fieldSetter]);
+
     return (
         <GenericFormCard
             title="Number of Adults"
@@ -87,11 +93,9 @@ const NumberAdultsCard: React.FC<ClientCardProps> = ({
                         additionalCondition: numberOfAdultsRange,
                     })}
                 />
-                {fields.adults.map((adult: Person, index: number) => {
-                    if (index >= fields.numberOfAdults) {
-                        return null;
-                    }
-                    return (
+                {fields.adults
+                    .slice(0, fields.numberOfAdults)
+                    .map((adult: Person, index: number) => (
                         <StyledCard key={adult.primaryKey ?? `new-adult-${index}`}>
                             <FormText>Adult {index + 1}</FormText>
                             <ControlledSelect
@@ -122,8 +126,7 @@ const NumberAdultsCard: React.FC<ClientCardProps> = ({
                                 )}
                             />
                         </StyledCard>
-                    );
-                })}
+                    ))}
             </>
         </GenericFormCard>
     );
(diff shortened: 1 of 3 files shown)


src/app/clients/getExpandedClientDetails.ts
```diff
@@ -82,7 +82,11 @@ const getRawClientDetails = async (clientId: string) => {
 };
 
 export const familyCountToFamilyCategory = (count: number): string => {
-    if (count <= 1) {
+    if (count == 0) {
+        return "No family";
+    }
```
claudia-golaes-dev, line 83: Maybe here you can either omit the word "Occupant" and keep "No family" or just 
`if (count == 0) {
        return "No ";
    }`

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -82,7 +82,11 @@ const getRawClientDetails = async (clientId: string) => {
 };
 
 export const familyCountToFamilyCategory = (count: number): string => {
-    if (count <= 1) {
+    if (count == 0) {
+        return "No family";
```
claudia-golaes-dev: how can you get now to count = 0?

anamariasuditu: > how can you get now to count = 0?

We currently have some clients in the database with no family, and we could also change from the Supabase the number of adults, even if in the `Clients Form` is no longer possible. These are the cases in which the count is 0.

src/app/clients/getExpandedClientDetails.ts
anamariasuditu: > Maybe here you can either omit the word "Occupant" and keep "No family" or just `if (count == 0) { return "No "; }`

Thank you, I modified it

src/app/clients/form/formSections/NumberAdultsCard.tsx
```diff
@@ -77,6 +88,9 @@ const NumberAdultsCard: React.FC<ClientCardProps> = ({
                     })}
                 />
                 {fields.adults.map((adult: Person, index: number) => {
+                    if (index >= fields.numberOfAdults) {
+                        return null;
+                    }
```
AlexandraIrinaMardale: Why do we need this?

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -82,7 +82,11 @@ const getRawClientDetails = async (clientId: string) => {
 };
 
 export const familyCountToFamilyCategory = (count: number): string => {
-    if (count <= 1) {
+    if (count == 0) {
+        return "No Family";
+    }
+
+    if (count == 1) {
```
AlexandraIrinaMardale: Can we use strict equality here, please?

src/app/clients/getExpandedClientDetails.ts
```diff
@@ -215,7 +219,8 @@ export const formatHouseholdFromFamilyDetails = (
     }
 
     const familyCategory = familyCountToFamilyCategory(family.length);
-    const occupantDisplay = `Occupant${adultCount + childCount > 1 ? "s" : ""}`;
+    const totalFamilyMembers = adultCount + childCount;
+    const occupantDisplay = `${totalFamilyMembers === 0 ? "" : "Occupant"}${totalFamilyMembers > 1 ? "s" : ""}`;
 
     return `${familyCategory} ${occupantDisplay} (${adultChildBreakdown.join(", ")})`;
 };
```
AlexandraIrinaMardale: This also needs to be updated as it will show `No Family ()` and we don't need the empty set of parentheses

src/app/clients/form/formSections/NumberAdultsCard.tsx
```diff
@@ -54,6 +54,17 @@ const NumberAdultsCard: React.FC<ClientCardProps> = ({
...
+        if (fields.numberOfAdults > fields.adults.length) {
+            const newAdults = [...fields.adults];
+            while (newAdults.length < fields.numberOfAdults) {
+                newAdults.push({});
+            }
+            fieldSetter({ adults: newAdults });
+        } else if (fields.numberOfAdults < fields.adults.length) {
+            fieldSetter({ adults: fields.adults.slice(0, fields.numberOfAdults) });
+        }
+    }, [fields.numberOfAdults, fields.adults, fieldSetter]);
```
AlexandraIrinaMardale: What do you think about creating a utility function for resizing the array? Also, could we use `Array.from()` with spread syntax instead of the while loop for better readability?

mihaicutaiulia: I've applied the changes here as well as in `NumberChildrenCard`.

src/app/clients/form/formSections/NumberAdultsCard.tsx
mihaicutaiulia: I swapped the `if` check inside the `map` for a `slice`

AlexandraIrinaMardale (approved):

PR #793, opened 2025-08-21
VFB-448: Selected items with empty extra fields not included

anamariasuditu:
## What's changed
Now, selected items should always appear in the summary as long as the checkbox is ticked, regardless of whether the extra information field is filled or not

### Before
https://github.com/user-attachments/assets/e5f6b6e1-2f83-4f60-aeaf-b0df681e2fab 

### After
https://github.com/user-attachments/assets/5cca1a52-2852-41c7-bfee-ff6e0e3b4ddd 

## Checklist

### DIFF
--- a/src/app/clients/form/submitFormHelpers.ts
+++ b/src/app/clients/form/submitFormHelpers.ts
@@ -49,12 +49,13 @@ export const formatClientRecord = (
             fields.dietaryRequirements !== null
                 ? checkboxGroupToArray(fields.dietaryRequirements)
                 : null,
-        hygiene_tampons: fields.hygieneProductsTampons,
-        hygiene_pads: fields.hygieneProductsPads,
+        hygiene_tampons:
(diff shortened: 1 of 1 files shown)


src/app/clients/form/submitFormHelpers.ts
stefanghenescu-work, line 1: This is more of an informative comment as this behaviour applies to all fields, not only the ones you modified. Currently, there’s no check for `undefined`. Do we expect any cases where a field could actually be `undefined`?

stefanghenescu-work: If we want to verify `undefined` behaviour as well, you can use the `??` operator to make it easier to read.

AlexandraIrinaMardale: Is there a reason not to go with the nullish coalescing operator?

AlexandraIrinaMardale (approved):

PR #795, opened 2025-08-21
VFB-451 Restructure the Dietary Requirements table

mihaicutaiulia:
## What's changed
- **Restructured Dietary Requirements table**: The data model has moved from a wide-table approach (`dietary_requirements` with many columns per diet) to a more normalized structure using two new tables: `diets` (for diet types) and `dietary_rules` (for the assignments of items as included/excluded for each diet).
- **New Supabase migrations**:
  - Created `diets` table and seeded with standard diet types.
  - Created `dietary_rules` table to map food items to dietary rules.
  - Added migration scripts to transform/migrate the existing data into the new structure.
  - Added a new view `dietary_rules_plus` for easier querying.
  - Updated names in the `diets` table for better display.
  - Enabled realtime updates for the new tables.
- **Frontend refactor**:
  - Updated all logic and UI in the admin Dietary Requirements table to use the new data structure.
  - The Dietary Requirements table now loads diet types dynamically from the DB, rather than hardcoding them.
  - "Included" and "Excluded" items are now managed through the new rules table.
  - The edit modal now shows only food-related items
- **Data migration**:
  - All existing dietary requirement data is migrated into the new, normalized structure.
  - Special handling to migrate "pet food" as a new item type where needed.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/admin/dieteryRequirementsTable/DietaryRequirementsActions.ts
```diff
@@ -47,66 +20,41 @@ type FetchDietaryRequirementsResult =
...
+                .map((rule) => rule.item_name || "Unnamed Item");
+            return {
+                id: diet.primary_key,
+                dietary_requirement: diet.name,
+                included,
+                excluded,
+            };
+        });
+
+        return { data: rows, error: null };
```
stefanghenescu-work: Can you try and add some newlines here to make the code easier to read?

src/app/admin/dieteryRequirementsTable/DietaryRequirementsActions.ts
```diff
@@ -47,66 +20,41 @@ type FetchDietaryRequirementsResult =
...
+        if (rulesError) {
+            const logId = await logErrorReturnLogId("Failed to fetch dietary rules", {
+                error: rulesError,
+            });
+            return { data: null, error: { type: "failedToFetchDietaryRequirements", logId } };
+        }
 
-        return { data: formattedData, error: null };
+        const rows: DietTableRow[] = (diets || []).map((diet) => {
+            const rulesForDiet = (rules || []).filter((rule) => rule.diet_id === diet.primary_key);
```
stefanghenescu-work: Instead of using the `||` operator I believe is better to have `??`.  This ensures edge cases are handled correctly: `||` treats any false value, which could mask valid data.
You can look here and tell me what you think is better for this specific case. [`||` vs `??`](https://stackoverflow.com/questions/61480993/when-should-i-use-nullish-coalescing-vs-logical-or)

src/app/admin/dieteryRequirementsTable/EditDietaryRequirementsModal.tsx
```diff
@@ -67,45 +62,56 @@ function checkArraysAreEqual(
...
+            .from("dietary_rules")
+            .select()
+            .eq("diet_id", dietId);
+
+        const included = (rulesData || [])
+            .filter((rule) => rule.status === "included")
+            .map((rule) => rule.item_id);
+        const excluded = (rulesData || [])
+            .filter((rule) => rule.status === "excluded")
+            .map((rule) => rule.item_id);
```
stefanghenescu-work: Same for `||` vs `??`

src/app/admin/dieteryRequirementsTable/EditDietaryRequirementsModal.tsx
```diff
@@ -23,25 +21,22 @@ interface Props {
...
+    diet_id: string;
+    item_id: string;
+    status: "included" | "excluded" | "not_specified";
 };
 
-type DietaryRequirementsTableRow = BaseDietaryRequirements & {
+type Item = {
     id: string;
+    item_name: string | null;
+    item_type: string | null;
```
stefanghenescu-work: When will the `item_name` and `item_type` be `null`? From what I can see, when a new item is added on the Lists page, both `description` and `type` are required, so it should not be possible to create a new item with these fields set to null or empty.

stefanghenescu-work (commented):
Before the specific comments regarding the code, I have one question: what happened to the `Pet Food` diet? I can’t seem to find it anywhere here.

mihaicutaiulia:
> Before the specific comments regarding the code, I have one question: what happened to the `Pet Food` diet? I can’t seem to find it anywhere here.

`Pet Food` is no longer classified as a diet — it has been changed to an item type. In the migration script `20250830184058_migrate_dietary_requirements_data.sql`, the data was updated so that items previously associated with the Pet Food diet are now assigned the Pet Food item type.

src/app/admin/dieteryRequirementsTable/EditDietaryRequirementsModal.tsx
mihaicutaiulia: we've updated the system so that both `item_name` and `item_type` are now required when editing or creating new items. However, for items that were created previously (before these fields became required), it's still possible for `item_name` or `item_type` to be null.

PR #802, opened 2025-08-27
VFB-500: Add Concurrency Documentation

anamariasuditu:
## What's changed
Document how concurrent edits behave for different pages/components

### DIFF
--- a/docs/concurrency.md
+++ b/docs/concurrency.md
@@ -70,4 +70,3 @@ Concurrency arises when multiple users try to update the same record at the same
   * The user cancels, exiting edit mode.
   * The user reloads the page.
   * The user successfully submits another row (for tables).
-* Affected rows stay in a consistent state (either rolled back with a fresh DB fetch, or marked as “conflict”).
--- a/docs/concurrency.md
+++ b/docs/concurrency.md
@@ -15,7 +15,7 @@ Concurrency arises when multiple users try to update the same record at the same
 5. **Conflict Handling & Messaging**
    
    If an update fails because the `last_updated` no longer matches, we show the user a consistent warning:
-   *Record has been edited recently – please refresh the page. Log ID: \${error.logId}*
+   *Record has been edited recently – please refresh the page.*
    
 ---
 
@@ -57,11 +57,8 @@ Concurrency arises when multiple users try to update the same record at the same
 * If `count === 0` → no row matched → concurrency conflict.
 * Conflict response:
 
-  * Generate warning log via `logWarningReturnLogId`.
-  * Record audit log entry with `wasSuccess: false`.
-  * Return typed error `{ type: "concurrentUpdateConflict", logId }`.
-* If DB update succeeds → fetch latest record to refresh state.
-* Ensures audit log for every attempt (success/failure).
+  * Show standard concurrency warning.
+* If DB update succeeds or is canceled → fetch latest record to refresh state.
 
 ---
 
@@ -74,9 +71,3 @@ Concurrency arises when multiple users try to update the same record at the same
   * The user reloads the page.
   * The user successfully submits another row (for tables).
 * Affected rows stay in a consistent state (either rolled back with a fresh DB fetch, or marked as “conflict”).
-
----
-
-## Audit Logs 
-
-* Every failed or blocked action logs a warning/error with a unique log ID, ensuring traceability.
--- a/docs/concurrency.md
+++ b/docs/concurrency.md
@@ -20,3 +20,26 @@ Concurrency is a common concern across the site. While the handling may differ b
 ### 5. Maintain Data Consistency
 - If a concurrency conflict is detected, display the following message:
   *Record has been edited recently - please refresh the page. Log ID: ${error.logId}*
+
+## Parcels Page Actions
+
+### 1. Add `last_updated` Field
+- Create a migration to add a `last_updated` column to the table.
+- Use the type `timestamp with time zone` (`timestamptz`) with a default value of `CURRENT_TIMESTAMP`.
+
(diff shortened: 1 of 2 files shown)


docs/concurrency.md
AlexandraIrinaMardale, line 1: This should be more of "how to handle concurrency issues" rather than explaining all the changes we've made so far. 

Steps 1, 2 and 5 are the same and these will always be necessary. Where we can make the distinction is for case like notes vs collection centres - the tricky part being that in the collection centre table, the data gets updated and we may loose the original last updated if we don't use refs.

We should also document how we decided to handle displaying errors, when we clear them, how the rest of the table behaves when a row has run into a concurrent edit

docs/concurrency.md
```diff
@@ -0,0 +1,81 @@
+# How to Handle Concurrency Issues
+
+Concurrency arises when multiple users try to update the same record at the same time. To prevent overwriting changes and to keep data consistent across the app, we follow a few common patterns.
+
+## General Principles
+
+1. **`last_updated` Field**
+   
+   Every table that supports editing should include a `last_updated` column of type `timestamp with time zone` (`timestamptz`)  with default `CURRENT_TIMESTAMP`.
```
AlexandraIrinaMardale: Nitpick: 2 spaces between ``(`timestamptz`)`` and `with`

docs/concurrency.md
```diff
@@ -0,0 +1,81 @@
...
+---
+
+## Error Display & Clearing Strategy
+
+* Errors are shown as floating toasts.
+* They clear automatically when:
+
+  * The user cancels.
+  * No rows remain in edit mode.
+* Affected rows stay in a consistent state (either rolled back with a fresh DB fetch, or marked as “conflict”).
```
AlexandraIrinaMardale: What does this mean?

docs/concurrency.md
```diff
@@ -0,0 +1,81 @@
...
+* Before running a bulk action, re-fetch the selected rows.
+* For each row, check whether `last_updated` has changed since selection.
+* If **any** selected row is outdated, abort the whole action.
+
+  * *Avoids partial updates that leave the system in an inconsistent state.*
+* Show the same standardized concurrency warning.
+* Refs in Admin Tables:
+   * When inserting or editing rows that include refs (like `time_slots` for `collection_centres`), store the original `last_updated` of the parent row in memory (`originalTimestampsRef`) and include it on update.
+
+   * On conflict, show the standard concurrency message and reset state by re-fetching both the parent and its ref rows.
```
AlexandraIrinaMardale: The data is not automatically refreshed - it happens when we reload or cancel our edit

docs/concurrency.md
```diff
@@ -0,0 +1,81 @@
...
+
+---
+
+## Error Display & Clearing Strategy
+
+* Errors are shown as floating toasts.
+* They clear automatically when:
+
+  * The user cancels.
+  * No rows remain in edit mode.
```
AlexandraIrinaMardale: The error is cleared when the specific row is no longer in edit mode (i.e. we cancel the edit) - we don't have any checks for how many rows are in edit mode

docs/concurrency.md
```diff
@@ -0,0 +1,81 @@
...
+  * Return typed error `{ type: "concurrentUpdateConflict", logId }`.
+* If DB update succeeds → fetch latest record to refresh state.
+* Ensures audit log for every attempt (success/failure).
+
+---
+
+## Error Display & Clearing Strategy
+
+* Errors are shown as floating toasts.
+* They clear automatically when:
```
AlexandraIrinaMardale, line 68: I think the error also disappears when we successfully submit another row - if we are editing in a table (also worth checking what happens in the info page )

docs/concurrency.md
anamariasuditu: > What does this mean?

It means that if a row is involved in a concurrent edit, firstly it shows the error, then it fetches again data from DB (after exiting edit mode), meaning these are the 2 states a row goes through in a concurrent edit.

docs/concurrency.md
anamariasuditu: > The data is not automatically refreshed - it happens when we reload or cancel our edit

Yes, I will rephrase it

docs/concurrency.md
anamariasuditu: > The error is cleared when the specific row is no longer in edit mode (i.e. we cancel the edit) - we don't have any checks for how many rows are in edit mode

I will rephrase it

docs/concurrency.md
anamariasuditu: > I think the error also disappears when we successfully submit another row - if we are editing in a table (also worth checking what happens in the info page )

Yes, this is the case for the other concurrent edits. I just noticed that for the info page it doesn't behave in the same way, I will update it in the other ticket to be consistent and add this in the documentation.

docs/concurrency.md
AlexandraIrinaMardale: But it doesn't do that automatically - it shows the error and the user can refresh/cancel so that the data is updated 🤔

docs/concurrency.md
AlexandraIrinaMardale: Have you added this?

docs/concurrency.md
anamariasuditu: > But it doesn't do that automatically - it shows the error and the user can refresh/cancel so that the data is updated 🤔

Alright, I will remove it

docs/concurrency.md
anamariasuditu: > Have you added this?

Yes, here:
The user successfully submits another row (for tables). (line 73)

PR #803, opened 2025-08-27
VFB-511: Drivers Page

claudia-golaes-dev:
### Jira Ticket
- [**VFB-511: Add and manage driver Circuit IDs in the Drivers Tab**](https://softwiretech.atlassian.net/browse/VFB-511)

## What's changed
- Added a new page, called `Drivers`, that can only be accessed by users with organizational roles
- Created the `drivers` table
- Generated seeds for the Drivers Table
- Added audit logs for editing the Drivers table (add/update/delete)

## Screenshots
image
image

## Videos
drivers_page1
drivers_page2

## Checklist

    - Checkout the dev branch
    - Checkout your branch


src/app/drivers/driversTable/DriversTable.tsx
```diff
@@ -0,0 +1,444 @@
...
+    useEffect(() => {
+        setErrorMessage(null);
+        fetchDrivers()
+            .then((response) => setRows(response))
+            .catch((error) => {
+                void logErrorReturnLogId("Error with fetch: Drivers", error);
+                setErrorMessage("Error fetching data, please reload");
+            })
+            .finally(() => setIsLoading(false));
+    }, []);
```
LeonardCiobanu-Softwire, line 114: I think it would be better to use dependencies, rather than letting it re-run after every re-render

src/app/drivers/driversTable/DriversTable.tsx
```diff
@@ -0,0 +1,444 @@
...
+                    setErrorMessage("Error fetching data, please reload");
+                } else {
+                    setErrorMessage(null);
+                }
+            });
+
+        return () => {
+            void supabase.removeChannel(subscriptionChannel);
+        };
+    }, []);
```
LeonardCiobanu-Softwire, line 160: Here too

src/app/drivers/driversTable/DriversTable.tsx
```diff
@@ -0,0 +1,444 @@
...
+        const originalLastUpdated = originalTimestampsRef.current[row.id];
+        const rowWithOriginal = {
+            ...row,
+            originalLastUpdated,
+        };
+
+        try {
+            if (row.isNew) {
+                const sorted = [
+                    ...rows.filter((rowToBeCompared) => rowToBeCompared.id !== row.id),
```
LeonardCiobanu-Softwire, line 207: In what cases would this filter's check be false?

src/app/drivers/driversTable/DriversTable.tsx
```diff
@@ -0,0 +1,444 @@
...
+                    row,
+                ].sort((first, second) => {
+                    const newFirst = first.name ?? "";
+                    const newSecond = second.name ?? "";
+                    return newFirst.localeCompare(newSecond);
```
LeonardCiobanu-Softwire, line 212: I think just using "<" would be enough

src/app/drivers/driversTable/DriversTable.tsx
```diff
@@ -0,0 +1,444 @@
...
+                    setErrorMessage(`Failed to add driver. Log ID: ${insertDriversError.logId}`);
+                    setRows((prevRows) => prevRows.filter((prevRow) => prevRow.id !== row.id));
+                    void sendAuditLog({
+                        ...baseAuditLog,
+                        wasSuccess: false,
+                        logId: insertDriversError.logId,
+                    });
+                    throw new Error("Insert failed");
+                } else {
+                    setRows((prevRows) => prevRows.filter((prevRow) => prevRow.id !== row.id));
```
LeonardCiobanu-Softwire, line 235: Doesn't this remove the row that you wanted to add from the table?
Or is this just a check to see that there are no other rows with that id?

src/app/drivers/driversTable/DriversTable.tsx
```diff
@@ -0,0 +1,444 @@
...
+        }
+    };
+
+    const handleRemoveClick = (id: GridRowId) => () => {
+        const editedRow = rows.find((row) => row.id === id);
+        if (editedRow != undefined) {
+            deleteDbDrivers(editedRow).then(() =>
+                setRows((oldRows) => oldRows.filter((row) => row.id !== id))
+            );
+            const baseAuditLog = getBaseAuditLogForDriversAction("delete a driver", editedRow);
```
LeonardCiobanu-Softwire, line 305: Small typo: 'Delete a driver' instead of 'delete a driver'

LeonardCiobanu-Softwire (commented):
I left some comments 👍

src/app/drivers/driversTable/DriversTable.tsx
claudia-golaes-dev: Yes, it removes the row, but only by design (the row that is in edit mode). The actual row is expected to appear on table refresh, after inserting it in the db

src/app/drivers/driversTable/DriversTable.tsx
claudia-golaes-dev: It’s false only when the IDs are equal

src/app/drivers/driversTable/DriversTable.tsx
claudia-golaes-dev: The sort method needs to have a negative/zero/positive result, and `a<b` would only have a zero/postive result

src/app/drivers/driversTable/DriversTable.tsx
LeonardCiobanu-Softwire: I figured as much, but it's unclear to me if the id is the one in supabase or in the table. Also in each case, why is this check necessary?

PR #814, opened 2025-09-02
Vfb 489 create logs page

LeonardCiobanu-Softwire:
## What's changed
Added new page: **Logs**

- Added to navbar
- Removed the audit logs panel from the **Admin** page
- Only accessible by admin accounts

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image | image |

## Checklist
- [ ] I have self-reviewed this PR


src/app/logs/AuditLogTable/types.ts
```diff
@@ -53,13 +73,16 @@ export interface AuditLogRow {
     statusOrder: string;
     wasSuccess: boolean | null;
     websiteData: string;
-    dietaryRequirement: string;
 }
 
 export const convertAuditLogPlusRowsToAuditLogRows = (
     auditLogResponse: DbAuditLogRow[]
```
alexandrutrifu-urzica, line 79: Can you please rename this to `auditLogPlusRows` to respect the format below?

src/app/logs/AuditLogTable/AuditLogPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const [areFiltersLoadingForFirstTime, setAreFiltersLoadingForFirstTime] =
+        useState<boolean>(false);
+
+    const [urlParamsHaveBeenProcessed, setUrlParamsHaveBeenProcessed] = useState<boolean>(false);
+
+    const [errorMessage, setErrorMessage] = useState<string | null>(null);
+
+    const [selectedLogId, setSelectedLogId] = useState<string | null>(null);
+
+    const currentlyAppliedFilters: any[] = [];
```
alexandrutrifu-urzica: This is never written to - besides, the `Audit Log` page doesn't currently use filters, so it would be safe to remove all related states and table props, including the `CircularProgress` component below. 
Perhaps they could be added in the future if someone decides they are necessary.

src/app/logs/AuditLogTable/fetchAuditLogData.ts
```diff
@@ -0,0 +1,209 @@
...
+
+const getAuditLogQuery = (
+    supabase: Supabase,
+    sortState: AuditLogSortState,
+    filters: any[],
+    selectString = "*"
+): DbQuery<DbAuditLogRow> => {
+    let query = supabase.from("audit_log_plus").select(selectString) as DbQuery<DbAuditLogRow>;
+
+    // here is where the filters would be applied to the query
```
alexandrutrifu-urzica, line 24: Can you remove this?

src/app/logs/AuditLogTable/types.ts
```diff
@@ -27,8 +45,10 @@ export type AuditLogCountResponse =
           error: AuditLogCountError;
       };
 
+export type AuditLogErrorType = "abortedFetch" | "failedAuditLogFetch" | "failedToFetchAuditLogs";
```
alexandrutrifu-urzica: What's the difference between `"failedAuditLogFetch" | "failedToFetchAuditLogs"`?

src/app/logs/AuditLogTable/fetchAuditLogData.ts
```diff
@@ -0,0 +1,209 @@
...
+): Promise<AuditLogRow[]> => {
+    const query = supabase
+        .from("audit_log_plus")
+        .select("*")
+        .in("primary_key", auditLogIds) as DbQuery<DbAuditLogRow>;
+
+    return runAuditLogQueryAndConvertToAuditLogRows(query);
+};
+
+export const getSingleAuditLogById = async (
```
alexandrutrifu-urzica: This is redundant - the method above should suffice. The same applies to the `runAuditLogQuery` methods below, since you can work with a single-element array.

alexandrutrifu-urzica (changes requested):

src/app/logs/AuditLogTable/AuditLogPage.tsx
LeonardCiobanu-Softwire: I think the `CircularProgress` component should be kept, just to better visualize that the params are loading.

src/app/logs/AuditLogTable/fetchAuditLogData.ts
LeonardCiobanu-Softwire: It would be better if I kept it in, just for future reference.

src/app/logs/AuditLogTable/types.ts
LeonardCiobanu-Softwire: Thanks for pointing this out, I'll remove it

PR #812, opened 2025-09-01
VFB-409: Table Refactoring

alexandrutrifu-urzica:
## What's changed
[VFB-409](https://softwiretech.atlassian.net/jira/software/c/projects/VFB/boards/1130?assignee=712020%3A6bec2b54-13b2-4f60-bfb1-44b31b4bfe1f&selectedIssue=VFB-409) The table infrastructure has been updated to use the [Material React Table](https://www.material-react-table.com/) library.

This has enabled us to implement the originally-requested **drag-and-drop functionality** inside the `Lists Page`.

## Screenshots / Videos
| Before     | After      |
|------------|------------|
| image| image|
| image| image|
| image | image|
| image| image |
| image| image|
| image| image|
| image| image|

## Checklist
- [x] The ticket is up-to-date - Please document any deviations from the original approach if there is any.
- [ ] I have documented the testing steps for QA
- [x] I have self-reviewed this PR
- [x] Make sure you've verified it works via `npm run dev`
- [x] Make sure you've verified it works via `npm run build` and `npm run start`
- [x] Make sure you've fixed all linting problems with `npm run lint_fix`
- [x] Make sure you've tested via `npm run test`

PR #879, opened 2026-03-20
Vfb 409 Replace table library for draggable rows on Lists page

stonelink:
## What's changed
This PR replaces https://github.com/LambethFoodbank/foodbankapp/pull/812 - cherry picking those changes onto a branch from latest `dev` to use the material-react-table everywhere, then making the row dragging for the Lists page work correctly.

## Screenshots / Videos
See https://github.com/LambethFoodbank/foodbankapp/pull/812. Layouts have been tweaked since then but not majorly.

## Checklist


docs/table-components.md
```diff
@@ -0,0 +1,270 @@
...
+    columnVisibility,
+  },
+  onColumnVisibilityChange: setColumnVisibility,
+});
+```
+
+We do not use the default `Show/Hide Menu` in the MRT library, mainly because of its limitations regarding column
+groups (e.g. toggling multiple columns through a single handler). To some extent, these drawbacks are specific to the
+library version we are currently using (v2.3.0).
+Given this situation, we opted for our own `ColumnTogglePopup` component, which interacts with MRT's `visibility` state.
```
stonelink, line 53: Look into whether the Referral Details column group is appropriate - probably better to have a concatenated Referral Details single column, which would then enable us to use the MRT Show/Hide menu.

src/app/admin/usersTable/styles.ts
stonelink, line 1: From comparing with `dev`:
* The actions column is much too wide
* Row height is too large

src/app/parcels/parcelsTable/styles.ts
stonelink, line 1: Initial observations comparing with `dev`:
* icons column is too wide
* column height is too much

src/components/Tables/materialTable/materialTableMethods.tsx
```diff
@@ -0,0 +1,66 @@
...
+    TableHeaders,
+} from "@/components/Tables/materialTable/tableTypes";
+
+import React from "react";
+
+const remToPx = (rem: string | undefined): number | undefined => {
+    if (!rem) {
+        return undefined;
+    }
+    return parseFloat(rem) * 16;
```
stonelink: This is assuming that the default font size in the browser is 16pt. It should probably ask the browser.

src/components/Tables/materialTable/tableStyles.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const muiTableHeadCellProps = (theme: DefaultTheme): TableCellProps => ({
+    sx: {
+        backgroundColor: theme.main.background[2],
+        color: theme.main.foreground[2],
+        fontSize: "1rem",
+        fontWeight: "bold",
+        borderColor: theme.main.border,
+        whiteSpace: "normal",
+        wordBreak: "break-word",
+        overflowWrap: "anywhere",
```
stonelink: On `dev` the header text can wrap - see the Reports page

src/app/reports/reportsTable/styles.ts
```diff
@@ -1,5 +1,24 @@
-export const reportsTableColumnStyleOptions = {
-    weekCommencing: {
-        minWidth: "8rem",
```
stonelink, line 3: This is important - the Week Commencing column should have this min width

stonelink: After removing maxWidth from that column and letting headers wrap, standard table layout result is good

docs/table-components.md
stonelink: I'll leave a comment for Dan to consider raising another ticket

src/components/Tables/materialTable/materialTableMethods.tsx
stonelink: Getting `window` at this point is fiddly, so making the assumption explicit for now.

